PR #1, opened 2022-07-19
Improve readme

matteocontrini:
- Fix key and cert parameter names for the server
- Add instructions on how to launch the player when using a custom domain for the Warp server


kixelated:
Oh sorry I didn't see this PR until riiight now. Yeah the README is a mess, let me pull in your changes.

kixelated:
I changed the readme to use `127.0.0.1` instead of `localhost.warp.demo` to simplify the README a bit. I've also asked the WebTransport folks to support self-signed root CAs to get around the Chrome Canary requirement.

PR #13, opened 2023-05-02
Fix for audio breaking after a certain time into the stream

wanjohiryan:
After some time into the audio stream, the audio breaks and an error message is displayed in the console stating that the frame offset exceeds the total frames.  ``
caught RangeError: Failed to execute 'copyTo' on 'AudioData': Frame offset exceeds total frames (1024 >= 1024). ``
This is my proposed solution to the issue.  I have tested and verified that my fix works and the audio now runs smoothly without any break.

I hope it helps :)

### DIFF
--- a/player/src/audio/ring.ts
+++ b/player/src/audio/ring.ts
@@ -59,11 +59,15 @@ export class Ring {
                     frameCount: first.length,
                 })
 
+               //For some reason this breaks audio... and this is my temporary fix
+               //console.log("frame offset", first.length , "frame count", second.length) to test
+               if (first.length < second.length) {
                 frame.copyTo(second, {
                     planeIndex: i,
                     frameOffset: first.length,
                     frameCount: second.length,
                 })
+              }
             }
         }
 
--- a/player/src/audio/ring.ts
+++ b/player/src/audio/ring.ts
@@ -50,7 +50,9 @@ export class Ring {
                     planeIndex: i,
                     frameCount: count,
                 })
-            } else {
+            //audio seems to be breaking whenever endIndex is 0
+            //this works, without "chopiness" 
+            } else if (startIndex  >=  endIndex && endIndex != 0) {
                 const first = channel.subarray(startIndex)
                 const second = channel.subarray(0, endIndex)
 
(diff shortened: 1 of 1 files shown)


kixelated:
Oh hey @wanjohiryan , sorry I missed this PR until now. Hmm, I think the proper fix is something slightly different.


I think you would get this error message when:
```
frameOffset == 1024
frameCount == 0
```

Which would happen when:
```
first.length == 1024
second.length == 0
```

Which would happen when:
```
startIndex == 0
endIndex == 0
```


Hmm, I think you can fix this by changing the `<` on line 45 to `<=`.

wanjohiryan:
Hey @kixelated 

Thanks for your response.

I will update the pull request to reflect this.

If you don't mind me asking, why is does the audio sound "choppy"? It has some background noise i cannot seem to get rid of... the only solution was to add to change the use a higher value at 
https://github.com/wanjohiryan/warp-1/blob/43af19ba94dae9c9ea72f50164c2128c79a39c02/player/src/audio/index.ts#LL18C13-L18C61
``js
ring: new RingInit(2, 4410), // 100ms at 44.1khz
``
where i changed it to 
``js
ring: new RingInit(2, 4410 * 5), // 500ms at 44.1khz
``
Any idea on how to solve this?

kixelated:
The audio rendering is certainly bugged. I'm going to spend some time today trying to debug.

kixelated:
Oh, actually it should be choppy if you're still using this PR. `first.length < second.length` has a 50% chance to happen when the ring buffer wraps around, and then you drop the majority of audio samples that iteration.

Quick math,  I think it results in an average of 9ms dropped every 100ms. Increasing the buffer size reduces the number of times the buffer wraps around so it's an average of 9ms every 500ms, so it's less noticeable.

wanjohiryan:
After doing my local test(s), the audio works fine after i get rid of my "patch", however we are back to square one - where audio cuts off after sometime into the stream

And the console logs show the error:
``
caught RangeError: Failed to execute 'copyTo' on 'AudioData': Frame offset exceeds total frames (1024 >= 1024).
 ``

wanjohiryan:
This seems to work... 

The problematic component seems to be ``endIndex``... that's as far as I could go with my brute force approach. 

Please investigate it

kixelated:
I'll merge this but properly fix it. The issue is that startIndex=0 and endIndex=0 can be interpreted in two different ways (full or empty), so there's a bit more refactoring needed.

Thanks for the PR!

wanjohiryan:
Thanks :)

``The issue is that startIndex=0 and endIndex=0 can be interpreted in two different ways (full or empty)``

What do you propose I do? To be honest I have no idea on how to go about this.

PR #42, opened 2023-07-12
Generic transport trait

francoismichel:
Here is a crate containing a generic transport trait to be used as the transport for MoQ.

This is largely inspired by h3::quic, with a few changes to make it simpler.

Currently, the traits are not required to be `Send`, but it is required by MoQ, so maybe we could also add that constraint on the traits.

Furthermore, we could merge `SendStream` and `SendStreamUnframed` to make it simpler.

I have a quiche-based implementation of this trait and an update of the relay that work correctly.

A next step would be to add setting priorities/urgency through the trait

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,4 +4,5 @@ members = [
 	"moq-transport-quinn",
     "moq-demo",
     "moq-warp",
+    "transport",
 ]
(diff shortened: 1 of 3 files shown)


francoismichel:
Merged in https://github.com/kixelated/webtransport-rs

PR #43, opened 2023-07-15
Make moq-transport totally generic

francoismichel:
Let's iterate on that PR.

This PR makes the moq-transport crate totally generic in the quic implementation viewpoint.
It removes every mention of quinn in `moq-transport` and the warp relay.
It now entirely relies on the `webtransport-generic` crate.

To use moq with a transport implementation, the application must instanciate and stablish a webtransport session
that implements the `Connection` trait of the `webtransport-generic` crate.

the `moq-demo-quinn` crate shows an example of application using quinn underneath.

Let me know your thoughts and don't hesitate to suggest or apply changes.


kixelated:
I took at stab at this with #51. It was a pain but I think I figured out the required traits to avoid wrapping everything in `Box` like this PR.

PR #47, opened 2023-07-22
Remove the file source, since it doesn't loop.

kixelated:
ffmpeg -> moq-warp coming soon


englishm:
LGTM

https://github.com/kixelated/moq-rs/issues/23 -> WONTFIX

PR #49, opened 2023-07-23
Fly.io

englishm:
Adds support for deploying the `moq-demo` relay server to fly.io

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,31 @@
+FROM rust:latest as builder
+
+# Make a fake Rust app to keep a cached layer of compiled crates
+RUN USER=root cargo new app
+WORKDIR /usr/src/app
+COPY Cargo.toml Cargo.lock ./
+# Needs at least a main.rs file with a main function
+RUN mkdir src && echo "fn main(){}" > src/main.rs
+# Will build all dependent crates in release mode
+RUN --mount=type=cache,target=/usr/local/cargo/registry \
+    --mount=type=cache,target=/usr/src/app/target \
+    cargo build --release
+
+# Copy the rest
+COPY . .
+# Build (install) the actual binaries
+RUN cargo install --path .
+
+# Runtime image
+FROM debian:bullseye-slim
+
+# Run as "app" user
+RUN useradd -ms /bin/bash app
+
+USER app
+WORKDIR /app
+
+# Get compiled binaries from builder's cargo install directory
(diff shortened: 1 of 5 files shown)


englishm:
Working now on getting the relay server demo running in GKE, but this might be useful to revisit at some point.

cert/fly-secrets
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
+ cat cert/moq-demo.englishm.net.key | base64 | flyctl secrets set --stage MOQ_KEY=-
```
kixelated: I don't think this file should be checked in?

entrypoint.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env sh
+
+mkdir cert
+echo "$MOQ_CRT" | base64 -d > cert/moq-demo.crt
+echo "$MOQ_KEY" | base64 -d > cert/moq-demo.key
+
+# while true; do
```
kixelated: heh

Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+# Copy the rest
+COPY . .
+# Build (install) the actual binaries
+RUN cargo install --path moq-demo
+
+# Runtime image
+FROM debian:bullseye-slim
+
+# Run as "app" user
+RUN useradd -ms /bin/bash app
```
kixelated, line 35: It's a good practice, but personally I wouldn't use a separate user within a docker container. It just causes permissions pain that doesn't accomplish anything because of the nature of containers.

Dockerfile
```diff
@@ -0,0 +1,41 @@
+FROM rust:latest as builder
+
+# Make a fake Rust app to keep a cached layer of compiled crates
+RUN USER=root cargo new app
```
kixelated, line 4: Is the `USER=root` required?

Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+WORKDIR /usr/src/app
+COPY Cargo.toml Cargo.lock ./
+
+RUN mkdir -p moq-transport/src moq-transport-quinn/src moq-demo/src moq-warp/src
+COPY moq-transport/Cargo.toml moq-transport/Cargo.toml
+COPY moq-transport-quinn/Cargo.toml moq-transport-quinn/Cargo.toml
```
kixelated: Won't exist when you update.

kixelated (commented):
gud Dockerfile, just needs to be updated with the latest branch.

PR #54, opened 2023-08-17
WIP: moq-pub

englishm:
Draft PR just for visibility, not really ready for review yet. 
Branch still includes a bunch of debugging that I'm slowly pulling back out now that I have the basic functionality in place.
I might cherry-pick some of the bits worth keeping out into separate PRs.


englishm:
(Merge conflicts are probably mostly from where I was sprinkling in printf-style debugging with `dbg!` in a bunch of places)

englishm:
Rebasing this and working through the updates it'll need to get back in sync with the latest from `main` now.

englishm:
Oof, forgot that I do still need to switch over the the JSON catalog. Spent so long poking around in quinn today trying to chase down those panics around prioritization and mutex poisoning I almost forgot about that.

moq-pub/src/client.rs
```diff
@@ -0,0 +1,125 @@
...
+}
+
+pub struct Client {
+	session: moq_transport_quinn::Session,
+	source: Arc<MapSource>,
+
+	recv_tasks: JoinSet<anyhow::Result<()>>,
+}
+
+impl Client {
```
kixelated: I don't think this is used?

moq-pub/src/main.rs
```diff
@@ -0,0 +1,61 @@
...
+
+	let config = Config {
+		addr: args.bind_address,
+		uri: args.uri,
+	};
+
+	let mut media = Media::new().await?;
+	let session_runner = SessionRunner::new(config).await?;
+	let mut log_viewer = LogViewer::new(session_runner.get_incoming_receivers().await).await?;
+	let mut media_runner = MediaRunner::new(
```
kixelated, line 36: nit: that's a lot of async constructors and awaits.

moq-pub/src/media.rs
```diff
@@ -0,0 +1,409 @@
...
+			anyhow::ensure!(size_large >= 16, "impossible extended box size: {}", size_large);
+
+			reader.take(size_large - 16)
+		}
+
+		2..=7 => {
+			anyhow::bail!("impossible box size: {}", size)
+		}
+
+		size @ 272033 => {
```
kixelated: xd

moq-pub/src/media_runner.rs
```diff
@@ -0,0 +1,138 @@
...
+		// ANNOUNCE the namespace
+		self.outgoing_ctl_sender
+			.send(Message::Announce(Announce {
+				track_namespace: namespace.to_string(),
+			}))
+			.await?;
+
+		// wait for the go ahead
+		loop {
+			if let Message::AnnounceOk(_) = self.incoming_ctl_receiver.recv().await? {
```
kixelated: Unfortunately this will race with any other incoming messages. I don't think it will break currently but it's going to a blocker eventually.

kixelated: Also what about `AnnounceError`?

kixelated (approved):
rough around the edges but yolo

moq-pub/src/media.rs
englishm: 🤦‍♂️

moq-pub/src/media_runner.rs
englishm: I do need to handle the error case here, yes, but the race shouldn't be an issue because we just loop if we get something other that what we're waiting for. `incoming_ctl_receiver` is a broadcast receiver so that if code elsewhere wants to handle a different message, it should still be able to get it.

- [x] Handle AnnounceError properly

moq-pub/src/main.rs
englishm: True, I should see if some of these might be better to just be synchronous. I guess maybe all of these could be given how I'm using them here. 🤷‍♂️ 

- [ ] ~~Remove unnecessary async fns?~~

moq-pub/src/client.rs
englishm: Oops, yes, this whole file was re-committed accidentally from an earlier version of the code.

- [x] Remove client.rs

moq-pub/README.md
```diff
@@ -0,0 +1,56 @@
+# moq-pub
+
+WARNING: Work in Progress!!!
```
englishm: - [x] Update `moq-pub` README

moq-pub/src/main.rs
englishm: Actually, there are reasons this is simpler - mainly because I want to share some code in these constructors with fns that I _do_ want to be async later on. Could maybe do some more clean up later, but I'm not going to do that now.

moq-pub/src/media_runner.rs
englishm: Re-reading this part of the code to add error handling, I realized that you might still be right about control message races here... `incoming_ctl_receiver` _is_ a broadcast receiver (essentially the inverse of an `mpsc` channel), **but** some of the other receivers are created by resubscribing to **this same instance** _after_ it would have potentially already consumed unrelated messages. Leaving a TODO to revisit that later. :-/

PR #60, opened 2023-08-30
moq-pub: JSON catalog support, bugfixes

englishm:
Fixes some bugs around subscription handling and adds support for the new JSON catalog format

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -490,6 +490,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "four-cc"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


englishm:
cc: @zafergurel. Didn't quite get this working yet, but it's a start!

zafergurel:
@englishm If I can do something to help, please let me know.

englishm:
Ok, after fixing some bugs around subscriber-sent track ids, this hard-coded catalog works for the particular BBB test video I happen to be using. Next I need to make it actually generate real catalogs dynamically by parsing moov headers.

englishm:
Since I've learned the player side isn't currently dependent on all of these fields for playback (it actually still reads them from the init segment), I'm less stressed about working out how to parse all of them from the media itself right now.

Planning to thread through som command line args to set the remaining fields and call this good.

zafergurel:
I got it working on my machine. Thanks! @englishm

Issue #65, opened 2023-09-06
moq-pub: Avoid namespace conflict


englishm:
We should either (or both): 

- Allow users to specify a particular namespace as a command line argument
- Generate UUIDs as part of the namespaces (or maybe just as the default if nothing is specified) to avoid conflict

englishm:
The current value is simply hardcoded here:
https://github.com/kixelated/moq-rs/blob/90818ac848959c7c90f1096a79300944f09dc6cd/moq-pub/src/main.rs#L44

zafergurel:
I got this.

englishm:
Resolved by https://github.com/kixelated/moq-rs/pull/66. Thanks @zafergurel!

PR #66, opened 2023-09-06
moq-pub: Avoid namespace conflict

zafergurel:
namespace as a command line argument
or assign a unique (uuid4) value.

### DIFF
--- a/moq-pub/src/main.rs
+++ b/moq-pub/src/main.rs
@@ -42,7 +42,6 @@ async fn main() -> anyhow::Result<()> {
 	join_set.spawn(async { session_runner.run().await.context("failed to run session runner") });
 	join_set.spawn(async move { log_viewer.run().await.context("failed to run media source") });
 
-	// TODO: generate unique namespace with UUID and/or take a command line arg
 	if config.namespace.len() == 0 {
 		config.namespace = Uuid::new_v4().to_string();
 	}
--- a/moq-pub/src/main.rs
+++ b/moq-pub/src/main.rs
@@ -27,6 +27,10 @@ async fn main() -> anyhow::Result<()> {
 
 	let mut config = Config::parse();
 
+	if config.namespace.len() == 0 {
+		config.namespace = format!("quic.video/{}", Uuid::new_v4());
+	}
+
 	let mut media = Media::new(&config).await?;
 	let session_runner = SessionRunner::new(&config).await?;
 	let mut log_viewer = LogViewer::new(session_runner.get_incoming_receivers().await).await?;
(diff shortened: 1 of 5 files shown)


moq-pub/src/main.rs
```diff
@@ -41,7 +43,10 @@ async fn main() -> anyhow::Result<()> {
 	join_set.spawn(async move { log_viewer.run().await.context("failed to run media source") });
 
 	// TODO: generate unique namespace with UUID and/or take a command line arg
```
englishm: We can check this off now :)

moq-pub/src/main.rs
```diff
@@ -41,7 +43,10 @@ async fn main() -> anyhow::Result<()> {
...
-	media_runner.announce("quic.video/moq-pub-foo", media.source()).await?;
+	if config.namespace.len() == 0 {
+		config.namespace = Uuid::new_v4().to_string();
```
englishm: I wonder if we might want to concatenate this onto a default hostname-like prefix like `quic.video/<uuid>` to convey a bit more of what we expect the structure/semantics of namespaces to be. Fine as-is, but that's my only nit here.

moq-pub/src/main.rs
```diff
@@ -17,13 +17,15 @@ use media::*;
...
 // TODO: clap complete
 
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
 	env_logger::init();
 
-	let config = Config::parse();
+	let mut config = Config::parse();
 
 	let mut media = Media::new(&config).await?;
```
englishm, line 34: We need to move the uuid namespace generation logic above here so we don't end up with an empty namespace in the catalog.

englishm (changes requested):
Tested! Works with the caveat of one issue of possible empty namespaces ending up in the catalog. Commented on a couple other tiny nits, but otherwise this looks pretty good! Thank you!

moq-pub/src/main.rs
englishm: I realize now that I had introduced that second hardcoded namespace string in `src/media.rs` after leaving this TODO comment so it ended up being a bit misleading, sorry!

moq-pub/src/main.rs
zafergurel: Done

moq-pub/src/main.rs
zafergurel: I missed this one :). Removed it.

moq-pub/src/main.rs
zafergurel: Sounds good, done.

zafergurel:
Fixes #65

englishm (approved):
LGTM!

PR #74, opened 2023-09-18
Implement prioritization in moq-pub

kixelated:
Here's the main change in webtransport-quinn 0.5.3: https://github.com/kixelated/webtransport-rs/commit/ec553fa340895aad3505487f4624440e111e3384

I haven't run into any errors so I don't know what was broken before @englishm. I'm hoping that setting the stream priority to max when writing the stream header avoids the issue? Otherwise we need to go bug diving.


englishm:
The panics I was seeing before originated here:
https://github.com/quinn-rs/quinn/blob/a7a3c4da6a94504a53ef34ffe19eb862fe3c4820/quinn-proto/src/connection/streams/state.rs#L495-L509 but I never did wrap my head around how it could get into that state.

kixelated:
Let's try this and see what happens.

PR #83, opened 2023-09-26
(DO NOT MERGE) Quick fix to use local webtransport crates

englishm:
Until a new version of webtransport-proto can be pushed to crates.io

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2119,8 +2119,6 @@ dependencies = [
 [[package]]
 name = "webtransport-generic"
 version = "0.5.0"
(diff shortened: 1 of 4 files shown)


englishm:
Assumes a checkout of [webtransport-rs](https://github.com/kixelated/webtransport-rs) alongsize moq-rs

englishm:
Address an issue with WebTransport handshakes that appears with Chrome 117+.

englishm:
Fixed in 5e4eb420c0922c6b82230775f5ead3e6b11950e6

PR #112, opened 2023-10-24
Add support for mp4a audio with moq-pub.

kixelated:
I would refactor a bit but that can wait until we have the proper warp catalog spec.


kixelated:
I can't quite figure out the issue with ffmpeg and audio timestamps.

Issue #108, opened 2023-10-23
Relay - Cache Network


kixelated:
The relays form a mesh network. This won’t scale past a few nodes.

PR #118, opened 2023-11-06
moq-api controlled routing of subscriptions between relays

Markcanfly:
Implement a possiblity for arbitrary routing of subscriptions between relays, controlled by **moq-api**.

The default behavior stays the same, relays are connected as a full mesh. Meaning if you SUBSCRIBE to a track on a relay, it will directly call out to the origin relay of that track, and SUBSCRIBE to the track there.

If however you supply one or more `--next-relay ${RELAY_URL}` parameters when starting the relay, the relay will send these next_relay URLs to **moq-api** as a URL parameter. If **moq-api** finds the track in its KV store, instead of returning the real origin of the track it will choose from the supplied `next_relays`, and return that as the origin.

This results in the ability to restrict traffic within nodes to arbitrary topologies. See examples in [dev/topologies/star/topo.md](https://github.com/Markcanfly/moq-rs/blob/b2aa06ed5033a2ba14c9c497f86d52a732303194/dev/topologies/star/topo.md) and [dev/topologies/spine-and-leaf/topo.md](https://github.com/Markcanfly/moq-rs/blob/b2aa06ed5033a2ba14c9c497f86d52a732303194/dev/topologies/spine-and-leaf/topo.md)

By default the choice of the next relay is random, to serve as a basic load balancing mechanism, but smarter routing patterns can be implemented there.


Markcanfly:
This is a solution for #108

kixelated (changes requested):
The code looks great, but I'm not quite sure how we would use it. Can you give an example using some real-world data-centers for my small brain?

For example, let's say we have us-east, us-central, us-west, and eu-west.

If us-central is configured with `--next-relays us-west --next-relays us-east`, how does it fetch from an origin in `eu-west`? If I understand correctly, it will randomly choose between east and west, but one of them is clearly wrong and headed in the wrong direction. If us-central was a spine instead, well you have the same problem but with us-east getting to us-west.

I think the --next-relays is absolutely a great way to configure adjacent neighbors. But I think moq-api has to do some form of path solving, otherwise rolling the dice will just result in poor routing.

Markcanfly:
For now it doesn't have the stateful routing, this is more of a skeleton for that, but it's still suitable for a simple tree-like structure. I won't have time to work on figuring out an actual application-layer routing for this in the next 2 months, so if anyone wants to use this as a base, please do.

kixelated:
> For now it doesn't have the stateful routing, this is more of a skeleton for that, but it's still suitable for a simple tree-like structure. I won't have time to work on figuring out an actual application-layer routing for this in the next 2 months, so if anyone wants to use this as a base, please do.

Yeah of course.

I think the play is `PUT relay/<name>` on startup and keep refreshing that entry while the relay is alive. It would contain the paths to other relays via `--next-relay`. You could then make a simple path-finding algorithm to figure out how to go from one relay to another.

But I don't think we can run this code until then. Rolling the dice on which non-equal route to take is just going to make the experience worse.

PR #125, opened 2023-11-16
Remove the custom deploy stuff for moq-pub.

kixelated:
It should be in a separate repo/image.

@englishm if you have a chance, could you also move the fly stuff into a separate repo/image? I think this repo should just provide compiled binaries without deploy/cloud gunk included.

Here's what I'm doing for the code that publishes BBB and the clock: https://github.com/kixelated/quic.video/blob/main/pub/Dockerfile

It does mean you have to build the final image separately, but it's not that bad and gives you more control over what you actually want to deploy.


kixelated:
context: I'm adding the word "chicago:" in front of the time on BBB. That doesn't make any sense to be included in a generic moq-rs image, as it's very specific to the hosting in GCP.

PR #127, opened 2023-11-22
feat: add docker-compose.yml

rfwatson:
Add a docker-compose.yml for local testing/development.

Also updated the Dockerfile a bit, image is now ~150mb instead of 1.5gb and removed some stuff that is unused (I think). Don't see a reason this shouldn't work on fly.io but I can't test that part.


kixelated (approved):
Thanks!

Fun fact: I started making something with docker-compose... but I wanted to try out something else out. I started using [podman to create a k8s cluster](https://docs.podman.io/en/v4.2/markdown/podman-play-kube.1.html) but I got bored and just made those crude shell scripts instead.

rfwatson:
Nice! Podman looks interesting, need to find some time to play with it.

I might do a separate PR to allow the services to accept env vars instead of (or perhaps as well as) the command line flags, if you have no objections?

PR #133, opened 2024-02-09
feat: add moq-sub to subscribe to media from moq relays

Frando:
This adds a simple `moq-sub` tool which allows to connect to a moq relay and subscribe to a broadcast. It expects the format as published by `moq-pub`, and will then dump the first video and first audio track to stdout. By piping `moq-sub` into e.g. `ffplay` you can play media from a moq relay.

Also includes a `dev/sub` script that works without further arguments when using `dev/relay` and `dev/pub`.

Also added an optional `AUDIO=1` env var to `dev/pub` to not skip audio tracks.


moq-sub/src/media.rs
```diff
@@ -0,0 +1,160 @@
...
+		debug!("track {name}: finish");
+		Ok(())
+	}
+
+	async fn recv_segment(mut segment: segment::Subscriber, out: Arc<Mutex<O>>) -> anyhow::Result<()> {
+		trace!("segment={} start", segment.sequence);
+		while let Some(fragment) = segment.fragment().await? {
+			trace!("segment={} fragment={}", segment.sequence, fragment.sequence);
+			let buf = Self::recv_fragment(fragment).await?;
+			out.lock().await.write_all(&buf).await?;
```
kixelated, line 108: Seems fine although we will need to handle this better eventually. MoQ is meant to be lossy.

kixelated (approved):
thanks @Frando these are well done

moq-sub/src/media.rs
Frando: The segments are already received in parallel. Your suggestion would be to also receive fragments in parallel?

kixelated: Unfortunately, the fragments need to be written in order for playback to work. We either need separate files for each segment (like HLS/DASH), or drop the previous segment on each new segment. The current code will interleave fragments from different segments which won't playback.

Frando:
Feel free to merge this and my other PR :) I am not working on these further for the time being.

kixelated:
Working on updating this to the new API.

PR #134, opened 2024-02-12
feat: expose moq-pub as library with generic input

Frando:
This PR has two changes, which allow to reuse the code in `moq-pub` from other libraries or apps:
* adds a `lib.rs` to `moq-pub` to expose the `media` module
* lets the `Media` struct accept an `input: impl AsyncRead` argument instead of being hardcoded to `stdin`

### DIFF
--- a/moq-pub/src/lib.rs
+++ b/moq-pub/src/lib.rs
@@ -1,0 +1,1 @@
+pub mod media;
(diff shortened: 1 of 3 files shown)


kixelated (approved):

Frando:
The failure seems unrelated to me?
```

error: unused import: `string::*`
Error:   --> moq-transport/src/coding/mod.rs:10:9
   |
10 | pub use string::*;
   |         ^^^^^^^^^
   |
   = note: `-D unused-imports` implied by `-D warnings`
```

kixelated:
Sorry for the delay!

PR #151, opened 2024-04-06
Replace webtransport-generic with web-transport

kixelated:
Makes things exponentially simpler and makes supporting WASM more feasible.


kixelated:
Also the hyphen is because somebody is squatting on the `webtransport` crate. But not the `web-transport` crate so I renamed everything to match.

PR #166, opened 2024-05-30
filter DNS query results to only include addresses that our quic endpoint can use

Xaeroxe:
Minor bug fix. Without this change it's possible that the connection will attempt to use an IP address which the internal `quinn::Endpoint` can't actually connect to.

### DIFF
--- a/moq-native/src/quic.rs
+++ b/moq-native/src/quic.rs
@@ -197,7 +197,7 @@ impl Client {
 		let addr = tokio::net::lookup_host((host.clone(), port))
 			.await
 			.context("failed DNS lookup")?
-			.filter(|s| own_protocol_ipv4 == s.is_ipv4())
+			.filter(|s| own_protocol_ipv4 == s.ip().to_canonical().is_ipv4())
 			.next()
 			.context("no DNS entries")?;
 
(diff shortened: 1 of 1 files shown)


Xaeroxe:
I went ahead and fixed the clippy warning. I also adjusted the "context" of the error to more accurately reflect the actual problem.

kixelated:
Yo, reverting this because it breaks production. I think the moq-pub client binds via ipv6 but connects to a ipv4 address, which should still be valid?

kixelated:
But yeah the IPv4 vs IPv6 stuff is a mess.

Xaeroxe:
I suspect I didn’t account for the IPV6_V6ONLY flag being set to false. The default value for this flag varies by operating system.

This would be easier if moq-rs created the quinn::Endpoint internally rather than accepting one from the caller. Then in that case we could use a method like libc’s getaddrinfo

kixelated:
> I suspect I didn’t account for the IPV6_V6ONLY flag being set to false. The default value for this flag varies by operating system.
> 
> This would be easier if moq-rs created the quinn::Endpoint internally rather than accepting one from the caller.

moq-native creates the quinn::Endpoint, because yeah it's a bit advanced to create your own.

PR #170, opened 2024-06-17
moq-sub (updated)

englishm:
Updated version of https://github.com/kixelated/moq-rs/pull/133 rebased on the latest from main and updated to work with the current API.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -29,6 +29,8 @@ jobs:
 
       # Check for semvar violations
       - uses: obi1kenobi/cargo-semver-checks-action@v2
+        with:
+          exclude: moq-sub
 
       # Used to publish any new crates
       - uses: katyo/publish-crates@v2
--- a/dev/pub
+++ b/dev/pub
@@ -36,12 +36,6 @@ URL="${URL:-"$SCHEME://$ADDR"}"
 # Default to a source video
 INPUT="${INPUT:-dev/bbb.mp4}"
 
-# Only inlcude audio track if AUDIO=1 is set
-# TODO enable audio by default again once fixed.
-if [[ ! -v AUDIO ]]; then
-	NO_AUDIO=1
-fi
-
 # Print out the watch URL
 echo "Watch URL: https://quic.video/watch/$NAME?server=$ADDR"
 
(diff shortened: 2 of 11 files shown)


.github/workflows/pr.yml
```diff
@@ -29,6 +29,8 @@ jobs:
 
       # Check for semvar violations
       - uses: obi1kenobi/cargo-semver-checks-action@v2
+        with:
+          exclude: moq-sub
```
englishm, line 33: I probably need to land this on `main` in a separate PR first. Bah.

englishm: Oh no, that worked. 
(Policy-wise that seems like a weird default to use the Action config from the branch you're on, but OK, whatever)

dev/pub
```diff
@@ -43,6 +43,6 @@ echo "Watch URL: https://quic.video/watch/$NAME?server=$ADDR"
 ffmpeg -hide_banner -v quiet \
 	-stream_loop -1 -re \
 	-i "$INPUT" \
-	-c copy \
+	-c copy -an \
```
kixelated, line 46: No audio?

moq-sub/src/main.rs
```diff
@@ -0,0 +1,92 @@
...
+
+	tokio::select! {
+		res = session.run() => res.context("session error")?,
+		res = media.run() => res.context("media error")?,
+	}
+
+	Ok(())
+}
+
+pub struct NoCertificateVerification {}
```
kixelated, line 48: I don't think this is needed any longer.

moq-sub/src/main.rs
```diff
@@ -0,0 +1,92 @@
...
+	/// The TLS configuration.
+	#[command(flatten)]
+	pub tls: moq_native::tls::Args,
+}
+
+fn moq_url(s: &str) -> Result<Url, String> {
+	let url = Url::try_from(s).map_err(|e| e.to_string())?;
+
+	// Make sure the scheme is moq
+	if url.scheme() != "https" {
```
kixelated, line 87: We support moqt:// now too.

kixelated (approved):
I'll clean it up later, good enough to merge whenever.

dev/pub
englishm: Wasn't working reliably when I tested, but could be because moq-sub is pretty naive about ordering still.

moq-sub/src/main.rs
englishm: Ack. Let's clean that up post-merge.

moq-sub/src/main.rs
englishm: Ah yeah, let's add that option post-merge also.

PR #172, opened 2024-06-17
Update to draft-04

englishm:
Update moq-rs to [draft-04](https://www.ietf.org/archive/id/draft-ietf-moq-transport-04.html)

Checklist (PRs to spec between -03 and -04 tags):
- [x] 405 ContentExists alignment (already matches existing implementation)
- Changes that impact the Subscribe message wire format:
  - https://www.ietf.org/archive/id/draft-ietf-moq-transport-03.html#name-subscribe-format
  - https://www.ietf.org/archive/id/draft-ietf-moq-transport-04.html#name-subscribe-format
  - [x] 451 Rename Track Request Parameters to Subscribe Parameters
  - [x] 439 Simplify/Clarify Subscribe by adding 4 modes
  - [x] 432 Make SUBSCRIBE start and end (typically) absolute
  - [x] 444 The end MUST be after the start
     - N/A - we don't really do things with these ranges yet... 
  - [x] 454 Remove special value for SUBSCRIBE StartObject
     - N/A - we don't really do things with these ranges yet... 
    
- Object Status
  - [x] 429 Add object status used to indicate lost or nonexistent objects
- Track Status
  - [x] 425 Add a way to get the current group ID
    - Track Status Request
    - Track Status
- [x] 435 SUBSCRIBE_UPDATE
  - [x] 457 Add a message code for SUBSCRIBE_UPDATE

For a few things here, implementations are only partial (full handling isn't always implemented, and not every new feature is exposed to library consumers), but this should cover at least all of the wire image changes so that basic interop can be achieved.


kixelated (approved):
Gotta update moq-js at least to add ObjectStatus before we can merge.

englishm:
Yep, this is just blocked on https://github.com/kixelated/moq-js/pull/107 and a minor rebase/cleanup to deal with version bumps that landed on main since I last touched this.

PR #173, opened 2024-07-05
Add sslkeylogfile envvar for debugging

1majom:
While trying to inspect the traffic more closely I have found out that with these two lines we can use the SSLKEYLOGFILE environment variable to later decrypt pcap files and see more of the QUIC and WebTransport traffic. It might not be useful. 

About envvar from the quinn repo:

> If the SSLKEYLOGFILE environment variable is set, the tests will emit UDP packets for inspection using external protocol analyzers like Wireshark, and NSS-compatible key logs for the client side of each connection will be written to the path specified in the variable.


kixelated (approved):
Secret ENV variables like this are gross but I suppose it's the only option.

PR #181, opened 2024-08-12
really small correction with url.join

1majom:
I know this is not an important change, but I came across it while changing the api for my own project and it might help others who want to do the same.

The url's join method overwrites the previous join. So e.g. url.join("origin/").join("vaj") won't make an url like kiscica.hu/origin/vaj, but kiscica.hu/vaj.

### DIFF
--- a/moq-api/src/client.rs
+++ b/moq-api/src/client.rs
@@ -17,7 +17,7 @@ impl Client {
 	}
 
 	pub async fn get_origin(&self, namespace: &str) -> Result<Option<Origin>, ApiError> {
-		let url = self.url.join("origin/")?.join(namespace)?;
+		let url = self.url.join(&format!("origin/{namespace}"))?;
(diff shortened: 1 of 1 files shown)


kixelated:
TIL `URL::join` is weird. But shouldn't the existing code work because of the trailing slash? That's probably why I added it in the first place.

1majom:
The existing code will work because this is consistent, so for all paths it will be only moq.relay/<uuid> not moq.relay/origin/<uuid>. I just wanted to correct this because when I made more complicated paths I saw that double joins won't get us the wanted outcome.

But if you prefer it that way instead of sticking to the origin/<uuid> part the affected lines could be replaced simply with "self.url.join(namespace)?". That way it will do the same as now without the double joins.

kixelated:
I just wasn't sure if anything was broken. Thanks for the improvement!

PR #219, opened 2024-10-29
add flake.nix, export moq packages, overlay and nixosModule

therishidesai:
For people who use nix this provides a nix dev shell, nix moq packages and a nixosModule for the relay server so you can run the relay server on a nixos system pretty easily.

### DIFF
--- a/nix/packages/moq.nix
+++ b/nix/packages/moq.nix
@@ -1,20 +1,22 @@
 { self, nixpkgs, flake-utils, ... }:
 flake-utils.lib.eachDefaultSystem (system:
   let
-    pkgs = import nixpkgs {
-      inherit system;
-    };
+    pkgs = nixpkgs.legacyPackages.${system};
+    moq-relay-version = (pkgs.lib.importTOML ../../moq-relay/Cargo.toml).package.version;
   in
     with pkgs;
     {
       packages = rec {
         moq-relay = rustPlatform.buildRustPackage rec {
           pname = "moq-relay";
-          version = "0.6.7";
+          version = moq-relay-version;
 
           src = ../../.;
 
-          cargoHash = "sha256-kOAulF1OqR1VBKSy15RURJEk2ZpgZIFxPwrr03RbvPk=";
+          cargoLock = {
+            lockFile = ../../Cargo.lock;
+            allowBuiltinFetchGit = true;
+          };
 
           nativeBuildInputs = [ pkg-config ];
 
--- a/nix/packages/moq.nix
+++ b/nix/packages/moq.nix
@@ -1,0 +1,32 @@
+{ self, nixpkgs, flake-utils, ... }:
+flake-utils.lib.eachDefaultSystem (system:
+  let
+    pkgs = import nixpkgs {
+      inherit system;
+    };
+  in
+    with pkgs;
+    {
+      packages = rec {
+        moq-relay = rustPlatform.buildRustPackage rec {
+          pname = "moq-relay";
+          version = "0.6.7";
+
+          src = ../../.;
+
+          cargoHash = "sha256-kOAulF1OqR1VBKSy15RURJEk2ZpgZIFxPwrr03RbvPk=";
+
+          nativeBuildInputs = [ pkg-config ];
+
+          buildInputs = [ libressl ];
+
+          meta = {
+            description = "Media Over QUIC relay server";
+            mainProgram = "moq-relay";
+            homepage = "https://quic.video/";
+            changelog = "https://github.com/kixelated/moq-rs/releases/tag/moq-relay-v${version}";
+          };
+        };
+      };
+    }
+)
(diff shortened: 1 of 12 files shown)


nix/packages/moq.nix
```diff
@@ -0,0 +1,32 @@
...
+    pkgs = import nixpkgs {
+      inherit system;
+    };
+  in
+    with pkgs;
+    {
+      packages = rec {
+        moq-relay = rustPlatform.buildRustPackage rec {
+          pname = "moq-relay";
+          version = "0.6.7";
```
kixelated: Can you make this pull from Cargo.toml automatically?

therishidesai: Unfortunately I can't right now. May be able to move to `crate2nix` and it can figure this stuff out at eval time. Can look more into this in a future PR

nix/packages/moq.nix
```diff
@@ -0,0 +1,32 @@
...
+
+          src = ../../.;
+
+          cargoHash = "sha256-kOAulF1OqR1VBKSy15RURJEk2ZpgZIFxPwrr03RbvPk=";
```
niklaskorz: This should load Cargo.lock directly instead, otherwise the hash will have to be updated whenever Cargo.lock changes

nix/packages/moq.nix
niklaskorz: Use `builtins.fromTOML` to extract the version field from `Cargo.toml`. No need for crate2nix to accomplish this.

nix/packages/moq.nix
```diff
@@ -0,0 +1,32 @@
...
+
+          nativeBuildInputs = [ pkg-config ];
+
+          buildInputs = [ libressl ];
```
niklaskorz: I'd be curious to know why libressl instead of openssl?

nix/shell.nix
```diff
@@ -0,0 +1,20 @@
+{ self, nixpkgs, flake-utils, ... }:
+flake-utils.lib.eachDefaultSystem (system:
+  let
+    pkgs = import nixpkgs {
+      inherit system;
+    };
```
niklaskorz:
```suggestion
    pkgs = nixpkgs.legacyPackages.${system};
```

With import, you're re-evaluating nixpkgs, so it should be avoided when possible.

nix/packages/moq.nix
kixelated: I can't manually bump this version especially because we're using release-plz. What about this snippet I found from a quick search?

```
{ pkgs ? import <nixpkgs> { } }:
let manifest = (pkgs.lib.importTOML ./Cargo.toml).package;
in
pkgs.rustPlatform.buildRustPackage rec {
  pname = manifest.name;
  version = manifest.version;
  cargoLock.lockFile = ./Cargo.lock;
  src = pkgs.lib.cleanSource ./.;
}
```

niklaskorz: Yup that should work

kixelated: > Use `builtins.fromTOML` to extract the version field from `Cargo.toml`. No need for crate2nix to accomplish this.

ha yeah I didn't see your reply until just now

niklaskorz: > > Use `builtins.fromTOML` to extract the version field from `Cargo.toml`. No need for crate2nix to accomplish this.
> 
> ha yeah I didn't see your reply until just now

`lib.importTOML` (from your snippet) is preferred here anyway because it combines reading the file + parsing, while `builtins.fromTOML` takes a string to parse as input.

nix/packages/moq.nix
therishidesai: libressl is a clean re-write of openssl with supposedly fewer bugs. The API is equivalent to openssl so it works the same. It is also the default ssl impelementation used in upstream nixpkgs

nix/packages/moq.nix
therishidesai: oh TIL about `builtins.fromTOML` will update it

nix/packages/moq.nix
niklaskorz: https://github.com/englishm/moq-rs/blob/cfd556680b05c811d21e72a07953b05f64a1a4a6/package.nix#L27-L30

therishidesai:
@kixelated @niklaskorz Everything should be cleaned up now from your comments

niklaskorz:
Fails to build on macOS as frameworks are missing, see:

https://github.com/englishm/moq-rs/blob/cfd556680b05c811d21e72a07953b05f64a1a4a6/package.nix#L41-L44

therishidesai:
> Fails to build on macOS as frameworks are missing, see:
> 
> https://github.com/englishm/moq-rs/blob/cfd556680b05c811d21e72a07953b05f64a1a4a6/package.nix#L41-L44

try now

kixelated (approved):

PR #272, opened 2024-12-23
feat: request for the fingerprint anytime an http url is passed

therishidesai:


moq-web/src/demo/index.html
```diff
@@ -8,7 +8,7 @@
 </head>
 
 <body>
-	<moq-video src="https://localhost:4443/demo/bbb" />
+	<moq-video src="http://localhost:4443/test-zed" />
```
kixelated: can you make this bbb again

moq-web/src/session.rs
```diff
@@ -3,23 +3,22 @@ use url::Url;
 
 use crate::{Error, Result};
 
-pub async fn connect(addr: &Url) -> Result<moq_transfork::Session> {
+pub async fn connect(addr: &mut Url) -> Result<moq_transfork::Session> {
```
kixelated: IMO Don't make this `&mut`. Instead either:
1. Clone the `&Url` if needed (http only).
2. Take ownership of the `Url` (caller will need to `clone()`).

The current code destructively changes the protocol of `self.src` which might cause a bug if we ever try to reconnect.

moq-relay/src/web.rs
```diff
@@ -1,7 +1,7 @@
 use std::{net, sync::Arc};
```
kixelated: You can clean up these unused imports too.

kixelated (approved):
Approach is gud, just a few PR nits.

moq-web/src/session.rs
therishidesai: I have to transform this URL  to https so that it works with WebTransport. I guess I can do that in the place where you create the MoQ connection

kixelated: Instead of:

```rust
let _ = addr.set_scheme("https");
```

You can do:
```rust
// Make a copy of the address, changing it from HTTP to HTTPS for WebTransport:
let mut addr = addr.clone();
let _ = addr.set_scheme("https");
```

therishidesai: Oh yeah that makes sense. I didn't realize the actual MoQ session gets created in the connect function lol

therishidesai:
Fixed the nits, and tested that everything works as expected

PR #355, opened 2025-03-15
Moq karp server

LiamF-2261667:
Adds a server version to moq-karp, so it doens't need a relay running to work.


moq-karp/src/cmaf/import.rs
```diff
@@ -284,10 +279,27 @@ impl Import {
 	}
 
 	// Read the media from a stream until processing the moov atom.
-	pub async fn init_from<T: AsyncRead + Unpin>(&mut self, input: &mut T) -> Result<()> {
-		let _ftyp = mp4_atom::Ftyp::read_from(input).await?;
+	pub async fn init_from<T: AsyncRead + Unpin>(&mut self, input: &mut T, broadcast: &mut BroadcastProducer) -> Result<()> {
+		// TODO: Do we need to read this?
+		let _ftyp = match mp4_atom::Ftyp::read_from(input).await { _ => {} };
```
LiamF-2261667: Forgot to undo this, was for testing purposes.

moq-karp/src/cmaf/import.rs
```diff
@@ -14,9 +13,6 @@ pub struct Import {
 	// Any partial data in the input buffer
 	buffer: BytesMut,
 
-	// The broadcast being produced
-	broadcast: BroadcastProducer,
```
kixelated: `BroadcastProducer` is `Clone`, so why is this required?

moq-karp/src/broadcast.rs
```diff
@@ -1,116 +1,175 @@
...
-	pub session: Session,
+	listeners: Vec<BroadcastListener>,
 	pub path: String,
 	id: u64,
-	catalog: Lock<CatalogProducer>,
+	context: Context,
 }
 
 impl BroadcastProducer {
-	pub fn new(mut session: Session, path: String) -> Result<Self> {
```
kixelated, line 18: +1

moq-karp/src/broadcast.rs
```diff
@@ -1,116 +1,175 @@
...
+
+		Ok(())
+	}
+}
+
 #[derive(Debug, Clone)]
 #[debug("{:?}", path)]
 pub struct BroadcastProducer {
-	pub session: Session,
+	listeners: Vec<BroadcastListener>,
```
kixelated: IMO:
```rust
#[derive(Debug, Clone)]
struct BroadcastProducerState {
    catalog: CatalogProducer,
    tracks: Vec<TrackProducer>,
    subscribers: Vec<Session>,
}

#[derive(Debug, Clone)]
#[debug("{:?}", path)]
pub struct BroadcastProducer {
  pub path: String,
  id: u64,
  state: Lock<BroadcastProducerState>,
}
```

moq-karp/src/broadcast.rs
```diff
@@ -1,116 +1,175 @@
...
 use moq_async::{spawn, Lock};
 use moq_transfork::{Announced, AnnouncedConsumer, AnnouncedMatch, Session};
+use crate::track::TrackConsumer;
 
 use derive_more::Debug;
 
+#[derive(Clone)]
+struct BroadcastListener {
+	pub session: Session,
+	pub catalog: Lock<CatalogProducer>,
```
kixelated: Why do you need this? I think you delete it and move it to `BroadcastProducerState` instead of giving each listener it's own copy of the catalog producer.

moq-karp/src/broadcast.rs
kixelated: Also, you should have a way to remove a `BroadcasterListener` when the session closes. Otherwise it's a memory leak.

And yeah that would involve a background task unfortunately.

kixelated: An alternative API would be something like `BroadcastProducer/BroadcastConsumer` pair in moq_transfork with a:

```rust
impl BroadcastProducer {
  pub fn publish(&mut self, track: TrackConsumer) {
    // inserts a new track and notifies all consumers
  }
}  

impl BroadcastConsumer {
  pub fn next(&mut self) -> Option<TrackConsumer> {
    // .. waits for a new track
  }
}
```

That way you wouldn't need a Vec of subscribers. You can instead spawn a background task to block on `broadcast.next()` and `session.closed()`.

kixelated (commented):
Overall the right direction. If `BroadcastProducer` (properly) remains `Clone`, with a copy given to `Import`, then we can merge this and I can try cleaning up.

moq-karp/src/broadcast.rs
```diff
@@ -1,116 +1,175 @@
...
+					catalog.publish().unwrap();
+				});
+			}
+		}
+
+		Ok(())
+	}
+}
+
 #[derive(Debug, Clone)]
```
kixelated, line 13: Yeah this `Clone` is why you had to remove `BroadcastProducer` from `Import`. That's because the listeners `Vec` would not be shared between the multiple handles. If you put `listeners` behind a `Lock` then it should just work.

kixelated: We could remove the `Clone` but someone wanted a way to get access to the current catalog. I think we keep it for now.

moq-karp/src/broadcast.rs
LiamF-2261667: What do you mean by this?

moq-karp/src/broadcast.rs
```diff
@@ -1,122 +1,137 @@
...
-
-use moq_async::{spawn, Lock};
+use crate::{Audio, Catalog, Error, Result, Track, TrackProducer, Video};
+use moq_async::{Lock};
 use moq_transfork::{Announced, AnnouncedConsumer, AnnouncedMatch, Session};
-
+use crate::track::TrackConsumer;
 use derive_more::Debug;
 
+#[derive(Clone)]
```
kixelated: This state shouldn't be Clone. We only want to Clone the mutex.

moq-karp/src/broadcast.rs
kixelated: Sorry I meant this is the right direction.

moq-karp/src/broadcast.rs
```diff
@@ -1,122 +1,137 @@
...
-		self.session.publish(consumer)?;
+		// Add the session to the list of subscribers
+		state.subscribers.push(session.clone());
 
-		let mut catalog = self.catalog.lock();
-		catalog.current.video.push(info.clone());
-		catalog.publish()?;
+		// If the session closes, remove it from the list of subscribers
+		let state = self.state.clone();
+		tokio::spawn(async move {
```
kixelated: Use the `moq-async` spawn so it'll work on WASM.

moq-karp/Cargo.toml
```diff
@@ -37,6 +37,14 @@ tokio = { version = "1.43", features = ["macros"] }
 
 web-time = "1"
 
+# Web server
```
kixelated: We might have to feature flag this stuff if it doesn't compile on WASM.

kixelated (approved):

moq-karp/Cargo.toml
LiamF-2261667: I'm having issues building moq-web, up untill now I've just been using the release version (which does work for some reason). So I can't test if this would be a problem.

kixelated: If the tests pass then no problem.

PR #360, opened 2025-03-26
fixed/bug in fetch method

sreyan-ghosh:
Fixed the bug occuring in the `fetch` method of the relay's web server.


kixelated:
Fixed in a different way in #361 and #362. There's no point having a leading slash... I'm not sure what I was thinking.

PR #454, opened 2025-06-30
feat: adding back our nix modules

gordysunsaronic:
adding back out nix modules


flake.nix
```diff
@@ -18,12 +18,19 @@
...
+    {
+      nixosModules = rs.nixosModules;
+      overlays = rs.overlays;
+    }
+    // flake-utils.lib.eachDefaultSystem (system: {
       devShells.default = nixpkgs.legacyPackages.${system}.mkShell {
         inputsFrom = [
           rs.devShells.${system}.default
-          js.devShells.${system}.default
+          js.devShells.${system}.default
```
therishidesai: nit: whitespace

flake.nix
```diff
@@ -18,12 +18,19 @@
...
+    }
+    // flake-utils.lib.eachDefaultSystem (system: {
       devShells.default = nixpkgs.legacyPackages.${system}.mkShell {
         inputsFrom = [
           rs.devShells.${system}.default
-          js.devShells.${system}.default
+          js.devShells.${system}.default 
         ];
+        shellHook = ''
+          echo "🎯 Using Gordy's Top-Level MoQ Flake (Rust environment only)"
```
therishidesai: nit: remove your name

rs/flake.nix
therishidesai, line 1: you should combine this flake.nix with the top level flake.nix so there is one unified nix flake in the moq repo

rs/nix/modules/moq-relay.nix
```diff
@@ -0,0 +1,72 @@
...
+      before = [ "moq-relay.service" ];
+    };
+
+    systemd.services.moq-relay = {
+      description = "Media over QUIC relay server (Gordy's Module)";
+      wantedBy = [ "multi-user.target" ];
+
+      serviceConfig = {
+        ExecStart = "${pkgs.moq-relay}/bin/moq-relay --bind [::]:${builtins.toString cfg.port}" + 
+          (if cfg.dev.enable then " --tls-generate ${cfg.dev.tls_url}" else "");
```
therishidesai: I don't think these args are up to date with the new moq-relay

rs/nix/packages/moq.nix
```diff
@@ -0,0 +1,34 @@
+{ fenix, naersk, flake-utils, ... }:
```
therishidesai: why naersk? We used to use crate2nix in the original moq. I think we should stick to crate2nix as it allows for better caching

rs/nix/shell.nix
```diff
@@ -0,0 +1,40 @@
+{ self, nixpkgs, flake-utils, fenix, ... }:
```
therishidesai: move the dev shell from the top level nix into this

therishidesai (changes requested):
lets remove all of the lines that have "Gordy" in it

rs/nix/packages/moq.nix
kixelated: I switched over to naersk primarily because it doesn't require a separate lockfile (don't want to mandate Nix). I don't know the differences otherwise.

rs/flake.nix
gordysunsaronic: just the rs/flake.nix or also the js/flake.nix?

gordysunsaronic: also - problem here is if i combine this with the top level flake.nix, it gives me problems with the js stuff which isn't in the moq-relay server but is related to the webplayer and front end

rs/nix/packages/moq.nix
therishidesai: I think that can get updated now, but naersk also works fine. This is a smaller project so shouldn't be an issue with the caching

rs/nix/shell.nix
gordysunsaronic: the dev shells have both rs and js, so i gotta keep it at the top level

rs/nix/modules/moq-relay.nix
gordysunsaronic: updated them

therishidesai (approved):
This looks good to me on the nix side of things. In future PRs let's add a way of validating the config file at eval time I case the underlying rust code gets changed. Let's also add a [NixOS VM test](https://nix.dev/tutorials/nixos/integration-testing-using-virtual-machines.html) for the relay in a future PR.

@kixelated thoughts on this PR?

js/hang-demo/src/index.ts
```diff
@@ -8,6 +8,11 @@ export { HangWatch, HangSupport };
 const watch = document.querySelector("hang-watch") as HangWatch;
 
 // If query params are provided, use it as the broadcast name.
+
+
```
kixelated: ?

js/hang-demo/src/index.ts
```diff
@@ -8,6 +8,11 @@ export { HangWatch, HangSupport };
...
+// ?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJwYXRoIjoidGVzdC1tYWMvIiwicHViIjpudWxsLCJzdWIiOiIvdGVzdC1tYWMvKiIsInN1Yj8iOnRydWUsImV4cCI6bnVsbCwiaWF0IjpudWxsfQ.Kssxm6hltd9tAJNxT-4fW8bICs9ancrcR_IRaCkvl1w`
 const urlParams = new URLSearchParams(window.location.search);
-const name = urlParams.get("name") ?? "demo/bbb";
-watch.setAttribute("url", `http://localhost:4443/${name}.hang`);
+const name = urlParams.get("name") ?? "jwt-test/bbb";
```
kixelated: ?

rs/flake.nix
```diff
@@ -83,7 +99,12 @@
             pkgs.cargo-shear
             pkgs.cargo-audit
           ];
+          
+          # Environment variables from moq-rs
+          shellHook = ''
+            export LIBCLANG_PATH="${pkgs.libclang.lib}/lib"
```
kixelated, line 117: What does this do?

rs/justfile
```diff
@@ -37,13 +37,41 @@ root:
 	if [ ! -f "dev/root.jwk" ]; then \
 		cargo run --bin moq-token -- --key "dev/root.jwk" generate; \
 	fi
+	
+	# Generate publisher-only token
+	if [ ! -f "dev/bum-publisher.jwt" ]; then \
```
kixelated: Remove these?

rs/nix/modules/moq-relay.nix
```diff
@@ -0,0 +1,238 @@
...
+  ...
+}:
+let
+  cfg = config.services.moq-relay;
+  
+  # Use TOML format for configuration
+  tomlFormat = pkgs.formats.toml {};
+  
+  # Build the configuration
+  configData = {
```
kixelated: pretty cool tbh

rs/nix/overlay.nix
```diff
@@ -0,0 +1,41 @@
...
+
+  hang = naersk'.buildPackage {
+    pname = "hang";
+    src = ../.;
+  };
+
+  moq-token = naersk'.buildPackage {
+    pname = "moq-token-cli";
+    src = ../.;
+    cargoBuildOptions = opts: opts ++ [ "-p" "moq-token-cli" ];
```
kixelated: We should also have these options for the other binaries.

rs/nix/shell.nix
```diff
@@ -0,0 +1,40 @@
...
+              combine [
+                latest.rustc
+                latest.cargo
+                targets.wasm32-unknown-unknown.latest.rust-std
+              ];
+          in
+            with pkgs;
+            mkShell {
+              nativeBuildInputs = [
+                go
```
kixelated: go?

kixelated (changes requested):

rs/flake.nix
gordysunsaronic: deals with some C dependencies in Rust code

PR #464, opened 2025-07-11
Fix Firefox audio not playing

masonblier:
This check for AudioContext state does not work in Firefox as unlike Chromium, Firefox does not automatically set the initial state of the AudioContext to `running`. Removing this check seems to cause no problems as the audio state is set to disabled by default.

More details about the difference of the initial AudioContext state between browsers is discussed here: https://github.com/WebAudio/web-audio-api/issues/2218

This fixes issue: https://github.com/kixelated/moq/issues/463

@kixelated


kixelated:
I don't know the ramifications of removing this check, but I'll test it and figure out an alternative if needed.

PR #477, opened 2025-07-19
Use JWT tokens for local development.

kixelated:
More complicated but much better for testing.

`just dev` and `just cluster` will use JWT tokens. `just web`, `just relay,` and `just pub` will be anonymous just because it makes things more complicated.  Easier to catch issues in local development this way.


kixelated:
@claude can you fix the "Bug: Glob Pattern Quoting Causes `rm` Command Failure" bug if you agree (removing quotes?), push the fix, then merge this PR?

PR #483, opened 2025-07-21
Add an ANNOUNCE_INIT message.

kixelated:
This message is the first reply received from the remote, acting kind of like an ANNOUNCE_OK. It lets the caller know all of the broadcasts currently available.

This is used to ensure that we have received all remote broadcasts before the connection is considered established. This is purely a library thing, the protocol itself does not require a client to know all broadcasts. I just think it's nice when a client knows everything available at all times.

This will require a new protocol version unfortunately.

### DIFF
--- a/rs/hang-gst/src/source/imp.rs
+++ b/rs/hang-gst/src/source/imp.rs
@@ -138,9 +138,7 @@ impl ElementImpl for HangSrc {
 				}
 				// Chain up first to let the bin handle the state change
 				let result = self.parent_change_state(transition);
-				if result.is_err() {
-					return result;
-				}
+				result?;
 				// This is a live source - no preroll needed
 				return Ok(gst::StateChangeSuccess::NoPreroll);
 			}


kixelated:
@claude could you verify this implements the draft changes: https://github.com/kixelated/moq-drafts/pull/10

While you're at it, could you rename AnnounceRequest to AnnouncePlease so it matches?

Finally, can you validate that both the rs and js halves are using the new protocol version number?

kixelated:
@claude This clippy warning:

error: this block may be rewritten with the `?` operator
   --> hang-gst/src/source/imp.rs:141:5
    |
141 | /                 if result.is_err() {
142 | |                     return result;
143 | |                 }
    | |_________________^ help: replace it with: `result?;`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#question_mark
    = note: `-D clippy::question-mark` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::question_mark)]`
    
    You'll need to use just check --workspace to see gstreamer errors.

PR #489, opened 2025-07-22
Use a size prefix for messages.

kixelated:
Matches moq-transport and is better at catching encoding bugs.

@claude Make sure this matches https://github.com/kixelated/moq-drafts/pull/11


kixelated:
@claude can you make the uninit change and push it?

Can you verify that Vec<u8> is not being used for decoding or encoding? Maybe just remove the generic Vec<T> to avoid accidents.

kixelated:
@claude Why does the Claude github action not enable bash by default if it's running in a sandbox already?

PR #494, opened 2025-07-23
Add an initial "preview" track for anonymous / zero-bitrate streaming

kixelated:
Useful to figure out who is in a conference room without actually joining. It should be a separate broadcast so different authentication can be applied, but tht will have to wait.


kixelated:
@claude Can you fix the great catches by cursor?

If effect ever returns a value, it's a bug. If it doesn't set anything, then it's a bug too.

kixelated:
@claude In the latest commit, can you replace lines that perform `if ... { signal.set(undefined); return }` with a cleanup instead. 

So any time we `x.set` a signal, we should follow it with a `effect.cleanup(() => x.set(undefined))`.

PR #500, opened 2025-07-25
Add a compatibility layer for moq-transport-07

kixelated:
Only implements moq-lite functionality. Allows connecting to a moq-rs-ietf relay if that's your jam.


kixelated:
@claude MAX_DATA should be enforced within #fill, adding an upper bound for all reads.

PR #510, opened 2025-07-31
Fix subscription termination bug

kixelated:
TrackProducers for requested tracks were being cached in the lookup HashMap, preventing them from being marked as unused when all consumers were dropped. This caused subscriptions to remain open indefinitely.

The fix introduces separate caches for published tracks (TrackConsumer) and requested tracks (TrackProducer). The requested track cache automatically cleans up when a TrackProducer becomes unused, allowing proper resource cleanup and subscription termination.

🤖 Generated with [Claude Code](https://claude.ai/code)


kixelated:
@claude can you update the code to just call remove after unused? It can be spawned inline now, there's no need for the cleanup function if it's just 2 lines long. And push the new commit

PR #524, opened 2025-08-05
Fix error announcing more than 127 broadcasts

CristobalAlvarez:
### Problem

I've been experiencing issues when trying to announce more than 127 broadcasts. Specifically:
- Beyond this number, some broadcasts do not appear in "/announced".
- It becomes impossible to establish new connections, although previously connected subscribers continue to receive data correctly.

The issue is due to the use of `now_or_never`, which causes continuous polling that eventually saturates Tokio.

### Solution

The `while` loop has been replaced with a `loop` using a timeout. This modification:
- Introduces a brief pause that minimizes saturation of Tokio using timeout, instead of using now_or_never

### DIFF
--- a/rs/moq-relay/src/web.rs
+++ b/rs/moq-relay/src/web.rs
@@ -81,9 +81,15 @@ async fn serve_announced(Path(prefix): Path<String>, cluster: Cluster) -> impl I
 	let mut origin = cluster.combined.consume_prefix(&prefix);
 	let mut broadcasts = Vec::new();
 
-	while let Some(Some(OriginUpdate { suffix, active })) = origin.next().now_or_never() {
-		if active.is_some() {
-			broadcasts.push(suffix);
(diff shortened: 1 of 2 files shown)


kixelated:
> It's not blocking — it seems that using `.now_or_never()` in a loop with `UnboundedReceiver::recv()` can cause it to stop yielding messages after around 127 items, even though more are still queued.
> 
> Related issue: [tokio-rs/tokio#7108](https://github.com/tokio-rs/tokio/issues/7108)

Oof, that's an unfortunate Tokio behavior. I understand the intent, but I feel like it can only introduce bugs like this...

The 10ms sleep would work but it's a little racy. I'll try reverting to a previous iteration where it used a Vec+notify instead of an unbounded channel.

kixelated:
Thanks for the report and fix!

PR #558, opened 2025-08-26
Support Webpack and fixes

skirsten:
On Next.js there is still some problems with the transformers library being loaded as node module somehow...

```
 ⚠ ./node_modules/.pnpm/@huggingface+transformers@3.7.2/node_modules/@huggingface/transformers/dist/transformers.node.mjs
Critical dependency: Accessing import.meta directly is unsupported (only property access or destructuring is supported)

Import trace for requested module:
./node_modules/.pnpm/@huggingface+transformers@3.7.2/node_modules/@huggingface/transformers/dist/transformers.node.mjs
./node_modules/.pnpm/@kixelated+hang@file+..+..+..+forks+moq+js+hang+dist+kixelated-hang-0.3.13.tgz_@types+react@19.1.11_react@19.1.1/node_modules/@kixelated/hang/publish/audio/captions-worker.js
```
but its not blocking anything.

Also turbopack just gives up without any error message and return code 0 :smile: 

## Summary by CodeRabbit

- New Features
  - Added speaking capability to Audio for integrated speaking state.
  - Improved multi-channel audio handling with explicit channel count support.

- Bug Fixes
  - Improved compatibility for audio and video worklet loading across build environments, reducing initialization failures.
  - Corrected support panel label from “Audio” to “Opus” for decoding details.

- Chores
  - Added Prettier and an async utility dependency to development/package tooling.

### DIFF
--- a/js/hang/package.json
+++ b/js/hang/package.json
@@ -38,11 +38,12 @@
 		"release": "tsx ../scripts/release.ts"
 	},
 	"dependencies": {
+		"@huggingface/transformers": "^3.7.0",
 		"@kixelated/moq": "workspace:^",
 		"@kixelated/signals": "workspace:^",
-		"zod": "^4.0.0",
+		"async-mutex": "^0.5.0",
 		"comlink": "^4.4.2",
-		"@huggingface/transformers": "^3.7.0"
+		"zod": "^4.0.0"
 	},
 	"devDependencies": {
 		"@types/audioworklet": "^0.0.77",
--- a/js/hang/src/publish/audio/captions.ts
+++ b/js/hang/src/publish/audio/captions.ts
@@ -1,8 +1,9 @@
 import * as Moq from "@kixelated/moq";
 import { Effect, Signal } from "@kixelated/signals";
+import type { Audio } from ".";
 import type * as Catalog from "../../catalog";
 import { u8 } from "../../catalog";
-import type { Audio } from ".";
+import { loadAudioWorklet } from "../../util/hacks";
 import type { Request, Result } from "./captions-worker";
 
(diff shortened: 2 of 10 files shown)


kixelated:
And yeah, the code rabbit review points out this could be racey. Maybe add a `loadWorklet` helper method that uses a mutex to make sure two (async) functions don't monkey patch `navigator.serviceWorker.register` at the same time and restores it even on exception. It won't actually matter at the moment, because we don't register serviceWorkers, but still.

kixelated:
also `just fix` to format stuff.

skirsten:
> also just fix to format stuff.

But my vscode formats on save and by default uses a different prettier version which causes a bunch of changes. If the prettier version is fixed in the project, it will always use that version -> no changes.

kixelated:
> > also just fix to format stuff.
> 
> But my vscode formats on save and by default uses a different prettier version which causes a bunch of changes. If the prettier version is fixed in the project, it will always use that version -> no changes.

Yeah I'm using biome, not prettier. It's faster and IMO does a better job at formatting.

kixelated:
yolo

I'll let you know if it breaks stuff for me.

skirsten:
What is the issue you are seeing? I just quickly ran
```
web-build url='http://localhost:4443/anon':
	pnpm -r i
	cd hang-demo && VITE_RELAY_URL="{{url}}" pnpm vite build src
```

and it seems fine (apart from only building the index.html file), but the assets for the workers are there.

skirsten:
Hmm, nevermind that does not look right...

```
new URL(
	"data:video/mp2t;base64,...",
	import.meta.url,
),
```

PR #578, opened 2025-09-05
feat(moq-native): support raw QUIC sessions with `moql://` URLs

Frando:
This adds support for raw quic connections, identified by the `moql://` URL scheme, to `moq-native`.

With this change I can subscribe to a raw QUIC stream with moq-native and hang by using a URL `moql://<remote-ip>:<remote-port>`. For example, when serving a stream with `just serve bbb`, it can be subscribed to with the URL `moql://localhost:4443`.

Authentication (as used by the relay) is not implemented (no token can be submitted when connecting over raw QUIC).

## Summary by CodeRabbit

* **New Features**
  * Server accepts both WebTransport and raw QUIC sessions automatically based on negotiated protocol (ALPN).
  * Client selects transport path from ALPN for better compatibility.
  * Relay supports sessions from both WebTransport and QUIC with unified path/token handling.
  * Server can report its local address and be cleanly shut down; requests can be accepted, rejected, or inspected uniformly.

### DIFF
--- a/rs/moq-native/src/server.rs
+++ b/rs/moq-native/src/server.rs
@@ -237,7 +237,10 @@ impl Request {
 	pub async fn close(self, status: http::StatusCode) -> Result<(), quinn::WriteError> {
 		match self {
 			Self::WebTransport(request) => request.close(status).await,
-			Self::Quic(request) => Ok(request.close(status)),
+			Self::Quic(request) => {
+				request.close(status);
+				Ok(())
+			}
 		}
 	}
 


rs/moq-relay/src/connection.rs
```diff
@@ -1,26 +1,30 @@
...
-		// Extract the path and token from the URL.
-		let path = self.request.url().path();
-		let token = self
-			.request
-			.url()
-			.query_pairs()
-			.find(|(k, _)| k == "jwt")
-			.map(|(_, v)| v.to_string());
-
+		let (path, token) = match &self.request {
```
kixelated, line 15: Do we need this match? The extraction should work the same.

Frando: Yeah I can remove it. For the raw QUIC connection the URL is a dummy with only host and scheme filled so both path and query string would be empty anyway.

kixelated: Going to test this. yolo

@coderabbitai Can you make a new PR that reverts this change? I'll merge this PR then merge the revert right afterwards.

kixelated: Lame. @Frando feel free to make a new PR if you want to revert the URL parsing or I can randomly do it one day. No need to block on this.

fluffy:
Great to see support for this - thanks

PR #601, opened 2025-09-23
Add a video jitter buffer.

kixelated:
And fix some existing bugs with the existing audio jitter buffer.

## Summary by CodeRabbit

* **New Features**
  * Latency is now reactive and accepts either a numeric value or a signal; HangWatch exposes a latency attribute/property (default 100 ms) that updates in real time to coordinate decode and render.
  * Video and audio playback use latency-aware consumers for buffering and timing.

* **Bug Fixes**
  * Improved frame ordering/skipping and more reliable resource cleanup.
  * Audio buffer now writes the latest samples window to maintain consistent output.

### DIFF
--- a/js/hang/src/frame.ts
+++ b/js/hang/src/frame.ts
@@ -1,5 +1,5 @@
 import type * as Moq from "@kixelated/moq";
-import { Effect } from "@kixelated/signals";
+import { Effect, Signal } from "@kixelated/signals";
 import * as Time from "./time";
 
 export interface Source {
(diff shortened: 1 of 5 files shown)


skirsten:
It's simply the video renderer that continues to clear the canvas, but because of my display its running at 240fps which causes the load.
image

Maybe this PR was not the cause of this regression. Anyway I am looking into a custom renderer anyway (that only requests an animation frame when there is a new frame).

PR #611, opened 2025-10-11
feat(hang): add support for annexb import

therishidesai:
## Summary by CodeRabbit

- **New Features**
  - Annex‑B H.264 import: reads Annex‑B streams, auto-publishes a video track on SPS detection, derives resolution/profile/level, and emits timed frames with keyframe flags.
  - CLI publish format option: choose Annex‑B or CMAF input when publishing.

- **Improvements**
  - Clearer, unified error reporting across Annex‑B and CMAF import flows.

- **Chores**
  - Added H.264 parsing dependency.

### DIFF
--- a/rs/hang-cli/src/client.rs
+++ b/rs/hang-cli/src/client.rs
@@ -1,5 +1,7 @@
+use crate::ImportType;
 use anyhow::Context;
-use hang::cmaf::Import;
+use hang::annexb;
+use hang::cmaf;
 use hang::moq_lite;
 use tokio::io::AsyncRead;
 use url::Url;
@@ -8,6 +10,7 @@ pub async fn client<T: AsyncRead + Unpin>(
 	config: moq_native::ClientConfig,
 	url: Url,
 	name: String,
+	format: ImportType,
 	input: &mut T,
 ) -> anyhow::Result<()> {
 	let broadcast = moq_lite::Broadcast::produce();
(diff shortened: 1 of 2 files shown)


rs/hang/src/annexb/error.rs
```diff
@@ -0,0 +1,22 @@
...
+	#[error("moq error: {0}")]
+	Moq(#[from] moq_lite::Error),
+
+	/// An error from the MP4 atom parsing library.
+	#[error("annexb parser error: {0}")]
+	AnnexB(#[from] h264_parser::Error),
+
+	/// An error from the hang media library.
+	#[error("karp error: {0}")]
+	Karp(#[from] crate::Error),
```
kixelated: lul it's called `Hang` now, and we should rename the other two references I missed.

rs/hang/src/annexb/import.rs
```diff
@@ -0,0 +1,101 @@
...
+		let mut buffer = BytesMut::new();
+
+		let mut tp: Option<TrackProducer> = None;
+
+		let now = std::time::Instant::now();
+
+		while input.read_buf(&mut buffer).await? > 0 {
+			parser.push(&buffer);
+			buffer.clear();
+			while let Ok(Some(au)) = parser.next_access_unit() {
```
kixelated:
```suggestion
			while let Some(au) = parser.next_access_unit()? {
```

rs/hang/src/annexb/import.rs
```diff
@@ -0,0 +1,101 @@
...
+						};
+						let track = video.track.clone().produce();
+						self.broadcast.insert_track(track.consumer);
+						self.catalog.add_video(video);
+						tp = Some(track.producer.into());
+						self.catalog.publish();
+					}
+				}
+
+				if let Some(ref mut track) = tp {
```
kixelated: This will always be true right?

IMO something like:

```rust
let track = match tp {
  Some(tp) => tp,
  None => {
    // create it
  }
};

// use it
track.write(frame);

// save it for next loop
tp = Some(track);

PR #647, opened 2025-10-29
rs: hang: Handle multiple TRUN boxes within one TRAF

jhiatt-verkada:
## Summary by CodeRabbit

* **Chores**
  * Updated media processing library dependency to version 0.9.0.

* **Bug Fixes**
  * Improved media file import handling for more accurate sample processing and data offset calculations.
  * Enhanced keyframe detection in video tracks with per-sample parameter resolution.
  * Added latency monitoring during media import with warnings for excessive processing delays.

### DIFF
--- a/rs/hang/src/cmaf/import.rs
+++ b/rs/hang/src/cmaf/import.rs
@@ -472,89 +472,94 @@ impl Import {
 			let default_sample_size = trex.map(|trex| trex.default_sample_size).unwrap_or_default();
 			let default_sample_flags = trex.map(|trex| trex.default_sample_flags).unwrap_or_default();
 
-			let tfhd = &traf.tfhd;
-			let trun = traf.trun.as_ref().ok_or(Error::MissingBox(Trun::KIND))?;
-
 			let tfdt = traf.tfdt.as_ref().ok_or(Error::MissingBox(Tfdt::KIND))?;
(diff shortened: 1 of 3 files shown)


jhiatt-verkada:
FYI most of the diff goes away if you check the "hide whitespace" box. For the most part it just moves code into a for-loop body.

rs/hang/src/cmaf/import.rs
```diff
@@ -472,89 +472,94 @@ impl Import {
...
+				if let Some(data_offset) = trun.data_offset {
+					let base_offset = tfhd.base_data_offset.unwrap_or_default() as usize;
+					// This is relative to the start of the MOOF, not the MDAT.
+					// Note: The trun data offset can be negative, but... that's not supported here.
+					let data_offset: usize = data_offset.try_into().map_err(|_| Error::InvalidOffset)?;
+					if data_offset < self.moof_size {
+						return Err(Error::InvalidOffset);
+					}
+					// Reset offset if the TRUN has a data offset
+					offset = base_offset + data_offset - self.moof_size - header_size;
```
jhiatt-verkada, line 496: TBH I wasn't really sure what the right behavior was here. I don't have access to the ISO spec so I don't know what combinations are considered legal (multiple TRUNs each with `data_offset` specified, some with/without, or all without).

PR #673, opened 2025-11-08
fix: wait for the first keyframe before sending frames

therishidesai:
Fixes occasional decode errors on the player if we send non-keyframes first

## Summary by CodeRabbit

* **Bug Fixes**
  * Frame processing now defers writing until the first keyframe is observed, preventing premature frame output and improving stream reliability.
* **Performance**
  * Reduced unnecessary data copying during payload handling, lowering memory use and improving processing efficiency.

### DIFF
--- a/rs/hang/src/annexb/import.rs
+++ b/rs/hang/src/annexb/import.rs
@@ -43,24 +43,24 @@ impl Import {
 			while let Some(au) = parser.next_access_unit()? {
 				match tp {
 					Some(ref mut track) => {
+						got_first_keyframe = got_first_keyframe || au.is_keyframe();
+						if !got_first_keyframe {
+							continue;
+						}
+
 						let ts = now.elapsed().as_micros();
 						let payload = match au.to_annexb_webcodec_bytes() {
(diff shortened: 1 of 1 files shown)


rs/hang/src/annexb/import.rs
```diff
@@ -35,17 +35,25 @@ impl Import {
...
+						got_first_keyframe = got_first_keyframe || au.is_keyframe();
+						if !got_first_keyframe {
+							continue;
+						}
+
 						let ts = now.elapsed().as_micros();
 						let payload = match au.to_annexb_webcodec_bytes() {
 							Cow::Borrowed(b) => Bytes::copy_from_slice(b),
-							Cow::Owned(b) => Bytes::copy_from_slice(&b),
+							Cow::Owned(b) => Bytes::from(b), // avoids a copy
```
kixelated, line 54: @therishidesai I couldn't help myself.

therishidesai: Oh lol, thanks

PR #715, opened 2025-11-23
feat: add file upload support

christriants:
This PR implements file upload as a media source for broadcasting, allowing users to upload video, audio, and image files from their computer and stream them through the MoQ relay.

Addresses: https://github.com/kixelated/moq/issues/702

### file.ts
This is a new file source class that decodes uploaded files:
- Video/audio: Uses `HTMLVideoElement.captureStream()` to extract `StreamTracks`
- Images: Renders to canvas at 30fps, then captures as video stream

### element.ts
We modify `js/hang/src/publish/element.ts` to support the new file source:
- Added `file` signal and `"file"` source type
- Implemented `#runSource` handler and `#renderFile` UI method
- Auto-enabled video/audio encoding when file is selected

Testing:
- User clicks the 📁 (file) button in the publish controls
- Browser file picker opens
- User selects a video, audio, or image file
- File is decoded and streamed to the MoQ relay
- Viewers can watch the uploaded content on `index.html?path=<broadcast-path>`

Future enhancements (?)
- Use `mediabunny` for formats browsers don't natively support

### DIFF
--- a/js/hang/src/publish/element.ts
+++ b/js/hang/src/publish/element.ts
@@ -275,12 +275,19 @@ export class HangPublishInstance {
 
 		if (source === "file") {
 			const fileSource = new Source.File({
-				enabled: new Signal(true),
+				enabled: new Signal(false),
 			});
 
 			effect.effect((effect) => {
 				const file = effect.get(this.parent.signals.file);
 				fileSource.setFile(file);
+				const audio = effect.get(this.parent.signals.audio);
+				const video = effect.get(this.parent.signals.video);
+				effect.set(
+					fileSource.enabled,
+					(audio || video) && Boolean(file),
+					false
+				);
 			});
 
 			fileSource.signals.effect((effect) => {


js/hang/src/publish/element.ts
```diff
@@ -502,6 +541,44 @@ export class HangPublishInstance {
...
+
+		effect.event(fileInput, "change", (e) => {
+			const file = (e.target as HTMLInputElement).files?.[0];
+			if (file) {
+				this.parent.file = file;
+				this.parent.source = "file";
+				this.parent.video = true;
+				this.parent.audio = true;
+			}
+		});
```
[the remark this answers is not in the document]
christriants: Good catch / fix.

js/hang/src/publish/element.ts
```diff
@@ -261,6 +272,33 @@ export class HangPublishInstance {
 
 			return;
 		}
+
+		if (source === "file") {
+			const fileSource = new Source.File({
+				enabled: new Signal(true),
```
kixelated: Ideally this would be toggled based on if audio/video broadcasting is enabled. See just above:

```ts
effect.set(screen.enabled, audio || video, false);
```

(we should probably make a single shared signal to avoid copy/pasting this)

js/hang/src/publish/element.ts
```diff
@@ -261,6 +272,33 @@ export class HangPublishInstance {
...
+			});
+
+			effect.effect((effect) => {
+				const file = effect.get(this.parent.signals.file);
+				if (file) {
+					fileSource.setFile(file);
```
kixelated: You probably want to unset this right if `!file`?

js/hang/src/publish/source/file.ts
```diff
@@ -0,0 +1,128 @@
...
+
+	constructor(config: FileSourceConfig) {
+		this.enabled = config.enabled;
+
+		this.signals.effect((effect) => {
+			const file = effect.get(this.#file);
+			const enabled = effect.get(this.enabled);
+
+			if (!file || !enabled) {
+				effect.set(this.source, {}, {});
```
kixelated:
```suggestion
				this.source.set({});
```

`effect.set` sets the value to the first arg and on cleanup, sets the value to the second arg. We don't need the latter.

kixelated: In fact you don't need this line because of:

```typescript
effect.set(this.source, { video: videoTrack as VideoStreamTrack }, {});
```

The second arg means we set the source to `{}` when any signal changes (cleanup).

christriants: Thanks, appreciate the callout.

js/hang/src/publish/element.ts
christriants: I think we could just call `setFile` regardless of whether or not `file` is set!

kixelated:
Thanks for much for the PR. I haven't had a chance to try it yet, but it's amazing how simple you made the implementation.

Issue #677, opened 2025-11-14
Non-vite support


kixelated:
The current codebase uses Vite specific imports for audio worklets. There does not seem to be a generic way to do it across bundlers (like WebWorkers).

We currently "compile" `@kixelated/hang` using tsc only. To support other bundlers and importing via a script tag, we should properly build the library via rollup or whatever.

I believe the main problem is how to specify the path to Worker and Worklet files. It will need to be configurable, and there will need to be instructions on how to hoist these files out of `node_modules` for those using bundlers?

Or a Javascript wizard can prove me wrong and just get it to work. Fucking legend.

We should have a repo of hello world working with various bundlers just to verify. Something like [this](https://github.com/kixelated/moq-web).

bhbs:
## **Proposal: Inline the AudioWorklet via a Blob URL**

### Summary

We can avoid all bundler-specific path resolution issues by inlining the AudioWorklet directly into the library bundle and loading it via a `blob:` URL at runtime.

### Approach

* **Bundle the AudioWorklet into a single JS file** at build time (e.g. with Rollup).
* **Embed that bundled file as a string** inside the main library bundle.
* **Load it via a Blob URL** instead of a file path.

Example:

```ts
import workletSource from '../dist/render-worklet.js?raw';

const blob = new Blob([workletSource], { type: 'text/javascript' });
const url = URL.createObjectURL(blob);

await audioContext.audioWorklet.addModule(url);
```

### Benefits

* No need for users to configure URLs or special loader plugins.
* No need to hoist the Worklet file out of `node_modules`.
* Works across all bundlers (Vite, Webpack, esbuild, Rollup, etc.) because Worklet code is just a string inside the main JS bundle.

### Prior Art

This is similar in spirit to what **`@ffmpeg/ffmpeg`** does with Web Workers and WASM
([see their usage docs](https://ffmpegwasm.netlify.app/docs/getting-started/usage/)).

They load the worker script and WASM binary from a CDN, convert them into Blobs, and then spawn a Worker from those Blob URLs to avoid cross-origin restrictions. 

### Caveat

One important limitation: you generally cannot load the library itself from a CDN if it spawns Workers, because the script that creates a Worker must be served from the same origin.

This is also why `@ffmpeg/ffmpeg` cannot be used directly via a `<script>` tag from a CDN ([ref](https://ffmpegwasm.netlify.app/docs/getting-started/installation)).

kixelated:
[...]

PR #717, opened 2025-11-24
Hang UI as a separate package - Issue 695

jdreetz:
# What does it change?

This PR addresses https://github.com/kixelated/moq/issues/695. It migrates the Watch and Publish UIs to a separate `hang-ui` package, that provides two Web Components (`hang-publish-ui` and `hang-watch-ui`). Internally, the Web Components use SolidJS to render and interact with the elements. 

## Web Components
The Web Components expect to wrap their respective `hang` Web Components (`hang-publish` for `hang-publish-ui` and `hang-watch` for `hang-watch-ui`). In order to use the existing Signal reactivity, the UI components will wait until they have a reference to the respective `Hang...Instance` (published as a `CustomEvent`: `watch-instance-available`, `publish-instance-available`), that provides information about the state of the watch/publish. To interact with the Signals on the Instances, we had to make some of their properties public: `video`, `audio`, `signals`, `file`.

## Context Provider
Once the Instance properties are ready, we use a SolidJS Context to expose the Instance properties to the UI components:
- `WatchUIContextProvider.tsx`
- `PublishUIContextProvider.tsx`

I'm open to using the Instance Signals directly in the components, but we still need something that waits for the Instance to become in the wrapped component (for some reason it's not available immediately) before the UI components can use them. Personally, I think it's helpful to keep the binding layers (Context) separate from the UI component implementation, as it reduces the noise in the components and allows the developer to focus on the markup and interactions, but I'm open to other design ideas.  

# How to test

- Run the dev command (`nix develop -c just dev`) to start your local dev
- Open the "publish" demo page
- Verify the controls are displayed under the preview element
- Verify the controls look and function the same as on `main`
- Open the index demo page
- Verify the controls are displayed under the preview element
- Verify the controls look and function the same as on `main`

### DIFF
--- a/js/hang-ui/src/Components/publish/PublishUIContextProvider.tsx
+++ b/js/hang-ui/src/Components/publish/PublishUIContextProvider.tsx
@@ -1,5 +1,6 @@
 import type HangPublish from "@kixelated/hang/publish/element";
 import type { HangPublishInstance, InstanceAvailableEvent } from "@kixelated/hang/publish/element";
+import { Effect } from "@kixelated/signals";
 import type { JSX } from "solid-js";
 import { createContext, createEffect, createSignal, onCleanup } from "solid-js";
 
@@ -93,7 +94,13 @@ export default function PublishUIContextProvider(props: PublishUIContextProvider
 	return <PublishUIContext.Provider value={value}>{props.children}</PublishUIContext.Provider>;
 
 	function onPublishInstanceAvailable(el: HangPublish, publishInstance: HangPublishInstance) {
-		publishInstance.signals.effect(function trackCameraDevices(effect) {
+		const localEffect = new Effect();
+
+		onCleanup(() => {
+			localEffect.close();
+		});
+
+		localEffect.effect(function trackCameraDevices(effect) {
 			const clearCameraDevices = () => setCameraMediaDevices([]);
 			const video = effect.get(publishInstance.video);
 
@@ -111,7 +118,7 @@ export default function PublishUIContextProvider(props: PublishUIContextProvider
 			setCameraMediaDevices(devices);
 		});
 
-		publishInstance.signals.effect(function trackMicrophoneDevices(effect) {
+		localEffect.effect(function trackMicrophoneDevices(effect) {
 			const clearMicrophoneDevices = () => setMicrophoneMediaDevices([]);
 			const audio = effect.get(publishInstance.audio);
 
@@ -135,17 +142,17 @@ export default function PublishUIContextProvider(props: PublishUIContextProvider
 			setMicrophoneMediaDevices(devices);
 		});
 
-		publishInstance.signals.effect(function trackNothingSourceActive(effect) {
+		localEffect.effect(function trackNothingSourceActive(effect) {
 			const selectedSource = effect.get(el.signals.source);
 			setNothingActive(selectedSource === undefined);
 		});
 
-		publishInstance.signals.effect(function trackMicrophoneSourceActive(effect) {
+		localEffect.effect(function trackMicrophoneSourceActive(effect) {
 			const audioActive = effect.get(el.signals.audio);
 			setMicrophoneActive(audioActive);
 		});
 
-		publishInstance.signals.effect(function trackVideoSourcesActive(effect) {
+		localEffect.effect(function trackVideoSourcesActive(effect) {
 			const videoSource = effect.get(el.signals.source);
 			const videoActive = effect.get(el.signals.video);
 
@@ -161,7 +168,7 @@ export default function PublishUIContextProvider(props: PublishUIContextProvider
 			}
 		});
 
-		publishInstance.signals.effect(function trackSelectedCameraSource(effect) {
+		localEffect.effect(function trackSelectedCameraSource(effect) {
 			const video = effect.get(publishInstance.video);
 
 			if (!video || !("device" in video)) return;
@@ -170,7 +177,7 @@ export default function PublishUIContextProvider(props: PublishUIContextProvider
 			setSelectedCameraSource(requested);
 		});
 
-		publishInstance.signals.effect(function trackSelectedMicrophoneSource(effect) {
+		localEffect.effect(function trackSelectedMicrophoneSource(effect) {
 			const audio = effect.get(publishInstance.audio);
 
 			if (!audio || !("device" in audio)) return;
@@ -179,7 +186,7 @@ export default function PublishUIContextProvider(props: PublishUIContextProvider
 			setSelectedMicrophoneSource(requested);
 		});
 
-		publishInstance.signals.effect(function trackPublishStatus(effect) {
+		localEffect.effect(function trackPublishStatus(effect) {
 			const url = effect.get(publishInstance.connection.url);
 			const status = effect.get(publishInstance.connection.status);
 			const audio = effect.get(publishInstance.broadcast.audio.source);
@@ -202,7 +209,7 @@ export default function PublishUIContextProvider(props: PublishUIContextProvider
 			}
 		});
 
-		publishInstance.signals.effect(function trackFileActive(effect) {
+		localEffect.effect(function trackFileActive(effect) {
 			const selectedSource = effect.get(el.signals.source);
 			setFileActive(selectedSource === "file");
 		});
--- a/js/hang-ui/src/Components/watch/WatchUIContextProvider.tsx
+++ b/js/hang-ui/src/Components/watch/WatchUIContextProvider.tsx
@@ -1,6 +1,7 @@
 import type { Time } from "@kixelated/hang";
 import type HangWatch from "@kixelated/hang/watch/element";
 import type { HangWatchInstance, InstanceAvailableEvent } from "@kixelated/hang/watch/element";
+import { Effect } from "@kixelated/signals";
 import type { JSX } from "solid-js";
 import { createContext, createEffect, createSignal, onCleanup } from "solid-js";
 
@@ -108,7 +109,13 @@ export default function WatchUIContextProvider(props: WatchUIContextProviderProp
 	return <WatchUIContext.Provider value={value}>{props.children}</WatchUIContext.Provider>;
 
 	function onWatchInstanceAvailable(watchEl: HangWatch, watchInstance: HangWatchInstance) {
-		watchInstance.signals.effect(function trackWatchStatus(effect) {
+		const localEffect = new Effect();
+
+		onCleanup(() => {
+			localEffect.close();
+		});
+
+		localEffect.effect(function trackWatchStatus(effect) {
 			const url = effect.get(watchInstance.connection.url);
 			const connection = effect.get(watchInstance.connection.status);
 			const broadcast = effect.get(watchInstance.broadcast.status);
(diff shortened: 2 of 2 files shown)


js/hang-demo/src/publish.html
```diff
@@ -22,10 +22,7 @@
 		Feel free to hard-code it if you have public access configured, like `url="https://relay.moq.dev/anon"`
 		NOTE: `http` performs an insecure certificate check. You must use `https` in production.
 	-->
-	<hang-publish url="%VITE_RELAY_URL%" path="me" controls>
-		<!-- It's optional to provide a video element to preview the outgoing media. -->
-		<video style="width: 100%; height: auto; border-radius: 4px; margin: 0 auto;" muted autoplay></video>
-	</hang-publish>
+	<hang-publish-ui url="%VITE_RELAY_URL%" path="me"></hang-publish-ui>
```
kixelated: I think we should have:

```html
<hang-publish-ui>
  <hang-publish><hang-publish>
<hang-publish-ui>
```

That way we don't need to duplicate all of the attributes, and can make attributes that only apply to the UI (ex. the location of various controls).

Of course it's more annoying to implement but that's web components baybee.

js/hang-ui/.prettierrc
```diff
@@ -0,0 +1,12 @@
+{
```
kixelated: Give biome a try; it's sooo much better than prettier IMO.

js/hang-ui/package.json
```diff
@@ -0,0 +1,36 @@
...
+        "build:types": "tsc -p tsconfig.json",
+        "clean": "rm -rf dist"
+    },
+    "devDependencies": {
+        "@rollup/plugin-node-resolve": "^16.0.3",
+        "@rollup/plugin-typescript": "^12.3.0",
+        "rollup": "^4.53.3",
+        "rollup-plugin-string": "^3.0.0",
+        "solid-element": "^1.9.1",
+        "solid-js": "^1.9.10",
```
kixelated: your choice if you use solids. I'd be down with trying out other frameworks

js/moq/package.json
```diff
@@ -6,8 +6,8 @@
 	"license": "(MIT OR Apache-2.0)",
 	"repository": "github:kixelated/moq",
 	"exports": {
-		".": "./src/index.ts",
-		"./zod": "./src/zod.ts"
+		".": "./dist/index.js",
```
kixelated: Ideally we can avoid this change because it makes local testing (via link) more difficult.

js/hang-demo/src/publish.html
jdreetz: I was rendering `<hang-publish>` [within the web component](https://github.com/kixelated/moq/pull/717/files#diff-d6c8f3286bf0267529a2fdd25563f8098ed984e4ce779b632b446bad178bd2c5R20) for `hang-publish-ui` because it's easier to get a ref to the `hang-publish` el, but I see what you're saying about having to duplicate all the attributes. I don't think it'll be an issue to get the ref using query selectors though.

js/moq/package.json
jdreetz: I was having difficulties getting the types for `HangPublish` in the UI without this change. Is there another way to do that? Maybe I can avoid using the types, but it'd make things easier to have access to them.

jdreetz: I got around this requirement by using esbuild to do the typescript handling in Rollup.

js/hang-ui/rollup.config.mjs
```diff
@@ -0,0 +1,57 @@
+// rollup.config.mjs / rollup.config.js
+import { readFileSync } from 'node:fs';
+import nodeResolve from '@rollup/plugin-node-resolve';
+import esbuild from 'rollup-plugin-esbuild';
+import solid from 'unplugin-solid/rollup';
+
+function inlineCss() {
+    return {
+        name: 'inline-css',
```
jdreetz, line 9: This seems like the most simple solution for now to include the CSS without having the user include a separate file, but as the UI grows, this made lead to issues with larger bundle sizes.

js/hang-ui/src/Components/publish/CameraSourceButton.tsx
```diff
@@ -0,0 +1,50 @@
+import { Show, useContext } from 'solid-js';
+import MediaSourceSourceSelector from './MediaSourceSelector';
+import { PublishUIContext } from './PublishUIContextProvider';
+
+export default function CameraSourceButton() {
+    const context = useContext(PublishUIContext);
+    const onClick = () => {
+        const hangPublishEl = context?.hangPublish();
+        if (!hangPublishEl) return;
```
jdreetz: to avoid having to do this check in a bunch of places, we could move the action into the Context, like `context?.toggleCameraSource()`. Or maybe there's a better way todo this directly with Signals?

js/hang-ui/src/Components/publish/FileSourceButton.tsx
```diff
@@ -0,0 +1,40 @@
...
+        HTMLInputElement | undefined
+    >();
+    const context = useContext(PublishUIContext);
+    const onClick = () => fileInputRef()?.click();
+    const onChange = (event: Event) => {
+        const castedInputEl = event.target as HTMLInputElement;
+        const file = castedInputEl.files?.[0];
+
+        if (file) {
+            context?.setFile(file);
```
jdreetz: alternatively, we could expose a setter on the HangPublish element and set it like that.

js/hang-ui/src/Components/publish/PublishUIContextProvider.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+    const value: PublishUIContextValue = {
+        hangPublish: props.hangPublish,
+        cameraDevices,
+        microphoneDevices,
+        publishStatus,
+        cameraActive,
+        screenActive,
+        microphoneActive,
+        fileActive,
```
jdreetz: I'm open to making these one property - `activeSources`, and then having the individual buttons check for membership in the Set/array.

js/hang-demo/vite.config.ts
```diff
@@ -16,6 +18,15 @@ export default defineConfig({
 			},
 		},
 	},
+	resolve: {
+		alias: {
+			"@kixelated/hang-ui/publish/element": path.resolve(
```
kixelated: kinda gross, but harmless I guess. You should leave a comment that this isn't needed when using the package from NPM.

This wasn't needed when `package.json` pointed to `src` instead of `dist`.

js/hang-ui/package.json
```diff
@@ -0,0 +1,32 @@
...
+	"repository": "github:kixelated/moq",
+	"main": "dist/publish-controls.esm.js",
+	"module": "dist/publish-controls.esm.js",
+	"exports": {
+		"./publish/element": "./dist/publish-controls.esm.js",
+		"./watch/element": "./dist/watch-controls.esm.js"
+	},
+	"sideEffects": true,
+	"scripts": {
+		"build": "npm run clean && rollup -c",
```
kixelated: And yeah, the other packages use a script to rewrite package.json to change `src` to `dist` for publishing. Your alias approach works too.

js/hang-ui/rollup.config.mjs
kixelated: We might want to do this anyway for `@kixelated/hang` because of worklets. I left some thoughts in #677

js/hang-ui/package.json
```diff
@@ -0,0 +1,32 @@
...
+		"check": "biome check src",
+		"clean": "rimraf dist",
+		"fix": "biome check src --fix"
+	},
+	"devDependencies": {
+		"@biomejs/biome": "^2.2.2",
+		"@kixelated/hang": "workspace:^0.7.0",
```
kixelated: Should this be a peerDependency?

js/hang-ui/src/Components/publish/PublishUIContextProvider.tsx
```diff
@@ -0,0 +1,193 @@
...
+				// @ts-expect-error ignore custom event - todo add event map
+				hangPublishEl.removeEventListener("publish-instance-available", onInstanceAvailable);
+			});
+		}
+	});
+
+	return <PublishUIContext.Provider value={value}>{props.children}</PublishUIContext.Provider>;
+
+	function onPublishInstanceAvailable(el: HangPublish, publishInstance: HangPublishInstance) {
+		publishInstance?.signals.effect(function trackCameraDevices(effect) {
```
kixelated: you should remove the `?` from all of these, it can't be null.

js/hang-ui/src/Components/publish/element.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { createSignal, onCleanup, onMount } from "solid-js";
+import PublishControls from "./PublishControls";
+import PublishControlsContextProvider from "./PublishUIContextProvider";
+import styles from "./styles.css?inline";
+
+customElement("hang-publish-ui", {}, function PublishControlsWebComponent(_, { element }) {
+	const [hangPublishEl, setHangPublishEl] = createSignal<HangPublish>();
+
+	const onInstanceAvailable = (event: CustomEvent) => {
+		const hangPublishEl = event.target as HangPublish;
```
kixelated, line 12: Not a fan of casting, the event should be typed.

js/hang-ui/src/Components/publish/element.tsx
```diff
@@ -0,0 +1,39 @@
...
+		setHangPublishEl(hangPublishEl);
+	};
+
+	onMount(() => {
+		const publishEl = element.querySelector("hang-publish");
+
+		if (publishEl) {
+			setHangPublishEl(publishEl);
+		} else {
+			element.addEventListener("publish-instance-available", onInstanceAvailable);
```
kixelated, line 22: You can use the signal instead if you want.

```js
const dispose = element.active.watch((instance?: PublishInstance) => {

});

onCleanup(dispose);
```

kixelated: Also we might be able to avoid the instance being undefined. The issue is the web component lifetime; we need to start work when inserted into the DOM and stop work when removed. Toggling enabled instead might do the trick.

js/hang-ui/package.json
jdreetz: I'll see if I can switch it to match how the other ones work. I don't like that the Vite config has to do special handling for that dependency.

js/hang-demo/vite.config.ts
jdreetz: I'll see if I can make it work like the other packages in the workspace.

js/hang-ui/src/Components/publish/element.tsx
jdreetz: for some reason, `active` isn't even defined when I first get a reference to the element in `onMount`, so I can't add the `watch` on `active`. Firing the event lets us know exactly when `active` is defined and has an instance AND it helps if the user adds `hang-watch` later on for some reason. 

I guess I could do something like poll for the `active` Signal, but that feels less ideal than just firing this event.

I don't understand what you mean about "Toggling enabled" though?

js/hang-ui/src/Components/publish/element.tsx
jdreetz: Added typings for both events.

js/hang-ui/src/Components/publish/PublishUIContextProvider.tsx
jdreetz: these should all be gone now.

js/hang-ui/package.json
jdreetz: Fixed.

js/hang-ui/package.json
jdreetz: Updated this to follow the pattern from the other packages. Works as far as I can tell.

js/hang-demo/vite.config.ts
jdreetz: removed this and followed the build pattern from the other packages.

PR #729, opened 2025-11-28
feat: Add audio track for hang-c library

emilsas:


kixelated:
Can you run `just fix` then we can merge this. There's a lot of things I want to improve with the rust code, but I can do that in a separate PR.

PR #738, opened 2025-12-05
Remove the `Instance` for Watch/Publish elements.

kixelated:
I think this is the best way of using WebComponent lifetimes. The main issue is that I need to call `close` before GC to avoid warnings. Well, why not call `close` when GC runs?

We avoid publishing when removed fro the DOM by toggling the enabled signal.

### DIFF
--- a/js/hang-demo/src/meet.ts
+++ b/js/hang-demo/src/meet.ts
@@ -1,4 +1,5 @@
 import "./highlight";
+import "@kixelated/hang-ui/publish/element";
 
 import HangMeet from "@kixelated/hang/meet/element";
 import HangPublish from "@kixelated/hang/publish/element";


js/hang-ui/src/Components/publish/CameraSourceButton.tsx
```diff
@@ -8,20 +8,23 @@ export default function CameraSourceButton() {
 		const hangPublishEl = context?.hangPublish();
 		if (!hangPublishEl) return;
 
-		if (hangPublishEl.source === "camera") {
+		if (hangPublishEl.source.peek() === "camera") {
```
kixelated, line 11: Not 100% sure if this is a good change, what do you think @christriants?

Basically by merging the element the `WatchInstance` and `Watch` we no longer need two separate namespaces. We could go back to hiding signals in a `signals` object or expose them as first class like this (ex. you can call `subscribe`).

js/hang-ui/src/Components/publish/CameraSourceButton.tsx
```diff
@@ -8,20 +8,23 @@ export default function CameraSourceButton() {
...
 			// Camera already selected, toggle video.
-			hangPublishEl.video = !hangPublishEl.video;
+			hangPublishEl.invisible.set(!hangPublishEl.invisible.peek());
 		} else {
-			hangPublishEl.source = "camera";
-			hangPublishEl.video = true;
+			hangPublishEl.source.set("camera");
+			hangPublishEl.invisible.set(false);
```
kixelated, line 16: I decided to flip the default around, so audio/video are enabled by default and you need to explicitly disable them.

js/hang-ui/src/Components/publish/CameraSourceButton.tsx
```diff
@@ -8,20 +8,23 @@ export default function CameraSourceButton() {
...
 		}
 	};
 
 	const onSourceSelected = (sourceId: MediaDeviceInfo["deviceId"]) => {
 		const hangPublishEl = context?.hangPublish();
 		if (!hangPublishEl) return;
 
-		hangPublishEl.videoDevice = sourceId;
```
kixelated, line 24: Inlined this setter. There's already other places in the code that do `"device" in video` so it's a bit more consistent.

js/hang-ui/src/Components/publish/PublishUIContextProvider.tsx
```diff
@@ -46,10 +44,7 @@ export default function PublishUIContextProvider(props: PublishUIContextProvider
 		const hangPublishEl = props.hangPublish();
 		if (!hangPublishEl) return;
 
-		hangPublishEl.file = file;
-		hangPublishEl.source = "file";
```
kixelated, line 50: I realized `source` can include `| File`.

js/hang-ui/src/Components/publish/PublishUIContextProvider.tsx
```diff
@@ -68,41 +63,12 @@ export default function PublishUIContextProvider(props: PublishUIContextProvider
...
-			onCleanup(() => {
-				hangPublishEl.removeEventListener("publish-instance-available", onInstanceAvailable);
-			});
-		}
-	});
-
-	return <PublishUIContext.Provider value={value}>{props.children}</PublishUIContext.Provider>;
-
-	function onPublishInstanceAvailable(el: HangPublish, publishInstance: HangPublishInstance) {
-		const localEffect = new Effect();
```
kixelated, line 97: I exposed `publish.signals` so this isn't needed any longer, although teeechnically `localEffect` is more correct. Up to you.

js/hang-ui/src/Components/publish/ScreenSourceButton.tsx
```diff
@@ -7,9 +7,9 @@ export default function ScreenSourceButton() {
 		const hangPublishEl = context?.hangPublish();
 		if (!hangPublishEl) return;
 
-		hangPublishEl.source = "screen";
-		hangPublishEl.audio = false;
```
kixelated, line 11: You can have audio when publishing a screen share.

js/hang-ui/src/Components/publish/element.tsx
```diff
@@ -1,30 +1,17 @@
...
-		if (publishEl) {
-			setHangPublishEl(publishEl);
-		} else {
-			element.addEventListener("publish-instance-available", onInstanceAvailable);
-		}
-	});
-
-	onCleanup(() => {
-		element.removeEventListener("publish-instance-available", onInstanceAvailable);
+		await customElements.whenDefined("hang-publish");
```
kixelated, line 13: This is required otherwise none of the properties are available.

js/hang/src/publish/element.ts
```diff
@@ -3,228 +3,94 @@ import { Effect, Signal } from "@kixelated/signals";
...
-		});
-		observer.observe(this.parent, { childList: true, subtree: true });
-		this.signals.cleanup(() => observer.disconnect());
+		cleanup.register(this, this.signals);
 
 		this.connection = new Moq.Connection.Reload({
-			enabled: true,
-			url: this.parent.signals.url,
+			url: this.url,
+			enabled: this.#enabled,
```
kixelated, line 54: We only connect while in the DOM, emulating the previous `Instance` behavior.

js/hang/src/publish/element.ts
```diff
@@ -3,228 +3,94 @@ import { Effect, Signal } from "@kixelated/signals";
...
-export type InstanceAvailableEvent = CustomEvent<{ instance: HangPublishInstance }>;
+type SourceType = "camera" | "screen";
 
-declare global {
-  interface GlobalEventHandlersEventMap {
-	'publish-instance-available': InstanceAvailableEvent;
-  }
-}
+// Close everything when this element is garbage collected.
+const cleanup = new FinalizationRegistry<Effect>((signals) => signals.close());
```
kixelated, line 15: This is why I think we can remove `Instance`. The cleanup isn't as immediate as the old code, but we disable everything when disconnected from the DOM so it should be good enough.

Issue #736, opened 2025-12-05
Close the VideoFrame


kixelated:
If you open the demo, scroll down to the bottom of the page, and then scroll back up, you'll get this warning after ~30s:

```
A VideoFrame was garbage collected without being closed. Applications should call close() on frames when done with them to prevent stalls.
```

We stop downloading video when the page is not visible but apparently don't clean up pending frames. Super easy fix; find the frame and call `.close()` on it.

pzanella:
Hi @kixelated!
I’ve updated PR #742 with flush method.
Let me know what you think. Thank you!

PR #742, opened 2025-12-07
fix: drain any remaining frames in the queue to prevent memory leaks

pzanella:
# Fix VideoFrame garbage collection warning on video pause/resume

## Description:
When pausing and resuming video playback, unclosed VideoFrames were being garbage collected, resulting in the warning:
`A VideoFrame was garbage collected without being closed. Applications should call close() on frames when done with them to prevent stalls.`

The #runTrack() method creates a queue of VideoFrames that are processed by the main frame-rendering loop. When playback is paused (effect is closed), the async frame-reading loop could be interrupted before all frames in the queue were consumed and closed. This left VideoFrames in memory without being properly cleaned up.

Closes: #736

### DIFF
--- a/js/hang/src/watch/video/source.ts
+++ b/js/hang/src/watch/video/source.ts
@@ -200,7 +200,20 @@ export class Source {
 		effect.cleanup(() => writer.close());
 
 		const reader = queue.readable.getReader();
-		effect.cleanup(() => reader.cancel());
+		effect.cleanup(async () => {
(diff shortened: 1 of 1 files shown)


kixelated:
I thiiiink this will work except when the writer is aborted (with an error). [flush](https://developer.mozilla.org/en-US/docs/Web/API/TransformStream/TransformStream#flushcontroller) might be the more correct way to accomplish the same thing?

PR #756, opened 2025-12-10
Add extended AAC support for variable-length AudioSpecificConfig

davegullo:
- Support extended audioObjectType (object_type 31, allowing types 32-63)
- Add explicit sample rate support (freq_index == 15 with 24-bit rate)
- Handle variable-length AudioSpecificConfig with extension data
- Improve channel configuration handling (support implicit and extended configs)
- Consume remaining extension bytes (SBR/PS) for robust parsing
- Update documentation to reflect variable-length config support

This enables compatibility with a wider range of AAC encoders including OBS and other sources that may use extended formats or explicit sample rates.

### DIFF
--- a/rs/hang/src/import/aac.rs
+++ b/rs/hang/src/import/aac.rs
@@ -24,141 +24,70 @@ impl Aac {
 		// - SBR/PS extensions
 		// - Explicit sample rates (freq_index == 15)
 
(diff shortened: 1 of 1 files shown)


kixelated:
@davegullo I did some minor refactoring if it looks gud to you

PR #759, opened 2025-12-10
feat: debug stats

### LINKED ISSUE #692: Demo debug stats
[The demos](https://github.com/kixelated/moq/tree/main/js/hang-demo/src) are pretty simple, limited to just embedding the web component and letting it rip with the built-in controls.

We should use the Javascript API to get more detailed stats about the connection. For example, `watch.video.catalog` is a reactive signal that lists the available video tracks and their properties; we should render that below the fold.

pzanella:
# Stats component

Introduces a comprehensive **Stats component** for real-time monitoring of media streaming performance metrics in the `hang-ui` library (now, it's only enabled in watch element).
Closes #692 

### Metrics Tracking
Four specialized metric handlers:
1. **Network** - Bandwidth, latency, packet loss
2. **Video** - Bitrate, resolution, frame rate
3. **Audio** - Bitrate, sample rate, channels
4. **Buffer** - Queue depth, jitter buffer status

### Architecture
- **Handlers Pattern** - Extensible base handler class for custom metrics
- **Type-Safe** - Full TypeScript support with generic context typing
- **Testable** - Comprehensive unit tests for all handlers and the main component
- **Styled** - Modular SCSS with responsive design

### DIFF
--- a/js/hang-ui/src/Components/stats/providers/registry.ts
+++ b/js/hang-ui/src/Components/stats/providers/registry.ts
@@ -22,7 +22,7 @@ export const providers: Record<KnownStatsProviders, ProviderConstructor> = {
 
 /**
  * Get provider class for a metric type
- * @param icon - Metric type identifier
+ * @param statProvider - Metric type identifier
  * @returns Provider constructor or undefined if not found
  */
 export function getStatsInformationProvider(statProvider: KnownStatsProviders): ProviderConstructor | undefined {
--- a/js/hang-ui/src/css.d.ts
+++ b/js/hang-ui/src/css.d.ts
@@ -2,8 +2,3 @@ declare module "*.css?inline" {
 	const content: string;
 	export default content;
 }
-
-declare module "*.scss?inline" {
-	const content: string;
-	export default content;
-}


js/hang-ui/src/Components/stats/Stats.tsx
```diff
@@ -0,0 +1,43 @@
...
+			<Show when={statsReady()}>
+				<StatsContext.Provider
+					value={{
+						audio: props.getElement(contextValue)?.audio,
+						video: props.getElement(contextValue)?.video,
+					}}
+				>
+					<StatsWrapper />
+				</StatsContext.Provider>
+			</Show>
```
kixelated: The syntax (from memory) is undoubtably wrong but you can do something like this:

```suggestion
			<Show when={props.getElement(contextValue)}>
			    { (element) => {
				    <StatsContext.Provider
					    value={{
						    audio: element.audio,
					    	video: element.video,
					    }}
				    >
				} }
					<StatsWrapper />
				</StatsContext.Provider>
			</Show>
```

js/hang/src/watch/audio/source.ts
```diff
@@ -165,6 +168,11 @@ export class Source {
 		});
 		effect.cleanup(() => consumer.close());
 
+		let totalBytes = 0;
+		let lastTimestamp = 0;
+		let lastCalculation = 0;
+		const bitrateSmoothingTime = 250;
```
kixelated: I don't think this is a good place to calculate this. The smoothing window is very specific to the stats window. Likewise I don't want to call `performance.now()` when it's not being used.

Can you instead create: 
```ts
export interface Stats {
  bytesReceived: number
}

#stats = new Signal<Stats | undefined>(undefined);
readonly stats: Getter<Stats | undefined> = this.#stats;
```

The UI layer can then create a timer to fire every 250ms (or whatever) to poll the current value and subtract the previous value.

js/hang/src/watch/video/source.ts
```diff
@@ -215,6 +219,10 @@ export class Source {
 			}
 		});
 
+		let lastFrameTimestamp: number | null = null;
+		let recentFpsValues: number[] = [];
+		const fpsSmoothing = 5;
```
kixelated: Same deal with video, can you export the cumulative number of frames received (and bytes)? The UI can choose how to best bucket/render them.

js/hang-ui/src/Components/stats/Stats.tsx
pzanella: Sorry, it's my first time using SolidJS, let me correct that.

js/hang-ui/src/Components/stats/handlers/audio.ts
```diff
@@ -46,13 +50,35 @@ export class AudioHandler extends BaseHandler {
...
 			this.context.setDisplayData("N/A");
 			return;
 		}
 
+		let bitrate: string | undefined;
+		if (stats && this.previousBytesReceived > 0) {
+			const bytesDelta = stats.bytesReceived - this.previousBytesReceived;
+			// Only calculate bitrate if there's actual data change
+			if (bytesDelta > 0) {
+				const bitsPerSecond = bytesDelta * 8 * (1000 / AudioHandler.POLLING_INTERVAL_MS);
```
kixelated: For best accuracy, also store `this.previousWhen = performance.now()` and subtract the actual time spent. Javascript won't fire intervals at exactly `AudioHandler.POLLING_INTERVAL_MS`, especially when the tab is backgrounded.

js/hang/src/watch/audio/source.ts
```diff
@@ -169,9 +173,6 @@ export class Source {
 		effect.cleanup(() => consumer.close());
 
 		let totalBytes = 0;
```
kixelated:
```typescript
effect.cleanup(() => this.#stats.set(undefined));
```

js/hang/src/watch/video/source.ts
```diff
@@ -246,24 +247,13 @@ export class Source {
...
-					}
-				}
-				lastFrameTimestamp = now;
+				// Track frame count and timestamp for FPS calculation in the UI
+				const currentStats = this.#stats.peek();
+				this.#stats.set({
+					frameCount: (currentStats?.frameCount ?? 0) + 1,
+					timestamp: frame.timestamp,
+					bytesReceived: currentStats?.bytesReceived ?? 0,
+				});
```
kixelated:
```suggestion
				this.#stats.update((current) => {
					frameCount: (current?.frameCount ?? 0) + 1,
					timestamp: frame.timestamp,
					bytesReceived: current?.bytesReceived ?? 0,
				});
```

js/hang/src/watch/audio/source.ts
kixelated: Actually, maybe we should just not allow `stats` to be `undefined`. Add to the previous number of `totalBytes` (like video) instead of resetting it to zero on a new rendition.

js/hang-ui/src/Components/stats/handlers/base.ts
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	/**
+	 * Get current value from signal
+	 * @param signal - Signal to peek
+	 * @returns Current signal value or undefined
+	 */
+	protected peekSignal<T>(signal: { peek: () => T | undefined } | undefined): T | undefined {
+		return signal?.peek?.();
+	}
```
kixelated: IMO this is unnecessary boilerplate.

js/hang-ui/src/Components/stats/utils/subscription.ts
```diff
@@ -0,0 +1,37 @@
...
+	/** Array of unsubscribe functions */
+	private subscriptions: (() => void)[] = [];
+
+	/**
+	 * Subscribe to a signal and track unsubscribe function
+	 * @param signal - Signal to subscribe to
+	 * @param callback - Function called when signal changes
+	 */
+	subscribe(signal: { subscribe?: (callback: () => void) => () => void } | undefined, callback: () => void): void {
+		const unsubscribe = signal?.subscribe?.(callback);
```
kixelated: All of the `?` and `undefined` scares me a bit. Seems easy to misuse this helper, and it's generally a code-smell to have functions that can take undefined and do nothing. The caller should check for undefined instead.

js/hang-ui/src/Components/stats/utils/subscription.ts
```diff
@@ -0,0 +1,37 @@
+/**
+ * Manages signal subscriptions for handlers
+ */
+export class SubscriptionManager {
```
kixelated: btw this is very similar to `Effect` in `@kixelated/signals`. It has the advantage of being able to subscribe to multiple signals, unsubscribing when one changes and running the callback again (which may then resubscribe).

pzanella: I agreed, I've removed SubscriptionManager, thanks.

js/hang-ui/src/Components/stats/providers/buffer.ts
```diff
@@ -22,9 +20,22 @@ export class BufferHandler extends BaseHandler {
 			return;
 		}
 
-		this.subscribe(video.source.syncStatus, this.updateDisplay);
-		this.subscribe(video.source.bufferStatus, this.updateDisplay);
-		this.subscribe(video.source.latency, this.updateDisplay);
+		// Subscribe to signal changes using Effect
+		this.signals.effect((effect) => {
```
kixelated, line 24: Here's how I would handle this:

```typescript
// This function will run once on start and whenever any of the status change
this.signals.effect((effect) => {
	const syncStatus = effect.get(this.props.video.source.syncStatus);
	const bufferStatus = effect.get(this.props.video.source.bufferStatus);
	const latency = effect.get(this.props.video.source.latency);

    // inline updateDisplayData
});
```

js/hang/src/watch/audio/source.ts
```diff
@@ -172,7 +172,9 @@ export class Source {
 		});
 		effect.cleanup(() => consumer.close());
 
-		let totalBytes = 0;
+		// Persist totalBytes across rendition changes
+		const currentStats = this.#stats.peek() ?? { bytesReceived: 0 };
+		let totalBytes = currentStats.bytesReceived;
```
kixelated: To be safer and support rendition switching (in the future), can you replace this local variable with:

```typescript
this.#stats.update((stats) => {
  bytesReceived: (stats?.bytesReceived ?? 0) + byteLength,
});
```

just like how video works

js/hang/src/watch/video/source.ts
```diff
@@ -248,12 +248,11 @@ export class Source {
 				if (!frame) break;
 
 				// Track frame count and timestamp for FPS calculation in the UI
-				const currentStats = this.#stats.peek();
-				this.#stats.set({
-					frameCount: (currentStats?.frameCount ?? 0) + 1,
+				this.#stats.update((current) => ({
+					frameCount: (current?.frameCount ?? 0) + 1,
```
kixelated: Why increment frameCount here but bytesReceived further down? Can you combine them?

js/hang-ui/src/Components/stats/Stats.test.tsx
```diff
@@ -0,0 +1,310 @@
...
+					codec: "aac",
+				}),
+				subscribe: vi.fn(() => vi.fn()),
+			},
+			stats: {
+				peek: () => ({
+					bytesReceived: 0,
+				}),
+				subscribe: vi.fn(() => vi.fn()),
+			},
```
jdreetz: it'd be nice if there was a some helper function or something to make generating these signal mocks easier, but this is probably fine for now.

js/hang-ui/src/Components/stats/Stats.test.tsx
```diff
@@ -0,0 +1,310 @@
+import { createContext, createSignal } from "solid-js";
+import { render } from "solid-js/web";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { Stats } from "./Stats";
+import type { HandlerProps } from "./types";
+
+/**
+ * Mock HandlerProps with audio and video data
+ */
+const createMockHandlerProps = (): HandlerProps => ({
```
jdreetz: what do you think about making this function configurable, so it can return either video or audio, and that would make the calls more explicit about what it's providing (video and audio vs just audio vs just video)?

jdreetz: I guess the default could be provide both.

js/hang-ui/src/Components/stats/Stats.tsx
```diff
@@ -0,0 +1,36 @@
...
+	context: Context<T>;
+	getElement: (ctx: T) => HandlerProps | undefined;
+}
+
+/**
+ * Stats component for displaying real-time media streaming metrics
+ * Accepts a generic context and a function to extract the media element
+ */
+export const Stats = <T = unknown>(props: StatsProps<T>) => {
+	const contextValue = useContext(props.context);
```
jdreetz, line 16: for watch and publish, I switched to using a wrapper hook around the context to ensure the context is never used if it's undefined. not sure if you want to do something similar, or is that what `useMetrics` is intended to?

js/hang-ui/src/Components/stats/Stats.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+	return (
+		<div class="stats">
+			<style>{styles}</style>
+			<Show when={props.getElement(contextValue)}>
```
jdreetz: why not wrap the whole component with the `Show`?

js/hang-ui/src/Components/stats/Stats.tsx
```diff
@@ -0,0 +1,36 @@
...
+				{(element) => (
```
jdreetz: What do you think about using an `_` for the name to indicate it's intended to be unused.

js/hang-ui/src/Components/stats/components/Button.tsx
```diff
@@ -0,0 +1,29 @@
+/**
+ * Props for toggle button component
+ */
+interface ButtonProps {
+	/** Whether stats panel is currently visible */
+	isVisible: boolean;
```
jdreetz: the name should indicate the visibility is referring to the `panel` not the button itself. so like `isPanelVisible`.

js/hang-ui/src/Components/stats/components/StatsItem.tsx
```diff
@@ -0,0 +1,53 @@
...
+			audio: props.audio,
+			video: props.video,
+		});
+
+		handler.setup({ setDisplayData });
+
+		onCleanup(() => {
+			handler.cleanup();
+		});
+	});
```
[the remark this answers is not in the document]
jdreetz: I had the same concern. I guess if it's just coming from our own code, it's fine for now.

js/hang-ui/src/Components/stats/handlers/registry.ts
```diff
@@ -0,0 +1,24 @@
...
+	buffer: BufferHandler,
+	network: NetworkHandler,
+};
+
+/**
+ * Get handler class for a metric type
+ * @param icon - Metric type identifier
+ * @returns Handler constructor or undefined if not found
+ */
+export function getHandlerClass(icon: Icons): HandlerConstructor | undefined {
```
jdreetz: is "Handler" the right way of describing these entities? It's more like a `StatsInformationProvider`, right?

js/hang-ui/src/Components/stats/handlers/registry.ts
```diff
@@ -0,0 +1,24 @@
...
+	return handlerRegistry[icon];
```
jdreetz: do you think using the icon name is the right way to map to the "handler"? It seems like we should have separate constants for that: `type KnownStatsProviders = 'video' | 'audio' | 'buffer' | 'network' ...add more over time`.

That feels clearer than re-using the icon name for the mapping. Icon names and handler types are two distinct concepts even though they're essentially the same at the moment. I think they should be separated.

js/hang-ui/src/Components/stats/components/StatsPanel.tsx
```diff
@@ -0,0 +1,22 @@
...
+ */
+interface StatsPanelProps extends HandlerProps {}
+
+/**
+ * Panel displaying all metrics in a grid layout
+ */
+export const StatsPanel = (props: StatsPanelProps) => {
+	return (
+		<div class="stats__panel">
+			<For each={Object.entries(PANEL_SVGS)}>
```
jdreetz: I mentioned this in the `getHandlerClass` file, but it feels kind of weird to have the list of icons be the data model for what stats we display. It feels like we should have a separate data structure for that. Like

```
const statsDetailItems = [
    {
        name: 'Buffer',
        icon: 'icon name',
        svg: <path to svg>,
        ...any other meta data we want to include in the future.
    }
]

File: js/hang-ui/src/Components/stats/components/StatsWrapper.tsx [hunk 23]
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Root wrapper component managing visibility state and layout
+ */
+export const StatsWrapper = () => {
+	const [isVisible, setIsVisible] = createSignal(false);
+	const metrics = useMetrics();
+
+	return (
+		<div class="stats__wrapper">
+			<Button isVisible={isVisible()} onToggle={setIsVisible} icon={BUTTON_SVG} />
```
jdreetz: mentioned this on the button, but `isVisible` should be like `isPanelVisible` otherwise the natural assumption (IMO) is that this is describing whether the _button_ is visible.

js/hang-ui/src/Components/stats/components/__tests__/StatsPanel.test.tsx
```diff
@@ -0,0 +1,116 @@
...
+			() => (
+				<StatsContext.Provider value={mockAudioVideo}>
+					<StatsPanel />
+				</StatsContext.Provider>
+			),
+			container,
+		);
+
+		const items = container.querySelectorAll(".stats__item");
+		expect(items.length).toBe(4);
```
jdreetz: Do you think it's a bad idea to import the icon list and use the length of that for the count to check against? 

Like - `expect(items.length).toBe(Object.entries(icons).length);`

js/hang-ui/src/Components/stats/components/icons.ts
```diff
@@ -0,0 +1,45 @@
...
+        <path d="M5 21v-6"/>
+        <path d="M12 21V3"/>
+        <path d="M19 21V9"/>
+    </svg>
+`;
+
+/**
+ * SVG icons for metric types
+ */
+export const PANEL_SVGS: Record<Icons, string> = {
```
jdreetz: any reason to use plain strings instead of JSX/TSX?

js/hang-ui/src/Components/stats/context.ts
```diff
@@ -0,0 +1,16 @@
...
+ * Context providing stream sources to child components
+ */
+export const StatsContext = createContext<HandlerProps>({});
+
+/**
+ * Hook to access stream sources from context
+ * @returns Stream properties (audio and video)
+ */
+export const useMetrics = () => {
+	const context = useContext(StatsContext);
```
jdreetz: for the other wrappers, we `throw` if context is undefined. Do you think we need to do that here?

js/hang-ui/src/Components/stats/handlers/audio.ts
```diff
@@ -0,0 +1,112 @@
...
+	setup(context: HandlerContext): void {
+		this.context = context;
+		const audio = this.props.audio;
+
+		if (!audio) {
+			context.setDisplayData("N/A");
+			return;
+		}
+
+		this.updateInterval = window.setInterval(this.updateDisplay, AudioHandler.POLLING_INTERVAL_MS);
```
jdreetz: you're using the wrapped `updateDisplay` to retain scope, right? We could also do `updateDisplayData.bind(this)`, right?

js/hang-ui/src/Components/stats/handlers/base.ts
```diff
@@ -0,0 +1,33 @@
+import { Effect } from "@moq/signals";
+import type { HandlerContext, HandlerProps, IStatsHandler } from "../types";
+
+/**
+ * Base class for metric handlers providing common utilities
+ */
+export abstract class BaseHandler implements IStatsHandler {
```
jdreetz: do we need `IStatsHandler` if we have a `BaseHandler`? What's the intent in having both?

js/hang-ui/src/Components/stats/handlers/network.ts
```diff
@@ -0,0 +1,167 @@
...
+		this.networkInfo = nav.connection ?? nav.mozConnection ?? nav.webkitConnection;
+
+		if (this.networkInfo?.addEventListener) {
+			this.networkInfo.addEventListener("change", this.updateDisplay);
+		}
+
+		window.addEventListener("online", this.onlineStatusChanged);
+		window.addEventListener("offline", this.onlineStatusChanged);
+
+		this.updateInterval = window.setInterval(this.updateDisplay, NetworkHandler.POLLING_INTERVAL_MS);
```
jdreetz: do we want to set this up if `networkInfo` is undefined?

js/hang-ui/src/Components/stats/style/colors.scss
jdreetz, line 1: why SCSS variables instead of native CSS variables?

js/hang-ui/src/Components/stats/style/spacing.scss
jdreetz, line 1: same question re: SCSS variables vs native CSS variables, why not use CSS variables instead? I'm not sure what browsers we're targeting exactly, but I'm assuming it's modern ones that support CSS variables?

js/hang-ui/src/Components/stats/providers/buffer.ts
pzanella: Thanks for the suggestion! I’ve applied this approach in commit [84f46ed](https://github.com/moq-dev/moq/pull/759/commits/84f46ed28a23bd9e8bba795ead6e06b97afc8209).
Please let me know if this looks good or if you’d like any changes.

js/hang/src/watch/video/source.ts
pzanella: I’ve combined them in the latest commit.
Please let me know if everything looks correct.

js/hang-ui/src/Components/stats/Stats.test.tsx
pzanella: Thanks for the suggestion!
I totally agree, a helper would make this much cleaner.
I can add it now if you want, just let me know.

js/hang-ui/src/Components/stats/Stats.test.tsx
pzanella: Yes, absolutely! Thanks for the suggestion!
Would something like this work?
```js
interface MockHandlerPropsOptions {
  audio?: boolean;
  video?: boolean;
}

const createMockHandlerProps = (options?: MockHandlerPropsOptions): HandlerProps => {
  const { audio = true, video = true } = options ?? {};
  
  return {
    ...(audio && { audio: { source: { ... } } }),
    ...(video && { video: { source: { ... } } }),
  };
};
```
Then:
```js
createMockHandlerProps();                    // audio and video
createMockHandlerProps({ video: false });    // audio only
createMockHandlerProps({ audio: false });    // video only
```
Let me know what you think!

js/hang-ui/src/Components/stats/Stats.tsx
pzanella: Would something like this work?
```js
export const useMetrics = () => {
  const context = useContext(StatsContext);
  if (!context?.audio || !context?.video) {
    throw new Error("useMetrics must be used inside StatsWrapper when both audio and video are available");
  }
  return context;
};
```

js/hang-ui/src/Components/stats/Stats.tsx
pzanella: Yes, absolutely! The Stats component was originally created differently and not inside hang-ui, so I’ll wrap it with the <Show> tag.
Thanks!

js/hang-ui/src/Components/stats/handlers/audio.ts
pzanella: You're absolutely right! Both approaches work. Since `updateDisplay` is only used once in the `setInterval` call, using `.bind(this)` directly would be simpler and avoid the extra property. I'll update it to inline the binding.

js/hang-ui/src/Components/stats/handlers/base.ts
pzanella: Good point! Currently, `IStatsHandler` is somewhat redundant since all implementations extend `BaseHandler`. If we don't plan to have handlers that implement the interface without using the base class, we could remove the interface and use `BaseHandler` directly as the type contract. What's your preference?

js/hang-ui/src/Components/stats/handlers/network.ts
pzanella: I'll add the code:
```js
if (!this.networkInfo) {
   context.setDisplayData("N/A");
   return;
}
```
as it's done in the other handlers. Thanks!

js/hang-ui/src/Components/stats/style/spacing.scss
pzanella: This wasn't refactored earlier because it was originally a standalone web component. Now that it's part of the `hang-ui` component. Should we define CSS variables locally within the stats folder, or centralize them at the `hang-ui` level for consistency? Do you think it's better to define CSS variables using :root, :host, or by specifying them in the .stats class? 
In my opinion, the CSS variables could be centralized and defined at the :host scope at the `hang-ui` level.  
What do you think? Can I proceed?
Thank you!

js/hang-ui/src/Components/stats/handlers/registry.ts
pzanella: I'm changing `getHandlerClass` in `getStatsInformationProvider`, thank you for suggestion.

js/hang-ui/src/Components/stats/Stats.test.tsx
jdreetz: Up to you. not a blocker

js/hang-ui/src/Components/stats/Stats.test.tsx
jdreetz: Works for me. Thanks!

js/hang-ui/src/Components/stats/Stats.tsx
jdreetz: thanks!

js/hang-ui/src/Components/stats/handlers/registry.ts
pzanella: Yes, I agreed. I'm creating a new type like you suggest me. Thanks!

js/hang-ui/src/Components/stats/handlers/base.ts
jdreetz: Yea, let's just use `BaseHandler`.

js/hang-ui/src/Components/stats/handlers/network.ts
jdreetz: Sounds good, thanks!

js/hang-ui/src/Components/stats/Stats.tsx
jdreetz: in the other wrappers, we just checked if `context` is defined and threw otherwise. That's all we need here IMO.

js/hang-ui/src/Components/stats/handlers/registry.ts
jdreetz: nice!

js/hang-ui/src/Components/stats/style/spacing.scss
jdreetz: Yea, I think we should centralize the css variables per component (Stats, Watch, Publish). So for Stats, I'd do `hang-ui/Components/stats/variables.css` and scope them to the Stats container element (not `:root` or `:host`). 

If we find that we're repeating styles across components, then I'd make some cross-component variables, but I don't think we need it _yet_ given how primitive the UI is. 

Let me know if you have concerns about this though. I'm not sure how this UI will evolve (or how long we'll updating it), so I think we should keep it simple until we know more.

js/hang-ui/src/Components/stats/Stats.tsx
pzanella: I totally agree and I'm adding this check. Thanks!

js/hang-ui/src/Components/stats/components/icons.tsx
```diff
@@ -0,0 +1,116 @@
...
+		class="stats__icon"
+	>
+		<title>Open statistics</title>
+		<path d="M5 21v-6" />
+		<path d="M12 21V3" />
+		<path d="M19 21V9" />
+	</svg>
+);
+
+export const statsDetailItems: { name: string; statProvider: KnownStatsProviders; svg: () => JSX.Element }[] = [
```
jdreetz: this should be probably be in a different file now, like `statsItems.tsx` or maybe we just rename this file.

js/hang-ui/src/Components/stats/style.css
```diff
@@ -0,0 +1,209 @@
...
+	width: 100%;
+	height: 100%;
+	display: flex;
+	align-items: center;
+	justify-content: flex-start;
+	flex-direction: column;
+}
+
+.stats .stats__button {
+	position: absolute;
```
jdreetz: I'm hesitant to make this is an overlay on the video. It looks really nice, but I wouldn't be surprised if users didn't like it blocking the video (even if just a tiny bit). 

What do you think about moving the button to the "control bar", like to the right of the fullscreen button? I think the panels are fine as an overlay though. I know we'd have to shrink the icon to make it look consistent with the other buttons, but I think putting it there makes more sense as that group (control bar buttons) represents ways to "interact" with the video, and a method to show stats about the video is a type of interaction (sort of).

js/hang-ui/src/Components/stats/components/icons.tsx
pzanella: Good point, I'm thinking about moving `statsButtonIcon` inside the Button.tsx component and `statsDetailItems` inside the StatsItem.tsx component, and then deleting icons.tsx. What do you think? Thank you!

js/hang-ui/src/Components/stats/style.css
pzanella: Good point, thank you for the suggestion.
I'm thinking about creating another signal inside WatchUIContextProvider.tsx, like:
```js
const [isStatsPanelVisible, setIsStatsPanelVisible] = createSignal<boolean>(false);
```
to manage the isStatsPanelVisible state.
I would then create a `StatsButton` component and use it inside WatchControls.tsx:
```js
<div class="watchControlsContainer">
   <div class="playbackControlsRow">
	<PlayPauseButton />
	<VolumeSlider />
	<WatchStatusIndicator />
	<FullscreenButton />
	<StatsButton />
   </div>
   <div class="latencyControlsRow">
	<LatencySlider />
	<QualitySelector />
   </div>
</div>
```
In element.tsx, I would import the `StatsPanel` component like this:
```js
<Show when={isStatsPanelVisible()}>
   <StatsPanel audio={...} video={...} />
</Show>
```
There are some changes involved, but it shouldn’t be a problem.
What do you think? 
Let me know, thanks!

js/hang-ui/src/Components/stats/components/icons.tsx
jdreetz: Work for me!

js/hang-ui/src/Components/stats/style.css
jdreetz: That sounds great!

kixelated (approved):
@jdreetz hit the merge button when everything looks good on your end. You guys have free reign on the UI stuff; I'll mostly focus on the `hang` and `moq` packages.

jdreetz (approved):

PR #760, opened 2025-12-10
feat: quality selector component

cotid-qualabs:
This pull request introduces the foundational elements for a video quality selection feature. It includes a new UI component for users to choose video renditions, integrates this component into the existing watch controls, and updates the underlying context and core video source logic to support managing and applying manual rendition selections.

### DIFF
--- a/js/hang-ui/src/Components/watch/WatchUIContextProvider.tsx
+++ b/js/hang-ui/src/Components/watch/WatchUIContextProvider.tsx
@@ -82,7 +82,10 @@ export default function WatchUIContextProvider(props: WatchUIContextProviderProp
 		const hangWatchEl = props.hangWatch();
 
 		if (hangWatchEl) {
-			hangWatchEl.video.source.target.set({ rendition: name });
+			hangWatchEl.video.source.target.update((prev) => ({
+				...prev,
+				rendition: name,
+			}));
 		}
 	};
 


cotid-qualabs:
@coderabbitai generate docstrings

js/hang/src/watch/video/source.ts
```diff
@@ -65,6 +65,9 @@ export class Source {
 	// Used as a tiebreaker when there are multiple tracks (HD vs SD).
 	target = new Signal<Target | undefined>(undefined);
 
+	// Optional manual override for the selected rendition name.
+	#manualSelection = new Signal<string | undefined>(undefined);
```
kixelated: Hmm, can you add this to `Target` instead? The plan was to add more criteria, like bitrate, and the track name makes sense in there too.

js/hang/src/watch/video/source.ts
```diff
@@ -107,6 +110,11 @@ export class Source {
 		this.#signals.effect(this.#runBuffer.bind(this));
 	}
 
+	// Allow external callers to select a specific rendition by name.
+	setActiveRendition(name: string | undefined): void {
```
kixelated: This could then be replaced with:

```typescript
watch.video.target.set({
  rendition: "foobar"
});
```

As you may have noticed, I'm going pretty hard on signals instead of explicit get/set methods. They have the advantage of being subscribable/reactive.

js/hang-ui/src/Components/watch/WatchUIContextProvider.tsx
```diff
@@ -66,6 +78,14 @@ export default function WatchUIContextProvider(props: WatchUIContextProviderProp
 		}
 	};
 
+	const setActiveRenditionValue = (name: string | undefined) => {
+		const hangWatchEl = props.hangWatch();
+
+		if (hangWatchEl) {
```
kixelated, line 84: These sorts of conditions worry me, because it means we can silently skip signal updates.

@jdreetz I think this happens a lot in the code base. I think we should gate the entire UI behind `HangWatch` so it's not optional. I don't think every UI element should handle what to do if the core element doesn't exist.

jdreetz: @kixelated I agree. I'll make a separate PR for that unless you think we should do it in this PR.

kixelated: Separate PR sounds gud.

jdreetz: https://github.com/moq-dev/moq/pull/764

js/hang/src/watch/video/source.ts
cotid-qualabs: Sure, implemented on commit  https://github.com/moq-dev/moq/pull/760/commits/585c29b36380e2ae7eb40c149906ce771a209768 and ready for review

js/hang/src/watch/video/source.ts
cotid-qualabs: implemented on commit https://github.com/moq-dev/moq/commit/585c29b36380e2ae7eb40c149906ce771a209768

js/hang-ui/src/Components/watch/WatchUIContextProvider.tsx
cotid-qualabs: Hi! I can wait for James PR to be merged and then update this implementation, if that works better for you

jdreetz: Mine should be ready soon, but looks like this is already merged so I'll refactor to support it.

jdreetz: https://github.com/moq-dev/moq/pull/764 includes the quality selector changes and updates them to remove the `hangWatch` defined checks.

PR #761, opened 2025-12-10
feat: hls import

sebastianpiq:
This PR introduces HLS import functionality. The import module now supports a manifest-based import type for HLS streams (with potential DASH support in the future).

#### HLS Import Implementation

- Added manifest type support to the import module for HLS streams
- Added media type support for the current existing imports (aac, fmp4, and anexb)
- Created a generic interface for manifest import (like the generic for media)
- Implemented a basic HLS importer that parses manifest data and pushes it using the existing fmp4 import implementation
- The importer continuously pulls segments from the manifest in a loop for a specified duration

#### CLI Support

Added CLI command for HLS import: `just ingest-hls`

e.g `nix develop -c just ingest-hls http://example.com/manifest.m3u8 example`

The HLS importer provides a first basic implementation that parses manifest data and uses the existing fmp4 import. The manifest polling loop uses the segment target duration to determine the refresh interval, though this timing mechanism could be further refined.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,9 @@ node_modules
 *.jwk
 *.jwt
 
+# HLS stream output (just hls-stream)
+dev/hls_*/
+
 # We're using bun (and sometimes deno)
 package-lock.json
 yarn.lock
--- a/justfile
+++ b/justfile
@@ -290,3 +290,74 @@ update:
 build:
 	bun run --filter='*' build
 	cargo build
+
+# Generate and serve an HLS stream from a video for testing pub-hls
+hls-stream name="bbb" port="8000":
+	#!/usr/bin/env bash
+	set -euo pipefail
+
+	just download "{{name}}"
+
+	INPUT="dev/{{name}}.mp4"
+	OUT_DIR="dev/hls_{{name}}"
+
+	rm -rf "$OUT_DIR"
+	mkdir -p "$OUT_DIR/video_1080p" "$OUT_DIR/video_144p" "$OUT_DIR/audio"
+
+	START_NUMBER=$(date +%s)
+
+	cat > "$OUT_DIR/master.m3u8" << 'EOF'
+	#EXTM3U
+
+	#EXT-X-MEDIA:TYPE=AUDIO,NAME="Stereo",GROUP-ID="audio",DEFAULT=YES,AUTOSELECT=YES,URI="audio/audio.m3u8"
+
+	#EXT-X-STREAM-INF:BANDWIDTH=4500000,CODECS="avc1.640029,mp4a.40.2",RESOLUTION=1920x1080,AUDIO="audio"
+	video_1080p/video.m3u8
+
+	#EXT-X-STREAM-INF:BANDWIDTH=350000,CODECS="avc1.640029,mp4a.40.2",RESOLUTION=256x144,AUDIO="audio"
+	video_144p/video.m3u8
+	EOF
+
+	sed -i 's/^\t//' "$OUT_DIR/master.m3u8"
+
+	echo ">>> Starting HLS stream generation..."
+	echo ">>> Master playlist: http://localhost:{{port}}/master.m3u8"
+
+	cleanup() {
+		echo "Shutting down..."
+		kill $(jobs -p) 2>/dev/null || true
+		exit 0
+	}
+	trap cleanup SIGINT SIGTERM
+
+	ffmpeg -loglevel warning -re -stream_loop -1 -i "$INPUT" \
+		-map 0:v:0 -r 25 -preset veryfast -g 50 -keyint_min 50 -sc_threshold 0 \
+		-c:v libx264 -profile:v high -level 4.1 -pix_fmt yuv420p -tag:v avc1 \
+		-bsf:v dump_extra -b:v 4M -vf "scale=1920:-2" \
+		-f hls -hls_time 2 -hls_list_size 12 \
+		-hls_flags independent_segments+delete_segments \
+		-hls_segment_type fmp4 -hls_fmp4_init_filename init.mp4 \
+		-hls_segment_filename "$OUT_DIR/video_1080p/segment_%09d.m4s" \
+		-start_number "$START_NUMBER" "$OUT_DIR/video_1080p/video.m3u8" &
+
+	ffmpeg -loglevel warning -re -stream_loop -1 -i "$INPUT" \
+		-map 0:v:0 -r 25 -preset veryfast -g 50 -keyint_min 50 -sc_threshold 0 \
+		-c:v libx264 -profile:v high -level 4.1 -pix_fmt yuv420p -tag:v avc1 \
+		-bsf:v dump_extra -b:v 300k -vf "scale=256:-2" \
+		-f hls -hls_time 2 -hls_list_size 12 \
+		-hls_flags independent_segments+delete_segments \
+		-hls_segment_type fmp4 -hls_fmp4_init_filename init.mp4 \
+		-hls_segment_filename "$OUT_DIR/video_144p/segment_%09d.m4s" \
+		-start_number "$START_NUMBER" "$OUT_DIR/video_144p/video.m3u8" &
+
+	ffmpeg -loglevel warning -re -stream_loop -1 -i "$INPUT" \
+		-map 0:a:0 -c:a aac -b:a 128k \
+		-f hls -hls_time 2 -hls_list_size 12 \
+		-hls_flags independent_segments+delete_segments \
+		-hls_segment_type fmp4 -hls_fmp4_init_filename init.mp4 \
+		-hls_segment_filename "$OUT_DIR/audio/segment_%09d.m4s" \
+		-start_number "$START_NUMBER" "$OUT_DIR/audio/audio.m3u8" &
+
+	sleep 2
+	echo ">>> HTTP server: http://localhost:{{port}}/"
+	cd "$OUT_DIR" && python3 -m http.server {{port}}


rs/hang-cli/src/client.rs
```diff
@@ -18,32 +21,39 @@ pub async fn client<T: AsyncRead + Unpin>(
...
-	import
-		.init_from(input)
-		.await
-		.context("failed to initialize from media stream")?;
-
-	// Announce the broadcast as available once the catalog is ready.
 	origin.producer.publish_broadcast(&name, broadcast.consumer);
 
-	// Notify systemd that we're ready.
 	let _ = sd_notify::notify(true, &[sd_notify::NotifyState::Ready]);
```
kixelated, line 20: Can you try to call this only after `init_from`? It's nice because it means `systemctl start` will block until the media is initialized. Ideally the same for `publish_broadcast` but that doesn't matter as much.

rs/hang-cli/src/client.rs
```diff
@@ -18,32 +21,39 @@ pub async fn client<T: AsyncRead + Unpin>(
 	tracing::info!(%url, %name, "connecting");
 	let session = client.connect(url).await?;
 
-	// Create an origin producer to publish to the broadcast.
```
kixelated, line 21: my comments :O

rs/hang-cli/src/import.rs
```diff
@@ -41,3 +45,29 @@ impl Import {
 		self.inner.decode_from(input).await
 	}
 }
+
+pub struct ImportManifest {
+	inner: hang::import::manifest::ImportManifest,
+}
+
+impl ImportManifest {
```
kixelated: Is this wrapper needed? It seems to delegate (almost) everything.

rs/hang/src/import/manifest/mod.rs
```diff
@@ -0,0 +1,41 @@
...
+
+/// A generic interface for importing a manifest (playlist) into a hang broadcast.
+pub struct ImportManifest {
+	protocol: Protocol,
+}
+
+impl ImportManifest {
+	pub fn new(broadcast: BroadcastProducer, format: &str, url: Url, client: Client) -> Option<Self> {
+		let protocol = match format {
+			"hls" => {
```
kixelated: IMO this too generic. Just keep a separate HLS module so it's more type safe. We can make a separate DASH module later, no need for this wrapper.

In general we want to avoid switching based on `format: &str` in Rust, opting to use an enum or concrete types instead. It only exists in `Generic` because C doesn't have a great type system.

rs/hang/src/import/media/fmp4.rs
```diff
@@ -71,6 +71,37 @@ impl Fmp4 {
 		}
 	}
 
+	/// Create a new fMP4 importer that shares an existing catalog.
+	///
+	/// This allows multiple importers (e.g., for different HLS renditions) to
+	/// contribute to the same catalog track.
+	pub fn with_catalog(broadcast: hang::BroadcastProducer, catalog: CatalogProducer) -> Self {
```
kixelated: The new `hang::BroadcastProducer` contains `catalog: CatalogProducer`. You have two catalogs right now.

(yes the name is confusing because `moq_lite::BroadcastProducer` is something else)

rs/hang/src/import/media/fmp4.rs
```diff
@@ -71,6 +71,37 @@ impl Fmp4 {
...
+	}
+
+	/// Parse fMP4/CMAF bytes (init segment or media segment).
+	///
+	/// This is a convenience wrapper around `decode()` for use with byte slices.
+	pub fn parse(&mut self, bytes: &[u8]) -> anyhow::Result<()> {
+		use std::io::Cursor;
+		let mut cursor = Cursor::new(Bytes::copy_from_slice(bytes));
+		self.decode(&mut cursor)
+	}
```
kixelated: `copy_from_slice` will always allocate a copy on the heap. We want to avoid that if possible.

```rust
	pub fn parse(&mut self, mut bytes: &[u8]) -> anyhow::Result<()> {
		self.decode(&mut bytes)?;
		 // optional, but if it's not true we'd have to return `n` bytes read
		assert!(bytes.is_empty());
		Ok(())
	}
```

I know `bytes::Buf` is confusing. It takes a `&mut &[u8]` (kind of like a pointer to a pointer) because changes the size of the slice to indicate how much was read. I think this helper method is probably more confusing, and it'll be less efficient than `decode` when used with a `bytes::Bytes` (increments the ref count, doesn't allocate/copy).

rs/hang-cli/src/main.rs
```diff
@@ -63,6 +63,10 @@ pub enum Command {
 		/// The format of the input media.
 		#[arg(long, value_enum, default_value_t = ImportType::Cmaf)]
 		format: ImportType,
+
+		/// URL of an HLS playlist to ingest (for --format hls).
+		#[arg(long)]
+		hls_url: Option<Url>,
```
kixelated: Hmm, I think we should have a separate action. This is not actually optional when HLS is used, and I want to avoid hundreds of `--hls` prefixed flags in the future.

Ideally the CLI could support nesting, something like:
```
hang import hls --url ...
```

I can set it up if you would like. Clap can be confusing but it should be relatively simple. I think we could make a deprecated but backwards compatible `--format`.

rs/hang/src/import/mod.rs
```diff
@@ -1,9 +1,5 @@
...
-mod generic;
+pub mod manifest;
+pub mod media;
 
-pub use aac::*;
-pub use annexb::*;
-pub use fmp4::*;
-pub use generic::*;
+pub use manifest::*;
+pub use media::*;
```
kixelated: I think you remove this split.

rs/hang/src/import/media/fmp4.rs
```diff
@@ -71,6 +71,37 @@ impl Fmp4 {
...
+			tracks: HashMap::default(),
+			last_keyframe: HashMap::default(),
+			moov: None,
+			moof: None,
+			moof_size: 0,
+		}
+	}
+
+	/// Get the catalog producer for sharing with other importers.
+	pub fn catalog(&self) -> CatalogProducer {
```
kixelated: Not needed with `hang::BroadcastProducer`

rs/hang/src/import/media/mod.rs
```diff
@@ -15,7 +21,7 @@ enum Decoder {
 /// A generic interface for importing media into a hang broadcast.
 ///
 /// If you know the format in advance, use the specific decoder instead.
-pub struct Generic {
+pub struct ImportMedia {
```
kixelated: Keep it `Generic`, or maybe we should rename it to `Stream` or something to indicate that it needs a single stream of bytes. (unlike HLS)

rs/justfile
```diff
@@ -107,6 +107,9 @@ pub name url="http://localhost:4443/anon" *args:
 		-f mp4 -movflags cmaf+separate_moof+delay_moov+skip_trailer+frag_every_frame \
 		- | cargo run --bin hang -- publish --url "{{url}}" --name "{{name}}" {{args}}
 
+# Ingest a live HLS media playlist and publish it via hang (full ladder).
+ingest-hls url name="demo" relay="http://localhost:4443/anon":
```
kixelated: `pub-hls` IMO

rs/hang/src/import/manifest/hls.rs
```diff
@@ -0,0 +1,565 @@
...
+}
+
+impl HlsConfig {
+	pub fn new(playlist: Url) -> Self {
+		Self { playlist }
+	}
+}
+
+/// Result of a single ingest step.
+pub struct StepOutcome {
```
kixelated: no `pub`

rs/hang/src/import/manifest/hls.rs
```diff
@@ -0,0 +1,565 @@
...
+///
+/// Provides `init()` to prime the ingest with initial segments, and `service()`
+/// to run the continuous ingest loop.
+pub struct Hls {
+	/// Broadcast that all CMAF importers write into.
+	broadcast: BroadcastProducer,
+
+	/// Shared catalog handle used when multiple CMAF importers contribute to
+	/// the same `catalog.json` track.
+	shared_catalog: Option<CatalogProducer>,
```
kixelated: Not needed, use `self.broadcast.catalog`

rs/hang/src/import/manifest/hls.rs
```diff
@@ -0,0 +1,565 @@
...
+					audio = audio_url.as_deref().unwrap_or("none"),
+					"selected master playlist renditions"
+				);
+
+				return Ok(());
+			}
+		}
+
+		// Fallback: treat the provided URL as a single media playlist.
+		self.video.push(TrackState::new(MediaType::Video, self.cfg.playlist.clone()));
```
kixelated: Maybe try parsing it? And don't assume it's video?

rs/hang/src/import/manifest/hls.rs
```diff
@@ -0,0 +1,565 @@
...
+	async fn ensure_tracks(&mut self) -> Result<()> {
+		// Tracks already discovered.
+		if !self.video.is_empty() {
+			return Ok(());
+		}
+
+		let body = self.fetch_bytes(&self.cfg.playlist).await?;
+		if let Ok((_, master)) = m3u8_rs::parse_master_playlist(&body) {
+			let variants = select_variants(&master);
+			if !variants.is_empty() {
```
kixelated:
```suggestion
			if variants.is_empty() {
			    return Ok(()); // or throw an error
			}
```

removes an indentation level

rs/hang/src/import/manifest/hls.rs
```diff
@@ -0,0 +1,565 @@
...
+				// Create a video track state for every usable variant.
+				for variant in &variants {
+					let video_url = resolve_uri(&self.cfg.playlist, &variant.uri)?;
+					self.video.push(TrackState::new(MediaType::Video, video_url));
```
kixelated: Yeah we should check properly.

rs/hang/src/import/manifest/hls.rs
```diff
@@ -0,0 +1,565 @@
...
+				}
+
+				// Choose an audio rendition based on the first variant with an audio group.
+				if let Some(variant) = variants.iter().find(|v| v.audio.is_some()) {
+					if let Some(group_id) = variant.audio.as_deref() {
```
kixelated: Should be able to combine both of these. 

```rust
if let Some(group_id) = variants.iter().find_map(|v| v.audio.as_deref()) {
```

rs/hang/src/import/manifest/hls.rs
kixelated: In fact, do we even need `MediaType`? The fMP4 import will figure it out based on the init segment.

Unless it's valid to have an init segment that doesn't match the playlist. ie. there are multiple tracks in the init segment but we only want one of them.

rs/hang/src/import/hls.rs
```diff
@@ -0,0 +1,565 @@
...
+		kind: TrackKind,
+		track: &mut TrackState,
+		playlist: &MediaPlaylist,
+	) -> Result<usize> {
+		self.ensure_init_segment(kind, track, playlist).await?;
+
+		let mut consumed = 0usize;
+		let mut sequence = playlist.media_sequence;
+
+		for segment in &playlist.segments {
```
kixelated: a fun OVER-OPTIMIZATION, up to you

```rust
// skip the first N segments that we've already seen.
// might be off-by-one, too lazy to check
let skip = track.next_sequence.unwrap_or(0).saturating_sub(playlist.media_sequence);
for segment in playlist.segments.iter().skip(skip) {
	self.push_segment(kind, track, segment, sequence).await?;
	consumed += 1;
	sequence += 1;
}
```

rs/hang/src/import/media/fmp4.rs
kixelated: And yeah I read more, `fetch_bytes` returns `Bytes` so this `parse` function is indeed making an extra allocation/copy. Just call `decode` on the returned `Bytes` and it'll work like magic.

kixelated:
And yeah, I really love the reuse of the `fMP4` import. It feels almost like magic.

rs/hang-cli/src/client.rs
sebastianpiq: ups 🙃

rs/hang-cli/src/import.rs
sebastianpiq: You're right. Probably not... I wanted to abstract the generic 'media import' from the 'manifest import' but it doesn't really have any use. I will create a separate HLS module.

rs/hang/src/import/manifest/mod.rs
sebastianpiq: Noted! Same here about the hls module

rs/hang/src/import/media/mod.rs
sebastianpiq: Ok, I only renamed it to distinguish 'Media Generic' from 'Manifest Generic'. But if I'm creating a separate HLS module, it makes more sense to leave it as it was

kixelated: I just renamed it to `Decoder` because I agree, `Generic` is pretty bad.

kixelated:
(I would push straight to the PR branch but I don't have permission, it's an organization setting)

PR #764, opened 2025-12-11
fix: gate UI behind hang element

jdreetz:
This PR updates `PublishUIContextProvider` and `WatchUIContextProvider` to not render until the required element (`hang-publish` and `hang-watch` respectively) are available. This allows any components (including contexts) that use the elements to avoid having to check if they're defined before using them.


jdreetz:
> K I re-enabled it

thanks!

js/hang-ui/src/Components/publish/CameraSourceButton.tsx
```diff
@@ -5,23 +5,17 @@ import { PublishUIContext } from "./PublishUIContextProvider";
 export default function CameraSourceButton() {
 	const context = useContext(PublishUIContext);
 	const onClick = () => {
-		const hangPublishEl = context?.hangPublish();
-		if (!hangPublishEl) return;
-
-		if (hangPublishEl.source.peek() === "camera") {
+		if (context?.hangPublish.source.peek() === "camera") {
```
kixelated: Can we avoid `context` being `undefined`? Straight from the docs: https://docs.solidjs.com/reference/component-apis/use-context#recommended-usage

kixelated: And my minor preference would be to have each component take:

```typescript
export default function CameraSourceButton(props: { context: PublishUIContext }) {
```

Just to avoid this wrapper and the implicit way `useContext` works. But it doesn't really matter.

jdreetz: Updated everything to use the context "wrapper" as recommended in the docs. I'm hesitant to pass the context in as a prop, because then you have to prop drill it down from everywhere, right, which kind defeats the purpose of using a context I thought?

kixelated: Yeah, I just don't like the magic of global variables like `useContext` but whatever works 👍

PR #771, opened 2025-12-13
frame caching for rendering on pause

### LINKED ISSUE #765: Keep showing the last VideoFrame when paused
Feels better than a black screen.

gdavila:
- Added `#lastFrame` signal to cache the most recent frame
- When not paused: updates cache with new frames and renders them
- When paused: uses the cached frame for rendering
- Fixes #765

### DIFF
--- a/js/hang/src/watch/video/renderer.ts
+++ b/js/hang/src/watch/video/renderer.ts
@@ -46,7 +46,7 @@ export class Renderer {
 
 		// Only update if dimensions actually changed (setting canvas.width/height clears the canvas)
 		// TODO I thought the signals library would prevent this, but I'm too lazy to investigate.
-		if (canvas.width !== display.width && canvas.height !== display.height) {
+		if (canvas.width !== display.width || canvas.height !== display.height) {
 			canvas.width = display.width;
 			canvas.height = display.height;
 		}


js/hang/src/watch/video/renderer.ts
```diff
@@ -76,22 +78,33 @@ export class Renderer {
 		const ctx = effect.get(this.#ctx);
 		if (!ctx) return;
 
+		const currentFrame = effect.get(this.source.frame);
```
kixelated: We shouldn't call this when `paused`. Otherwise we're waking up on each new frame... in theory. In practice there won't be any new frames when paused, but it's pretty easy to fix.

```js
let frameToRender;

// NOTE: we always .clone() because we call .close() in cleanup().
// VideoFrames are ref counted so this is cheap
if (paused) {
  frameToRender = this.#lastFrame?.clone();
} else {
  frameToRender = effect.get(this.source.frame).clone();
  this.#lastFrame?.close();
  this.#lastFrame = frameToRender?.clone()
}
```

kixelated: And yeah `#lastFrame` shouldn't be a signal.

gdavila: I tested the suggested change, but I’m seeing an issue when moving effect.get(this.source.frame) inside the isPlaying condition.

With that change in place, the cached frame is never rendered so paused always show a black screen

From troubleshooting, I was able to confirm the following:
	•	#lastFrame is being cached correctly. When transitioning from playing → paused, it contains the latest frame rendered during playback.
	•	When the paused state is entered, #render() is still being invoked with the proper lastFrame - no black frame,  as expected.
	•	Despite that, no frame is actually drawn to the canvas or something else is drawn on top of it. 
	
Thoughts?

kixelated: Lemme clone and check it out myself real quick.

kixelated:
k so I figured out the issue. The canvas was getting cleared when the canvas was resized. I have no idea why the signal to resize the canvas was being refired, but eh it's good enough.

Issue #708, opened 2025-11-16
TLS certificate rotation


kixelated:
`moq-relay` supports loading TLS certificates from disk. This is cool, but certificates expire and need to be rotated every month or so.

It would be nice if there way a way to reload these certificates without restarting the process. I'm thinking either:

1. Reload the configuration on `SIGUSR1` or something. `systemctl reload moq-relay` could send it.
2. Watch the filesystem and reload certificates on change.

kixelated:
Also I can share my latest setup that uses Nix + systemd + certbot on Linode. Otherwise it'll be difficult to test.

einrobin:
Hey @kixelated I'd like to work on this feature, can you please assign the issue to me?

PR #774, opened 2025-12-14
Certificate reloading

einrobin:
Addresses #708 

The certificates will get refreshed when the SIGUSR1 signal is sent. This adds the possibility to configure ex. systemd reload for updating certificates.

Tested on Linux manually and in a Kubernetes container using cert-manager to rotate certificates and simply running `pkill -USR1 -f moq-relay`.

I also tried a lot around watching the files for changes (polling, inotify on linux) but I think it doesn't make much sense, so I removed it again. I found it was just too unpredictable, my personal use-case in Kubernetes would require watching the parent directory and not the certificate file itself, and I guess that wouldn't stay the only exception to this. Also having two different triggers for reloading seems not very intuitive to me.

### DIFF
--- a/rs/moq-native/src/server.rs
+++ b/rs/moq-native/src/server.rs
@@ -9,8 +9,6 @@ use rustls::sign::CertifiedKey;
 use std::fs;
 use std::io::{self, Cursor, Read};
 use std::sync::{Arc, RwLock};
-#[cfg(unix)]
-use tokio::signal::unix::{signal, SignalKind};
 use url::Url;
 use web_transport_quinn::{http, ServerError};
 
@@ -107,25 +105,7 @@ impl Server {
 		let certs = Arc::new(certs);
 
 		#[cfg(unix)]
-		{
-			let certs = certs.clone();
-			tokio::spawn(async move {
-				let tls_config = config.tls.clone();
-
-				match signal(SignalKind::user_defined1()) {
-					Ok(mut signal) => loop {
-						if signal.recv().await.is_some() {
-							tracing::info!("reloading server certificates");
-
-							if let Err(err) = certs.load_certs(&tls_config) {
-								tracing::warn!(%err, "failed to reload server certificates");
-							}
-						}
-					},
-					Err(err) => tracing::warn!(%err, "failed to setup server certificate reloading"),
-				}
-			});
-		}
+		tokio::spawn(Self::reload_certs(certs.clone(), config.tls.clone()));
 
 		let mut tls = rustls::ServerConfig::builder_with_provider(provider)
 			.with_protocol_versions(&[&rustls::version::TLS13])?
@@ -159,6 +139,22 @@ impl Server {
 			accept: Default::default(),
 			certs,
 		})
+	}
+
+	#[cfg(unix)]
+	async fn reload_certs(certs: Arc<ServeCerts>, tls_config: ServerTlsConfig) {
+		use tokio::signal::unix::{signal, SignalKind};
+
+		// Dunno why we wouldn't be allowed to listen for signals, but just in case.
+		let mut listener = signal(SignalKind::user_defined1()).expect("failed to listen for signals");
+
+		while listener.recv().await.is_some() {
+			tracing::info!("reloading server certificates");
+
+			if let Err(err) = certs.load_certs(&tls_config) {
+				tracing::warn!(%err, "failed to reload server certificates");
+			}
+		}
 	}
 
 	// Return the SHA256 fingerprints of all our certificates.
--- a/rs/moq-relay/src/web.rs
+++ b/rs/moq-relay/src/web.rs
@@ -24,8 +24,6 @@ use clap::Parser;
 use moq_lite::{OriginConsumer, OriginProducer};
 use serde::{Deserialize, Serialize};
 use std::future::Future;
-#[cfg(unix)]
-use tokio::signal::unix::{signal, SignalKind};
 use tower_http::cors::{Any, CorsLayer};
 
 use crate::{Auth, Cluster};
@@ -121,7 +119,7 @@ impl Web {
 			let config = hyper_serve::tls_rustls::RustlsConfig::from_pem_file(cert.clone(), key.clone()).await?;
 
 			#[cfg(unix)]
-			setup_reload(config.clone(), cert, key);
+			tokio::spawn(reload_certs(config.clone(), cert, key));
 
 			let server = hyper_serve::bind_rustls(listen, config);
 			Some(server.serve(app))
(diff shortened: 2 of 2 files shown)


rs/moq-relay/src/reload.rs
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	pub fn watch_changes<F>(&self, listener: F)
+	where
+		F: Fn() + Send + Sync + 'static,
+	{
+		self.listeners.lock().unwrap().push(Arc::new(listener));
+	}
+
+	pub fn start_background_task(self: Arc<Self>) {
```
kixelated: IMO call this automatically in `new()`.

rs/moq-relay/src/reload.rs
```diff
@@ -0,0 +1,68 @@
...
+		tokio::spawn(async move {
+			let clone = self.clone();
+			if let Err(err) = clone.start_watching().await {
+				tracing::warn!(%err, "failed to watch for configuration changes");
+			}
+		});
+	}
+
+	pub async fn start_watching(&self) -> anyhow::Result<()> {
```
kixelated: Should not be `pub`

rs/moq-relay/src/reload.rs
```diff
@@ -0,0 +1,68 @@
...
+		#[cfg(unix)]
```
kixelated: idk the whole file should be gated or something instead.

rs/moq-relay/src/reload.rs
```diff
@@ -0,0 +1,68 @@
...
+		let mut sigusr1 = signal(SignalKind::user_defined1())?;
+
+		#[cfg(unix)]
+		loop {
+			let reload = tokio::select! {
```
kixelated: You don't need a select for one branch. This is the same as:

```rs
sigusr1.recv().await;
```

rs/moq-relay/src/reload.rs
```diff
@@ -0,0 +1,68 @@
+use std::sync::{Arc, Mutex};
+#[cfg(unix)]
+use tokio::signal::unix::{signal, SignalKind};
+
+type ReloadListener = dyn Fn() + Send + Sync;
+
+pub struct ConfigReloader {
```
kixelated: I don't think you need this struct at all. It makes things more complicated IMO.

I woild replace it with:

```rs
#[cfg(unix)]
tokio::spawn(async move {
    let mut sigusr1 = tokio::signal::unix::signal(SignalKind::user_defined1())?;
    while let Some(_) = sigusr1.recv() {
       // contents of listener
    }
});
```

Even if we had multiple reloadable configurations, you could spawn a task for each. Tokio fans out signals.

rs/moq-native/src/server.rs
```diff
@@ -301,20 +338,20 @@ impl QuicRequest {
 
 #[derive(Debug)]
 struct ServeCerts {
-	certs: Vec<Arc<CertifiedKey>>,
+	certs: RwLock<Vec<Arc<CertifiedKey>>>,
```
kixelated: Why is this a `RwLock` if `ServerReloader` replaces the Arc, instead of modifying it inline?

rs/moq-native/src/server.rs
```diff
@@ -301,20 +338,20 @@ impl QuicRequest {
...
 		Self {
-			certs: Vec::new(),
+			certs: RwLock::new(Vec::new()),
 			provider,
 		}
 	}
 
-	// Load a certificate and corresponding key from a file
-	pub fn load(&mut self, chain: &PathBuf, key: &PathBuf) -> anyhow::Result<()> {
+	// Load a certificate and corresponding key from a file, but don't add it to the certs
```
kixelated, line 353: Why not? Why is this is `&mut self` if it doesn't even mutate `ServeCerts`?

rs/moq-native/src/server.rs
```diff
@@ -78,10 +79,36 @@ impl ServerConfig {
 	}
 }
 
+#[derive(Clone)]
+pub struct ServerReloader {
+	certs: Arc<ServeCerts>,
+}
+
+impl ServerReloader {
+	pub fn reload(&self, config: &ServerTlsConfig) -> anyhow::Result<()> {
```
kixelated: Another option is to add this method to `ServeCerts` instead, so it will load and replace without this `ServerReloader`. In that scenario, then you would need a RWMutex.

rs/moq-relay/src/main.rs
```diff
@@ -2,24 +2,46 @@ mod auth;
...
+		let tls_config = tls_config.clone();
+
+		reloader_arc.clone().watch_changes(move || {
+			let server_reloader = server_reloader.clone();
+			let tls_config = tls_config.clone();
+			if let Err(err) = server_reloader.reload(&tls_config) {
+				tracing::warn!(%err, "failed to reload server certificate");
+			}
+		});
+	}
```
kixelated: Yeah this should be moved inside of `server.init()`, and there's no need for `ConfigReload`. No point bubbling up config reloading this high in the stack.

rs/moq-relay/src/web.rs
```diff
@@ -118,10 +123,22 @@ impl Web {
 		};
 
 		let https = if let Some(listen) = self.config.https.listen {
-			let cert = self.config.https.cert.as_ref().expect("missing certificate");
-			let key = self.config.https.key.as_ref().expect("missing key");
-
-			let config = hyper_serve::tls_rustls::RustlsConfig::from_pem_file(cert, key).await?;
+			let cert = self.config.https.cert.clone().expect("missing https.cert");
+			let key = self.config.https.key.clone().expect("missing https.key");
+			let config = hyper_serve::tls_rustls::RustlsConfig::from_pem_file(cert.clone(), key.clone()).await?;
```
kixelated, line 119: Why are we calling clone twice?

rs/moq-relay/src/web.rs
```diff
@@ -118,10 +123,22 @@ impl Web {
...
+
+			let config_clone = config.clone();
+			self.reloader.watch_changes(move || {
+				let config_clone = config_clone.clone();
+				let cert = cert.clone();
+				let key = key.clone();
+
+				tokio::spawn(async move {
```
kixelated: And yeah, this is an example of why the `reloader` helper isn't all that helpful. This should be replaced with:

```rust
#[cfg(unix)]
tokio::spawn(async move {
    let mut sigusr1 = tokio::signal::unix::signal(SignalKind::user_defined1())?;
    while let Some(_) = sigusr1.recv() {
       if let Err(err) = config_clone.reload_from_pem_file(cert, key).await {
           tracing::warn!(%err, "failed to reload web certificate");
        }
    }
});
```

rs/moq-native/src/server.rs
einrobin: It was `&mut self` but I've removed that because on reload it first loads all the certificates/keys and then replaces them. This prevents that certificates get unloaded when there is an issue with the new certificates like a key that doesn't match the certificate.

kixelated:
Hit the ready for review button when you're ready. So AI SLOP can also review.

PR #778, opened 2025-12-15
libmoq-consume additions

davegullo:
Works in MacOS && Linux with:
https://github.com/davegullo/obs-hang


kixelated:
Going to merge this so you get credit, then immediately make some improvements lul.

Issue #763, opened 2025-12-10
Add support for "raw" container


kixelated:
Currently, we VarInt encode the PTS in the simple `hang` container. This is quite overkill and makes it slightly more difficult to build, as media layers need to implement QUIC VarInts now for a single value.

This is actually a good test for a new `container` enum we should add to the catalog. We should create a `container: "raw"` that uses a u64 while `container: "legacy"` uses a varint. In the (near) future we would add `fmp4` and eventually `loc`.

The first step would be to add decode support for a new [container](https://github.com/moq-dev/moq/blob/0788fe9d72028d96a221b6248a4810ca13865d69/js/hang/src/catalog/video.ts#L12) field. If the field is detected, we change [this line](https://github.com/moq-dev/moq/blob/0788fe9d72028d96a221b6248a4810ca13865d69/js/hang/src/frame.ts#L33) to read 8 bytes instead of a varint number of bytes. Same on the [Rust side](https://github.com/moq-dev/moq/blob/0788fe9d72028d96a221b6248a4810ca13865d69/rs/hang/src/model/group.rs#L68) although that is less immediately important (used by gstreamer).

The implementation should be generic enough that we can add multiple containers in the future.

gdavila:
@kixelated I could take a look at this. At least the JS part for now.

PR #784, opened 2025-12-17
feat: Add raw container format with fixed u64 timestamp encoding (JS)

gdavila:
Introduces support for a new "raw" container format that uses fixed 8-byte u64 encoding for frame timestamps. 

- Added container  field to VideoConfig and AudioConfig catalog schemas
- Created new container type system: "legacy" (VarInt), "raw" (u64), "fmp4" (future)
- Publishers (JS) default to "raw" Container format
- Subscribers fallback to "legacy" Container when field is absent (backward compatibility with Rust implementation)

Implements partially (js only)  #763

### DIFF
--- a/js/hang/src/publish/audio/encoder.ts
+++ b/js/hang/src/publish/audio/encoder.ts
@@ -27,6 +27,8 @@ export type EncoderProps = {
 	// The size of each group. Larger groups mean fewer drops but the viewer can fall further behind.
 	// NOTE: Each frame is always flushed to the network immediately.
 	maxLatency?: Time.Milli;
+
+	container?: Catalog.Container;
 };
 
 export class Encoder {
@@ -57,13 +59,13 @@ export class Encoder {
 
 	#signals = new Effect();
 
-	constructor(props?: EncoderProps, container?: Catalog.Container) {
+	constructor(props?: EncoderProps) {
 		this.source = Signal.from(props?.source);
 		this.enabled = Signal.from(props?.enabled ?? false);
 		this.muted = Signal.from(props?.muted ?? false);
 		this.volume = Signal.from(props?.volume ?? 1);
 		this.maxLatency = props?.maxLatency ?? (100 as Time.Milli); // Default is a group every 100ms
-		this.#container = container ?? DEFAULT_CONTAINER;
+		this.#container = props?.container ?? DEFAULT_CONTAINER;
 
 		this.#signals.effect(this.#runSource.bind(this));
 		this.#signals.effect(this.#runConfig.bind(this));
--- a/js/hang/src/publish/video/encoder.ts
+++ b/js/hang/src/publish/video/encoder.ts
@@ -10,6 +10,7 @@ import type { Source } from "./types";
 export interface EncoderProps {
 	enabled?: boolean | Signal<boolean>;
 	config?: EncoderConfig | Signal<EncoderConfig | undefined>;
+	container?: Catalog.Container;
 }
 
 // TODO support signals?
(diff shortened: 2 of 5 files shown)


js/hang/src/catalog/audio.ts
```diff
@@ -13,6 +14,9 @@ export const AudioConfigSchema = z.object({
 	// See: https://w3c.github.io/webcodecs/codec_registry.html
 	codec: z.string(),
 
+	// Container format for timestamp encoding
+	container: ContainerSchema.optional(),
```
kixelated: Is there a way to have this default to `DEFAULT_CONTAINER` instead? And remove `undefined` as an option.

js/hang/src/container/codec.ts
```diff
@@ -0,0 +1,155 @@
...
+}
+
+/**
+ * Gets the size in bytes of an encoded timestamp for the given container format.
+ * For variable-length formats, returns the maximum size.
+ *
+ * @param container - The container format
+ * @returns Size in bytes
+ */
+export function getTimestampSize(container: Catalog.Container = DEFAULT_CONTAINER): number {
```
kixelated, line 55: Is this used? I don't think it's useful.

js/hang/src/publish/audio/encoder.ts
```diff
@@ -122,11 +124,13 @@ export class Encoder {
 		const worklet = effect.get(this.#worklet);
 		if (!worklet) return;
 
+		const container = this.#container ?? "raw";
```
kixelated: We should default to "legacy" for now.

js/hang/src/publish/audio/encoder.ts
```diff
@@ -55,12 +56,13 @@ export class Encoder {
 
 	#signals = new Effect();
 
-	constructor(props?: EncoderProps) {
+	constructor(props?: EncoderProps, container?: Catalog.Container) {
```
kixelated: Can you add `container` to `EncoderProps`?

js/hang/src/publish/video/encoder.ts
```diff
@@ -57,11 +58,17 @@ export class Encoder {
 	// True when the encoder is actively serving a track.
 	active = new Signal<boolean>(false);
 
-	constructor(frame: Getter<VideoFrame | undefined>, source: Signal<Source | undefined>, props?: EncoderProps) {
+	constructor(
+		frame: Getter<VideoFrame | undefined>,
+		source: Signal<Source | undefined>,
+		props?: EncoderProps,
+		container?: Catalog.Container,
```
kixelated: Same deal, should be in `EncoderProps`

js/hang/src/container/codec.ts
gdavila: Yeah. true. Removed.

js/hang/src/catalog/audio.ts
gdavila: Done. The fallback to the default values were a mess in general. Hope it is clear now.

js/hang/src/publish/audio/encoder.ts
gdavila: Done. The fallback to the default values were a mess in general. Hope it is clear now.

js/hang/src/catalog/audio.ts
gdavila: FYI: DEFAULT_CONTAINER is now used elsewhere, and the previously hardcoded values have been removed from various places.

However at the moment, I’m overriding the default value in publish/element.ts  during the Broadcast class initialization, to ensure that new web publishers use raw by default.

js/hang/src/publish/audio/encoder.ts
gdavila: Done. I doubted initially  if container should be part of EncoderProps. It is unrealistic to set the container type for audio and video but I think it is ok for now :)

kixelated: IMO seems perfectly fine to use separate containers for audio/video. Raw makes a lot of sense for audio, while something else (or custom) could make sense for video.

PR #789, opened 2025-12-19
Add HLS import module

kixelated:
#761 But with merge conflicts fixed.

@sebastianpiq I changed the `pub-hls` so it's simpler (one ffmpeg command). I think there's still a lot of improvements that can be made but this is a good start; let's merge it now to prevent more conflicts.


kixelated:
Also I added support to load from disk instead of only over HTTP, so we don't need the python simple HTTP server.

PR #793, opened 2025-12-20
feat: add shared/button component

pzanella:
### Introduce shared UI components folder

This pull request creates the folder directory to organize common UI components. Centralizing shared elements improves code reuse, maintainability, and consistency across the user interface.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1273,6 +1273,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "135b12329e5e3ce057a9f972339ea52bc954fe1e9358ef27f95e89716fbc5424"
 
 [[package]]
+name = "humantime-serde"
+version = "1.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "57a3db5ea5923d99402c94e9feb261dc5ee9b4efa158b0315f788cf549cc200c"
+dependencies = [
+ "humantime",
+ "serde",
+]
+
+[[package]]
 name = "hyper"
 version = "1.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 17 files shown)


js/hang-ui/src/Components/watch/FullscreenButton.tsx
```diff
@@ -1,18 +1,58 @@
+import Button from "../shared/components/button";
+import type { IconSet } from "../shared/types/icons";
 import useWatchUIContext from "./useWatchUIContext";
 
+const fullscreenIcons: IconSet = {
```
kixelated: Could we make these files on disk instead of inlined Javascript?

Either the bundler could inline them (meh) or they could be packaged into a `public` folder that would need to be served by the application (cachable).

kixelated:
> Hi @kixelated, I tried to follow your suggestions. The style is loaded as `link` tag inside shadowRoot, and the icons are loaded as separate assets. Please let me know if this is what you intended or if you had another idea. I’m happy to adjust it if needed. Thank you!

Okay, my suggestion is probably wrong.

Yes, loading CSS and SVG out-of-band is going to result in a faster video load, but we're probably optimizing the wrong thing here. We really should be prioritizing the developer experience when it comes to this stock UI; nobody *should* actually use it in production.

So I think `?inline` is the right approach for `hang-ui`. I might have to do `setBasePath` for `hang` but that's a separate issue. Check out #819

kixelated:
> Hi @kixelated! Thank you for your feedback. Sorry, I think I misunderstood your suggestion. I will look at PR #819. If you agree, I can close this PR and work on the new one, ok? Let me know, thanks.

No no my suggestion was just wrong. The shoelace style could make sense if we're taking this super seriously, shipping the hang-ui in production, so the page can load milliseconds faster (by not blocking JS on SVG/CSS). We'll get to that bridge later?

PR #794, opened 2025-12-20
feat: iroh support

Frando:
This adds support for [iroh](https://github.com/n0-computer/iroh/) connections to `moq-native`, `moq-relay`, and `hang-cli`.
The integration was pretty straightforward overall, which IMO really validates the design of the moq crates. (The large number of added lines is mostly Cargo.lock).

## Demo

```sh


# Terminal 1: Start a relay server
just relay
# Copy the endpoint id printed at "iroh listening"

# Terminal 2: Publish via moq-lite over raw iroh QUIC
# (replace ENDPOINT_ID with the relay's endpoint id)
just pub bbb iroh://ENDPOINT_ID anon
# Alternatively you can use WebTransport over HTTP/3 over iroh,
# which allows to set a path prefix in the URL:
just pub bbb h3+iroh://ENDPOINT_ID/anon

# Terminal 3: Start web server
just web
````

Open [locahost:5173](http://localhost:5173) and see beautiful Big Buck Bunny playing in your browser. The browser plays from the relay via WebTransport as usual, but the source is connected to the relay via iroh.

I also updated the `hang serve` command to support serving via iroh. This repo does not include a native viewer but you can use the `watch` example from [iroh-live](https://github.com/n0-computer/iroh-live) to watch bbb published via iroh.

## Notes and open questions

* The iroh integration is feature-flagged and non-default
* This uses `web-transport-iroh`, which implements the web-transport traits for iroh connections. The crate  lives in the iroh-live repo](https://github.com/n0-computer/iroh-live/tree/main/web-transport-iroh) and is now published as well.
* `web-transport-iroh` supports both raw QUIC connections, and WebTransport over HTTP/3  over iroh. It matches `web-transport-quinn` in most regards. Quoting from the section the PR adds to README:

> When the iroh feature is enabled, you can connect to iroh endpoints with these URLs:
> 
> * `iroh://<ENDPOINT_ID>`: Connect via moq-lite over raw QUIC.
> * `moql+iroh://<ENDPOINT_ID>`: Connect via moq-lite over raw QUIC (same as above)
> * `moqt+iroh://<ENDPOINT_ID>`: Connect via IETF MoQ over raw QUIC
> * `h3+iroh://<ENDPOINT_ID>/optional/path?with=query`: Connect via WebTransport over HTTP/3.

* When the `iroh` feature is enabled, moq-native and downstream crates will include *two* quic implementations, because iroh uses `iroh-quinn`, not `quinn` directly. `iroh-quinn` is a fork of quinn with additional features (so far only minor ones, but its main branch and next iroh release will include full support for QUIC multipath). The public API for iroh-quinn is (nearly) identical to `quinn`, so it could be an option to switch regular moq from `quinn` to `iroh-quinn`, which would save quite some dependency cost when using both regular QUIC/WebTransport and iroh in the relay. tbd!

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,10 +191,10 @@ just relay --iroh-enabled
 # We set an `anon/` prefix to match the broadcast name the web ui expects
 # Because moq-lite does not have headers if using raw QUIC, only the hostname
 # in the URL can be used.
-just pub-iroh bbb iroh://ENDPOINT_ID anon/
+just pub bbb iroh://ENDPOINT_ID/anon --iroh-enabled
 # Alternatively you can use WebTransport over HTTP/3 over iroh,
 # which allows to set a path prefix in the URL:
-just pub-iroh bbb h3+iroh://ENDPOINT_ID/anon
+just pub bbb h3+iroh://ENDPOINT_ID/anon --iroh-enabled
 
 # Terminal 3: Start web server
 just web
--- a/justfile
+++ b/justfile
@@ -148,19 +148,7 @@ pub name url="http://localhost:4443/anon" prefix="" *args:
 	# Publish the media with the hang cli.
 	just ffmpeg-cmaf "dev/{{name}}.fmp4" |\
 	cargo run --bin hang -- \
-		publish --url "{{url}}" --name "{{prefix}}{{name}}" fmp4 {{args}}
-
-# Publish a video using ffmpeg to the localhost relay server
-# Enables iroh support so you can use an iroh:// URL of the relay
-pub-iroh name url="http://localhost:4443/anon" prefix="" *args:
-	# Download the sample media.
-	just download "{{name}}"
-	# Pre-build the binary so we don't queue media while compiling.
-	cargo build --bin hang
-	# Publish the media with the hang cli.
-	just ffmpeg-cmaf "dev/{{name}}.fmp4" |\
-	cargo run --bin hang -- \
-		--iroh-enabled publish --url "{{url}}" --name "{{prefix}}{{name}}" fmp4 {{args}}
+		publish --url "{{url}}" --name "{{prefix}}{{name}}" {{args}} fmp4
 
 # Generate and ingest an HLS stream from a video file.
 pub-hls name relay="http://localhost:4443/anon":

### CODE REVIEW

File: rs/hang-cli/src/client.rs [hunk 1]
```diff
@@ -1,14 +1,38 @@
...
+	#[cfg(feature = "iroh")] iroh: moq_native::iroh::EndpointConfig,
+	url: Url,
+	name: String,
+	publish: Publish,
+) -> anyhow::Result<()> {
 	tracing::info!(%url, %name, "connecting");
-	let session = client.connect(url).await?;
+	match url.scheme() {
+		#[cfg(feature = "iroh")]
+		"iroh" => {
```
Frando: URL format is tbd, as implemented you  pass `https://moq-relay.example.org` for a QUIC connection, and `iroh://<endpoint-id-in-hex>` for an iroh connection. The `iroh://` scheme is not well-defined elsewhere yet. Once I add h3 support to web-transport-iroh, it might make sense to support `iroh+moql://` for raw moq-lite over iroh (direct QUIC, no WebTransport/h3) and `iroh+h3://` for moq-lite wrapped in WebTransport over iroh (which would allow setting headers for auth on the request).

rs/moq-native/src/iroh/client.rs
```diff
@@ -0,0 +1,43 @@
...
+		// Connect to the server using the addr we just resolved.
+		let conn = self.endpoint.connect(addr, web_transport_iroh::ALPN.as_bytes()).await?;
+
+		// Connect with the connection we established.
+		Ok(web_transport_iroh::Session::raw(conn, url))
+	}
+
+	pub async fn close(&self) {
+		self.endpoint.close().await
+	}
```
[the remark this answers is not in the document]
Frando: Yeah, good catch, but switching from `&mut self` to `&self` puts a `Sync` bound on `self` which is not given due to a `FuturesUnordered` of boxed futures with are `Send` but not `Sync`. Could be solved with [sync-wrapper](https://docs.rs/sync_wrapper/latest/sync_wrapper/) instead though, but don't think this is worth the dependency rn.

justfile
```diff
@@ -128,23 +128,31 @@ download-url name:
...
-		-i "dev/{{name}}.fmp4" \
-		-c copy \
-		-f mp4 -movflags cmaf+separate_moof+delay_moov+skip_trailer+frag_every_frame \
-		- | cargo run --bin hang -- publish --url "{{url}}" --name "{{name}}" fmp4 {{args}}
+# Publish a video file using ffmpeg to a relay server over iroh
+# NOTE: The default url (iroh endpoint id) matches the secret key set in dev/relay.toml
+pub-iroh name='bbb' url='iroh://a73123fce41108f024a196a399edadbba8060be166c779aa50bf4731931492d3' *args:
+	cargo build --bin hang
+	# Publish the media with the hang cli.
+	just ffmpeg-cmaf "dev/{{name}}.fmp4" | cargo run --bin hang -- publish --url "{{url}}" --name "anon/{{name}}" fmp4 {{args}}
```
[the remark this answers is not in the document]
Frando: The AI bot is mistaken:

* The key does match. The key in the justfile is the matching pubkey for the secret key set in `dev/relay.toml`.
* The `anon/{{name}}` is correct. The other pub command uses WebTransport and  includes a path `/anon` in the URL. The `moq-relay` adds the URL path as prefix to all broadcasts for WebTransport connections. However for raw QUIC or iroh connections there's no HTTP requests involved, thus no way to send a path, which means there's never a prefix added. To match the broadcast path expected by the demo web ui, we publish on `anon/{{name}}` directly here.

dev/relay.toml
```diff
@@ -25,3 +25,11 @@ listen = "[::]:4443"
...
 # Allow anonymous access to everything.
 public = ""
+
+[iroh]
+# 32 byte secret key for the iroh endpoint. Defines the endpoint ID of the iroh endpoint.
+# The matching endpoint id is set in the justfile as a default for the `pub-iroh` command.
+# Can also be set via MOQ_IROH_SECRET env variable if you don't want to put secrets into the config file.
+# Alternatively, set `secret_key_path` to a path to load the secret key from a file.
+# If the file does not exist, a random key will be generated and written to the path.
+secret_key = "b724c6a37a004ae3d14c3efed84ccc145ee78f884e40f1b47063bdfd479ebed1"
```
kixelated: Should this be checked in?

rs/hang-cli/Cargo.toml
```diff
@@ -14,6 +14,10 @@ categories = ["multimedia", "network-programming", "web-programming"]
 [[bin]]
 name = "hang"
 path = "src/main.rs"
+
+[features]
+default = ["iroh"]
```
kixelated, line 19: How much extra compile time does this add roughly? I suppose it's probably not much if `rustls` is shared between quinn and the iroh fork.

I definitely think we should enable the feature for any pre-compiled binary, just not sure if it should be enabled by default.

rs/hang-cli/src/client.rs
```diff
@@ -1,14 +1,38 @@
...
-pub async fn client(config: moq_native::ClientConfig, url: Url, name: String, publish: Publish) -> anyhow::Result<()> {
-	let client = config.init()?;
-
+pub async fn client(
+	config: moq_native::ClientConfig,
+	#[cfg(feature = "iroh")] iroh: moq_native::iroh::EndpointConfig,
+	url: Url,
+	name: String,
+	publish: Publish,
+) -> anyhow::Result<()> {
```
kixelated: heh I think it's time we made a struct for this

rs/hang-cli/src/client.rs
kixelated: I think `iroh://` is fine because we'll have ALPN to detect if HTTP/3 and WebTransport is supported.

rs/hang-cli/src/main.rs
```diff
@@ -25,6 +25,16 @@ pub enum Command {
...
 
+		/// Optionally enable serving via iroh.
+		#[cfg(feature = "iroh")]
+		#[clap(long)]
+		iroh: bool,
+
+		/// Configuration for the iroh endpoint.
+		#[cfg(feature = "iroh")]
+		#[command(flatten)]
+		iroh_config: moq_native::iroh::EndpointConfig,
```
kixelated: Does this work in clap?

```rust
#[cfg(feature = "iroh")]
#[clap(long, flatten)]
iroh: Option<moq_native::iroh::EndpointConfig>,
```

I don't remember. Would be nice to avoid the boolean flag if possible. At the very least, clap should enforce one is not set without the other.

rs/moq-native/Cargo.toml
```diff
@@ -41,6 +42,7 @@ tokio = { workspace = true, features = ["full"] }
 tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 url = "2"
+web-transport-iroh = { git = "https://github.com/n0-computer/iroh-live", optional = true }
```
kixelated: publish it lul

Feel free to make a PR to `moq-dev/web-transport` if you want me to publish it instead.

rs/moq-native/src/iroh.rs
```diff
@@ -0,0 +1,85 @@
...
+	/// Path to a secret key for the iroh endpoint.
+	///
+	/// If the file doesn't exist, a random key will be generated and stored there.
+	#[arg(
+		id = "iroh-secret-key-path",
+		long = "iroh-secret-key-path",
+		env = "MOQ_IROH_SECRET_PATH",
+		conflicts_with = "iroh-secret-key"
+	)]
+	secret_key_path: Option<PathBuf>,
```
kixelated: Can you make this `pub`, so the struct can be manually constructed?

rs/hang-cli/Cargo.toml
Frando: Yeah I'll change it do a non-default feature.

dev/relay.toml
Frando: Actually it shouldn't because if it is different people doing dev will connect to each other haha. Will remove and add instructions instead.

Frando: Fixed

rs/hang-cli/Cargo.toml
Frando: Fixed

rs/moq-native/Cargo.toml
Frando: Published `web-transport-iroh@0.1.0`

rs/moq-native/src/iroh.rs
Frando: Fixed

rs/hang-cli/src/main.rs
Frando: It does, but would still not allow to optionally enable iroh support. All fields in `EndpointConfig` are optional, because we can legitimately create an iroh endpoint without any configuration. So if we want to make binding the iroh endpoint and serving via it optional, I fear a separate boolean option is the only way. We could say that the `serve` command *always* serves via iroh, too, if the `iroh` feature is enabled. However, I think having it configurable is better. 

I did not add the configuration for the `publish` command, because there the iroh endpoint will just be unused if not using an iroh URL. Ideally, parsing the URL could happen before even binding the endpoint, however that would need some more changes to moq-native which I wasn't sure about yet.

Frando:
I pushed a round of updates:

* I released `web-transport-iroh` and this PR no longer has git dependencies
* Refactored the iroh integration to be directly embedded into `moq_native::Client` and `moq_native::Server`. This made the code simpler, and followed the recent changes on `main` to have the client support different means of connecting.
* Respectively, added optional iroh configuration to `moq_native::ClientConfig` and `moq_native::ServerConfig`
* Fixed the relay configuration in the dev examples

Open questions:

* Overall, the feature-gating for iroh support turned out quite reasonable now IMO. One side effect is that `ClientConfig::init` and `ServerConfig::init` are now async, because binding an iroh endpoint is async (other than binding a quinn endpoint, which is sync). I think it's fine because those init methods are only ever called from async contexts in the codebase, but with the iroh feature disabled (the default) those `init` methods are now technically unused-async. I think it's OK. If others disagree, I'll find a way to change it but haven't found something nice-and-typesafe yet.

rs/moq-relay/src/connection.rs
```diff
@@ -12,18 +12,14 @@ pub struct Connection {
...
-					.query_pairs()
-					.find(|(k, _)| k == "jwt")
-					.map(|(_, v)| v.to_string());
+				let path = url.path();
+				let token = url.query_pairs().find(|(k, _)| k == "jwt").map(|(_, v)| v.to_string());
 				(path, token)
 			}
-			Request::Quic(_conn) => ("", None),
+			None => ("", None),
 		};
```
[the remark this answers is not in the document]
Frando: This is all unchanged from main.
(see the Request::Quic arm on main. Raw QUIC connections have no support for URLs, headers or tokens.)

README.md
```diff
@@ -164,6 +164,40 @@ just web      # Terminal 3: Start web server
...
+# which allows to set a path prefix in the URL:
+just pub bbb h3+iroh://ENDPOINT_ID/anon
+
+# Terminal 3: Start web server
+just web
+```
+
+Then open [localhost:5173](http://localhost:5173) and watch BBB, pushed from terminal 1 via iroh to the relay running in terminal 2, from where the browser fetches it over regular WebTransport.
+
+`just serve` serves a video via iroh alongside regular QUIC (it enables the `iroh` feature). This repo currently does not provide a native viewer, so you can't subscribe to it directly. However, you can use the [watch example from iroh-live](https://github.com/n0-computer/iroh-live/blob/main/iroh-live/examples/watch.rs) to view a video published via `hang-native` (which `just serve-iroh` expands to).
```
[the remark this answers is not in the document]
Frando: > Additionally, line 200 references just serve-iroh, but this target doesn't exist in the justfile—only just serve is available. Update this reference accordingly.

Fixed.

The rest is correct as-is, your justfile suggestions are wrong. Added more comments.

rs/hang-cli/src/client.rs
Frando: After the latest refactor there's no new args needed here anymore.

PR #803, opened 2025-12-23
crypto panic https web silently failing

davegullo:
FIX for:

thread 'tokio-runtime-worker' (8) panicked at /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rustls-0.23.35/src/crypto/mod.rs:249:14:

Could not automatically determine the process-level CryptoProvider from Rustls crate features.
Call CryptoProvider::install_default() before this point to select a provider manually, or make sure exactly one of the 'aws-lc-rs' and 'ring' features is enabled.


kixelated:
Fixed in #804

PR #807, opened 2025-12-27
Include sd-notify only on unix

einrobin:
Since sd-notify uses unix stuff, it's not possible to build it on anything that's not unix. This PR adds cfg(unix) where sd-notify is being used.

Tested the moq-relay on Debian 13 in a systemd service to verify it's still working.


kixelated:
I figured it was a no-op on other platforms, but this works too.

PR #808, opened 2025-12-29
moq-native: Add websocket fallback option

jhiatt-verkada:
This depends on a new crate introduced by https://github.com/moq-dev/web-transport/pull/138 .

### DIFF
--- a/rs/moq-clock/src/main.rs
+++ b/rs/moq-clock/src/main.rs
@@ -48,7 +48,7 @@ async fn main() -> anyhow::Result<()> {
 
 	tracing::info!(url = ?config.url, "connecting to server");
 
-	let session = client.connect(config.url).await?;
+	let session = client.connect_with_fallback(config.url).await?;
 
 	let track = Track {
 		name: config.track,


rs/moq-native/src/client.rs
```diff
@@ -135,10 +204,74 @@ impl Client {
...
+			transport,
+			websocket_delay: config.websocket.delay.map(Into::into),
+		})
+	}
+
+	pub async fn connect(&self, url: Url) -> anyhow::Result<web_transport_quinn::Session> {
+		self.connect_quic(url).await
+	}
+
+	pub async fn connect_with_fallback(&self, url: Url) -> anyhow::Result<WebTransportSessionAny> {
```
kixelated, line 219: I wonder if instead of `WebTransportSessionAny`, we could return an enum for `Quinn` or `WebSocket`. `moq_lite` is generic over both of these so it would be faster than doing an enum dispatch.

rs/moq-native/src/client.rs
```diff
@@ -30,6 +58,41 @@ pub struct ClientTls {
...
+pub struct ClientWebSocket {
+	/// Delay in milliseconds before attempting WebSocket fallback (default: 200)
+	/// If WebSocket won the previous race for a given server, this will be 0.
+	#[arg(
+		id = "websocket-delay",
+		long = "websocket-delay",
+		env = "MOQ_CLIENT_WEBSOCKET_DELAY",
+		default_value = "200ms"
+	)]
+	#[serde(deserialize_with = "deserialize_humantime")]
```
kixelated, line 72: https://docs.rs/humantime-serde/latest/humantime_serde/

Issue #734, opened 2025-12-04
JWK set support


einrobin:
It would be great if the relay had an option to load JWKs from a JWK Set.

```toml
[auth]
# Either provide a key from file
# key = "public.jwk"
# Or load keys from a URL
jwks = "https://<my service>/jwks.json"
```

JWK Sets are defined in [RFC7517 section 5](https://datatracker.ietf.org/doc/html/rfc7517#section-5).
A Set could look something like this: https://login.microsoftonline.com/common/discovery/v2.0/keys

The JWKs should be refreshed regularly, say every 24 hours. For verification it would just pick the correct JWK according to the `kid` field in the JWT. That way the service hosting the JWKs could easily rotate the keys without any downtime.

I guess it would also make sense to enforce https on the JWKs URL.

PR #809, opened 2025-12-31
JWK sets

einrobin:
Addresses #734 

Still a draft since I wanna do some more testing myself, add unit tests and also write some documentation on it.

The config looks something like this:
```toml
    [auth]
    jwks_uri = "https://my-website/keys.json"
    jwks_refresh_interval = 3600 # comment out to disable continous refreshing
    # key = "/app/public.jwk"
    public = "anon"
```

### DIFF
--- a/doc/guide/authentication.md
+++ b/doc/guide/authentication.md
@@ -128,6 +128,6 @@ Configure the relay:
 # public = "anon"                                               # Optional: allow anonymous access to anon/**
 
 key = "https://auth.example.com/keys.json"                      # JWK set URL for authenticated paths
-jwks_refresh_interval = 86400                                   # Optional: refresh the JWK set every N seconds, no refreshing if omitted
+refresh_interval = 86400                                   # Optional: refresh the JWK set every N seconds, no refreshing if omitted
 ```
 :::
--- a/rs/moq-relay/src/auth.rs
+++ b/rs/moq-relay/src/auth.rs
@@ -44,8 +44,8 @@ pub struct AuthConfig {
 	/// How often to refresh the JWK set (in seconds), will be ignored if the `key` is not a valid URI.
 	/// If not provided, there won't be any refreshing, the JWK set will only be loaded once at startup.
 	/// Minimum value: 30, defaults to None
-	#[arg(long = "jwks-refresh-interval", env = "MOQ_AUTH_JWKS_REFRESH_INTERVAL")]
-	pub jwks_refresh_interval: Option<u64>,
+	#[arg(long = "auth-refresh-interval", env = "MOQ_AUTH_REFRESH_INTERVAL")]
+	pub refresh_interval: Option<u64>,
 
 	/// The prefix that will be public for reading and writing.
 	/// If present, unauthorized users will be able to read and write to this prefix ONLY.
@@ -68,7 +68,7 @@ pub struct AuthToken {
 	pub cluster: bool,
 }
 
-const JWKS_REFRESH_ERROR_INTERVAL: Duration = Duration::from_mins(5);
+const REFRESH_ERROR_INTERVAL: Duration = Duration::from_mins(5);
 
 #[derive(Clone)]
 pub struct Auth {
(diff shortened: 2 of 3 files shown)


rs/moq-relay/src/auth.rs
```diff
@@ -36,9 +37,27 @@ impl axum::response::IntoResponse for AuthError {
 pub struct AuthConfig {
 	/// The root authentication key.
 	/// If present, all paths will require a token unless they are in the public list.
-	#[arg(long = "auth-key", env = "MOQ_AUTH_KEY")]
+	#[arg(long = "auth-key", env = "MOQ_AUTH_KEY", conflicts_with = "jwks_uri")]
```
kixelated: Could we detect if this starts with `http` instead? Seems cleaner than `conflicts_with`.

rs/moq-token/src/set.rs
```diff
@@ -0,0 +1,177 @@
+use crate::{Claims, Key, KeyOperation};
+use anyhow::Context;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use std::path::Path;
+use std::sync::{Arc, RwLock};
+
+pub trait KeyProvider {
```
kixelated: I don't think you need a `trait`. `Arc<Mutex<KeySet>>` should be good enough, with the (HTTP) background task periodically refreshing the keys.

rs/moq-token/src/set.rs
```diff
@@ -0,0 +1,177 @@
...
+	fn get_keys(&self) -> anyhow::Result<KeySet> {
+		Ok(self.clone())
+	}
+}
+
+/// JWK Set Loader that allows refreshing of a JWK Set
+#[cfg(feature = "jwks-loader")]
+pub struct KeySetLoader {
+	jwks_uri: String,
+	keys: RwLock<Option<KeySet>>,
```
kixelated: I don't like how this is `Option`. We should gate program start on the keys being loaded.

doc/guide/authentication.md
```diff
@@ -0,0 +1,136 @@
...
+      "kty": "OKP",
+      "crv": "Ed25519",
+      "x": "2FSK2q_o_d5ernBmNQLNMFxiA4-ypBSa4LsN30ZjUeU"
+    }
+  ]
+}
+```
+
+:::tip The following must be considered:
+- The endpoint must be HTTPS (unless you know what you're doing, then you can set `dangerously_allow_insecure_jwks = true` to allow HTTP)
```
kixelated: IMO let them configure HTTP if they want. It's probably some local service that provides the keys.

Flags like `dangerously_allow_insecure_jwks` are typically to disable HTTPS peer validation, not allow HTTP. I understand the extra scrutiny because these are auth keys, but eh it's a config file.

rs/moq-relay/src/auth.rs
```diff
@@ -36,9 +37,27 @@ impl axum::response::IntoResponse for AuthError {
...
+
+	#[arg(
+		long = "dangerously-allow-insecure-jwks",
+		env = "MOQ_AUTH_DANGEROUSLY_ALLOW_INSECURE_JWKS",
+		conflicts_with = "key"
+	)]
+	pub dangerously_allow_insecure_jwks: Option<bool>,
+
+	/// How often to refresh the JWK set (in seconds), if not provided the JWKs won't be refreshed.
+	/// Minimum value: 30
```
kixelated: What's the default? Nothing?

rs/moq-token/src/set.rs
```diff
@@ -0,0 +1,177 @@
...
+
+		key.decode(token)
+	}
+}
+
+/// JWK Set to spec https://datatracker.ietf.org/doc/html/rfc7517#section-5
+#[derive(Default, Clone)]
+pub struct KeySet {
+	/// Vec of an arbitrary number of Json Web Keys
+	pub keys: Vec<Arc<Key>>,
```
kixelated, line 12: `HashMap` based on `kid`?

rs/moq-relay/src/auth.rs
```diff
@@ -173,6 +283,9 @@ mod tests {
 		// Test anonymous access to /anon path
 		let auth = Auth::new(AuthConfig {
 			key: None,
+			jwks_uri: None,
```
kixelated: `#[derive(Default)]` so you can:

```rs
let auth = Auth::new(AuthConfig {
  public: Some("anon".to_string()),
  ..Default::default()
});

File: rs/moq-relay/src/connection.rs [hunk 8]
```diff
@@ -1,12 +1,13 @@
...
+use std::sync::Arc;
 
 use moq_native::Request;
 
 pub struct Connection {
 	pub id: u64,
 	pub request: Request,
 	pub cluster: Cluster,
-	pub auth: Auth,
+	pub auth: Arc<Auth>,
```
kixelated: I generally make classes `Clone` for aesthetic reasons, keeping `Arc` an internal implementation detail, but this works too.

rs/moq-token/src/set.rs
einrobin: Not sure about that, I'd keep it this way especially since the `kid` in a Key is optional and doesn't necessarily have to be unique. I don't think this matters for performance since we're probably talking about < 10 keys.

rs/moq-relay/src/auth.rs
```diff
@@ -60,15 +68,119 @@ pub struct AuthToken {
 	pub cluster: bool,
 }
 
+const JWKS_REFRESH_ERROR_INTERVAL: Duration = Duration::from_mins(5);
```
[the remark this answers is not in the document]
einrobin: Well it certainly does compile

PR #812, opened 2026-01-02
moq-native: WebSocket fallback with explicit enum

jhiatt-verkada:
This is an alternate approach to #808 . In this version the return type of `connect_with_fallback()` is an enum with `Quinn` + `WebSocket` variants, and it's up to the caller to explicitly dispatch to `moq_lite` over the different types.

### Advantages
* Overall fewer lines of new code required
* No need to synchronize with changes in other repository 
* Better performance, as other version introduced indirection in trait implementation

### Disadvantages
* `moq_native::Client::connect_with_fallback()` is not a simple replacement for `moq_native::Client::connect()` as the resulting `moq_lite::Session` instances are distinct types
* Future modifications to fallback logic (e.g. introducing another possible transport) could require changes at the callsites

### DIFF
--- a/rs/moq-native/src/client.rs
+++ b/rs/moq-native/src/client.rs
@@ -7,6 +7,7 @@ use std::path::PathBuf;
 use std::sync::{LazyLock, Mutex};
 use std::{fs, io, net, sync::Arc, time};
 use url::Url;
+use web_transport_ws::{tokio_tungstenite, tungstenite};
 
 // Track servers (hostname:port) where WebSocket won the race, so we won't give QUIC a headstart next time
 static WEBSOCKET_WON: LazyLock<Mutex<HashSet<(String, u16)>>> = LazyLock::new(|| Mutex::new(HashSet::new()));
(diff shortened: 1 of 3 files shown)


rs/moq-native/src/client.rs
```diff
@@ -30,6 +57,23 @@ pub struct ClientTls {
...
+	/// Delay in milliseconds before attempting WebSocket fallback (default: 200)
+	/// If WebSocket won the previous race for a given server, this will be 0.
+	#[arg(
+		id = "websocket-delay",
+		long = "websocket-delay",
+		env = "MOQ_CLIENT_WEBSOCKET_DELAY",
+		default_value = "200ms",
+		value_parser = humantime::parse_duration,
+	)]
+	#[serde(with = "humantime_serde")]
```
jhiatt-verkada, line 51: I'd initially avoided this crate because it's not clear that it's still maintained. It hasn't been updated in about 4 years so I wasn't sure if people would look askance at taking that dependency. However it does save a few lines of code here and in general I think `humantime` has nice ergonomics.

rs/moq-clock/src/main.rs
```diff
@@ -102,11 +99,32 @@ async fn main() -> anyhow::Result<()> {
...
+	match session {
+		moq_native::WebTransportSessionAny::Quinn(session) => {
+			let session = moq_lite::Session::connect(session, publish, subscribe).await?;
+			session.closed().await
+		}
+		moq_native::WebTransportSessionAny::WebSocket(session) => {
+			let session = moq_lite::Session::connect(session, publish, subscribe).await?;
+			session.closed().await
+		}
+	}
```
jhiatt-verkada: Here's where we need to explicitly handle the possible transport implementations.

kixelated:
@jhiatt-verkada Let me know if this doesn't work for you, thanks for the PR!

Issue #740, opened 2025-12-05
OBS: Hook up the remaining codecs


kixelated:
We added support for h.264 and AAC imports in the `hang-c` library. The codebase supports AV1, H.265, OPUS, VP9, etc however they need to be hooked up and tested. 

Go forth, get OBS running, and make sure we can parse/playback these other codecs!

PR #813, opened 2026-01-04
hev1 decoder

mcintyrehh:
Related to: https://github.com/moq-dev/moq/issues/740
MoQ OBS Plugin update: https://github.com/moq-dev/obs/pull/19

This PR adds an HEVC `hev1` decoder, following the format of the `avc1` import. The decoders are largely identical with the exception of:
* `init` - the avc3 decoder uses the h264_parser crate to handle parsing the SPS Nal unit, this PR adds a new `scuffle_h265` crate to the same end. A new method `get_vui_data` pulls out `framerate` and `display_ratio_width/height` from the VUI Parameters of the SPS. Framerate ends up getting passed through to the catalog! The FFmpeg HEVC encoder in OBS doesn't pass through an `aspect_ratio_idc`, so I haven't been able to confirm the display ratio fields. I am the babiest of Crab-lings so shout out to Claude for the Rust syntax help here!
* `decode_nal` - similar to the avc3 parser. Ignores the second NAL header byte containing `nuh_layer_id` and `nuh_temporal_id_plus1`. Not sure if there is anything we need to do with these?
  * matches NAL type by Non-VCL (SPS, then AUD/PPS/SEIs), then keyframe containing slices, then finally non-keyframe slices 
 
 HEVC with framerate and Opus! :exploding_head: 
Screenshot from 2026-01-04 00-01-51

Other changes:
* moved the the `NalIterator` and its child functions (`after_start_code`, `find_start_code`) into a new `annexb.rs` module and added a new `test_nal_iterator_realistic_h265`

### DIFF
--- a/rs/hang/src/import/decoder.rs
+++ b/rs/hang/src/import/decoder.rs
@@ -31,10 +31,6 @@ impl FromStr for DecoderFormat {
 				Ok(DecoderFormat::Avc3)
 			}
 			"hev1" => Ok(DecoderFormat::Hev1),
-			"h265" | "hevc" => {
-				tracing::warn!("format '{s}' is deprecated, use 'hev1' instead");
-				Ok(DecoderFormat::Hev1)
-			}
 			"fmp4" | "cmaf" => Ok(DecoderFormat::Fmp4),
 			"aac" => Ok(DecoderFormat::Aac),
 			"opus" => Ok(DecoderFormat::Opus),
--- a/rs/hang/src/import/hev1.rs
+++ b/rs/hang/src/import/hev1.rs
@@ -5,7 +5,7 @@ use anyhow::Context;
 use buf_list::BufList;
 use bytes::{Buf, Bytes};
 use moq_lite as moq;
-use scuffle_h265::SpsNALUnit;
+use scuffle_h265::{NALUnitType, SpsNALUnit};
 
 /// A decoder for H.265 with inline SPS/PPS.
 /// Only supports single layer streams, ignores VPS.
@@ -41,7 +41,7 @@ impl Hev1 {
 
 	fn init(&mut self, sps: &SpsNALUnit) -> anyhow::Result<()> {
 		let profile = &sps.rbsp.profile_tier_level.general_profile;
-		let vui_data = get_vui_data(&sps.rbsp.vui_parameters);
+		let vui_data = sps.rbsp.vui_parameters.as_ref().map(VuiData::new).unwrap_or_default();
 
 		let config = hang::catalog::VideoConfig {
 			coded_width: Some(sps.rbsp.cropped_width() as u32),
@@ -179,10 +179,10 @@ impl Hev1 {
 
 		// Bits 1-6: nal_unit_type
 		let nal_unit_type = (header >> 1) & 0b111111;
-		let nal_type = HevcNalType::try_from(nal_unit_type).ok();
+		let nal_type = NALUnitType::from(nal_unit_type);
 
 		match nal_type {
-			Some(HevcNalType::Sps) => {
+			NALUnitType::SpsNut => {
 				self.maybe_start_frame(pts)?;
 
 				// Try to reinitialize the track if the SPS has changed.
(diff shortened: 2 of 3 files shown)


rs/hang/src/import/hev1.rs
```diff
@@ -0,0 +1,396 @@
...
+		let nal_type = HevcNalType::try_from(nal_unit_type).ok();
+
+		match nal_type {
+			Some(HevcNalType::Sps) => {
+				self.maybe_start_frame(pts)?;
+
+				// Try to reinitialize the track if the SPS has changed.
+				let sps = SpsNALUnit::parse(&mut &nal[..]).context("failed to parse SPS NAL unit")?;
+				self.init(&sps)?;
+			}
```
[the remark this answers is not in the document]
mcintyrehh: This is not correct, SpsNALUnit::parse() calls NALUnitHeader::parse() which validates the SPS unit type:

```rust
impl SpsNALUnit {
    /// Parses an SPS NAL unit from the given reader.
    pub fn parse(mut reader: impl io::Read) -> io::Result<Self> {
        let nal_unit_header = NALUnitHeader::parse(&mut reader)?;
        if nal_unit_header.nal_unit_type != NALUnitType::SpsNut {
            return Err(io::Error::new(io::ErrorKind::InvalidData, "nal_unit_type is not SPS_NUT"));
        }

        let rbsp = SpsRbsp::parse(reader, nal_unit_header.nuh_layer_id)?;

        Ok(SpsNALUnit { nal_unit_header, rbsp })
    }
}
```

rs/hang/src/import/hev1.rs
```diff
@@ -0,0 +1,396 @@
...
+	Vps = 32,
+	Sps = 33,
+	Pps = 34,
+	Aud = 35,
+	EndOfSequence = 36,
+	EndOfBitstream = 37,
+	Filler = 38,
+	SeiPrefix = 39,
+	SeiSuffix = 40,
+} // ITU H.265 V10 Table 7-1 – NAL unit type codes and NAL unit type classes
```
mcintyrehh: Alternatively, we could use the scuffle [NALUnitType](https://docs.rs/scuffle-h265/0.2.2/scuffle_h265/struct.NALUnitType.html) enum.

rs/hang/src/import/hev1.rs
```diff
@@ -0,0 +1,396 @@
...
+		scuffle_h265::AspectRatioIdc::Aspect15_11 => Some((15, 11)),
+		scuffle_h265::AspectRatioIdc::Aspect64_33 => Some((64, 33)),
+		scuffle_h265::AspectRatioIdc::Aspect160_99 => Some((160, 99)),
+		scuffle_h265::AspectRatioIdc::Aspect4_3 => Some((4, 3)),
+		scuffle_h265::AspectRatioIdc::Aspect3_2 => Some((3, 2)),
+		scuffle_h265::AspectRatioIdc::Aspect2_1 => Some((2, 1)),
+		scuffle_h265::AspectRatioIdc::ExtendedSar => None,
+		_ => None, // Reserved
+	}
+}
```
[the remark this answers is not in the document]
mcintyrehh: Good call Robo Rabbit, will add!

rs/hang/src/import/decoder.rs
```diff
@@ -31,10 +31,6 @@ impl FromStr for DecoderFormat {
 				Ok(DecoderFormat::Avc3)
 			}
 			"hev1" => Ok(DecoderFormat::Hev1),
-			"h265" | "hevc" => {
-				tracing::warn!("format '{s}' is deprecated, use 'hev1' instead");
-				Ok(DecoderFormat::Hev1)
-			}
```
mcintyrehh, line 37: @kixelated the format that comes through the C api is `"hevc"`, so OBS bonks without this. Should we add this back, or is it better to fix upstream?

Image

kixelated: Hmm so it depends on how pedantic we should be.

"hevc" would involve the SPS/PPS/VPS in initialize, which I think OBS does. "hev1" would mean the SPS/PPS/VPS occurs before each keyframe, which I think OBS also does. So it supports both modes?

But since we're not populating the `description` (correct me if I'm wrong, on my phone), the current import code only supports `hev1`. We could absolutely add support for `hevc`, but that *should* also involve removing the SPS/PPS/VPS from before each i-frame.

I think the "correct" behavior is to have the OBS plugin change the format string from "hevc" to "hev1". Same for "h264" to "avc3" so we stop logging warnings.

kixelated: and we should technically not support/call `initialize` when using hev1 or avc3, as the entire point that it's not needed. In fact we might currently be writing the SPS/PPS/VPS twice for the first i-frame?

mcintyrehh: The buffer passed into `initialize` contains a VPS and an SPS unit,  but we're only sending VPS/SPS/PPS once before the first (and subsequent) IDR frames. So everything Should Just Work ™️ if we change the OBS plugin to send `"hev1"`, but if we know it is `avc3`/`hev1` we could just skip initialize and jump straight into `decode_stream`.

Image

Image

kixelated: > The buffer passed into `initialize` contains a VPS and an SPS unit, but we're only sending VPS/SPS/PPS once before the first (and subsequent) IDR frames. So everything Should Just Work ™️ if we change the OBS plugin to send `"hev1"`, but if we know it is `avc3`/`hev1` we could just skip initialize and jump straight into `decode_stream`.
> 
> Image Image

Yeah technically, `hevc` should not contain SPS/PPS/VPS before each keyframe. Of course, the decoder will ignore the redundant information but (*gasp*)  OBS will write an extra 100 bytes or so before each keyframe. It obviously doesn't matter but for pedantic reasons, we should have OBS report `hev1` until we can properly strip out the SPS/PPS/VPS.

kixelated: But I'm also fine supporting a `h265` mode that switches between `hevc` and `hvc1` based on the context.

PR #814, opened 2026-01-04
feat(hang-demo): add connection settings form with broadcast auto-dis…

### LINKED ISSUE #767: Text input for broadcast URL/path
Currently, the `hang-demo` can be launched with a `RELAY_URL` via `just web`. Then you can use the `?path` query parameter to select individual broadcasts.

That's kind of lame. We should have a simple form (below the fold) so it's easy to target arbitrary relays, broadcast names, etc. Bonus points if the broadcast names are auto-discovered via MoQ (drop-down instead of input text), so `just pub tos` would show both `bbb` and `tos` as valid options.

bas0N:
Add connection settings form with broadcast auto-discovery

Closes #767

Summary

Adds a collapsible "Connection Settings" panel to the watch demo that allows users to:

- Enter a custom relay URL without restarting the dev server
- Discover available broadcasts via MoQ SUBSCRIBE_NAMESPACE
- Select broadcasts from a dropdown instead of manually editing URL query params
- Share links with ?url=...&path=... query parameters

Demo

https://github.com/user-attachments/assets/049921a5-0f83-4358-a49a-2869cbd427a2

Changes

- js/hang-demo/src/index.html - Added collapsible settings form
- js/hang-demo/src/index.css - Styled form to match moq.dev theme
- js/hang-demo/src/index.ts - Added discovery logic using @moq/lite

### DIFF
--- a/js/hang-demo/src/config.ts
+++ b/js/hang-demo/src/config.ts
@@ -110,10 +110,12 @@ export default class HangConfig extends HTMLElement {
 	}
 
 	#dispatchChange() {
-		this.dispatchEvent(new CustomEvent("change", {
-			detail: { url: this.url, path: this.path },
-			bubbles: true,
-		}));
+		this.dispatchEvent(
+			new CustomEvent("change", {
+				detail: { url: this.url, path: this.path },
+				bubbles: true,
+			}),
+		);
 	}
 
 	#scheduleDiscovery() {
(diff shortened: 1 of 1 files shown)


kixelated:
Thanks for the video.

Could you remove the collapse functionality? Always show the server URL and broadcast name as input fields. The name needs to be an input field because some MoQ relays don't support announce yet.

Can you also auto-discover broadcasts? Clicking on them will populate the input box.

And finally, could you move this code into a separate file? I want index.ts and index.css to be very simple because this is part demo, part hello world. It might be cool as a web component too?

bas0N:
@kixelated some fixed were required, ready to check now

kixelated:
I think it's a cool improvement. @pzanella and @jdreetz do we want to throw this into hang-ui?

pzanella:
Cool improvement, @bas0N!

@kixelated @jdreetz
I think we could add a settings button to the player UI. It could open an overlay (like the stats component) where we can set connection settings and other options. What do you think?

PR #819, opened 2026-01-07
feat: add shared/button components but inline

kixelated:
@pzanella 

I went with vite (library mode) instead of rollup because it does a great job. It uses rollup under the hood of course but automatically handles stuff like `?inline`

TODO:
- We should type the Icons and allow them to be tree-shaken.
- Figure out NAMING. I renamed some stuff because capital C `Components` annoys me, but you guys decide.
- Actually test a built package.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,10 +3,8 @@
 *.local
 *.log*
 
-# Claude Code config (only allow at repository root)
+# Ignore local Claude Code config
 *.local.json
-*/**/.claude/
-!/.claude/settings.json
 
 # IDE
 .idea


js/hang-ui/src/Components/shared/button.tsx
```diff
@@ -0,0 +1,55 @@
...
+	type?: "button" | "submit" | "reset";
+	title?: string;
+	onClick?: () => void;
+	class?: string;
+	children: JSX.Element | string;
+	ariaLabel?: string;
+	ariaDisabled?: boolean;
+	disabled?: boolean;
+	tabIndex?: number;
+};
```
jdreetz, line 26: anyway we can re-use the built-in types?

like
```
type NativeButton = JSX.IntrinsicElements['button'];

export type ButtonProps = {
	type?: NativeButton['type'];
	title?: NativeButton['title'];
	onClick?: NativeButton['onClick'];
	class?: NativeButton['class'];
	children: JSX.Element | string;
	ariaLabel?: NativeButton['aria-label'];
...
```

pzanella:
Hi @kixelated! Sounds great! Yes, I know ViteJS and I love it! 😊

@kixelated  @jdreetz 
I’ll work on the open points and give you an update as soon as I can, ok?
About the Components folder, I’m in the middle and agree with both of you. I was thinking maybe a feature-first structure or a simplified Atomic Design, but the second might be too much for now. What do you think? Thanks!

kixelated:
K I reverted removing the `Components` folder. It's just very Java, having a folder that starts with a Capital and one contains only one nested folder. But my fault, removing it should be in a different PR anyway so it's easier to review.

I can also revert the files renamed to `index.tsx` too if you'd like. Same with changing the import path to no longer contain `element`, that one really depends on if we ever have plans of releasing a non-web-component UI.

jdreetz:
@kixelated can you elaborate on what you're looking for with this TODO - `Actually test a built package.`? You don't mean something like [Storybook](https://storybook.js.org/), do you?

PR #822, opened 2026-01-09
Cmaf Native Support 

jpbusta10:
## Add passthrough mode for CMAF/HLS import and MSE playback support

This PR adds comprehensive CMAF support with passthrough mode for ingestion and Media Source Extensions (MSE) for browser playback.

#### Rust/CLI Side
- Added `--passthrough` flag to `hang publish hls` command
- Updated `just pub-hls` command to support passthrough mode via optional parameter
- Passthrough mode transports complete CMAF fragments (moof+mdat) without decomposition, improving efficiency for certain use cases

#### TypeScript/Browser Side
- **New MSE-based playback**: Implemented `SourceMSE` class for both audio and video tracks
- **Automatic routing**: When container type is `cmaf`, playback automatically uses MSE instead of WebCodecs
- **Fragment handling**: Properly handles CMAF fragments with init segment (moov) detection and GOP grouping
- **Queue management**: Implements fragment queuing with size limits to prevent memory issues in live streaming
- **Fallback support**: Gracefully falls back to WebCodecs if MSE initialization fails
- **Type safety improvements**: Added proper typing for `setAudioSync` method, eliminating `any` type usage

### Technical Details

**Passthrough Mode:**
The passthrough flag enables transport of complete CMAF fragments without re-encoding or decomposing them. This is particularly useful when you want to preserve the original CMAF structure.

**Container Format:**
- `native`: Uses QUIC VarInt encoding (1-8 bytes, variable length), raw frame payloads (default)
- `raw`: Uses fixed u64 encoding (8 bytes, big-endian), raw frame payloads  
- `cmaf`: Fragmented MP4 container - frames contain complete moof+mdat fragments

**MSE Implementation:**
- Uses browser's native `MediaSource` API for CMAF playback
- Handles init segments (moov atoms) separately from media fragments
- Groups fragments by MOQ group before appending to SourceBuffer
- Implements proper buffering and playback state management
- Exposes audio/video elements for direct control (volume, mute, etc.)

### Usage

# Publish HLS with passthrough mode
just pub-hls bbb passthrough 

# Publish HLS without passthrough (default)
just pub-hls bbb 

The browser will automatically use MSE for playback when the container is `cmaf`, providing more efficient decoding and lower latency compared to WebCodecs for CMAF content.

### DIFF
--- a/js/hang/src/util/mp4-mime.ts
+++ b/js/hang/src/util/mp4-mime.ts
@@ -1,13 +1,13 @@
 import type * as Catalog from "../catalog";
 
 /**
- * Builds a MIME type string for MediaSource from a codec string.
+ * Builds an MP4 MIME type string for MediaSource from a codec string.
  *
  * @param codec - The codec string from the catalog (e.g., "avc1.42E01E", "mp4a.40.2")
  * @param type - "video" or "audio"
- * @returns MIME type string (e.g., "video/mp4; codecs=\"avc1.42E01E\"")
+ * @returns MP4 MIME type string (e.g., "video/mp4; codecs=\"avc1.42E01E\"")
  */
-export function buildMimeType(codec: string, type: "video" | "audio"): string {
+function buildMp4MimeType(codec: string, type: "video" | "audio"): string {
 	// For MP4 containers, we use the standard MIME type format
 	// Most codecs are already in the correct format for MSE
 	return `${type}/mp4; codecs="${codec}"`;
@@ -24,13 +24,13 @@ export function isMimeTypeSupported(mimeType: string): boolean {
 }
 
 /**
- * Builds and validates a MIME type for video from catalog config.
+ * Builds and validates an MP4 MIME type for video from catalog config.
  *
  * @param config - Video configuration from catalog
- * @returns MIME type string or undefined if not supported
+ * @returns MP4 MIME type string or undefined if not supported
  */
-export function buildVideoMimeType(config: Catalog.VideoConfig): string | undefined {
-	const mimeType = buildMimeType(config.codec, "video");
+export function buildMp4VideoMimeType(config: Catalog.VideoConfig): string | undefined {
+	const mimeType = buildMp4MimeType(config.codec, "video");
 	if (isMimeTypeSupported(mimeType)) {
 		return mimeType;
 	}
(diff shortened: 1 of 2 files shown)


js/hang/src/container/codec.ts
```diff
@@ -11,12 +11,14 @@ import type * as Time from "../time";
...
 export function encodeTimestamp(timestamp: Time.Micro, container: Catalog.Container = DEFAULT_CONTAINER): Uint8Array {
 	switch (container) {
-		case "legacy":
+		case "native":
 			return encodeVarInt(timestamp);
 		case "raw":
 			return encodeU64(timestamp);
-		case "fmp4":
-			throw new Error("fmp4 container not yet implemented");
+		case "cmaf": {
```
kixelated, line 18: We shouldn't be encoding the timestamp here. The entire frame payload should be a fMP4 fragment.

js/hang/src/util/mime.ts
```diff
@@ -0,0 +1,52 @@
+import type * as Catalog from "../catalog";
+
+/**
+ * Builds a MIME type string for MediaSource from a codec string.
+ *
+ * @param codec - The codec string from the catalog (e.g., "avc1.42E01E", "mp4a.40.2")
+ * @param type - "video" or "audio"
+ * @returns MIME type string (e.g., "video/mp4; codecs=\"avc1.42E01E\"")
+ */
+export function buildMimeType(codec: string, type: "video" | "audio"): string {
```
kixelated: This is very MP4 specific, but it has a generic name.

js/hang/src/watch/audio/source.ts
```diff
@@ -41,6 +41,10 @@ export class Source {
 	// Downcast to AudioNode so it matches Publish.Audio
 	readonly root = this.#worklet as Getter<AudioNode | undefined>;
 
+	// For MSE path, expose the HTMLAudioElement for direct control
+	#mseAudioElement = new Signal<HTMLAudioElement | undefined>(undefined);
+	readonly mseAudioElement = this.#mseAudioElement as Getter<HTMLAudioElement | undefined>;
```
kixelated, line 48: I don't think we expose this. We should expose an `AudioNode` instead so it works regardless of the WebCodecs/MSE backend used.

js/hang/src/watch/video/source.ts
```diff
@@ -192,12 +205,83 @@ export class Source {
...
+			});
+
+			this.#signals.effect((eff) => {
+				const stats = eff.get(mseSource.stats);
+				eff.set(this.#stats, stats);
+			});
+
+			// Expose method to set audio element for synchronization
+			this.setAudioSync = (audioElement: HTMLAudioElement | undefined) => {
+				mseSource.setAudioSync(audioElement);
```
kixelated: I think this approach is just wrong. There's no way you're going to get decent synchronization via `currentTime`

We should have a single `MediaSource` and add/remove audio/video `SourceBuffers`.

js/hang/src/watch/audio/source-mse.ts
```diff
@@ -0,0 +1,377 @@
...
+					break;
+				}
+
+				// Stop processing if MediaSource is closed
+				if (this.#mediaSource?.readyState === "closed") {
+					break;
+				}
+
+				const isMoovAtom = hasMoovAtom(frame.data);
+				const isInitSegment = isMoovAtom && !initSegmentReceived;
```
kixelated: We should send init segments as a separate track, or embedded inside the catalog. Sending it at the start of each group is technically valid but goes against the spirit of them.

In fact, the `description` field consists of a avcC box which can be used to build a valid MP4 init segment I believe. @littlespex would know more

kixelated (commented):
I think this needs another attempt.

1. The contents of each MoQ frame should be a fMP4 fragment (aka CMAF segment). No timestamp prefix allowed.
2. We should let MSE perform A/V sync.
3. We shouldn't push init segments over the data tracks. Either a separate track, or reconstructed from the catalog.

kixelated:
Replaced by #896

PR #841, opened 2026-01-13
feat: refactoring hang-ui folder

pzanella:
Reorganized the `hang-ui` source folder structure.


js/hang-ui/README.md
```diff
@@ -47,3 +47,78 @@ Here's how you can use them (see also @moq/hang-demo for a complete example):
...
+- **WatchControls**: Main control panel for the video player
+- **PlayPauseButton**: Play/pause toggle
+- **VolumeSlider**: Audio volume control
+- **LatencySlider**: Adjust playback latency
+- **QualitySelector**: Switch between quality levels
+- **FullscreenButton**: Toggle fullscreen mode
+- **BufferingIndicator**: Visual feedback during buffering
+- **StatsButton**: Toggle statistics panel
+
+#### **stats/**
```
jdreetz: Stats is only used by the Watch UI, right? Why isn't it a component of `watch`?

jdreetz (approved):
Looks great! One question: why is Stats outside of the Watch UI?

js/hang-ui/README.md
pzanella: Hey @jdreetz! Thank you for your feedback! The Stats component is currently used only by the Watch UI. I was thinking we could also use it for real-time stats in the Publish UI later. Do you think it makes sense to move it to the shared folder? If not, no worries, I can just keep it under the Watch UI. Let me know. Thanks!

jdreetz: If you think we'd use it elsewhere, then yea, I'd move it to shared components.

js/hang-ui/package.json
```diff
@@ -6,14 +6,14 @@
 	"license": "(MIT OR Apache-2.0)",
 	"repository": "github:moq-dev/moq",
 	"exports": {
-		"./publish": "./src/publish.tsx",
```
kixelated, line 9: I don't think this will work unfortunately. `vite build` dumps each entrypoint into the root. There might be a work-around but I couldn't figure one out.

I would have left a comment but JSON is a terrible format.

js/hang-ui/package.json
```diff
@@ -6,14 +6,14 @@
...
-		"./watch": "./src/watch.tsx",
-		"./stats": "./src/stats.tsx"
+		"./publish": "./src/publish/index.tsx",
+		"./watch": "./src/watch/index.tsx",
+		"./stats": "./src/shared/components/stats/index.tsx"
```
kixelated: Do we even need stats as an entrypoint..?

js/hang-ui/vite.config.ts
```diff
@@ -7,9 +7,9 @@ export default defineConfig({
 	build: {
 		lib: {
 			entry: {
-				publish: resolve(__dirname, "src/publish.tsx"),
-				watch: resolve(__dirname, "src/watch.tsx"),
-				stats: resolve(__dirname, "src/stats.tsx"),
+				publish: resolve(__dirname, "src/publish/index.tsx"),
```
kixelated: `publish:` results in `dist/publish.js`. Maybe "publish/index" might work but I haven't tried it.

js/hang-ui/package.json
pzanella: Thanks for the feedback @kixelated! I updated the `package.json` and changed the ViteJS configuration to have a folder structure. Please let me know what you think, and if this addresses your concern. Thanks!

kixelated (approved):
Can you try building the package and make sure dist looks correct? Would be nice to add a check like that to make sure the package builds correctly. In a separate PR of course

pzanella:
> Can you try building the package and make sure dist looks correct? Would be nice to add a check like that to make sure the package builds correctly. In a separate PR of course

Yes, I ran the build command and the following files are generated in `dist`. 
```
dist/
   publish/
      index.js
      index.js.map
   watch/
      index.js
      index.js.map
   icon-<HASH>.js
   icon-<HASH>.js.map
   README.md
   package.json
```
Please let me know if this looks correct to you, or if you have any suggestions. Thanks!

kixelated:
> Oh, it looks like the bundle expects all these to be on the page already:
> 
> ```
> 		rollupOptions: {
> 			external: ["@moq/hang", "@moq/lite", "@moq/signals"],
> 		},
> ```

Yeah, I'm not sure if we should bundle those too or keep the UI separate from the library. I can see both scenarios.

PR #844, opened 2026-01-13
Fix bugs with WebSocket fallback

jhiatt-verkada:
* Ensure default delay applies when constructing config via `ClientConfig::default()`
* Respect TLS configuration parameters

### DIFF
--- a/rs/moq-native/Cargo.toml
+++ b/rs/moq-native/Cargo.toml
@@ -50,7 +50,7 @@ tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 url = "2"
 web-transport-iroh = { workspace = true, optional = true }
 web-transport-quinn = { workspace = true }
-web-transport-ws = { workspace = true }
+web-transport-ws = { workspace = true, features = ["rustls-tls-webpki-roots"] }
 
 [dev-dependencies]
 anyhow = "1"
--- a/rs/moq-native/src/client.rs
+++ b/rs/moq-native/src/client.rs
@@ -351,21 +351,31 @@ impl Client {
 		}
 
 		// Convert URL scheme: http:// -> ws://, https:// -> wss://
-		match url.scheme() {
+		let needs_tls = match url.scheme() {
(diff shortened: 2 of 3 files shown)


rs/moq-native/Cargo.toml
```diff
@@ -45,6 +45,7 @@ serde = { version = "1", features = ["derive"] }
 serde_with = "3"
 time = "0.3"
 tokio = { workspace = true, features = ["full"] }
+tokio-tungstenite = { version = "0.24", features = ["rustls-tls-webpki-roots"] }
```
jhiatt-verkada: It would probably be better if we could continue to use this crate via `web-transport`'s re-export, so we can go that route if preferred. We need to enable this feature to enable `rustls` support. The `tokio-tungstenite` crate only exposes that via two higher-level features, either `rustls-tls-native-roots` or `rustls-tls-webpki-roots`, so you have to pick one or the other. I chose this one because the `webpki` crates are already in the dependency tree elsewhere so I thought that was the lesser of two evils.

rs/moq-native/src/client.rs
```diff
@@ -6,10 +6,10 @@ use std::collections::HashSet;
 use std::path::PathBuf;
 use std::sync::{LazyLock, Mutex};
 use std::{fs, io, net, sync::Arc, time};
+use tokio_tungstenite::tungstenite;
 use url::Url;
 #[cfg(feature = "iroh")]
 use web_transport_iroh::iroh;
-use web_transport_ws::{tokio_tungstenite, tungstenite};
```
jhiatt-verkada, line 12: See comment in Cargo.toml for why this is necessary.

rs/moq-native/Cargo.toml
kixelated: I don't think we can set feature flags when using re-exports unfortunately. You'd have to add these features to `web_transport_ws` and proxy them through that way, which is _probably_ better because then you won't run into versioning issues.

Issue #783, opened 2025-12-16
Buffer visualization


kixelated:
It would be nice (in the stats page?) to visualize the current buffer size. I was thinking horizontal bars with the left side being 0 and the right side being the target latency.

MoQ is interesting because segments/groups can arrive out of order. Instead of a single buffered amount, we need to support multiple ranges. Audio and video need to be represented separately too.

jdreetz:
> MoQ is interesting because segments/groups can arrive out of order. Instead of a single buffered amount, we need to support multiple ranges. Audio and video need to be represented separately too.

@kixelated do you have any recommendations on how to measure this? Right now [we're tracking the buffer size](https://github.com/jdreetz/moq/blob/feat/expose-buffer-duration/js/hang/src/watch/video/source.ts#L279) (in the Stats UI) as how long we "sleep" for between rendering frames. 

I've been trying to [measure the buffer size by looking at the min/max timestamps of frames](https://github.com/jdreetz/moq/blob/feat/expose-buffer-duration/js/hang/src/frame.ts#L320) in `this.#groups.frames` inside the frame `Consumer`, but that timespan is always zero, unless I do something like [wait for the `TransformStream` writer to be ready](https://github.com/jdreetz/moq/blob/feat/expose-buffer-duration/js/hang/src/watch/video/source.ts#L326) before we decode the frame, but adding that wait caused the buffer size to start to creep up over time. 

**What did you have in mind for measuring this?** If we use the "sleep" time, that's usually only about 40ms, even if I change the latency slider to a higher value, so I don't think is giving us something representative of the latency value, or is it? 

Also, I'm not sure if I should be testing this differently, like adding an artificial delay somehow to simulate poor network conditions?

PR #845, opened 2026-01-14
feat: Expose buffer information in Player UI

jdreetz:
image

# Description
This PR completes [Issue 783](https://github.com/moq-dev/moq/issues/783) by exposing two new properties on the Frame Consumer that calculate the earliest and latest frames in the buffer and presenting them as the buffer amount in the Player UI. The latency slider has been moved to the buffer UI, since the latency control should affect how much buffer is available for playback. 

# How to Test
Since it's hard to force buffering without actually having latency, I'm leaving in a utility to simulate latency in the Frame consumer when a frame is read. To enable it, set `window.simulateLatency` to true and increase the latency slider. This should cause a larger buffer to be created and the UI to display the amount.

### DIFF
--- a/js/hang-ui/src/watch/components/BufferOverlay.tsx
+++ b/js/hang-ui/src/watch/components/BufferOverlay.tsx
@@ -1,50 +1,35 @@
-import { For, Show } from "solid-js/web";
+import { Show } from "solid-js/web";
 import LatencySlider from "./LatencySlider";
-import useWatchUIContext from "./useWatchUIContext";
-
-type BufferDetails = {
-	content: "Video" | "Audio";
-	duration: number | undefined;
-	noDataMessage: string;
-};
+import BufferTimeline from "./BufferTimeline";
+import useWatchUIContext from "../hooks/use-watch-ui";
 
 /**
  * Overlay displaying buffer information for video and audio
  */
 export default function BufferOverlay() {
 	const context = useWatchUIContext();
-	const bufferDetails = (): BufferDetails[] => [
-		{
-			content: "Video",
-			duration: context.videoBufferDuration(),
-			noDataMessage: "Video may be paused or not available",
-		},
-		{
-			content: "Audio",
-			duration: context.audioBufferDuration(),
-			noDataMessage: "Audio may be muted or not available",
-		},
-	];
 
 	return (
 		<Show when={context.isBufferOverlayVisible()}>
 			<div class="bufferOverlay">
-				<div class="bufferOverlayHeader">Buffer Stats</div>
-				<For each={bufferDetails()}>
-					{(section) => (
-						<div class="bufferOverlaySection">
-							<div class="bufferOverlayLabel">{section.content}</div>
-							<div
-								class="bufferOverlayValue"
-								title={section.duration === undefined ? section.noDataMessage : undefined}
-							>
-								{section.duration !== undefined
-									? `${section.duration?.toFixed(1)}ms buffered`
-									: "No data"}
-							</div>
-						</div>
-					)}
-				</For>
+				<div class="bufferOverlayHeader">Buffer Information</div>
+				<div class="bufferOverlaySection">
+					<div class="bufferOverlayLabel">Video</div>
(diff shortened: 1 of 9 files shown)


js/hang/src/frame.ts
```diff
@@ -195,61 +206,57 @@ export class Consumer {
...
 		// Wake up any consumers waiting for a new frame.
 		this.#notify?.();
 		this.#notify = undefined;
 	}
 
+	#updateBufferRange() {
+		const { earliestTime, latestTime } = getBufferedRangeForGroup(this.#groups);
+		this.#earliestBufferTime.set(earliestTime);
+		this.#latestBufferTime.set(latestTime);
+	}
```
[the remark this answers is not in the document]
jdreetz: > but not on the hot path where decode() actually shifts frames out of the buffer (Line 254). If the UI displays “buffer amount”, it should decrease as playback consumes frames.

I implemented this change locally, but it seemed to break any buffer tracking.

js/hang-ui/src/watch/components/BufferOverlay.tsx
```diff
@@ -0,0 +1,51 @@
...
+import useWatchUIContext from "./useWatchUIContext";
+
+/**
+ * Overlay displaying buffer information for video and audio
+ */
+export default function BufferOverlay() {
+	const context = useWatchUIContext();
+
+	return (
+		<Show when={context.isBufferOverlayVisible()}>
```
pzanella, line 13: I noticed that the video and audio sections share a very similar structure.  
Would it make sense to extract this into a small data structure and render it with SolidJS’s `<For>`?
Something along these lines:

```ts
const sections = (): BufferSectionData[] => [
    {
        label: "Video",
        duration: context.videoBufferDuration(),
        noDataMessage: "Video may be paused or not available",
    },
    {
        label: "Audio",
        duration: context.audioBufferDuration(),
        noDataMessage: "Audio may be muted or not available",
    },
];

return (    
   ...     
   <For each={sections()}>
      ...
   </For>
);
```

jdreetz: Great idea! I made that change in the overlay.

pzanella (approved):

js/hang-ui/src/Components/watch/index.css
```diff
@@ -202,3 +202,60 @@
 	background: #1a1a1a;
 	color: #fff;
 }
+
+/* Buffer Overlay Styles */
+.bufferOverlay {
+	position: absolute;
+	top: 8px;
+	right: 8px;
+	background: rgba(0, 0, 0, 0.85);
```
pzanella: Sorry, quick suggestion: could we use variables.css for color and spacing to stay consistent with the rest of the codebase? Nothing critical, just a thought.

jdreetz: good call, updated!

js/hang/src/frame.ts
```diff
@@ -282,3 +293,39 @@ export class Consumer {
 		this.#groups.length = 0;
 	}
 }
+
+function getBufferedRangeForGroup(groups: Group[]) {
```
kixelated: the `ForGroup` suffix is confusing, because this works on all groups.

js/hang/src/frame.ts
```diff
@@ -260,7 +268,10 @@ export class Consumer {
 			}
 
 			const wait = new Promise<void>((resolve) => {
-				this.#notify = resolve;
+				this.#notify = () => {
+					this.#updateBufferRange();
```
kixelated: I think this is redundant? Basically we should call `this.#updateBufferRange();` before any `this.#notify` call if it could change the result.

kixelated (commented):
The new names are fine, but it definitely made it harder to review because not much had changed. 

@jdreetz This is a good start but ideally you can report the min/max timestamp of each group, not the min/max timestamp of all groups combined. That way we can visualize gaps. It's really cool when it works because you *should* see newer groups delivered first during congestion.

PR #848, opened 2026-01-15
rs: moq-native: Implement QUIC-LB compatible CID generation

jhiatt-verkada:
This implementation was developed by trial-and-error testing against AWS's Network Load Balancer with QUIC support. They seem to have some internal requirements that go beyond what's specified by the draft RFC; in particular the resulting Connection ID must be EXACTLY 16 bytes in length or else it will not work.

### DIFF
--- a/rs/moq-native/src/server.rs
+++ b/rs/moq-native/src/server.rs
@@ -5,7 +5,6 @@ use crate::crypto;
 #[cfg(feature = "iroh")]
 use crate::iroh::IrohQuicRequest;
 use anyhow::Context;
-use base64::Engine;
 use moq_lite::Session;
 use rand::Rng;
 use rustls::pki_types::{CertificateDer, PrivatePkcs8KeyDer};
@@ -22,31 +21,6 @@ use web_transport_quinn::http;
 use futures::FutureExt;
 use futures::future::BoxFuture;
 use futures::stream::{FuturesUnordered, StreamExt};
-
-/// Server ID for QUIC-LB support.
-#[serde_with::serde_as]
-#[derive(Clone, serde::Serialize, serde::Deserialize)]
-pub struct QuicLbServerId(#[serde_as(as = "serde_with::base64::Base64")] Vec<u8>);
-
-impl QuicLbServerId {
-	fn len(&self) -> usize {
-		self.0.len()
-	}
-}
-
-impl std::fmt::Debug for QuicLbServerId {
-	fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-		f.debug_tuple("QuicLbServerId").field(&hex::encode(&self.0)).finish()
-	}
-}
-
-impl std::str::FromStr for QuicLbServerId {
-	type Err = base64::DecodeError;
-
-	fn from_str(s: &str) -> Result<Self, Self::Err> {
-		base64::engine::general_purpose::STANDARD.decode(s.as_bytes()).map(Self)
-	}
-}
 
 /// TLS configuration for the server.
 ///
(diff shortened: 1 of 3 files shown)


rs/moq-native/src/server.rs
```diff
@@ -21,6 +23,42 @@ use futures::FutureExt;
...
+		if decoded_len != out.len() {
+			anyhow::bail!(
+				"invalid server ID length: expected {} bytes, got {}",
+				out.len(),
+				decoded_len
+			);
+		}
+		Ok(out)
+	}
+}
```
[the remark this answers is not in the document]
jhiatt-verkada: I agree this is weird. I think I'll change this to implement `Debug` instead because I only need it for logging.

rs/moq-native/src/server.rs
```diff
@@ -21,6 +23,42 @@ use futures::FutureExt;
 use futures::future::BoxFuture;
 use futures::stream::{FuturesUnordered, StreamExt};
 
+/// Server ID for QUIC-LB support.
+///
+/// This is fixed to 8 bytes to match the AWS implementation as of Jan 2026.
+#[serde_with::serde_as]
+#[derive(Clone, serde::Serialize, serde::Deserialize)]
+pub struct QuicLbServerId(#[serde_as(as = "serde_with::base64::Base64")] [u8; 8]);
```
kixelated: I'm biased towards hex but this works too.

rs/moq-native/src/server.rs
```diff
@@ -21,6 +23,42 @@ use futures::FutureExt;
 use futures::future::BoxFuture;
 use futures::stream::{FuturesUnordered, StreamExt};
 
+/// Server ID for QUIC-LB support.
+///
+/// This is fixed to 8 bytes to match the AWS implementation as of Jan 2026.
```
kixelated: We should make this variable length so it'll work with other implementations too. I think we would need a separate argument for the nonce size, then it'll work?

rs/moq-native/src/server.rs
jhiatt-verkada: I originally implemented it with hex but then I realized that if you're trying to do this in Kubernetes, the server ID will be provided as base64: https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/guide/use_cases/quic/#:~:text=Which%20tells%20the%20LBC%20to%20inject%20a%20Server%20ID%20into%20the%20container%20environment.%20The%20envoy%20config%20is%20specified%20to%20look%20for%20the%20Server%20ID%20under%20the%20environment%20variable%20AWS_LBC_QUIC_SERVER_ID.

So in that case base64 is the preferred format.

rs/moq-native/src/server.rs
jhiatt-verkada: There are two things which at least in theory can have variable length but in practice are fixed by the NLB implementation:
* Server ID
* Nonce
This comment here is about the server ID, but yeah we could make it a parameter of some sort.

kixelated: I think take a (variable length) server ID and nonce length, then we can support any QUIC-LB scheme. AWS would be 8 and 7 respectfully.

jhiatt-verkada: Ok, I modified it so that the server ID can be any length (just a Vec<u8>) and there's a flag to specify overall connection ID length rather than nonce length, because I found 16 to be a more pleasing default than 7.

Also the NLB implementation is maybe a bit less sensitive than I originally thought. I did my original testing to discover what works by experimenting with Envoy's `quic_lb` implementation, but I think `quinn` might be a little smarter about padding out to a length that works.

rs/moq-native/src/server.rs
```diff
@@ -60,6 +87,23 @@ pub struct ServerConfig {
...
+	/// If set, connection IDs will be derived semi-deterministically.
+	#[arg(id = "server-quic-lb-id", long = "server-quic-lb-id", env = "MOQ_SERVER_QUIC_LB_ID")]
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub quic_lb_id: Option<QuicLbServerId>,
+
+	/// Connection ID length when QUIC-LB support is enabled
+	/// Defaults to 16 if not specfied.
+	#[arg(
+		id = "server-connection-id-len",
+		long = "server-connection-id-len",
```
kixelated: `server-quic-lb-id-len`

Can't be longer than 20. Also kind of confusing because it's not the ID length, but the ID + nonce + 1.

I prefer `--server-quic-lb-nonce 7`. It's fatal to make the nonce too small, and wasteful to make it too big. Users shouldn't have to think about this when changing the server ID.

PR #853, opened 2026-01-19
Minimal observability metrics (relay & client)

fcancela:
### Summary
This PR adds the **smallest reviewable observability surface** for the MoQ demo: a reusable application-level byte accounting hook in `moq-lite`, minimal relay-side counters exposed via a Prometheus-style `/metrics` endpoint, and minimal browser-side OTLP metrics for connection type + time-to-first-frame.

### What’s included
- **`moq-lite`**: optional `Stats` sink for application payload byte accounting
  - Adds `Stats` trait and `connect_with_stats` / `accept_with_stats` plumbing
  - Wires byte accounting into publisher/subscriber read/write hot paths
- **`moq-native`**: adds `Request::accept_with_stats(...)` to pass the sink through to `moq-lite`
- **`moq-relay`**: minimal counters + `/metrics`
  - `moq_relay_active_sessions{transport="webtransport|websocket"}` (gauge)
  - `moq_relay_app_bytes_sent_total{transport=...}` (counter)
  - `moq_relay_app_bytes_received_total{transport=...}` (counter)
- **`js/hang`**: metrics-only OTLP HTTP export (no tracing)
  - `moq_client_connections_total{transport=...}` (counter)
  - `moq_client_startup_time_seconds{track_type="audio|video"}` (histogram, **time-to-first-frame** from track subscription start)
- **`js/lite`**: adds a tiny `onConnectionType(cb)` callback hook so hang can record `webtransport` vs `websocket`
- **`js/hang` video stats**: `framesDecoded` is counted on the decoder output path; removed the unimplemented `framesDropped`

### What’s intentionally out of scope
- Grafana dashboards / Tempo / Loki / collector compose stack
- Traces/log export
- Higher-cardinality media QoE metrics (rebuffer, decode timings, etc.)

### How to test (local)
1. Run the demo as usual
2. Verify relay metrics are present and changing:

```bash
curl -sS http://localhost:4443/metrics | grep '^moq_relay_'
```

You should see `moq_relay_app_bytes_*` increasing while playback is active.

3. (Optional) If you have an OTLP HTTP collector at `:4318`, the browser will POST metrics to:
- `http://<collector>:4318/v1/metrics`

### Notes for reviewers
- The `moq-lite::Stats` hook is intentionally **generic** and **non-blocking** (callers can implement with atomics).
- Relay `/metrics` uses **low-cardinality labels only** (`transport`), matching Prometheus best practices.


js/hang/src/observability/index.ts
```diff
@@ -0,0 +1,122 @@
...
+	const meter = metrics.getMeter(serviceName);
+
+	clientMetricsInstance = new ClientMetrics(meter);
+	setupConnectionTracking();
+	initialized = true;
+}
+
+function setupConnectionTracking() {
+	// Dynamically import to avoid circular deps.
+	import("@moq/lite")
```
kixelated, line 71: ?

js/hang/src/watch/element.ts
```diff
@@ -69,12 +75,30 @@ export default class HangWatch extends HTMLElement {
 
 		cleanup.register(this, this.signals);
 
+		// Auto-initialize observability when URL is set
+		// Derives OTel endpoint from relay URL (same host, port 4318)
+		this.signals.effect((effect) => {
+			const url = effect.get(this.url);
+			if (url) {
+				const otelEndpoint = `http://${url.hostname}:4318`;
+				Observability.initObservability({ otlpEndpoint: otelEndpoint, sessionId: this.sessionId });
```
kixelated, line 84: This would absolutely need to be opt-in.

rs/moq-relay/src/metrics.rs
```diff
@@ -0,0 +1,112 @@
...
+	app_bytes_received_webtransport: Arc<AtomicU64>,
+	app_bytes_received_websocket: Arc<AtomicU64>,
+}
+
+impl MetricsTracker {
+	pub fn new() -> Self {
+		Self::default()
+	}
+
+	pub fn inc_active_sessions(&self, transport: Transport) {
```
kixelated, line 42: Can you make this return a guard instead of needing `dec_active_sessions`?

rs/moq-lite/src/session.rs
```diff
@@ -34,14 +34,15 @@ impl Session {
...
 
-	/// Perform the MoQ handshake as a client, negotiating the version.
+	/// Perform the MoQ handshake as a client, negotiating the version, with optional stats hooks.
 	///
-	/// Publishing is performed with [OriginConsumer] and subscribing with [OriginProducer].
-	/// The connection remains active until the session is closed.
-	pub async fn connect<S: web_transport_trait::Session>(
+	/// This is equivalent to [`Session::connect`] but allows providing a [`crate::Stats`] sink
+	/// for application-level byte accounting (ignores transport retransmissions).
+	pub async fn connect_with_stats<S: web_transport_trait::Session>(
```
kixelated, line 41: We should probably make a new builder API for this. I'll try making something.

kixelated (commented):
Can we remove the JS side for now? I'm a little uneasy because at Twitch/Discord we never used prometheus for client metrics. At the very least it needs to be opt-in.

The Rust side looks good though.

Bonus points for Nix instead of Docker lul. I'll see if Claude can figure it out.

rs/moq-lite/src/session.rs
kixelated: #862

kixelated:
Going to try doing this with MoQ.

fcancela:
Hey Luke, sorry I didn't finish this piece of work. 

I totally agree this should be done entirerly with MoQ. @pzanella has been doing some work on his own to finish this up. But I haven't been able to find some time to help him. 

I'm pretty sure he would be eager to help with this effort.

Issue #678, opened 2025-11-14
Higher latency delivery


kixelated:
Both the [Javascript player](https://github.com/kixelated/moq/blob/61a1d90161e6a1149551e46446996812b10ff367/js/hang/src/frame.ts#L74) and [Rust player](https://github.com/kixelated/moq/blob/61a1d90161e6a1149551e46446996812b10ff367/rs/hang/src/model/track.rs#L90) have a configurable jitter buffer. They support buffering new groups for a number of milliseconds while old content arrives asynchronously. It should support going into the multiple seconds range, however:

The Rust publisher is hard-coded to only allow [2 groups in flight at most](https://github.com/kixelated/moq/blob/61a1d90161e6a1149551e46446996812b10ff367/rs/moq/src/lite/publisher.rs#L193). For video this translates to 2 GoPs. For audio this is nebulous as the size of groups is configurable potentially a single frame (so only 2 allowed in flight). I don't think the Javascript publisher has any logic; it relies on the subscriber to cancel which is not great.

We should implement this properly with a `Delivery Timeout` field on the wire, similar to MoqTransport. This would be a hint from the subscrber -> publisher (and publisher -> library) on when to drop groups. The expiration timer should start when a newer group is received, which miiiight not exactly match the IETF draft.

The Javascript and Rust player would send this timeout over the network so the remote publisher will maintain a jitter buffer with a similar duration.

PR #856, opened 2026-01-21
WIP: Delivery timeout

ac-freeman:
I implemented simplified delivery timeouts for a research work 6 months ago. I asked Claude to help bring those changes into the latest version of the repo. This hasn't been completely tested yet, but my original implementation did work for my purposes. I haven't touched anything on the JavaScript side yet.

Paper (soon to be updated): https://arxiv.org/abs/2507.15673
Relevant issue: #678

### DIFF
--- a/rs/hang/src/model/broadcast.rs
+++ b/rs/hang/src/model/broadcast.rs
@@ -75,12 +75,12 @@ pub struct BroadcastConsumer {
 
 impl BroadcastConsumer {
 	pub fn new(inner: moq_lite::BroadcastConsumer) -> Self {
-		let catalog = inner.subscribe_track(&Catalog::default_track()).into();
+		let catalog = inner.subscribe_track(&Catalog::default_track(), None).into();
 		Self { inner, catalog }
 	}
 
 	pub fn subscribe(&self, track: &moq_lite::Track, latency: std::time::Duration) -> TrackConsumer {
-		TrackConsumer::new(self.inner.subscribe_track(track), latency)
+		TrackConsumer::new(self.inner.subscribe_track(track, None), latency)
 	}
 }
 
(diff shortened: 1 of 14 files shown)


kixelated:
Hey @ac-freeman I appreciate the PR. I've implemented this probably 4 times already and I'm never happy with the result, lul.

I'd like to make delivery timeout work with fanout. So when the relay gets a subscription with a timeout of 1s, it subscribes upstream with 1s timeout. But when another request comes in with 5s timeout, it updates the existing subscription to 5s. That requires piping a lot of stuff through the model so we actually have some feedback on what consumers want, not just what producers are making.

kixelated:
> So, upstream_timeout = max(downstream_timeouts)? The upstream timeout would need to be updated when a subscriber drops, if it had the largest timeout. Wouldn't this also cascade with large fan-out? If you have one relay five layers deep that subscribes with a 60s timeout, that would bubble all the way up. Should the non-relay subscriber timeouts also affect upstream subscriptions? I just see a risk for a rogue subscriber hurting the latency for everyone.
> 
> The original publisher timeout should probably also take precedence over the first relay's subscription timeout. They might want to have some say in what the latency regime is.

Yeah, my plan is to have the publisher announce a TTL for each track. The relay *should* keep the data around for this amount of time, even if there are no active subscribers, just so new subscribers can get access. A relay could drop the cache early (LRU) if they're filling to refill it.

Each subscriber would have its own TTL upper bound to the publisher's TTL. And note this TTL would be based on when the content was received/produced, not when the subscription started.

> I just see a risk for a rogue subscriber hurting the latency for everyone.

That's my issue with relying on TTLs; you can't handle publisher congestion if there's a VOD worker with a target latency of 30s alongside a real-time user with a target latency of 1s. I use priorities (descending group order) instead so the publisher deliver the latest content first (for the 1s viewer) and later backfills old content (for the VOD worker).

For me, Delivery Timeout is an optimization to avoid transmitting old data that will be immediately dropped by the receiver. It's not meant to drain the queue during congestion because the queue is being actively prioritized.

PR #857, opened 2026-01-22
feat: restyling Publish UI

pzanella:
Restyled the publish UI and cleaned up the code.
Here are some image examples:
image
image
image

### DIFF
--- a/.github/workflows/libmoq.yml
+++ b/.github/workflows/libmoq.yml
@@ -25,6 +25,8 @@ jobs:
             os: macos-15-intel
           - target: aarch64-apple-darwin
             os: macos-latest
+          - target: universal-apple-darwin
+            os: macos-latest
           - target: x86_64-pc-windows-msvc
             os: windows-latest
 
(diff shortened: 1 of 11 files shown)


kixelated (approved):
Seems fine to me, although I don't understand the `muted` changes. Just make sure you've tested all of the permutations.

js/hang-ui/src/publish/components/CameraSourceButton.tsx
```diff
@@ -24,10 +24,10 @@ export default function CameraSourceButton() {
 	};
 
 	return (
-		<div class="publishSourceButtonContainer">
+		<div class="publish-ui__source-button-wrapper flex--center">
 			<Button
 				title="Camera"
-				class={`publishSourceButton ${context.cameraActive() ? "active" : ""}`}
+				class={`publish-ui__source-button flex--center ${context.cameraActive() ? "publish-ui__source-button--active" : ""}`}
```
jdreetz, line 30: this is really cool! we should file an issue to re-factor all of the UI components to use this format.

pzanella: Glad you like it! I agree we should use this format.

PR #858, opened 2026-01-22
fix(dev): make just recipes WSL2-friendly

fcancela:
## Summary
Fix WSL2 + Windows Chrome local dev where **WebTransport (QUIC/UDP) to `localhost` fails**, causing a **WebSocket fallback**.

## Problem / symptoms
- Browser console:
  - `Failed to establish a connection to https://localhost:4443/anon: net::ERR_CONNECTION_RESET`
  - `... using WebSocket fallback; the user experience may be degraded`
- Visiting `http://.../anon` shows `400 Connection header did not include 'upgrade'` (expected; `/anon` is not an HTTP endpoint).

## Root cause (validated)
- `http://<wsl-ip>:4443/certificate.sha256` worked (HTTP/TCP reachability ok)
- `ws://<wsl-ip>:4443/anon` worked (WebSocket/TCP ok)
- Only WebTransport to `https://localhost:4443/...` failed (WSL2 localhost forwarding doesn’t reliably support UDP/QUIC)
- Switching to the **WSL VM IP** and binding/certifying for it makes WebTransport succeed.

## Fix
Dev tooling now:
- Uses **WSL VM IP** instead of `localhost` for the relay URL when in WSL (`just web auto` / `just dev`)
- In WSL only, runs the relay with:
  - `--server-bind 0.0.0.0:4443` (QUIC reachable from Windows)
  - `--tls-generate localhost,<wsl-ip>` (WebTransport upgrades to `https://…`; cert SAN must match the IP host)
- Keeps non-WSL behavior unchanged.
- Runs Vite with `--host 0.0.0.0` so Windows can open the dev server via the WSL IP.

## Test plan
- In WSL: `just dev`
- Confirm WebTransport connects (no `ERR_CONNECTION_RESET`, no WebSocket fallback warning)

### DIFF
--- a/justfile
+++ b/justfile
@@ -29,13 +29,19 @@ dev:
 	bun run concurrently --kill-others --names srv,bbb,web --prefix-colors auto \
 		"just relay" \
 		"sleep 1 && just pub bbb http://localhost:4443/anon" \
-		"sleep 2 && just web http://localhost:4443/anon"
-
+		"sleep 2 && just web auto"
 
 # Run a localhost relay server without authentication.
 relay *args:
-	# Run the relay server overriding the provided configuration file.
-	TOKIO_CONSOLE_BIND=127.0.0.1:6680 cargo run --bin moq-relay -- dev/relay.toml {{args}}
+	@if [ -n "${WSL_DISTRO_NAME:-}" ]; then \
+		ip="$(hostname -I | awk '{print $1}')"; \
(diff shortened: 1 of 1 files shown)


kixelated:
@fcancela I switched over the local dev files to use `0.0.0.0`, can you try on Windows and hopefully it just works?

Otherwise we can try using the WSL IP but I'd like a better understanding of what is going on. This PR seemingly arbitrarily binds to 4 different IPs depending on the command.

PR #859, opened 2026-01-22
CMAF passthrough attempt v2

kixelated:
Replaces #822

Overview:
1. Added support for `--passthrough` (Rust only) when using HLS and fMP4 import to write the raw MP4 bytes.
2. MSE is a separate `Backend`, so it will be tree-shaken when not used. 
    - The same but reverse for WebCodecs. No need for a fMP4 encoder.
3. The `<hang-watch>` element takes either a `<video>` or `<canvas>` tag to switch between backends.
    - Won't tree-shake, use the JS API if you want to reduce the bundle size.
4. `Watch.Broadcast` no longer owns the video/audio component; they have to be constructed separately.
   - Allows tree-shaking `Watch.Audio.Source` and `Watch.Video.Source`.
   - Opens the door for more tree-shakable modules that take a `Broadcast`.
5. Removed support for the unused `raw` container.
    - In hind-sight, a new container format should use delta encoded timestamps anyway.
    - Additionally, timestamps are being added to the `moq-lite` layer that we could reuse.
    - Sorry @gdavila, I didn't think enough before filing that issue.
6. The "mse" container contains an `init_track`, used to initialize the decoder.
    - This is inlined in the `CMSF` draft, but I plan to use it via a separate track for HLS fallback.
    - It's more efficient this way when renditions share an init segment.
7. A bunch of API changes to support the generic `Backend` interface.
   - ex. `watch.audio.muted` instead of `watch.muted`.

TODO:
- [ ] WebCodecs support for CMAF (decode MP4)
- [ ] MSE support for hang (encode MP4)
- [ ] Fix fMP4 import `--passthrough` not working with bbb
- [ ] Fix the buffering signal
- [ ] Fix the stats panel
- [ ] Actually review the HLS import changes
- [ ] Send `jitter` in the catalog so we can increase the latency on demand?
- [ ] Remove `init_track` and build it in memory instead?

### DIFF
--- a/js/hang/src/watch/mse/audio.ts
+++ b/js/hang/src/watch/mse/audio.ts
@@ -167,7 +167,7 @@ export class Audio implements Backend {
 		effect.cleanup(() => data.close());
 
 		effect.spawn(async () => {
-			// Generate init segment from catalog config (always uses track_id=1)
+			// Generate init segment from catalog config (uses track_id from container)
 			const initSegment = Mp4.createAudioInitSegment(selected.config);
 			await this.#appendBuffer(sourceBuffer, initSegment);
 
@@ -177,8 +177,6 @@ export class Audio implements Backend {
 				const frame = await data.readFrame();
 				if (!frame) return;
 
-				// Rewrite track_id to 1 to match our generated init segment
-				Mp4.rewriteTrackId(frame);
 				await this.#appendBuffer(sourceBuffer, frame);
 
 				// Seek to the start of the buffer if we're behind it (for startup).
--- a/js/hang/src/catalog/container.ts
+++ b/js/hang/src/catalog/container.ts
@@ -18,6 +18,8 @@ export const ContainerSchema = z
 			kind: z.literal("cmaf"),
 			// Time units per second
 			timescale: u53Schema,
+			// Track ID used in the moof/mdat fragments
+			trackId: u53Schema,
 		}),
 	])
 	.default({ kind: "legacy" });
--- a/js/hang/src/mp4/encode.ts
+++ b/js/hang/src/mp4/encode.ts
@@ -232,16 +232,17 @@ function createAvc1Box(width: number, height: number, avcC: Uint8Array): Uint8Ar
  * ```
  */
 export function createVideoInitSegment(config: Catalog.VideoConfig): Uint8Array {
-	const { codedWidth, codedHeight, description } = config;
+	const { codedWidth, codedHeight, description, container } = config;
 	if (!codedWidth || !codedHeight || !description) {
 		// TODO: We could
 		throw new Error("Missing required fields to create video init segment");
 	}
 
-	const timescale = 1_000_000; // microseconds
+	// Use timescale from CMAF container, or microseconds for legacy
+	const timescale = container.kind === "cmaf" ? container.timescale : 1_000_000;
 
-	// Always use track_id=1. Incoming CMAF data segments are rewritten to match.
-	const trackId = 1;
+	// Use track_id from CMAF container, or default to 1 for legacy
+	const trackId = container.kind === "cmaf" ? container.trackId : 1;
 
 	// ftyp - File Type Box
 	const ftyp: FileTypeBox = {
@@ -441,10 +442,13 @@ export function createVideoInitSegment(config: Catalog.VideoConfig): Uint8Array 
  * Supports AAC (mp4a) and Opus codecs.
  */
 export function createAudioInitSegment(config: Catalog.AudioConfig): Uint8Array {
-	const { sampleRate, numberOfChannels, description, codec } = config;
+	const { sampleRate, numberOfChannels, description, codec, container } = config;
 
-	const timescale = 1_000_000; // microseconds
-	const trackId = 1;
+	// Use timescale from CMAF container, or microseconds for legacy
+	const timescale = container.kind === "cmaf" ? container.timescale : 1_000_000;
+
+	// Use track_id from CMAF container, or default to 1 for legacy
+	const trackId = container.kind === "cmaf" ? container.trackId : 1;
 
 	// ftyp - File Type Box
 	const ftyp: FileTypeBox = {
--- a/js/hang/src/watch/mse/source.ts
+++ b/js/hang/src/watch/mse/source.ts
@@ -133,9 +133,9 @@ export class Source implements Backend {
 					await new Promise((resolve) => sourceBuffer.addEventListener("updateend", resolve, { once: true }));
 				}
 
-				// Keep at least 1 second of buffered data.
-				if (element.currentTime > 1) {
-					sourceBuffer.remove(0, element.currentTime - 1);
+				// Keep at least 10 seconds of buffered data to avoid removing I-frames.
+				if (element.currentTime > 10) {
+					sourceBuffer.remove(0, element.currentTime - 10);
 				}
 			}
 		}, 1000);
--- a/js/hang/src/watch/mse/video.ts
+++ b/js/hang/src/watch/mse/video.ts
@@ -166,7 +166,7 @@ export class Video implements Backend {
 		effect.cleanup(() => data.close());
 
 		effect.spawn(async () => {
-			// Generate init segment from catalog config (always uses track_id=1)
+			// Generate init segment from catalog config (uses track_id from container)
 			const initSegment = Mp4.createVideoInitSegment(selected.config);
 			await this.#appendBuffer(sourceBuffer, initSegment);
 
@@ -176,8 +176,6 @@ export class Video implements Backend {
 				const frame = await data.readFrame();
 				if (!frame) return;
 
-				// Rewrite track_id to 1 to match our generated init segment
-				Mp4.rewriteTrackId(frame);
 				await this.#appendBuffer(sourceBuffer, frame);
 
 				// Seek to the start of the buffer if we're behind it (for startup).
--- a/rs/hang/src/catalog/container.rs
+++ b/rs/hang/src/catalog/container.rs
@@ -22,5 +22,8 @@ pub enum Container {
 	Cmaf {
 		/// Time units per second
 		timescale: u64,
+		/// Track ID used in the moof/mdat fragments
+		#[serde(rename = "trackId")]
+		track_id: u32,
 	},
 }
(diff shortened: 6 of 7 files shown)


js/hang/src/frame.ts
```diff
@@ -282,3 +275,86 @@ export class Consumer {
...
+		// NOTE: Precision loss above 2^52
+		v = Number(view.getBigUint64(0) & 0x3fffffffffffffffn);
+	} else {
+		throw new Error("impossible");
+	}
+
+	return [v, remain];
+}
+
+function encodeVarInt(v: number): Uint8Array {
```
jdreetz, line 333: should we have tests for any of this stuff?

js/hang-demo/src/index.html
```diff
@@ -28,10 +28,14 @@
 
 		NOTE: `reload` will detect when the broadcast goes offline/online and automatically reconnect.
 		TODO: Cloudflare doesn't support it yet (SUBSCRIBE_NAMESPACE), so make sure you remove it if you're using Cloudflare.
+
+		NOTE: You can use a <video> element instead of a <canvas> element to use the MSE backend.
```
jdreetz, line 30: I would put this comment above the video tag.

js/hang/src/watch/backend.ts
```diff
@@ -0,0 +1,175 @@
...
+	config = new Signal<Catalog.AudioConfig | undefined>(undefined);
+
+	// The stats of the audio.
+	stats = new Signal<Audio.Stats | undefined>(undefined);
+}
+
+/// A generic backend that supports either MSE or WebCodecs based on the provided element.
+///
+/// This is primarily what backs the <hang-watch> web component, but it's useful as a standalone for other use cases.
+export class Combined implements Backend {
```
jdreetz, line 106: Is `Combined` gonna make sense as a name long term? What about like `MultiSourceBackend` or something slightly more descriptive?

js/hang/src/watch/mse/cmaf.ts
```diff
@@ -0,0 +1,708 @@
...
+ * ```
+ */
+export function createVideoInitSegment(config: Catalog.VideoConfig): Uint8Array {
+	const { codedWidth, codedHeight, description } = config;
+	if (!codedWidth || !codedHeight || !description) {
+		// TODO: We could
+		throw new Error("Missing required fields to create video init segment");
+	}
+
+	const timescale = 1_000_000; // microseconds
```
jdreetz: should we use the `Time.Micro` type here?

js/hang/src/watch/mse/cmaf.ts
```diff
@@ -0,0 +1,708 @@
...
+	};
+
+	// dinf - Data Information Box
+	const dinf: DataInformationBox = {
+		type: "dinf",
+		boxes: [dref],
+	};
+
+	// Build codec-specific sample entry
+	const sampleEntry = createAudioSampleEntry(codec, sampleRate, numberOfChannels, description);
```
jdreetz: can you explain the point of a sample entry maybe?

js/hang/src/watch/mse/audio.ts
```diff
@@ -0,0 +1,185 @@
...
+		if (!mediaSource) return;
+
+		const broadcast = effect.get(this.broadcast);
+		if (!broadcast) return;
+
+		const active = effect.get(broadcast.active);
+		if (!active) return;
+
+		const selected = effect.get(this.#selected);
+		if (!selected) return;
```
jdreetz, line 124: it'd be cool if we could specify all these "defined" checks as a dependency for the method execution. Maybe something to add to the Signals lib.

js/hang/src/watch/mse/audio.ts
```diff
@@ -0,0 +1,185 @@
...
+		if (selected.config.container.kind === "cmaf") {
+			init = active.subscribe(selected.config.container.init_track.name, Catalog.PRIORITY.audio);
+			effect.cleanup(() => init?.close());
+		}
+
+		const data = active.subscribe(selected.track, Catalog.PRIORITY.audio);
+		effect.cleanup(() => data.close());
+
+		effect.spawn(async () => {
+			if (init) {
```
jdreetz: should we also check if `init.readFrame` is defined?

js/hang/src/watch/video/source.ts
```diff
@@ -1,61 +1,44 @@
...
-type SyncStatus = {
-	state: "ready" | "wait";
-	bufferDuration?: number;
-};
-
-export interface VideoStats {
-	frameCount: number;
-	timestamp: number;
-	bytesReceived: number;
-}
```
jdreetz, line 43: This is gonna cause the Stats panel to break for now, right? cc: @pzanella

js/hang/src/watch/video/source.ts
```diff
@@ -65,67 +48,76 @@ export class Source {
...
+	readonly frame: Getter<VideoFrame | undefined> = this.#frame;
 
 	// The target latency in milliseconds.
 	latency: Signal<Time.Milli>;
 
 	// The display size of the video in pixels, ideally sourced from the catalog.
-	display = new Signal<{ width: number; height: number } | undefined>(undefined);
-
-	// Whether to flip the video horizontally.
-	flip = new Signal<boolean | undefined>(undefined);
```
jdreetz, line 87: was this being used? did this get moved somewhere else?

js/hang/src/watch/video/source.ts
```diff
@@ -265,32 +257,23 @@ export class Source {
...
 					await new Promise((resolve) => setTimeout(resolve, sleep));
 				}
 
-				if (sleep > MIN_SYNC_WAIT_MS) {
-					// Include how long we slept if it was above the threshold.
-					this.syncStatus.set({ state: "ready", bufferDuration: sleep });
-				} else {
-					this.syncStatus.set({ state: "ready" });
-
+				if (sleep <= BUFFERING_MS) {
```
jdreetz, line 287: could we add a comment about why this is relevant?

js/hang-ui/src/watch/components/BufferControl.tsx
```diff
@@ -0,0 +1,136 @@
...
+		const leftPct = (visibleStartMs / maxRange()) * 100;
+		const widthPct = Math.max(0.5, ((visibleEndMs - visibleStartMs) / maxRange()) * 100);
+		const isOverflow = endMs > maxRange();
+		const overflowSec = isOverflow ? ((endMs - visibleStartMs) / 1000).toFixed(1) : null;
+		return {
+			style: `left: ${leftPct}%; width: ${widthPct}%; background: ${color};`,
+			isOverflow,
+			overflowSec,
+		};
+	};
```
jdreetz, line 32: can / should we type these?

js/hang-ui/src/watch/components/BufferControl.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+	// Determine color based on gap detection and buffering state
+	const rangeColor = (index: number, isBuffering: boolean) => {
+		if (isBuffering) return "#f87171"; // red
+		if (index > 0) return "#facc15"; // yellow
+		return "#4ade80"; // green
```
jdreetz, line 38: it'd be good to use css variables for this eventually, keep all that stuff in one place, make it easy to theme.

js/hang-ui/src/watch/components/BufferControl.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+	const updateBufferFromMouseX = (clientX: number) => {
+		if (!containerRef) return;
+		const rect = containerRef.getBoundingClientRect();
+		const trackWidth = rect.width - LABEL_WIDTH;
+		const x = Math.max(0, Math.min(clientX - rect.left - LABEL_WIDTH, trackWidth));
+		const ms = (x / trackWidth) * maxRange();
+		const snapped = Math.round(ms / RANGE_STEP) * RANGE_STEP;
+		const clamped = Math.max(MIN_RANGE, Math.min(maxRange(), snapped));
+		context.setLatencyValue(clamped);
```
jdreetz, line 56: we should rename this to `setBufferAmount` eventually.

js/hang/src/frame.ts
kixelated: IMO We should export this from `@moq/lite` and yeah add some tests.

js/hang/src/watch/backend.ts
kixelated: `MultiBackend` is nicer, yeah.

js/hang/src/watch/mse/cmaf.ts
kixelated: Also if you have any opinions on the timer helpers. It's a little annoying how much casting is required because the math operations convert them back to numbers.

js/hang/src/watch/mse/audio.ts
kixelated: A bulk way of doing this, yeah. Maybe something like: 
```js
this.signals.effect([
  this.element,
  this.mediaSource,
  this.broadcast,
  this.active,
  this.#selected
], (element, mediaSource, broadcast, active, selected) => {
  ...
});
```

My main goal is to avoid the ability to fuck up like React effects (forget to specify a dependency). I also don't like the magic of SolidJS signals; it's nice to explicitly call out what will cause the function to refire. Totally open to any opinions you have.

js/hang/src/watch/video/source.ts
kixelated: I was using this for `hang.live` but it's better to use `this.catalog.flip` instead.

js/hang-ui/src/watch/components/BufferControl.tsx
kixelated: Sure; this was vibe coded.

js/hang-ui/src/watch/components/BufferControl.tsx
kixelated: Yeah, and I would like to report the total latency too, not just the configurable buffer size.

js/hang/src/watch/mse/audio.ts
jdreetz: > My main goal is to avoid the ability to fuck up like React effects (forget to specify a dependency)

oh, I see what you mean. I approved the other PR, but it's up to you if you think there's enough value in not missing dependencies vs the additional boilerplate.

js/hang/src/watch/mse/cmaf.ts
jdreetz: > It's a little annoying how much casting is required because the math operations convert them back to numbers

I've used alias types for stuff like this too, e.g. `type Microseconds = numbers`. You don't have to cast everywhere (I think) but you get the benefit of additional readability.

PR #866, opened 2026-01-26
feat: add hang-ui types

pzanella:
Provides TypeScript types for Hang UI.


js/hang-ui/src/types/global.d.ts
```diff
@@ -0,0 +1,20 @@
+declare global {
```
kixelated: Afaik this is generally the "correct" way of registering WebComponents:
https://github.com/moq-dev/moq/blob/78bc805a7683afdcaa1135a7dbebec6359b8528a/js/hang/src/watch/element.ts#L215

As a bonus, I don't think any of this file is needed? It'll get rolled into the exported `.d.ts`. It's less prone to getting out of sync too.

js/hang-ui/package.json
```diff
@@ -5,9 +5,16 @@
 	"description": "Media over QUIC library UI components",
 	"license": "(MIT OR Apache-2.0)",
 	"repository": "github:moq-dev/moq",
+	"types": "./src/types/global.d.ts",
```
kixelated: I think this is wrong because we have separate `exports`.

js/hang-ui/src/types/global.d.ts
pzanella: Thanks for the feedback! I will move the type declarations to the component files using `HTMLElementTagNameMap` as you suggested. 
Something like this:
``` 
declare global {
	interface HTMLElementTagNameMap {
		"hang-watch-ui": HTMLElement;
	}
}
```
We don't have custom props, so `HTMLElement` should be enough. What do you think?
I'll also remove the unnecessary files. Thank you!

PR #871, opened 2026-01-28
Improve video rendering and sync

kixelated:
The b-frames "support" was well intentioned, but introduced a lot of latency. The main problem is that #ref is computed AFTER queuing, we only pop from the queue after sleeping for some amount. This means we were rendering frames significantly later than intended.

Instead of moving the ref computation before adding to the queue, I'd like to remove the queue. However, this means that frames are rendered in parallel. This is annoying but also means we support b-frames automatically. In fact maybe they just work now, who knows.

The `syncStatus` is racey because of this change, and it was pretty wrong in the first place. I believe it would only flag when a frame needed to be rendered in the future, which is the exact opposite of what we wanted it to do anyway. Any buffering detection can't be keyed off frames received anyway; it should instead spin if frames are not forthcoming.

### DIFF
--- a/js/hang/src/watch/sync.ts
+++ b/js/hang/src/watch/sync.ts
@@ -1,4 +1,4 @@
-import { Time } from "@moq/lite";
+import type { Time } from "@moq/lite";
 import { Effect, Signal } from "@moq/signals";
 import type * as Catalog from "../catalog";
 
@@ -87,9 +87,7 @@ export class Sync {
 	// Sleep until it's time to render this frame.
 	//
 	// Returns the amount of time we tried to sleep.
-	async wait(timestamp: Time.Milli): Promise<Time.Milli> {
-		let slept = Time.Milli.zero;
-
+	async wait(timestamp: Time.Milli): Promise<void> {
 		if (!this.#reference) {
 			throw new Error("reference not set; call update() first");
 		}
@@ -101,13 +99,15 @@ export class Sync {
 			const ref = (now - timestamp) as Time.Milli;
 
 			const sleep = this.#reference - ref + this.#latency.peek();
-			if (sleep <= 0) return slept;
+			if (sleep <= 0) return;
 			const wait = new Promise((resolve) => setTimeout(resolve, sleep)).then(() => true);
 
 			const ok = await Promise.race([this.#update, wait]);
-			slept = (slept + sleep) as Time.Milli;
-
-			if (ok) return slept;
+			if (ok) return;
 		}
+	}
+
+	close() {
+		this.signals.close();
 	}
 }
--- a/js/hang/src/watch/video/source.ts
+++ b/js/hang/src/watch/video/source.ts
@@ -224,6 +224,12 @@ export class Source {
 					const ok = await Promise.race([wait, effect.cancel]);
 					if (!ok) return;
 
+					if (timestamp < (this.#timestamp.peek() ?? 0)) {
+						// Late frame, don't render it.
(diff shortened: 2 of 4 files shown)


js/hang/src/watch/video/source.ts
```diff
@@ -83,24 +65,20 @@ export class Source {
...
-	// Additional buffer in milliseconds (on top of catalog's minBuffer).
-	buffer: Signal<Time.Milli>;
+	// The timestamp of the current frame.
+	#timestamp = new Signal<Time.Milli | undefined>(undefined);
+	readonly timestamp: Getter<Time.Milli | undefined> = this.#timestamp;
 
-	// Computed latency = catalog.minBuffer + buffer
-	#latency: Latency;
-	readonly latency: Getter<Time.Milli>;
+	// Additional buffer in milliseconds (on top of catalog's minBuffer).
```
jdreetz, line 72: is this comment accurate?

js/hang/src/watch/video/source.ts
```diff
@@ -227,83 +199,43 @@ export class Source {
...
+						// Late frame, don't render it.
+						return;
+					}
 
-				if (sleep > MIN_SYNC_WAIT_MS) {
-					this.syncStatus.set({ state: "wait", bufferDuration: sleep });
-				}
+					if (this.frame.peek() === undefined) {
+						// Render something while we wait for the sync to catch up.
+						this.frame.set(frame.clone());
```
jdreetz, line 220: 👍

js/hang/src/watch/video/source.ts
```diff
@@ -227,83 +199,43 @@ export class Source {
...
+					}
 
-				if (sleep > 0) {
-					// NOTE: WebCodecs doesn't block on output promises (I think?), so these sleeps will occur concurrently.
-					// TODO: This cause the `syncStatus` to be racey especially
-					await new Promise((resolve) => setTimeout(resolve, sleep));
-				}
+					const wait = this.sync.wait(timestamp).then(() => true);
```
jdreetz, line 223: i'd leave a comment about what's this is doing.

js/hang/src/watch/sync.ts
```diff
@@ -0,0 +1,113 @@
...
+import { Effect, Signal } from "@moq/signals";
+import type * as Catalog from "../catalog";
+
+export interface SyncProps {
+	buffer?: Time.Milli | Signal<Time.Milli>;
+	audio?: Catalog.AudioConfig | Signal<Catalog.AudioConfig | undefined>;
+	video?: Catalog.VideoConfig | Signal<Catalog.VideoConfig | undefined>;
+}
+
+export class Sync {
```
jdreetz, line 11: what is this syncing from/to? Broadcast? Should it be called BroadcastSync?

jdreetz (commented):
I _think_ these changes are good. Just left a few nits about tweaking names/comments.

js/hang/src/watch/sync.ts
kixelated: Right now yes, but for tree-shaking reasons I want to move stuff out of `Broadcast`. You can technically use this `Sync` class across multiple broadcasts too if they decide to use the same timestamp clock.

kixelated: ex. the `MSE.Video` and `WebCodecs.Video` modules would take `Catalog` and `Sync` modules. Inverting the hierarchy means we can shake.

jdreetz: > modules would take Catalog and Sync modules
Probably makes testing easier too with dependency injection.

PR #873, opened 2026-01-28
moq-relay: Skip jwt query param when no token configured

kixelated:
## Summary
- Don't append `?jwt=` to cluster URLs when no token is configured
- Results in cleaner URLs without an empty query parameter

## Test plan
- [ ] Verify relay starts without `--cluster-token` and connects to other nodes without `?jwt=` in URL
- [ ] Verify relay with `--cluster-token` still includes the JWT in the query parameter


kixelated:
Bonus fix for #868

PR #879, opened 2026-01-29
Reduce some boilerplate with getAll

kixelated:
A bulk way of checking if signals are null.

### DIFF
--- a/js/hang/src/publish/preview.ts
+++ b/js/hang/src/publish/preview.ts
@@ -24,16 +24,13 @@ export class Preview {
 		this.info = Signal.from(props?.info);
 
 		this.signals.effect((effect) => {
-			const enabled = effect.get(this.enabled);
-			if (!enabled) return;
-
+			if (!effect.get(this.enabled)) return;
 			effect.set(this.catalog, { name: Preview.TRACK, priority: Preview.PRIORITY });
 		});
 	}
 
 	serve(track: Moq.Track, effect: Effect): void {
-		const enabled = effect.get(this.enabled);
-		if (!enabled) return;
+		if (!effect.get(this.enabled)) return;
 
 		const info = effect.get(this.info);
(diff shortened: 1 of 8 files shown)


js/hang/src/publish/audio/encoder.ts
```diff
@@ -153,16 +153,12 @@ export class Encoder {
 	}
 
 	serve(track: Moq.Track, effect: Effect): void {
-		if (!effect.get(this.enabled)) return;
-
-		const source = effect.get(this.source);
```
kixelated, line 158: Wasn't being used, oops? `getAll`  can catch unused variables.

js/hang/src/publish/audio/encoder.ts
```diff
@@ -153,16 +153,12 @@ export class Encoder {
...
-		if (!source) return;
-
-		const worklet = effect.get(this.#worklet);
-		if (!worklet) return;
+		const enabled = effect.get(this.enabled);
+		if (!enabled) return;
 
-		const config = effect.get(this.#config);
-		if (!config) return;
+		const values = effect.getAll([this.#worklet, this.#config]);
```
kixelated, line 159: If we make this check for truthy instead of null/undefined, this can also check enabled. I'm not sure which to do.

js/hang/src/watch/audio/source.ts
```diff
@@ -145,17 +144,9 @@ export class Source {
...
-
-		const broadcast = effect.get(this.broadcast);
-		if (!broadcast) return;
-
-		const config = effect.get(this.config);
-		if (!config) return;
-
-		const active = effect.get(this.active);
-		if (!active) return;
+		const values = effect.getAll([this.catalog, this.broadcast, this.config, this.active]);
```
jdreetz, line 147: 🤌

PR #881, opened 2026-01-29
Export Varint module from @moq/lite

kixelated:
Add a new Varint module to @moq/lite that provides QUIC variable-length integer encoding/decoding functions. This eliminates duplicate varint implementations that were previously scattered across the codebase.

- Add js/lite/src/varint.ts with encode, decode, encodeTo, encode62To, and size functions
- Export as Varint namespace from @moq/lite
- Refactor stream.ts to use the new Varint module
- Update legacy.ts to import Varint from @moq/lite instead of local copy
- Add comprehensive unit tests for the varint module

https://claude.ai/code/session_016LVyh4Am8LVxV5ieLgYQrW

### DIFF
--- a/js/hang/src/container/legacy.ts
+++ b/js/hang/src/container/legacy.ts
@@ -35,7 +35,7 @@ export class Producer {
 	}
 
 	static #encode(source: Uint8Array | Source, timestamp: Time.Micro): Uint8Array {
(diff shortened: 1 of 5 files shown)


kixelated:
@coderabbitai review

PR #906, opened 2026-02-04
Use `moq` instead of `hang` for some crates

kixelated:
The goal is that `hang` is the catalog/container format, not the library. This broadens the scope to support more formats in the future.

- rename `hang-cli` to `moq-cli`
- split `hang` into `moq-mux`


kixelated:
@coderabbitai review

PR #919, opened 2026-02-06
Fix multi node clusters

jhiatt-verkada:
This is not intended to be mergeable but serves as a proof-of-concept where all cluster node announcements are published on the root node. The cluster leaves specify a query parameter indicating what hostname to publish.

### DIFF
--- a/dev/leaf.toml
+++ b/dev/leaf.toml
@@ -5,7 +5,7 @@ level = "debug"
 
 [server]
 # Listen for QUIC connections on UDP:4444
-listen = "0.0.0.0:4444"
+listen = "[::]:4444"
 
 # Generate a self-signed certificate for the server.
 # You should use a real certificate in production.
@@ -13,7 +13,7 @@ tls.generate = ["localhost"]
 
 [web.http]
 # Listen for HTTP and WebSocket (TCP) connections on the given address.
-listen = "0.0.0.0:4444"
+listen = "[::]:4444"
 
 # This clustering scheme is very very simple for now.
 #
@@ -64,5 +64,9 @@ tls.disable_verify = true
 # `just auth-key` will populate this file.
 key = "dev/root.jwk"
 
-# Allow anonymous publishing and subscribing for this prefix.
-public = "anon"
+# NOTE: Cannot have both key and public configured together in newer moq-relay
+# public = "anon"
+
+[iroh]
+# Disable iroh for local testing
+enabled = false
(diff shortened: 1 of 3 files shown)


rs/moq-token/src/claims.rs
```diff
@@ -62,7 +62,7 @@ pub struct Claims {
 	pub subscribe: Vec<String>,
 
 	/// The expiration time of the token as a unix timestamp.
-	#[serde(rename = "exp")]
+	#[serde(rename = "exp", skip_serializing_if = "Option::is_none")]
```
jhiatt-verkada, line 65: Not sure if anything changed but the token validation code was choking on tokens with `"exp": null`.

rs/moq-relay/src/cluster.rs
```diff
@@ -234,9 +243,14 @@ impl Cluster {
...
 	async fn run_remote(mut self, node: &str, token: String, origin: BroadcastConsumer) -> anyhow::Result<()> {
+		let cluster_node = if Some(node) == self.config.root.as_ref().map(|r| r.as_str()) {
+			self.config.node.clone().unwrap_or_default()
+		} else {
+			"".to_string()
+		};
 		let url = match token.is_empty() {
 			true => Url::parse(&format!("https://{node}/"))?,
-			false => Url::parse(&format!("https://{node}/?jwt={token}"))?,
+			false => Url::parse(&format!("https://{node}/?jwt={token}&node={cluster_node}"))?,
```
jhiatt-verkada, line 253: This should be properly escaped

rs/moq-relay/src/cluster.rs
```diff
@@ -137,7 +146,7 @@ impl Cluster {
 
 		// Announce ourselves as an origin to the root node.
 		// This goes into primary so it gets shared with other nodes via run_remote_once.
-		if let Some(myself) = self.config.node.as_ref() {
+		if false && let Some(myself) = self.config.node.as_ref() {
```
jhiatt-verkada, line 149: Hack to disable existing publish (should just be deleted)

rs/moq-relay/src/connection.rs
```diff
@@ -62,6 +66,15 @@ impl Connection {
 			.accept()
 			.await?;
 
+		let _node_announce = if let Some(cluster_node) = cluster_node
```
jhiatt-verkada, line 69: Gets dropped when session ends -> node is unpublished

PR #920, opened 2026-02-06
AV1 decoder

christriants:
Related to: #740 
MoQ OBS Plugin update: moq-dev/obs#20

This PR adds AV1 (av01) decoder support with inline sequence headers, following the format of the `avc3` decoder. Key implementation details:

Core Decoder (av01.rs):
- Dual format support: Handles both av1C (MP4 container format) and raw OBU streams
  - `initialize()` - Detects format by examining first bytes: 0x0a signature with 16-byte length indicates av1C, otherwise treats as raw OBUs
  - `init_from_av1c()` - Parses 16-byte `av1C` box to extract codec parameters (profile, level, tier, etc). Resolution is set to None and deferred to inline sequence headers since `av1C` doesn't contain dimension info
  - `decode_obu()` - Processes individual Open Bitstream Units, handling sequence headers, frame headers, temporal delimiters, and tile groups

- `av1C` initialization: Parses 16-byte av1C box to extract codec parameters (profile, level, tier, bit depth, chroma subsampling), defers resolution extraction to inline sequence headers
- Sequence header handling: Uses `scuffle_av1` crate to parse inline sequence headers for resolution and color configuration

- Frame detection:
  - Parses AV1 uncompressed header to identify frame types (bits 5-6 of first byte after OBU header)
  - Handles `show_existing_frame` flag (bit 7) which OBS uses for display reordering - preserves keyframe state from preceding sequence header rather than parsing referenced frame
  - `maybe_start_frame()` - Flushes completed frames to track when temporal delimiter or metadata OBU is encountered

- OBU iteration: Custom iterator that parses LEB128-encoded size fields to split OBU boundaries in streams

API Alignment:
- Implements existing `new()`, `initialize()`, `decode_stream()`, and `decode_frame() `methods

A couple of bugs I am trying to sort out / understand:
- Frontend console error: `invalidStateError: Failed to execute 'clone' on 'VideoFrame': Cannot clone closed VideoFrame.
    at #runRender (renderer.ts:90:29)` - I think we can fix this separately. 

Screenshot 2026-02-06 at 1 05 28 PM

### DIFF
--- a/rs/moq-mux/src/import/av01.rs
+++ b/rs/moq-mux/src/import/av01.rs
@@ -245,11 +245,11 @@ impl Av01 {
 		buf: &mut T,
 		pts: Option<hang::container::Timestamp>,
 	) -> anyhow::Result<()> {
-		let pts = self.pts(pts)?;
-
 		let obus = ObuIterator::new(buf);
 
 		for obu in obus {
+			// Generate PTS for each OBU to avoid reusing same timestamp
+			let pts = self.pts(pts)?;
 			self.decode_obu(obu?, Some(pts))?;
 		}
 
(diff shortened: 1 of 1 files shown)


rs/moq-mux/src/import/av01.rs
```diff
@@ -0,0 +1,443 @@
...
+				full_range: false,
+			}
+			.into(),
+			description: None,
+			framerate: None,
+			bitrate: None,
+			display_ratio_width: None,
+			display_ratio_height: None,
+			optimize_for_latency: None,
+			container: hang::catalog::Container::Legacy,
```
christriants, line 84: Is `Legacy` correct here?

kixelated: It's basically a big fat `TODO` to future me.

At the very least timestamps should be delta encoded.

kixelated:
> * Frontend console error: `invalidStateError: Failed to execute 'clone' on 'VideoFrame': Cannot clone closed VideoFrame. at #runRender (renderer.ts:90:29)` - I think we can fix this separately.

That's an existing race condition. I'm not sure what causes it, but it's harmless.

rs/moq-mux/src/import/decoder.rs
```diff
@@ -97,6 +107,8 @@ impl FromStr for StreamFormat {
 			"hev1" => Ok(StreamFormat::Hev1),
 			#[cfg(feature = "mp4")]
 			"fmp4" | "cmaf" => Ok(StreamFormat::Fmp4),
+			#[cfg(feature = "av1")]
+			"av01" | "av1" => Ok(StreamFormat::Av01),
```
kixelated: Maybe throw in `av1C` too since both are supported.

PR #931, opened 2026-02-07
docs: native client integration guide

alltheseas:
## Summary

- **`doc/rs/env/native.md`**: Fill in the empty file with a guide for building native MoQ clients in Rust — connecting, publishing, subscribing, reading catalogs/frames, platform decoders, and common pitfalls.
- **`doc/concept/layer/hang.md`**: Add video `description` field semantics (AVCC vs Annex B) and group-to-keyframe relationship.
- **`doc/app/relay/http.md`**: Add late-join pattern using HTTP group fetch (`?group=N`).

Written from experience integrating MoQ into a native Rust video player. All code examples reference the current API (`OrderedConsumer`, `CatalogConsumer`, `Origin::produce()`, etc.).

### DIFF
--- a/doc/app/relay/http.md
+++ b/doc/app/relay/http.md
@@ -1,38 +1,35 @@
 ---
-title: HTTP Endpoints
-description: Debug endpoints exposed by moq-relay
+title: HTTP
+description: HTTP endpoints exposed by moq-relay
 ---
 
 # HTTP Endpoints
 
-moq-relay exposes HTTP endpoints for debugging and diagnostics. These run on TCP alongside the QUIC/UDP server.
-
-::: warning
-The HTTP server is unencrypted and intended for local debugging only. Don't expose it to the public internet.
-:::
+moq-relay exposes HTTP/HTTPS endpoints via TCP too.
+These were initially added for debugging but are useful for many things, such as fetching old content.
 
 ## Configuration
 
+The relay supports both HTTP and HTTPS, configured independently:
+
 ```toml
 [web.http]
-# Listen for HTTP connections on TCP
-listen = "0.0.0.0:4443"
+# Listen for unencrypted HTTP connections on TCP
+listen = "0.0.0.0:80"
+
+[web.https]
+# Listen for encrypted HTTPS connections on TCP
+listen = "0.0.0.0:443"
+cert = "cert.pem"
+key = "key.pem"
 ```
 
-The default is the same port as the QUIC server, just on TCP instead of UDP.
+::: warning
+HTTP is unencrypted, which means any [authentication tokens](/app/relay/auth) will be sent in plaintext.
+It's recommended to only use HTTPS in production.
+:::
 
-## Endpoints
-
-### GET /certificate.sha256
-
-Returns the SHA-256 fingerprint of the TLS certificate. Useful for local development with self-signed certificates.
-
-```bash
-curl http://localhost:4443/certificate.sha256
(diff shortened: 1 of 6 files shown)


alltheseas:
Hope this is a good starting point to build the native docs. Thanks for all the guidance

doc/rs/env/native.md
```diff
@@ -0,0 +1,283 @@
...
+For AAC-LC audio, [symphonia](https://crates.io/crates/symphonia) decodes to PCM samples and [cpal](https://crates.io/crates/cpal) handles platform audio output.
+For Opus, symphonia also supports decoding, or use the `opus` crate directly.
+
+Use a ring buffer between the decoder and audio output to absorb network jitter.
+
+## Common Pitfalls
+
+### Missing Keyframe on Late Join
+
+When joining a live stream mid-broadcast, you may receive delta frames before a keyframe arrives.
```
kixelated: Not with `moq-lite`, we always start subscriptions at group boundaries and each group starts with a keyframe. There's no way to receive a frame without first receiving its dependencies.

But `moq-transport` works like this for some reason and I hate it.

doc/rs/env/native.md
```diff
@@ -0,0 +1,283 @@
...
+```
+
+See the [Authentication guide](/app/relay/auth) for how to generate tokens.
+
+## Publishing
+
+The [video example](https://github.com/moq-dev/moq/blob/main/rs/hang/examples/video.rs) demonstrates publishing end-to-end.
+The key pattern is: create an `Origin`, connect a session to it, then publish broadcasts:
+
+```rust
```
kixelated, line 66: I try to put examples in an `examples` folder so we make sure they actually compile.

doc/concept/layer/hang.md
```diff
@@ -91,3 +91,41 @@ Unfortunately, fMP4 is not quite designed for real-time streaming and incurs eit
...
+:::
+
+## Groups and Keyframes
+
+Each MoQ group aligns with a video Group of Pictures (GoP).
+A new group starts with a keyframe (IDR frame) that can be decoded independently.
+
+This has important implications:
+- **Skipping a group means skipping an entire GoP.** The relay can drop old groups without corrupting the decoder state.
+- **Late-join viewers** need the first frame of a group (the keyframe) to start decoding. Joining mid-group produces corrupted output until the next group boundary.
```
kixelated: Not possible to join mid group with `moq-lite`.

doc/app/relay/http.md
```diff
@@ -107,7 +107,36 @@ The HTTP endpoints can serve as basic health checks:
...
+```
+
+The `?group=N` parameter requests a specific group by its sequence number.
+Without it, you get the most recent group (the live edge).
+
+**Typical late-join flow:**
+
+1. Subscribe to the video track via MoQ to start receiving live groups.
+2. Note the first group sequence number you receive (e.g. `N`).
+3. Fetch group `N-1` via HTTP to get the keyframe for the previous GoP.
```
kixelated: Not needed unless you want a larger jitter buffer.

alltheseas:
> Thanks! I'm having claude go through and make some changes.
> 
> And yeah let me know if you're running into an issue with missing keyframes. We always serve the latest group and all frames in order, even if they were generated in the past. It shouldn't be possible to receive a delta frame without its keyframe.

Sounds good - I'll defer to you to update the docs as needed

PR #932, opened 2026-02-07
fix: add React-compatible webcomponent setters and type declarations

pzanella:
This PR makes `@moq/hang` and `@moq/hang-ui` web components work better with React.

- added setters for `url`, `path`, `jitter`. This allows React to pass data directly through props
- created separate type declarations for React and other frameworks

### DIFF
--- a/js/scripts/element-wrappers/README.md
+++ b/js/scripts/element-wrappers/README.md
@@ -6,7 +6,7 @@ Generates typed framework wrappers from [Custom Elements Manifest](https://custo
 
 ## Architecture
 
-```
+```text
 element-wrappers/
 ├── index.ts              # Orchestrator — runs all enabled generators
 ├── generators/


js/hang/src/publish/element.ts
```diff
@@ -229,4 +245,21 @@ declare global {
...
+
+declare module "react" {
+	namespace JSX {
+		interface IntrinsicElements {
+			"hang-publish": React.HTMLAttributes<HangPublish> & {
+				[K in Observed]?: string | number | boolean;
+			};
+		}
+	}
 }
```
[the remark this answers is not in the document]
pzanella: What about something like this?
```
declare module "react" {
	namespace JSX {
		interface IntrinsicElements {
			"hang-publish": import("react").HTMLAttributes<HangPublish> & {
				[K in Observed]?: string | number | boolean;
			};
		}
	}
}
```
Do you think this would fix the potential issue?

js/hang-ui/src/publish/index.tsx
```diff
@@ -1,27 +1,74 @@
 import type HangPublish from "@moq/hang/publish/element";
-import { customElement } from "solid-element";
```
jdreetz, line 2: Do we have to get rid of solid-element to support CEM? I thought it just needs the more detailed JS Docs comment? Although we're not making much use of it currently, solid-element makes handling setters and getters a lot easier than the vanilla Custom Element syntax, so I'd prefer to keep it if possible.

js/hang-ui/src/watch/index.tsx
```diff
@@ -1,26 +1,77 @@
 import type HangWatch from "@moq/hang/watch/element";
-import { customElement } from "solid-element";
```
jdreetz, line 2: same question as above re: removing solid-element.

js/hang-ui/src/watch/index.tsx
```diff
@@ -1,26 +1,77 @@
...
+class HangWatchComponent extends HTMLElement {
+	#root?: ShadowRoot;
+	#dispose?: () => void;
+	#connected = false;
 
-	onMount(async () => {
-		await customElements.whenDefined("hang-watch");
-		const watchEl = element.querySelector("hang-watch");
-		setNested(watchEl ? (watchEl as HangWatch) : undefined);
-	});
```
jdreetz, line 13: we don't need any of this anymore because of the `queueMicrotask` stuff below?

js/hang-ui/package.json
```diff
@@ -6,15 +6,26 @@
...
+			"default": "./src/publish/index.tsx"
+		},
+		"./watch": {
+			"types": "./watch/index.d.ts",
+			"default": "./src/watch/index.tsx"
+		},
+		"./react": {
+			"types": "./wrappers/react/index.d.ts",
+			"default": "./src/wrappers/react/index.ts"
+		}
```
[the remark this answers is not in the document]
jdreetz: @pzanella have you verified this?

js/scripts/element-wrappers/README.md
jdreetz, line 1: I think it's really cool we're adding support for all this, I'm just surprised there isn't another library that can replace all this transformation / code gen code. Maybe that's all coming in the future as people standardize on the methods though?

js/hang/README.md
```diff
@@ -218,6 +218,38 @@ useEffect(() => {
...
+This package uses Custom Elements Manifest (CEM) to automatically generate framework-specific wrappers.
+
+### React Wrappers
+
+Auto-generated from CEM, exported from `@moq/hang/react`:
+
+```tsx
+import { HangWatch, HangPublish } from '@moq/hang/react';
+
+<HangWatch url="https://relay.example.com" path="/stream" jitter={100} muted reload>
```
jdreetz, line 232:
```
HTML Attributes are strings.
Javacript properties are typed and reactive.

`<hang-watch volume="0.8" />` will work, but it's not type-safe.
You can use DOM callbacks to detect when the attribute changes but it's not as convenient.
```


^ Do we need to update any of the text above given the added CEM support? Or should we add a link or something to part of the document explaining how type safety is enabled through CEM?

js/hang-ui/README.md
```diff
@@ -37,15 +37,11 @@ Here's how you can use them (see also @moq/hang-demo for a complete example):
...
-			<video
-				style="width: 100%; height: auto; border-radius: 4px; margin: 0 auto;"
-				muted
-				autoplay
-			></video>
-		</hang-publish>
-	</hang-publish-ui>
+<hang-publish-ui>
+	<hang-publish url="<MOQ relay URL>" path="<relay path>">
+		<video style="width: 100%; height: auto; border-radius: 4px; margin: 0 auto;" muted autoplay></video>
```
jdreetz, line 42: we should add a comment above this explaining that using the `video` tag causes MSE to be used with MOQ.

js/hang-ui/vite.config.ts
```diff
@@ -1,19 +1,29 @@
...
+			entryRoot: "src",
+		}),
+	],
 	build: {
 		lib: {
 			entry: {
 				"publish/index": resolve(__dirname, "src/publish/index.tsx"),
 				"watch/index": resolve(__dirname, "src/watch/index.tsx"),
+				// Auto-generated by scripts/generate-wrappers.ts during prebuild phase
+				"wrappers/react/index": resolve(__dirname, "src/wrappers/react/index.ts"),
```
jdreetz, line 21: If a user adds another wrapper, will it get added here automatically during pre-build?

js/hang/src/watch/element.ts
```diff
@@ -17,7 +17,52 @@ type Observed = (typeof OBSERVED)[number];
...
+ * decoding, synchronization, and rendering. Supports configurable jitter buffer, volume control,
+ * pause/resume, and automatic reconnection.
+ *
+ * @attr {string} url - The WebTransport URL of the MOQ relay server (e.g., "https://relay.example.com")
+ * @attr {string} path - The broadcast path to subscribe to (e.g., "/live/stream")
+ * @attr {boolean} paused - Whether playback is paused
+ * @attr {number} volume - Audio volume level (0.0 to 1.0, default: 0.5)
+ * @attr {boolean} muted - Whether audio is muted
+ * @attr {boolean} reload - Whether to automatically reconnect on connection loss
+ * @attr {number} jitter - Target jitter buffer in milliseconds (default: 100)
```
jdreetz, line 34: Are these types correct? Aren't they returning primitive wrapped Signals?

js/scripts/element-wrappers/utils/codegen.ts
```diff
@@ -0,0 +1,167 @@
...
+ * Format a documentation section with consistent indentation
+ *
+ * @param title - Section title (e.g., "slots", "events")
+ * @param items - Array of items with name, description, and optional extra text
+ * @returns Array of formatted JSDoc lines
+ */
+function formatDocSection(
+	title: string,
+	items: Array<{ name: string; description?: string; extra?: string }>,
+): string[] {
```
jdreetz, line 37: can we type this any more specifically?

js/scripts/element-wrappers/utils/codegen.ts
```diff
@@ -0,0 +1,167 @@
...
+		...items.map((item) => {
+			const extra = item.extra ?? "";
+			const desc = item.description ? ` - ${escapeJSDoc(item.description)}` : "";
+			return ` * - \`${item.name}\`${extra}${desc}`;
+		}),
+	];
+}
+
+/**
+ * Generate JSDoc comment block from custom element metadata
```
jdreetz, line 51: this is generating CEM specific JSDoc comments, isn't it? Should we mention that?

js/scripts/element-wrappers/utils/codegen.ts
jdreetz, line 1: I'm surprised the functionality in this file isn't available in a library already yet, but maybe it's too new.

js/hang-ui/src/publish/index.tsx
pzanella: You're right, keeping `solid-element` for getters/setters makes sense. I'll revert those changes and focus on fixing the JSDoc comments for CEM support instead.

js/scripts/element-wrappers/README.md
pzanella: You're right, there are some NPM packages available, but they didn't really fit what I had in mind for our specific needs. Anyway, I’m waiting for your PR (#948) to be merged before I introduce the CEM support, since the structure has changed now. So, I'm going to close this PR for now and I'll open a fresh one once yours is merged. What do you think? Thank you!

js/hang/src/watch/element.ts
```diff
@@ -170,14 +215,26 @@ export default class HangWatch extends HTMLElement implements Backend {
...
+		value ? this.setAttribute("url", String(value)) : this.removeAttribute("url");
+	}
+
 	get path(): Signal<Moq.Path.Valid | undefined> {
 		return this.broadcast.path;
 	}
 
+	set path(value: string | Moq.Path.Valid | undefined) {
+		value ? this.setAttribute("path", String(value)) : this.removeAttribute("path");
+	}
```
kixelated, line 228: Currently, the usage pattern is:

```js
watch.path.set("foo")
watch.path.get()
watch.path.changed(() => { ... })
```

And you can use the React helper:
```js
import { react } from "@moq/signals"
const path = react(watch.path);
```

This PR would change it to:
```js
watch.path = "foo"
watch.path.get()
watch.path.set("foo") // still works
watch.path.changed(() => { ... })
```

I don't like the inconsistency there; I'd rather have an explicit `.set`.

js/scripts/element-wrappers/README.md
kixelated: Yeah I think this is too much. A react wrapper would be cool but this feels finicky and a maintenance overhead.

Maybe start with an explicit react component? If it's too much duplication, then we can venture into code generation.

jdreetz: The React wrapper and custom element types seem valuable to me, but yea, the code generation stuff feels like it might be too much at this point.

PR #948, opened 2026-02-12
Split UI components into dedicated packages

jdreetz:
## Split UI components into dedicated packages

Extracts watch, publish, and shared UI code out of `@moq/hang` into their own packages, establishing clearer boundaries between media logic and UI.

### Changes

- **`@moq/ui-core`** (new) — Shared UI primitives (Button, Icon, Stats panel) used by both watch and publish UIs. Uses structural interfaces for stats provider types to avoid depending on `@moq/watch` or `@moq/hang`.
- **`@moq/watch`** (new) — Watch/subscribe logic + optional `@moq/watch/ui` overlay (`<hang-watch-ui>`), built with Vite + SolidJS.
- **`@moq/publish`** (new) — Publish logic + optional `@moq/publish/ui` overlay (`<hang-publish-ui>`), built with Vite + SolidJS.
- **`@moq/hang`** — Removed all watch/publish/UI code. Now contains only the core media layer: catalog, container (CMAF + Legacy), and utilities.

### Build

Each package has two TypeScript configs:
- tsconfig.json — IDE + type-checking (includes UI)
- tsconfig.build.json — Declaration emit only (excludes UI, which Vite handles)

### Documentation

- Updated `doc/js/` pages and package READMEs to reflect the new structure
- Corrected UI web component examples to show the wrapping pattern
- Fixed license file paths

### Nix

- Added `git` and `cmake` to flake.nix `rustDeps` — required by `boring-sys` to build BoringSSL

### DIFF
--- a/js/hang/package.json
+++ b/js/hang/package.json
@@ -1,7 +1,7 @@
 {
 	"name": "@moq/hang",
 	"type": "module",
-	"version": "0.1.2",
+	"version": "0.1.3",
 	"description": "Media over QUIC library",
 	"license": "(MIT OR Apache-2.0)",
 	"repository": "github:moq-dev/moq",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -156,7 +156,7 @@ checksum = "3109e49b1e4909e9db6515a30c633684d68cdeaa252f215214cb4fa1a5bfee2c"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.114",
+ "syn 2.0.115",
  "synstructure",
 ]
 
@@ -168,7 +168,7 @@ checksum = "7b18050c2cd6fe86c3a76584ef5e0baf286d038cda203eb6223df2cc413565f7"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.114",
+ "syn 2.0.115",
 ]
 
 [[package]]
@@ -216,7 +216,7 @@ checksum = "9035ad2d096bed7955a320ee7e2230574d28fd3c3a0f186cbea1ff3c7eed5dbb"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.114",
+ "syn 2.0.115",
 ]
 
 [[package]]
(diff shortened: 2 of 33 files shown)


kixelated:
Oh sorry, the Nix stuff should have been fixed by #952, but it's still good to have. The quiche backend doesn't quite work so I disabled it by default.

js/hang/package.json
```diff
@@ -7,23 +7,20 @@
...
 		".": "./src/index.ts",
-		"./publish": "./src/publish/index.ts",
-		"./publish/element": "./src/publish/element.ts",
-		"./watch": "./src/watch/index.ts",
-		"./watch/element": "./src/watch/element.ts",
 		"./catalog": "./src/catalog/index.ts",
+		"./container": "./src/container/index.ts",
+		"./util/hex": "./src/util/hex.ts",
+		"./util/libav": "./src/util/libav.ts",
+		"./util/hacks": "./src/util/hacks.ts",
```
kixelated: Feels bad. Maybe just export `util`?

We could make separate packages for stuff like libav I guess.
Other stuff you could probably just copy-paste (ex. hex). Not a huge deal.

js/publish/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+	"extends": "../tsconfig.json",
+	"compilerOptions": {
+		"rootDir": ".",
+		"outDir": "dist",
+		"jsx": "preserve",
+		"jsxImportSource": "solid-js"
```
kixelated: hmm maybe this should be in `ui` instead.

js/ui-core/package.json
```diff
@@ -1,18 +1,17 @@
 {
-	"name": "@moq/hang-ui",
+	"name": "@moq/ui-core",
```
kixelated, line 2: nit: `@moq/theme`?

js/publish/vite.config.ts
```diff
@@ -0,0 +1,22 @@
...
+	build: {
+		lib: {
+			entry: {
+				"ui/index": resolve(__dirname, "src/ui/index.tsx"),
+			},
+			formats: ["es"],
+		},
+		rollupOptions: {
+			external: ["@moq/hang", "@moq/lite", "@moq/signals", "@moq/ui-core"],
+		},
```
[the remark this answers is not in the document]
kixelated: We don't expose solid-js or solid-element via the public API, so I'm not sure we should do this. Yes the bundle size might be larger if the user happens to also use the exact same solid-js version, but it's much easier. But I could be misunderstanding what external does.

js/hang/package.json
jdreetz: yea, I'll just export util.

jdreetz: Updated. The imports all use the namespace now too like the other items from that package.

js/publish/tsconfig.json
jdreetz: Updated.

js/ui-core/package.json
kixelated: Let's not for now.

PR #957, opened 2026-02-13
Fix building the @moq/publish and @moq/watch packages

kixelated:
1. AudioWorklets were not resolved at runtime. There's now a script to compile and inline them. In the long-term, we should serve stuff as public assets but there's no need right this moment.
2. Types were not being published.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -9,7 +9,6 @@
         "@types/bun": "^1.3.8",
         "concurrently": "^9.2.1",
         "publint": "^0.3.0",
-        "vite": "^7.3.1",
       },
     },
     "doc": {
@@ -66,9 +65,11 @@
       "devDependencies": {
         "@tailwindcss/typography": "^0.5.16",
         "@tailwindcss/vite": "^4.1.13",
+        "esbuild": "^0.27.0",
         "highlight.js": "^11.11.1",
         "tailwindcss": "^4.1.13",
         "typescript": "^5.9.2",
+        "vite": "^7.3.1",
         "vite-plugin-solid": "^2.11.10",
       },
     },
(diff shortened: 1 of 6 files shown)


js/common/package.ts
```diff
@@ -3,47 +3,59 @@
...
 
 if (pkg.exports) {
 	for (const key in pkg.exports) {
 		const val = pkg.exports[key];
 		if (typeof val === "string") {
-			pkg.exports[key] = rewritePath(val);
+			if (val.endsWith(".css")) {
+				// CSS exports are only needed for dev-time resolution;
+				// consumers inline them at build time via @import.
+				// We purposely do not copy them to the dist to help catch bugs.
```
kixelated, line 27: Eventually, we should roll them up and serve them out of `public` instead of inline.

js/common/package.ts
```diff
@@ -78,4 +90,19 @@ writeFileSync("dist/package.json", JSON.stringify(pkg, null, 2));
 console.log("📄 Copying README.md...");
 copyFileSync("README.md", join("dist", "README.md"));
 
+// Lint the package to catch publishing issues
+console.log("🔍 Running publint...");
+const { messages, pkg: lintPkg } = await publint({
```
kixelated, line 95: Catches bugs with the generated package.json file. ex. It was pointing to non-existent css files.

js/common/vite-plugin-worklet.ts
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * A Vite plugin that compiles AudioWorklet files and inlines them as blob URLs.
+ *
+ * Usage: import workletUrl from "./my-worklet.ts?worklet"
+ *
+ * The worklet file is compiled to JS with all dependencies bundled via esbuild,
+ * then inlined as a string. At runtime, a blob URL is created and exported.
+ * Pass the URL to audioWorklet.addModule().
+ */
+export function workletInline(): Plugin {
```
kixelated, line 15: Eventually, we should serve this out of `public` instead.

js/publish/tsconfig.json
```diff
@@ -1,10 +1,11 @@
 {
 	"extends": "../tsconfig.json",
 	"compilerOptions": {
-		"rootDir": ".",
-		"outDir": "dist"
+		"rootDir": "./src",
+		"jsx": "preserve",
```
kixelated, line 5: I didn't mean to make this change (claude pls) but I guess it's fine to have these JSX settings for the entire module.

PR #963, opened 2026-02-14
Rename `hang-` web components to `moq-`

kixelated:
Also split `hang-support` into pieces.

- `hang-watch` -> `moq-watch`
- `hang-watch-ui` -> `moq-watch-ui`
- `hang-publish` -> `moq-publish`
- `hang-publish-ui` -> `moq-publish-ui`
- `hang-support` ->
  - `moq-watch-support` 
  - `moq-publish-support`
- `hang-demo` -> `moq-demo`

### DIFF
--- a/js/demo/.env
+++ b/js/demo/.env
@@ -1,0 +1,1 @@
+VITE_RELAY_URL=http://localhost:4443/anon
--- a/js/demo/vite.config.ts
+++ b/js/demo/vite.config.ts
@@ -6,6 +6,7 @@ import { workletInline } from "../common/vite-plugin-worklet";
 
 export default defineConfig({
 	root: "src",
+	envDir: resolve(__dirname),
 	plugins: [tailwindcss(), solidPlugin(), workletInline()],
 	build: {
 		target: "esnext",
--- a/js/publish/vite.config.ts
+++ b/js/publish/vite.config.ts
@@ -11,6 +11,8 @@ export default defineConfig({
 				index: resolve(__dirname, "src/index.ts"),
 				element: resolve(__dirname, "src/element.ts"),
 				"ui/index": resolve(__dirname, "src/ui/index.tsx"),
+				"support/index": resolve(__dirname, "src/support/index.ts"),
+				"support/element": resolve(__dirname, "src/support/element.ts"),
 			},
 			formats: ["es"],
 		},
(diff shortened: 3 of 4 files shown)


js/demo/src/support.html
pzanella, line 1: Quick question: is `support.html` still needed now that each demo has its own support banner? And, `support.ts` is missing but still referenced by `support.html`.

kixelated: Nope we should remove that, good catch.

doc/js/@moq/hang/index.md
jdreetz, line 2: Should we get rid of `hang` here too? like `@moq/core` or something?

doc/js/env/web.md
```diff
@@ -64,33 +64,52 @@ Subscribe to and render a MoQ broadcast.
...
 ```html
 <script type="module">
-    import "@moq/hang/support/element";
+    import "@moq/watch/support/element";
 </script>
 
-<!-- Show only when publishing is not supported -->
-<hang-support mode="publish" show="partial"></hang-support>
+<!-- Show only when a polyfill/fallback is needed -->
+<moq-watch-support show="warning"></moq-watch-support>
```
jdreetz, line 93: Why are we splitting this up?

js/demo/src/mse.html
```diff
@@ -17,22 +17,21 @@
 
 		See index.html for more details on these components.
 	-->
-	<hang-support mode="watch" show="partial"></hang-support>
+	<moq-watch-support show="warning"></moq-watch-support>
 
-	<hang-watch-ui>
-		<hang-watch id="watch" url="%VITE_RELAY_URL%" path="bbb" muted jitter="200" reload>
+	<moq-watch-ui>
+		<moq-watch id="watch" url="%VITE_RELAY_URL%" path="bbb" muted jitter="200" reload>
```
jdreetz, line 23: did you see the warnings about this env variable in the PR checks?

doc/js/@moq/hang/index.md
kixelated: We should still keep `@moq/hang` as the container/catalog format. It won't be the only media format we will support.

doc/js/env/web.md
kixelated: `moq-watch-support` reports the compatibility of a specific version of `moq-watch`. It doesn't matter much in practice, but its misleading to version them independently.

ex. If we add WebGPU rendering to `moq-watch`, then we would also want that reflected in `moq-watch-support`.

jdreetz: Got it, thanks!

PR #964, opened 2026-02-14
Fix SolidJS not being found.

kixelated:
The devDependencies are weird because moq/demo sources the raw files, not the compiled files.

### DIFF
--- a/js/hang-demo/package.json
+++ b/js/hang-demo/package.json
@@ -16,11 +16,14 @@
 		"@moq/watch": "workspace:^",
 		"@moq/publish": "workspace:^"
 	},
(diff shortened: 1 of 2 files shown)


js/hang-demo/package.json
```diff
@@ -16,11 +16,14 @@
 		"@moq/watch": "workspace:^",
 		"@moq/publish": "workspace:^"
 	},
+	"//devDependencies": "solid-js and solid-element are only needed because Vite resolves raw .tsx source from workspace packages. They are NOT needed when using @moq/watch and @moq/publish as published npm packages.",
```
jdreetz, line 19: Never seen this way of commenting in JSON before, very cool!

PR #1055, opened 2026-03-05
Hang improvements

Qizot:
This PR makes couple of fixes to the hang package:

- fixes a deadlock in `buffer_until` that was caused by using `read` instead of `read_unbuffered`
- fixes a bug where frames could be discarded by a group because of future cancellation (buffer_until future is being dropped in a lot of places)
- fixes a bug where the `OrderedConsumer` could retuned unordered frames because of short-lived groups (each audio frame is a separate group, consuming it too fast and out-of-order would just pass it to the caller without reordering any frames)

### DIFF
--- a/rs/hang/src/container/consumer.rs
+++ b/rs/hang/src/container/consumer.rs
@@ -198,10 +198,12 @@ impl OrderedConsumer {
 					self.pending_timeout = None;
 				}
 				// Timeout waiting for a missing sequence — skip the gap.
-				() = async {
-					match &mut self.pending_timeout {
-						Some(sleep) => sleep.as_mut().await,
-						None => std::future::pending().await,
+				Some(()) = async {
+					if let Some(sleep) = &mut self.pending_timeout {
+						sleep.as_mut().await;
+						Some(())
+					} else {
+						None
 					}
 				} => {
 					drop(buffering);


rs/hang/src/container/consumer.rs
```diff
@@ -89,6 +98,27 @@ impl OrderedConsumer {
 	pub async fn read(&mut self) -> Result<Option<OrderedFrame>, Error> {
 		let latency = self.max_latency.try_into()?;
 		loop {
+			// Try to promote from pending to current when there's no gap.
+			if self.current.is_none() {
+				// Arm gap timeout if we're waiting on a missing sequence but have later groups queued.
+				if let Some(expected) = self.next_sequence {
+					let gap_exists = self.pending.front().is_some_and(|g| g.info.sequence > expected);
+					if gap_exists && self.pending_timeout.is_none() {
+						self.pending_timeout = Some(Box::pin(tokio::time::sleep(self.max_latency)));
```
kixelated, line 107: I think this needs to be `max_latency - timestamp_diff`.

Like we don't want to sleep a full 1s if max_latency is 1s, but the group we're blocked on is at least 900ms old already.

And this is the hard part, because `timestamp_diff` changes over time. If we receive a newer frame, the amount of time we need to sleep changes.

PR #1056, opened 2026-03-05
Uniffi libmoq bindings

Qizot:
This PR adds support for uniffi bindings that can be used for code generation for both iOS and android.

Besides the bindings the PR adds separate error codes instead of a single generic `NotFound` which makes it difficult to debug things sometimes.

### DIFF
--- a/rs/moq-ffi/src/api.rs
+++ b/rs/moq-ffi/src/api.rs
@@ -154,6 +154,8 @@ pub struct MoqCatalogStream {
 #[derive(uniffi::Object)]
 pub struct MoqTrack {
 	inner: Arc<tokio::sync::Mutex<hang::container::OrderedConsumer>>,
+	close: std::sync::Mutex<Option<tokio::sync::oneshot::Sender<()>>>,
+	close_rx: Arc<tokio::sync::Mutex<tokio::sync::oneshot::Receiver<()>>>,
 }
 
 #[derive(uniffi::Object)]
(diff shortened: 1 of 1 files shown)


rs/libmoq/src/consume.rs
```diff
@@ -167,8 +197,33 @@ impl Consume {
 		Ok(())
 	}
 
+	/// Returns audio config fields as owned Rust values (no C struct needed).
+	#[cfg(feature = "uniffi-api")]
+	pub fn audio_config_data(
+		&self,
+		catalog: Id,
+		index: usize,
+	) -> Result<(String, String, Option<Vec<u8>>, u32, u32), Error> {
```
kixelated: Definitely should return a struct instead of this.

rs/libmoq/src/consume.rs
```diff
@@ -43,7 +46,7 @@ impl Consume {
 	}
 
 	pub fn catalog(&mut self, broadcast: Id, mut on_catalog: OnStatus) -> Result<Id, Error> {
```
kixelated: Is this valid for uniffi?

```rs
pub fn catalog(&mut self, broadcast: Id, mut on_catalog: OnResult<hang::catalog::Root>) -> Result<Id, Error> {
```

If so, you don't need any other catalog methods. uniffi should probably be a separate crate in that case?

rs/libmoq/src/lib.rs
```diff
@@ -16,6 +18,10 @@
 //! All functions return negative error codes on failure or non-negative values on success.
 //! Resources are managed through opaque integer handles that must be explicitly closed.
 
+#[cfg(all(feature = "c-api", feature = "uniffi-api"))]
+compile_error!("Features `c-api` and `uniffi-api` are mutually exclusive. Enable only one.");
```
kixelated: 100% you should make `moq-ffi` instead of feature gating.

kixelated:
@coderabbitai review

PR #1057, opened 2026-03-05
Fix trun sample duration fallback to trex default

erikherz:
## Summary

- CMAF encoders (e.g. GPAC with `cdur=0.033`) may omit per-sample duration in `trun`, relying on `default_sample_duration` from the `trex` box in the init segment
- The decoder now falls back to the trex default when neither trun nor tfhd provide a sample duration, instead of throwing `Invalid sample duration 0`

## Changes

- Parse `default_sample_duration` from `trex` in `decodeInitSegment()` (manual box scan, no new library dependency)
- Add `defaultSampleDuration` to `InitSegment` interface and CMAF container schema
- Add optional `defaultSampleDuration` param to `decodeDataSegment()`
- Pass `defaultSampleDuration` through in video/audio decoders (`@moq/watch`)

## Context

Testing with a CMAF publisher using GPAC DASH-IF ingest (1s segments, 30 chunks/sec). GPAC sets `default_sample_duration=512` in `trex` (timescale 15360 = 33.3ms per frame) but does not set the `TR_SAMPLE_DURATION` flag in `trun` entries. The `tfhd` also omits `defaultSampleDuration`, so the only source of truth is `trex`.

## Test plan

- [ ] Verify streams from GPAC CMAF ingest no longer throw `Invalid sample duration 0`
- [ ] Verify streams that include per-sample duration in trun still work (no regression)
- [ ] Verify the trex parser handles init segments without a trex box (returns undefined)

### DIFF
--- a/js/hang/src/catalog/container.ts
+++ b/js/hang/src/catalog/container.ts
@@ -20,6 +20,8 @@ export const ContainerSchema = z
 			timescale: u53Schema,
 			// Track ID used in the moof/mdat fragments
 			trackId: u53Schema,
+			// Default sample duration from trex (fallback when trun/tfhd omit per-sample durations)
+			defaultSampleDuration: u53Schema.optional(),
 		}),
 	])
 	.default({ kind: "legacy" });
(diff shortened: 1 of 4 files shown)


kixelated:
So we should do #1059 instead. Or initData to keep CMSF compatibility. My current approach of building the init segment is not sustainable.

PR #1061, opened 2026-03-06
Improve libmoq C bindings

kixelated:
## Summary
- Add error code enum and documentation
- Fix catalog lifecycle (rename `catalog_snapshot_close` to `catalog_free`)
- Fix origin compilation and cleanup
- Simplify cfg directives

Author: @qizot

### DIFF
--- a/rs/libmoq/src/consume.rs
+++ b/rs/libmoq/src/consume.rs
@@ -15,6 +15,14 @@ struct ConsumeCatalog {
 	video_codec: Vec<String>,
 }
 
+/// A spawned task entry: close sender to signal shutdown, callback to deliver status.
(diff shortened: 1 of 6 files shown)


rs/libmoq/src/consume.rs
```diff
@@ -16,6 +15,14 @@ struct ConsumeCatalog {
 	video_codec: Vec<String>,
 }
 
+/// A spawned task entry: close sender to signal shutdown, callback to deliver status.
+/// Close revokes the callback by taking the entire entry.
+struct TaskEntry {
```
kixelated, line 20: @Qizot this is so we can't race and run the callback after `_close` is called.

PR #1071, opened 2026-03-10
Uniffi async objects

kixelated:
> ⏺ Done. Branch uniffi-async-objects is pushed. Net result: +403 / -1,350 lines — the crate went
  from 10 source files down to 4.

@Qizot vibe coded, won't have a chance to review/test before you wake up tomorrow. Any problems with an async-first approach?


Qizot:
Okay, so from my testing there are a couple of changes necessary.

1. Methods `catalog()` or `subscribe_track()` must be run in the async context since they spawn async jobs, without it the app crashes with tokio task being run without a reactor (which makes sense).

2. For the `MoqTrack` it would be great if you could add the `close` method, as right now there is no way of stopping the track while it is running (I've just added a oneshot channel and select in the next future iterator, and it worked fine).

rs/moq-ffi/src/consumer.rs
```diff
@@ -0,0 +1,259 @@
...
+pub struct MoqAudioRendition {
+	pub codec: String,
+	pub description: Option<Vec<u8>>,
+	pub sample_rate: u32,
+	pub channel_count: u32,
+	pub bitrate: Option<u64>,
+}
+
+#[derive(uniffi::Record)]
+pub struct MoqUser {
```
Qizot: I'm not sure if things that are specific to hang should be eventually be prefixed with `Moq...`

rs/moq-ffi/src/ffi.rs
```diff
@@ -0,0 +1,102 @@
...
+	/// If cancelled/aborted, the state is lost (which is fine — the object is being destroyed).
+	pub async fn run<R, F, Fut>(&self, f: F) -> Result<R, MoqError>
+	where
+		R: Send + 'static,
+		F: FnOnce(T) -> Fut + Send + 'static,
+		Fut: Future<Output = (T, Result<R, MoqError>)> + Send + 'static,
+	{
+		let join_handle = {
+			let mut state = self.state.lock().unwrap();
+			let inner = state.inner.take().ok_or(MoqError::Cancelled)?;
```
Qizot: We should probably mention in the docs that no more than one concurrent `run` can be executing or something, in this case if you try to run 2 in concurrenlty the second attempt will return `cancelled` which I'm not sure is that obvious or correct

rs/moq-ffi/src/session.rs
```diff
@@ -0,0 +1,156 @@
...
+		return Err(MoqError::Log("logging already initialized".into()));
+	}
+
+	log.init();
+
+	Ok(())
+}
+
+#[uniffi::export]
+impl MoqClient {
```
Qizot, line 47: What we are missing from the libmoq implementation is an ability to asynchronously notify about session state change e.g. closed/disconnected.

rs/moq-ffi/src/consumer.rs
kixelated: Oof yeah this should be removed.

kixelated: And I'm going to remove the `hang.live` user stuff from the catalog; I just need a generic way to add/get custom fields first.

rs/moq-ffi/src/ffi.rs
kixelated: I'm going to try to see if I can keep the mutex held during the entire call, so a second call would block not error.

rs/moq-ffi/src/session.rs
kixelated: There's a `closed()` method in `MoqSession`. `MoqClient` is only used to establish a new session.

PR #1295, opened 2026-04-12
Replace postMessage with SharedArrayBuffer for lock-free audio transfer

kixelated:
## Summary
Replaces the message-passing audio transfer mechanism with a lock-free shared ring buffer implementation using SharedArrayBuffer, eliminating sample drops and reducing latency overhead in audio worklet communication.

## Key Changes

- **New SharedRingBuffer implementation** (`shared-ring-buffer.ts`):
  - Lock-free ring buffer using atomic operations for thread-safe coordination between main thread and AudioWorklet
  - Handles out-of-order writes, gap filling with zeros, and automatic overflow management
  - Supports dynamic latency adjustment via `setLatency()`
  - Provides modular i32 arithmetic for correct wrapping behavior with large sample indices

- **Comprehensive test suite** (`shared-ring-buffer.test.ts`):
  - 527 lines of tests covering initialization, insertion, reading, stall behavior, ring wrapping, multi-channel audio, overflow handling, and edge cases
  - Tests verify gap filling, out-of-order writes, latency skipping, and timestamp tracking

- **Decoder refactoring** (`decoder.ts`):
  - Allocates SharedRingBuffer on worklet initialization with 1-second capacity
  - Writes decoded audio directly to shared memory via `insert()` instead of `postMessage()`
  - Polls shared buffer state (timestamp, stalled flag) every 50ms instead of receiving postMessage updates
  - Dynamically re-allocates buffer if capacity becomes insufficient for target latency

- **AudioWorklet simplification** (`render-worklet.ts`):
  - Removes AudioRingBuffer dependency and complex state tracking
  - Simplified `process()` to directly read from SharedRingBuffer
  - Eliminates underflow tracking and periodic state messages

- **Message protocol cleanup** (`render.ts`):
  - Removes `Data` and `Latency` message types
  - `Init` message now carries SharedRingBufferInit data (samples and control SharedArrayBuffers)
  - Eliminates `ToMain` state messages in favor of atomic polling

## Notable Implementation Details

- Uses atomic operations (`Atomics.load`, `Atomics.store`) for lock-free coordination
- Implements modular i32 arithmetic (`i32Max`, `slot()`) to handle wrapping with signed 32-bit integers
- Automatically advances READ pointer on overflow to prevent buffer exhaustion
- Implements latency skipping to maintain target latency distance between READ and WRITE pointers
- Handles fractional timestamps by rounding to nearest sample index

https://claude.ai/code/session_01DyHBcF8vEs1wtbJubRVExp


js/watch/src/audio/decoder.ts
```diff
@@ -55,6 +54,9 @@ export class Decoder {
 	#buffered = new Signal<BufferedRanges>([]);
 	readonly buffered: Getter<BufferedRanges> = this.#buffered;
 
+	// Audio buffer bridging main thread and worklet (shared memory or postMessage transport).
+	#buffer: AudioBuffer | undefined;
```
kixelated: #buffer is used a lot, maybe ring or samples or some other name.

js/watch/src/audio/decoder.ts
```diff
@@ -107,23 +109,26 @@ export class Decoder {
...
-					this.#stalled.set(event.data.stalled);
-					this.#trimDecodeBuffered(timestamp);
-				}
-			};
+			// Initial target latency in samples.
+			const latency = this.source.sync.buffer.peek();
+			const latencySamples = Math.ceil(sampleRate * Time.Second.fromMilli(latency));
+
+			// Let the factory pick the best transport (SharedArrayBuffer or postMessage).
+			this.#buffer = createAudioBuffer(worklet, channelCount, sampleRate, latencySamples);
```
kixelated: const buffer

So effect.cleanup has a concrete .close() target. If we need to assign to this.#buffer (do we?), that can be another line.

js/watch/src/audio/decoder.ts
```diff
@@ -107,23 +109,26 @@ export class Decoder {
...
+
+			// Poll buffer state for both transports. The shared path reads it from shared
+			// memory; the post path reads values cached from state messages.
+			effect.interval(() => {
+				if (!this.#buffer) return;
+				const ts = Time.Milli.fromMicro(this.#buffer.timestamp);
+				this.#timestamp.set(ts);
+				this.#stalled.set(this.#buffer.stalled);
+				this.#trimDecodeBuffered(ts);
+			}, 50);
```
kixelated: I feel like the interface should take or return a Signal for timestamp/stalled. This interval should be abstracted way, at the very least as part of createAudioBuffer

js/watch/src/audio/decoder.ts
```diff
@@ -140,16 +145,16 @@ export class Decoder {
 	}
 
 	#runLatency(effect: Effect): void {
+		// Gate on the worklet signal so this effect re-runs once the buffer is created.
 		const worklet = effect.get(this.#worklet);
 		if (!worklet) return;
 
-		const latency = effect.get(this.source.sync.buffer);
+		const buffer = this.#buffer;
+		if (!buffer) return;
```
kixelated: Uh this probably needs to be a Signal?

js/watch/src/audio/render-worklet.ts
```diff
@@ -1,54 +1,65 @@
 import type { Message, State } from "./render";
 import { AudioRingBuffer } from "./ring-buffer";
+import { SharedRingBuffer } from "./shared-ring-buffer";
 
 class Render extends AudioWorkletProcessor {
-	#buffer?: AudioRingBuffer;
+	// Exactly one of these is set after init, depending on which path the main thread chose.
+	#shared?: SharedRingBuffer;
+	#post?: AudioRingBuffer;
```
kixelated: maybe?
#backend?: SharedRingBuffer | AudioRingBuffer;

PR #1299, opened 2026-04-14
Add mTLS support for moq-relay

kixelated:


demo/relay/justfile
```diff
@@ -1,68 +1,93 @@
-set fallback
+set fallback := true
```
kixelated: Is this needed?

rs/moq-native/src/client.rs
```diff
@@ -87,6 +100,39 @@ pub struct ClientConfig {
...
+		let pem = std::fs::read(path).context("failed to read client identity")?;
+		let certs: Vec<CertificateDer<'static>> = rustls_pemfile::certs(&mut pem.as_slice())
+			.collect::<Result<_, _>>()
+			.context("failed to parse client identity certs")?;
+		let leaf = certs
+			.first()
+			.with_context(|| format!("no certificates found in {}", path.display()))?;
+		let (_, cert) = x509_parser::parse_x509_certificate(leaf.as_ref())
+			.map_err(|err| anyhow::anyhow!("failed to parse identity certificate: {err}"))?;
+		let san = cert.subject_alternative_name().ok().flatten().and_then(|san| {
```
kixelated: Return an error if this fails, isntead of using ok()?

rs/moq-native/src/client.rs
```diff
@@ -87,6 +100,39 @@ pub struct ClientConfig {
...
+		};
+		let pem = std::fs::read(path).context("failed to read client identity")?;
+		let certs: Vec<CertificateDer<'static>> = rustls_pemfile::certs(&mut pem.as_slice())
+			.collect::<Result<_, _>>()
+			.context("failed to parse client identity certs")?;
+		let leaf = certs
+			.first()
+			.with_context(|| format!("no certificates found in {}", path.display()))?;
+		let (_, cert) = x509_parser::parse_x509_certificate(leaf.as_ref())
+			.map_err(|err| anyhow::anyhow!("failed to parse identity certificate: {err}"))?;
```
kixelated: .context instead

rs/moq-native/src/client.rs
```diff
@@ -87,6 +100,39 @@ pub struct ClientConfig {
...
+		let Some(path) = self.identity.as_ref() else {
+			return Ok(None);
+		};
+		let pem = std::fs::read(path).context("failed to read client identity")?;
+		let certs: Vec<CertificateDer<'static>> = rustls_pemfile::certs(&mut pem.as_slice())
+			.collect::<Result<_, _>>()
+			.context("failed to parse client identity certs")?;
+		let leaf = certs
+			.first()
+			.with_context(|| format!("no certificates found in {}", path.display()))?;
```
kixelated: .context

rs/moq-native/src/client.rs
```diff
@@ -202,10 +248,28 @@ impl Client {
...
+			Some(path) => {
+				let pem = std::fs::read(path).context("failed to read client identity")?;
+				let chain: Vec<CertificateDer<'static>> = rustls_pemfile::certs(&mut pem.as_slice())
+					.collect::<Result<_, _>>()
+					.context("failed to parse client identity certs")?;
+				anyhow::ensure!(!chain.is_empty(), "no certificates found in {}", path.display());
+
+				let key = rustls_pemfile::private_key(&mut pem.as_slice())
+					.context("failed to parse client identity key")?
+					.with_context(|| format!("no private key found in {}", path.display()))?;
```
kixelated: In general, use .context instead. I don't care about printing the path in the error message.

rs/moq-native/src/client.rs
```diff
@@ -272,6 +336,15 @@ impl Client {
 		self
 	}
 
+	/// Returns the [`rustls::ClientConfig`] used for outbound TLS connections.
+	///
+	/// Callers can clone this into other TLS-consuming libraries (e.g. reqwest
+	/// via `ClientBuilder::use_preconfigured_tls`) to share the same cert
+	/// chain, root CAs, and identity without reparsing the PEMs.
+	pub fn tls(&self) -> &rustls::ClientConfig {
+		&self.tls
```
kixelated: Is this already pub?

rs/moq-native/src/client.rs
```diff
@@ -87,6 +100,39 @@ pub struct ClientConfig {
...
 }
 
+impl ClientTls {
+	/// Parse the configured identity PEM (if any) and return the first DNS
+	/// SAN on its leaf certificate.
+	///
+	/// Useful for sanity-checking that a caller's own cluster node name
+	/// matches the identity they will present. Returns `Ok(None)` if no
+	/// identity is configured.
+	#[cfg(feature = "quinn")]
```
kixelated: This might be a problem, since we use it in non-quinn paths. Doesn't make sense to feature gate it?

rs/moq-native/src/quinn.rs
```diff
@@ -137,6 +138,49 @@ impl QuinnClient {
 	}
 }
 
+// ── Peer identity extraction ───────────────────────────────────────
+
+fn extract_peer_identity(conn: &quinn::Connection) -> Option<PeerIdentity> {
```
kixelated: Should return a Result

rs/moq-native/src/quinn.rs
```diff
@@ -137,6 +138,49 @@ impl QuinnClient {
...
+			None
+		}
+	};
+
+	Some(PeerIdentity { dns_name })
+}
+
+// ── Root CA loading ────────────────────────────────────────────────
+
+fn load_root_ca(config: &ServerTlsConfig) -> anyhow::Result<rustls::RootCertStore> {
```
kixelated: can this be impl ServerTlsConfig?

rs/moq-native/src/server.rs
```diff
@@ -585,6 +665,42 @@ pub struct ServerTlsInfo {
 	pub fingerprints: Vec<String>,
 }
 
+impl ServerTlsInfo {
+	/// Returns the first DNS SAN on each configured end-entity (leaf) certificate.
+	///
+	/// Errors if any certificate exposes zero or more than one DNS SAN —
+	/// cluster peers are expected to advertise exactly one, matching the
+	/// cluster node name.
+	#[cfg(feature = "quinn")]
```
kixelated: Does this have to be gated to Quinn?

rs/moq-native/src/server.rs
```diff
@@ -585,6 +665,42 @@ pub struct ServerTlsInfo {
...
+	pub fn single_dns_sans(&self) -> anyhow::Result<Vec<String>> {
```
kixelated: The name sucks. 

`validate_sans` maybe? 

Also what is this even used for? I don't think we really care if the server uses a DNS san, and even if it did, we would require all certs advertise the same one.

rs/moq-native/Cargo.toml
```diff
@@ -23,6 +23,7 @@ quinn = [
     "dep:rcgen",
     "dep:reqwest",
     "dep:rustls-webpki",
+    "dep:x509-parser",
```
kixelated: IDK just make this required for all backends. Its temporary that we only support this for Quinn.

rs/moq-relay/src/auth.rs
```diff
@@ -274,8 +289,18 @@ impl PublicAccess {
...
+	/// `tls` is an already-built [`rustls::ClientConfig`] — typically the
+	/// one from the relay's cluster [`moq_native::Client`] — reused for
+	/// the HTTP client that fetches keys and public-access rules. Sharing
+	/// the config means cert chains and root CAs are loaded exactly once.
+	///
+	/// `mtls_enabled` tells Auth whether the server accepts mTLS peers
+	/// (i.e. `server.tls.root` is non-empty). When true, missing JWT/public
+	/// config is a warning (mTLS peers still authenticate); when false, it
+	/// is fatal because no one can authenticate.
+	pub async fn init(self, tls: &rustls::ClientConfig, mtls_enabled: bool) -> anyhow::Result<Auth> {
```
kixelated: Honestly, let's go back to when we didn't share the rustls::ClientConfig from client. Make a separate auth.tls config that is nearly identical to client.tls. It's just better for configuration, because we don't NEED to use the same cert/ca/etc.

And all of these parameters can be removed that way.

rs/moq-relay/src/auth.rs
```diff
@@ -274,8 +289,18 @@ impl PublicAccess {
...
+		Auth::new(self, tls, mtls_enabled).await
```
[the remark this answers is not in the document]
kixelated: +1

rs/moq-relay/src/main.rs
```diff
@@ -21,17 +21,59 @@ async fn main() -> anyhow::Result<()> {
...
+			.identity_dns_name()?
+			.context("client.tls.identity has no DNS SAN; cluster peers cannot authenticate")?;
+		match config.cluster.node.as_deref() {
+			None => {
+				tracing::info!(%san, "deriving cluster.node from client.tls.identity SAN");
+				config.cluster.node = Some(san);
+			}
+			Some(node) => {
+				anyhow::ensure!(
+					node == san || is_san_with_port(&san, node),
```
kixelated: Could this reuse `resolve_peer_node`? And maybe call it `validate_peer`

rs/moq-relay/src/main.rs
```diff
@@ -21,17 +21,59 @@ async fn main() -> anyhow::Result<()> {
...
+					"cluster.node {node:?} does not match client.tls.identity SAN {san:?}",
+				);
+			}
+		}
+	}
+
+	// Every server cert must carry exactly one DNS SAN, and (if cluster.node
+	// is set) that SAN must equal the hostname portion of the node name —
+	// the same rule we apply to peers that connect in via mTLS.
+	let server_sans = server.tls_info().read().unwrap().single_dns_sans()?;
```
kixelated: WE only care that the client cert matches the cluster.node. Obvious the server side should match too, but it's not a requirement.

kixelated:
@coderabbitai review

PR #1300, opened 2026-04-14
feat: add TrackConsumer::produce() for zero-copy track sharing

ksletmoe-aws:
## Motivation

When building a relay that re-publishes selected tracks from one broadcast into another (e.g., remapping namespaces with filtered catalog metadata), the current approach requires a forwarding loop:

```rust
let mut consumer = source_broadcast.subscribe_track(&track)?;
let mut producer = new_broadcast.create_track(track)?;

while let Some(mut group) = consumer.next_group().await? {
    let mut out = producer.append_group()?;
    while let Some(data) = group.read_frame().await? {
        out.write_frame(data)?;
    }
}
```

This copies every frame through an extra read/write cycle, adding latency.

## Solution

Add `TrackConsumer::produce()`, which upgrades a consumer back to a `TrackProducer` sharing the same underlying state. The relay can then `insert_track()` it into a different broadcast — zero copy, no forwarding loop:

```rust
let consumer = source_broadcast.subscribe_track(&track)?;
let producer = consumer.produce()?;
new_broadcast.insert_track(&producer)?;
```

Both broadcasts now serve the same underlying track data with no additional latency.

## Changes

**`rs/moq-lite/src/model/track.rs`** — Added `TrackConsumer::produce() -> Result<TrackProducer>` with doc comments covering shared ownership semantics, lifetime behavior, and error conditions. Preserves the abort error (consistent with other `TrackConsumer` error paths).

## Tests

4 unit tests: happy path with shared-state verification, dropped producer (`Error::Dropped`), aborted track (preserves `Error::Cancel`), and keep-alive property (upgraded producer prevents auto-close).

## Notes

- **Additive only** — no breaking changes to existing APIs.
- The returned `TrackProducer` shares state with the original: mutations (appending groups, finishing, aborting) through either producer affect all consumers. It keeps the track alive as long as it exists.
- Per discussion: `TrackConsumer::produce()` is the non-breaking approach for `main`. The breaking `insert_track(Consumer)` change on `dev` can supersede this later.


kixelated:
What about `TrackConsumer::produce() -> Result<TrackProducer>`?

kixelated:
The other option is to have insert_track take a Consumer, which it used to do, but I forget why that was changed. The underlying issue may have been with tokio::watch

kixelated:
> The other option is to have insert_track take a Consumer, which it used to do, but I forget why that was changed. The underlying issue may have been with tokio::watch

Ah right it's like that on the dev branch, where all of my breaking changes go, but this seems like the best option for now.

ksletmoe-aws:
That makes sense -- `TrackConsumer::produce() -> Result<TrackProducer>` is simpler. I'll get that changed; thanks!

ksletmoe-aws:
> Ah right it's like that on the dev branch, where all of my breaking changes go, but this seems like the best option for now.

Woops, I had missed your message. Which seems like the best option? The second, having insert_track take a Consumer?

kixelated:
> > Ah right it's like that on the dev branch, where all of my breaking changes go, but this seems like the best option for now.
> 
> Woops, I had missed your message. Which seems like the best option? The second, having insert_track take a Consumer?

produce() would be best right now. Maybe that changes with the breaking API changes, but those are maybe a month from getting merged so I don't want to block you.

kixelated (approved):

PR #1301, opened 2026-04-15
conducer: replace Weak<Waker> waker list with atomic-flag slots

kixelated:
Each Waiter now holds an Arc<AtomicBool> "active" flag that is shared
with every WaiterList slot it registers into. Dropping the Waiter flips
the flag to false, so we can detect dead slots in O(1) without the
previous Weak<Waker> upgrade dance.

WaiterList is backed by SmallVec<[Slot; 32]> instead of
VecDeque<Weak<Waker>>, keeping registrations on the stack for the
common case. (smallvec rather than tinyvec because Slot can't sensibly
implement Default — Waker has no default.)

register() now does a bounded scan: it checks up to two existing slots
and reuses the first inactive one in place, otherwise appends. This
keeps the per-call cost O(1) while still draining stale entries over
time in typical single- or few-waiter workloads.

### DIFF
--- a/rs/conducer/src/waiter.rs
+++ b/rs/conducer/src/waiter.rs
@@ -98,8 +98,8 @@ impl WaiterList {
 	}
 
 	/// Wake all live waiters, consuming the list.
-	pub fn wake(self) {
-		for waker in self.entries.into_iter().filter_map(|w| w.upgrade()) {
+	pub fn wake(mut self) {
+		for waker in self.entries.drain(..).filter_map(|w| w.upgrade()) {
 			waker.wake_by_ref();
 		}
 	}
--- a/rs/conducer/src/waiter.rs
+++ b/rs/conducer/src/waiter.rs
@@ -97,8 +97,9 @@ impl WaiterList {
 		}
 	}
 
-	/// Wake all live waiters, consuming the list.
-	pub fn wake(mut self) {
+	/// Wake all live waiters, draining the list.
+	pub fn wake(&mut self) {
+		self.cursor = 0;
 		for waker in self.entries.drain(..).filter_map(|w| w.upgrade()) {
 			waker.wake_by_ref();
 		}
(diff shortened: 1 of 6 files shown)


rs/conducer/src/waiter.rs
```diff
@@ -35,47 +44,65 @@ impl Waiter {
...
+				let i = self.cursor % self.entries.len();
+				if self.entries[i].strong_count() == 0 {
+					// Reuse the dead slot in place. Each Waiter owns a
+					// unique Arc<Waker>, so strong_count == 0 uniquely
+					// identifies a slot whose owner has been dropped —
+					// no will_wake / pointer comparison needed.
+					self.entries[i] = new_weak;
+					return;
+				}
+				self.cursor = self.cursor.wrapping_add(1);
```
kixelated: Do the modulo here instead? Marginally faster.

rs/conducer/src/waiter.rs
```diff
@@ -35,47 +44,65 @@ impl Waiter {
...
 			entries: std::mem::take(&mut self.entries),
+			cursor: 0,
 		}
 	}
 
 	/// Wake all live waiters, consuming the list.
-	pub fn wake(mut self) {
-		for waker in self.entries.drain(..).filter_map(|w| w.upgrade()) {
+	pub fn wake(self) {
+		for waker in self.entries.into_iter().filter_map(|w| w.upgrade()) {
```
kixelated: This should drain. And reset cursor back to 0.

rs/conducer/src/waiter.rs
```diff
@@ -35,47 +44,65 @@ impl Waiter {
...
 	pub fn register(&mut self, waiter: &Waiter) {
-		// Clean up dead entries at front that fail to upgrade
-		while let Some(front) = self.entries.pop_front() {
-			if front.strong_count() == 0 {
-				// Dead entry, skip
-				continue;
+		let new_weak = Arc::downgrade(&waiter.waker);
+
+		if !self.entries.is_empty() {
+			for _ in 0..2 {
```
kixelated: self.entries.len().min(2)

Avoids the is_empty check and avoids checking a 1 element array twice.

PR #1322, opened 2026-04-17
moq-lite/moq-relay: hop-based clustering

kixelated:
## Summary

Ports the hop-based clustering design from `origin/dev` (#1082 + #1152) onto `main`. Replaces the three-tier `primary`/`secondary`/`combined` origin and the `cluster: bool` token flag with a single `OriginProducer` per relay tagged with a stable `OriginId`, and carries the hop chain on every `Broadcast` so loops are refused and the shortest path wins.

### moq-lite

- New `OriginId` type — non-zero 62-bit varint, encoded as `u64` on the wire.
- `Broadcast` now carries `hops: Vec<OriginId>`; `BroadcastProducer` / `BroadcastConsumer` / `BroadcastDynamic` expose it via a `pub info: Broadcast` field plus a `Deref<Target = Broadcast>`. Backwards-compat: `Broadcast::produce()` still works and a new `produce_with_info` is added.
- `Lite04` `Announce` changes from `Vec<u64>` to `Vec<OriginId>`. `Lite03` still sends count-only, decoded as `UNKNOWN` placeholders. `MAX_HOPS` tightened from 256 to 32.
- `OriginProducer` gains `id`, `with_id`, `id()`; `OriginConsumer` tracks the origin id so the wire publisher can append it to outbound hops.
- `OriginProducer::publish_broadcast` refuses broadcasts whose hop chain already contains our id (loop detection) and prefers the shortest-hop entry as active; backups promote to active by shortest-hop on close.
- `lite::publisher` appends `origin.origin_id()` to the hop list before each outbound `Announce::Active`; `lite::subscriber` attaches the received hops to the new `BroadcastProducer`.

### moq-relay

- `cluster.rs` collapses to a single `origin: OriginProducer` initialised `with_id` from config. `primary`/`secondary`/`combined` are gone; the special `internal/origins/*` registration dance is gone.
- `ClusterConfig` replaces `--cluster-root`/`--cluster-node`/`--cluster-prefix` with `--cluster-connect` (repeat or comma-separated) for a full mesh, plus an optional `--cluster-origin-id` (env `MOQ_CLUSTER_ORIGIN_ID`) for deterministic IDs in tests/logs.
- `auth.rs` keeps main's HTTP/mTLS additions; only the `claims.cluster`-driven routing is dropped (`#[allow(deprecated)]` on the one call site that unavoidably reads it for back-compat). `AuthToken.register` is still used for mTLS SAN validation in `connection.rs`.
- `connection.rs`, `websocket.rs`, `main.rs` updated for the single-origin API.

### moq-token

- `Claims::cluster` is now `#[deprecated]`. Existing signed tokens still parse; the flag no longer affects routing. Test modules and the `basic` example opt into `#[allow(deprecated)]` so `-D warnings` still passes.

### js/lite

- Wire `MAX_HOPS` tightened to 32 to match Rust.
- `Publisher` generates a random 53-bit non-zero `originId` (via `crypto.getRandomValues`) in its constructor and appends it to the `hops` list on every outbound `Announce::Active`. Ended announces still go out with empty hops (peer matches on path).

### demo/relay

- `leaf0.toml` / `leaf1.toml` / `root.toml` switched to the mesh `connect = [...]` format with pinned per-node `origin_id`s (1 / 10 / 11) for readable logs in dev.

## Test plan

- [x] `cargo test -p moq-lite -p moq-token -p moq-relay` — all pass (257 + 100 + relay lib tests). Adjusted `test_duplicate` expectation: on equal hop lengths, the newer broadcast wins (was: always replace).
- [x] `cargo check --workspace --all-targets` — clean.
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean.
- [x] `bun run --filter='@moq/lite' check` + `bun biome check js/lite` — clean.
- [ ] Local smoke test against `demo/relay/{root,leaf0,leaf1}.toml`: publish from a client on leaf0, subscribe from a client on leaf1, confirm the broadcast arrives via the mesh and that a restart of one node triggers re-routing (not yet run).
- [ ] Interop: browser `@moq/lite` publisher against the refactored relay — protocol is wire-compatible but the JS origin ID plumbing is new.

## Notes

- `cargo-semver-checks` will flag this as a breaking change on `moq-lite` and `moq-relay`. release-plz is expected to pick up the right version bumps automatically; no manual version edits in this PR.
- Browser clients aren't full mesh peers — they only publish their own broadcasts and don't forward, so JS doesn't need the full `publish_broadcast`/loop-detection logic. A per-connection origin ID is enough for the relay to tag and deduplicate incoming announcements.


kixelated:
!coderabbitai review

kixelated:
@coderabbitai review

PR #1332, opened 2026-04-19
moq-native: resolve DNS hostnames in --server-bind

kixelated:
Accept host:port inputs like `fly-global-services:443` (used on Fly.io)
by changing ServerConfig::bind from SocketAddr to String and resolving
at bind time. Only the first resolved address is used since Quinn does
not support binding to multiple addresses.

https://claude.ai/code/session_01BJSqQsZ5iyuCtY9EuYSy5b

### DIFF
--- a/rs/moq-cli/src/main.rs
+++ b/rs/moq-cli/src/main.rs
@@ -110,7 +110,7 @@ async fn main() -> anyhow::Result<()> {
 
 	match cli.command {
 		Command::Serve { config, dir, name, .. } => {
-			let web_bind = config.bind.unwrap_or("[::]:443".parse().unwrap());
+			let web_bind = config.bind.clone().unwrap_or_else(|| "[::]:443".to_string());
 
 			let server = config.init()?;
 			#[cfg(feature = "iroh")]
@@ -120,7 +120,7 @@ async fn main() -> anyhow::Result<()> {
 
 			tokio::select! {
 				res = run_server(server, name, publish.consume()) => res,
-				res = run_web(web_bind, web_tls, dir) => res,
+				res = run_web(&web_bind, web_tls, dir) => res,
 				res = publish.run(stats_interval) => res,
 			}
 		}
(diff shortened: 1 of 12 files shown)


rs/moq-cli/src/web.rs
```diff
@@ -3,23 +3,22 @@ use axum::handler::HandlerWithoutStateExt;
...
 pub async fn run_web(
-	bind: SocketAddr,
+	bind: &str,
 	tls_info: Arc<RwLock<moq_native::ServerTlsInfo>>,
 	public: Option<PathBuf>,
 ) -> anyhow::Result<()> {
 	let listen = tokio::net::lookup_host(bind)
 		.await
-		.context("invalid listen address")?
+		.with_context(|| format!("invalid listen address '{bind}'"))?
```
kixelated: Use .context, I don't care about what value they entered.

rs/moq-cli/src/main.rs
```diff
@@ -110,7 +110,7 @@ async fn main() -> anyhow::Result<()> {
 
 	match cli.command {
 		Command::Serve { config, dir, name, .. } => {
-			let web_bind = config.bind.unwrap_or("[::]:443".parse().unwrap());
+			let web_bind = config.bind.clone().unwrap_or_else(|| "[::]:443".to_string());
```
kixelated, line 113: Any way to simplify this? Not sure why it changed.

rs/moq-native/src/server.rs
```diff
@@ -154,6 +159,31 @@ impl ServerConfig {
...
+	}
+}
+
+/// Resolve a `host:port` string to a single [`net::SocketAddr`].
+///
+/// Accepts both literal socket addresses (e.g. `[::]:443`) and DNS hostnames
+/// paired with a port (e.g. `fly-global-services:443`). Only the first
+/// resolved address is returned, matching the client-side resolver, since
+/// Quinn only supports binding/connecting to a single IP.
+pub(crate) fn resolve_host_port(addr: &str) -> anyhow::Result<net::SocketAddr> {
```
kixelated: Should this be a member? IDK seems like it should be a generic helper or utility.

PR #1343, opened 2026-04-23
relay: add subdomain-based slug routing for customer isolation

kixelated:
## Summary
- Add `--auth-domain` (env `MOQ_AUTH_DOMAIN`, TOML `domains`) to configure suffix lists for host-based routing.
- When a connection URL host is `<slug>.<suffix>`, the slug is prepended to the path before auth runs, so `customer.cdn.moq.dev/foo` is equivalent to `cdn.moq.dev/customer/foo`.
- Hosts that exactly match a suffix — or match none — fall back to plain path routing. Multi-label slugs (`a.b.<suffix>`) are rejected as `400 InvalidHost` to keep customer isolation unambiguous.
- `AuthError::InvalidHost` maps to `400 Bad Request` (distinct from credential errors, which stay `401`).

## Test plan
- [x] `cargo test -p moq-relay` — 69 passed, including new unit tests for slug prepend, exact-suffix, non-match, empty path, multi-label rejection, multiple suffixes, case-insensitive host match, no-domains-configured, JWT/register passthrough, and an end-to-end flow where a subdomain slug combined with a public prefix authorizes a connection (and a different subdomain is rejected).
- [ ] Manual smoke with a wildcard cert against a real relay deploy.

### DIFF
--- a/rs/moq-relay/src/auth.rs
+++ b/rs/moq-relay/src/auth.rs
@@ -34,8 +34,8 @@ impl AuthParams {
 	///
 	/// When the URL host matches one of `domains` as `<labels>.<suffix>`, the
 	/// labels are prepended to the URL path in DNS-reverse order (broadest
-	/// scope first), so `customer.usw.cdn.moq.dev/foo` with suffix
-	/// `cdn.moq.dev` routes to `/usw/customer/foo`. An exact-suffix or
+	/// scope first), so `team.customer.cdn.moq.dev/foo` with suffix
+	/// `cdn.moq.dev` routes to `/customer/team/foo`. An exact-suffix or
 	/// non-matching host is left as-is (plain path-based routing).
 	///
 	/// `domains` must be pre-canonicalized by [`Auth::new`] (lowercased and
@@ -66,8 +66,8 @@ impl AuthParams {
 /// the labels joined with `/` in DNS-reverse order so the broadest scope
 /// becomes the outermost path segment. With suffix `cdn.moq.dev`:
 ///
-/// - `customer.cdn.moq.dev`     → `Some("customer")`
-/// - `customer.usw.cdn.moq.dev` → `Some("usw/customer")`
+/// - `customer.cdn.moq.dev`      → `Some("customer")`
+/// - `team.customer.cdn.moq.dev` → `Some("customer/team")`
 ///
 /// An exact match against a suffix or a host that matches no suffix returns
 /// `None` (plain path-based routing).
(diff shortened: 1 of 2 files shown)


rs/moq-relay/src/auth.rs
```diff
@@ -44,8 +53,34 @@ impl AuthParams {
...
+/// nested customers can add the inner domain to the list.
+fn match_domain(host: Option<&str>, domains: &[String]) -> Result<Option<String>, AuthError> {
+	let Some(host) = host else {
+		return Ok(None);
+	};
+	// URL hosts are case-insensitive; match that way.
+	let host_lc = host.to_ascii_lowercase();
+	for suffix in domains {
+		let suffix_lc = suffix.to_ascii_lowercase();
+		if host_lc == suffix_lc {
```
kixelated: IMO do one strip_suffix call.

rs/moq-relay/src/auth.rs
```diff
@@ -44,8 +53,34 @@ impl AuthParams {
...
+			return Ok(None);
+		}
+		if let Some(slug) = host_lc.strip_suffix(&suffix_lc).and_then(|s| s.strip_suffix('.')) {
```
kixelated: Maybe add the leading . to the domain after parsing the config file?

rs/moq-relay/src/auth.rs
```diff
@@ -44,8 +53,34 @@ impl AuthParams {
...
+			if slug.is_empty() || slug.contains('.') {
```
kixelated: We could replace . with / to support multiple paths.

rs/moq-relay/src/auth.rs
```diff
@@ -577,7 +636,23 @@ impl Auth {
 			anyhow::bail!("no auth-key, auth-key-dir, or public path configured");
 		}
 
-		Ok(Self { resolver, public })
+		// Sort domain suffixes longest-first so overlapping configurations
+		// (e.g. ["moq.dev", "cdn.moq.dev"]) match the most specific suffix,
+		// rather than letting the configured order silently decide.
+		let mut domains = config.domains;
```
kixelated: We should also lowercase and add a . prefix here.

rs/moq-relay/src/auth.rs
```diff
@@ -577,7 +636,23 @@ impl Auth {
...
+		Ok(Self {
+			resolver,
+			public,
+			domains: Arc::from(domains.into_boxed_slice()),
+		})
+	}
+
+	/// Build [`AuthParams`] from an incoming connection URL, applying any
+	/// configured subdomain-based slug routing.
+	pub fn params_from_url(&self, url: &url::Url) -> Result<AuthParams, AuthError> {
```
kixelated: Why is this public? IDK seems like it's too specific.

kixelated (commented):
s

PR #1349, opened 2026-04-27
@moq/watch: add static catalog format

skirsten:
## Summary

- Adds a `"static"` catalog format on `@moq/watch` that lets callers pass a `Catalog.Root` directly instead of fetching it via the hang or MSF catalog track. Connection + broadcast name are still required so media tracks can be subscribed normally.
- Promotes `Broadcast.catalog` from `Getter` to writable `Signal<Catalog.Root | undefined>` and accepts an initial `catalog` prop. `<moq-watch>` exposes a matching `catalog` JS property.
- Adds a `demo/web/src/static.html` demo with a textarea + Apply button to manually drive `<moq-watch catalog-format="static">` against the default `bbb` broadcast, cross-linked from the existing watch/mse/publish demos.
- Documents the new format in `doc/js/@moq/watch.md`.

## Test plan

- [x] `bun run check` in `js/watch` passes.
- [x] `bun run build` in `demo/web` produces `dist/static.html`.
- [x] `just demo` running, open `static.html`, paste a real `Catalog.Root` (capture from `index.html` via `JSON.stringify(watch.catalog)`), click Apply, video plays without a `catalog.json` track subscription.
- [x] Default `hang` and `msf` flows on `index.html` still behave as before.

### DIFF
--- a/doc/js/@moq/watch.md
+++ b/doc/js/@moq/watch.md
@@ -69,7 +69,7 @@ a real bundler (the examples below).
 - `paused` — Pause playback (boolean)
 - `muted` — Mute audio (boolean)
 - `volume` — Audio volume (0–1, default: 1)
-- `catalog-format` — Catalog format: `"hang"` (default), `"msf"` (see [MSF](/concept/standard/msf)), or `"static"` (supply the catalog yourself)
+- `catalog-format` — Catalog format: `"hang"` (default), `"msf"` (see [MSF](/concept/standard/msf)), or `"manual"` (supply the catalog yourself)
 
 ## Catalog Formats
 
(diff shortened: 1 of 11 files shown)


js/watch/src/broadcast.ts
```diff
@@ -110,23 +127,30 @@ export class Broadcast {
 						return update ? toHang(update) : undefined;
 					};
 
+		// Capture the AbortSignal of *this* run; it flips to aborted on teardown
+		// (re-run or close) and lets us tell that apart from a natural stream end,
+		// so we never clobber a user-provided catalog Signal on a planned re-run.
```
kixelated: Do we need this? If the format is static, we return early and never reach this code.

doc/js/@moq/watch.md
```diff
@@ -101,6 +101,39 @@ const broadcast = new Watch.Broadcast({
 broadcast.catalogFormat.set("msf");
 ```
 
+### Static catalogs
+
+Use `catalog-format="static"` (or `catalogFormat: "static"`) to skip the catalog
```
kixelated: nit: Is `static` the best name if it can change? Maybe use `manual` instead?

kixelated (commented):
looks good, small nits

js/watch/src/broadcast.ts
skirsten: claude added this because of coderabbits review: Basically when changing format from hang to static and the other way it was not happy. But yea, I also don't like this logic. I will try to simplify it.

kixelated:
thanks @skirsten !

PR #1350, opened 2026-04-27
moq-relay: authenticate HTTPS callers via the cluster mTLS CA

kixelated:
## Summary

- The QUIC server already short-circuits to \`AuthToken::unrestricted()\` when a peer presents a client cert signed by \`--server-tls-root\` (\`connection.rs:34\`). The HTTPS web server (\`/announced\`, \`/fetch\`, \`/ws/*\`) didn't — it required a JWT in the query string. So on a relay host you couldn't hit \`/announced\` with the local \`cluster.crt\` even though the moq layer treats that cert as full access.
- This wires the same path through the HTTPS listener: when \`--server-tls-root\` is set, the listener installs a \`WebPkiClientVerifier\` (with \`.allow_unauthenticated()\` so JWT-only callers still work), and a verified peer cert produces \`AuthToken::unrestricted()\` via a new \`WebState::resolve_token\` helper.
- Implementation: a tiny \`MtlsAcceptor\` wraps \`RustlsAcceptor\` and, after the handshake, installs a per-connection tower middleware (\`SetMtlsExtension\`) that injects an \`Option<MtlsPeer>\` request extension. Handlers extract \`Option<Extension<MtlsPeer>>\` and route through \`resolve_token\`.
- Cert hot-reload via SIGUSR1 is preserved on the mTLS path (the existing \`reload_from_pem_file\` would silently strip client-cert verification, so the mTLS path re-runs the full builder via \`reload_from_config\`).

When \`--server-tls-root\` is unset, behavior is unchanged.

## Test plan

- [x] \`cargo test -p moq-relay\` — 63/63 pass (59 existing + 4 new in \`web::tests\`: \`build_mtls_config\` round-trip + error paths, \`SetMtlsExtension\` injection)
- [x] \`cargo clippy -p moq-relay --all-targets\` clean
- [ ] Operator end-to-end on a deployed relay:
      \`\`\`
      curl --cacert /etc/letsencrypt/live/cdn.moq.dev/fullchain.pem \\
           --cert /var/lib/moq/cluster.crt --key /var/lib/moq/cluster.key \\
           https://use.cdn.moq.dev/announced
      \`\`\`
      should return a newline-separated broadcast list with no \`?jwt=\`. Same curl without \`--cert/--key\` should 401.
- [ ] JWT path regression: \`/announced?jwt=<existing-token>\` from a non-mTLS client returns 200 unchanged.
- [ ] SIGUSR1 cert reload: rotate the HTTPS cert/key on disk, signal the relay, hit \`/announced\` with both auth modes again.


rs/moq-relay/src/web.rs
```diff
@@ -81,16 +90,43 @@ pub struct WebState {
 	pub conn_id: AtomicU64,
 }
 
+impl WebState {
+	/// Resolve the auth token for an incoming request. Returns
+	/// [`AuthToken::unrestricted`] when an mTLS-verified client cert is
+	/// present (mirrors `Connection::authenticate` on the QUIC side); falls
+	/// back to JWT verification otherwise.
+	pub(crate) async fn resolve_token(
```
kixelated: Does this need to be a method? IDK seems like it could be inlined.

rs/moq-relay/src/web.rs
```diff
@@ -81,16 +90,43 @@ pub struct WebState {
...
+	tls_roots: Vec<PathBuf>,
 }
 
 impl Web {
-	/// Creates a new web server with the given state and configuration.
-	pub fn new(state: WebState, config: WebConfig) -> Self {
-		Self { state, config }
+	/// Creates a new web server with the given state, configuration, and mTLS
+	/// CA roots. Pass an empty `tls_roots` to disable HTTPS client cert auth.
+	pub fn new(state: WebState, config: WebConfig, tls_roots: Vec<PathBuf>) -> Self {
```
kixelated: Instead of tls_roots, is there a TlsConfig or something we could pass instead?

Or should the web configure the roots separately?

rs/moq-relay/src/web.rs
```diff
@@ -145,8 +209,67 @@ impl Web {
 	}
 }
 
+/// Build a [`rustls::ServerConfig`] for the HTTPS listener that includes a
+/// client cert verifier. Mirrors the construction in
```
kixelated: Is it possible to reuse any of this code in moq-native? Or is it not worth exporting?

rs/moq-relay/src/web.rs
```diff
@@ -145,8 +209,67 @@ impl Web {
...
+		.expect("no default crypto provider installed");
+
+	let verifier = WebPkiClientVerifier::builder_with_provider(root_store, provider.clone())
+		.allow_unauthenticated()
+		.build()
+		.context("failed to build https client cert verifier")?;
+
+	let mut config = rustls::ServerConfig::builder_with_provider(provider)
+		.with_safe_default_protocol_versions()?
+		.with_client_cert_verifier(verifier)
```
kixelated: Is there any way we could share the mTLS and regular paths by making a custom validator when no roots are configured? I'm not a fan the mtls and non-mtls split.

rs/moq-relay/src/web.rs
kixelated: Inlined into the three handlers in 6ce82bd1; agree the helper added more indirection than it saved.

rs/moq-relay/src/web.rs
kixelated: Moved to a new `--web-https-client-ca` flag on `HttpsConfig` (6ce82bd1). Web is now self-configuring; `Web::new` is back to `(state, config)`.

rs/moq-relay/src/web.rs
kixelated: Considered exporting a `pub fn load_pem_roots` from moq-native, but the only sharable bit is the root-store loader (cert/key loading is web-specific because it uses the dedicated `--web-https-cert/--web-https-key` paths, not the QUIC server's). The duplication is ~10 lines of straight rustls-pki-types calls. Left it inline for this PR; happy to do the moq-native export as a follow-up if you want.

rs/moq-relay/src/web.rs
kixelated: Unified in 6ce82bd1: single `build_https_config` swaps the verifier between `WebPkiClientVerifier` and `with_no_client_auth` based on whether `client_ca` is set. One `MtlsAcceptor`, one reload task. The acceptor is a near-no-op when no CA is configured, but it keeps the wiring symmetric.

rs/moq-relay/src/web.rs
```diff
@@ -145,8 +209,67 @@ impl Web {
...
+		.with_single_cert(cert_chain, key_der)
+		.context("invalid https cert/key pair")?;
+
+	// axum_server's default config sets these; without them h2 negotiation breaks.
+	config.alpn_protocols = vec![b"h2".to_vec(), b"http/1.1".to_vec()];
+
+	Ok(config)
+}
```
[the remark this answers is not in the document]
kixelated: Addressed in 6ce82bd1 — dropped the cross-file line-number reference. The new doc just describes the verifier behavior locally.

rs/moq-relay/src/web.rs
```diff
@@ -67,6 +76,21 @@ pub struct HttpsConfig {
...
+	/// mirroring the QUIC server's `--server-tls-root` behavior. Clients that
+	/// don't present a cert continue through the normal JWT path.
+	#[arg(
+		long = "web-https-client-ca",
+		id = "web-https-client-ca",
+		value_delimiter = ',',
+		env = "MOQ_WEB_HTTPS_CLIENT_CA"
+	)]
+	#[serde(default, skip_serializing_if = "Vec::is_empty")]
+	pub client_ca: Vec<PathBuf>,
```
kixelated: This should follow a similar naming convention. I think I call it `roots` or something elsewhere?

rs/moq-relay/src/web.rs
```diff
@@ -88,7 +112,8 @@ pub struct Web {
 }
 
 impl Web {
-	/// Creates a new web server with the given state and configuration.
+	/// Creates a new web server with the given state and configuration. mTLS
+	/// for HTTPS is enabled when [`HttpsConfig::client_ca`] is non-empty.
```
kixelated: This comment doesn't seem needed.

PR #1354, opened 2026-04-28
Fix missing channel samples for audio encoder

Qizot:
It seems we can't actively rely on the `channelCount` passed to the audio encoder.

On iOS safari the line `channelCount: settings.channelCount ?? root.channelCount,` resolves to `2`, but afterwards we receive mono audio in `onmessage`.

Since the number of channels in `AudioData` must match the number of channels the encoder has been initialized with, we are fixing the `AudioData` by copying the active channel to the missing one.

### DIFF
--- a/js/publish/src/audio/encoder.ts
+++ b/js/publish/src/audio/encoder.ts
@@ -163,6 +163,7 @@ export class Encoder {
 			// We're using an async polyfill temporarily for Safari support.
 			await Util.Libav.polyfill();
 
+			let closed = false;
 			const encoder = new AudioEncoder({
 				output: (frame) => {
 					if (frame.type !== "key") {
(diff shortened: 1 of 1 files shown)


js/publish/src/audio/encoder.ts
```diff
@@ -179,30 +180,39 @@ export class Encoder {
...
+			});
 
 			console.debug("encoding audio", config);
 			encoder.configure(config);
 
 			worklet.port.onmessage = ({ data }: { data: Capture.AudioFrame }) => {
-				const channels = data.channels.slice(0, worklet.channelCount);
-				const joinedLength = channels.reduce((a, b) => a + b.length, 0);
-				const joined = new Float32Array(joinedLength);
+				if (closed || encoder.state !== "configured") return;
```
kixelated, line 197: Do we need the `closed` variable if we already check encoder.state?

js/publish/src/audio/encoder.ts
```diff
@@ -179,30 +180,39 @@ export class Encoder {
...
+
+				const channels = config.numberOfChannels;
+				const joined = new Float32Array(frames * channels);
 
-				channels.reduce((offset: number, channel: Float32Array): number => {
-					joined.set(channel, offset);
-					return offset + channel.length;
-				}, 0);
+				for (let i = 0; i < channels; i += 1) {
+					const channel = data.channels[i] ?? (data.channels.length === 1 ? data.channels[0] : undefined);
```
kixelated, line 206: I'm a bit confused, what is this doing? If channels is greater than 1, but data.channels is 1, we duplicate the channel?

kixelated (commented):
I'm pretty confused when it comes to this logic in general. Maybe document the browser behavior we're trying to work around?

Qizot:
This was wrong approach, we should have reconfigured the encoder instead.

PR #1359, opened 2026-04-28
feat(player): unify Consumer across container formats

ksletmoe-aws:
## Summary

Replace the two separate consumer implementations (Legacy and CMAF) with a single generic `Consumer` class that accepts a `ContainerFormat` strategy for frame parsing. This mirrors the Rust `moq-mux` `Consumer<F: Container>` pattern and eliminates ~90% code duplication between the two existing consumers.

Additionally, add a `sequential` delivery mode flag to fix audio stuttering caused by inter-group serialization. When `sequential: false`, frames are delivered from any buffered group as soon as available (lowest-PTS first), avoiding the stall at group boundaries where `next()` would block waiting for the transport to signal group EOF. Audio codecs (AAC/Opus) are all-keyframe and don't need ordering guarantees, so this is safe. Video retains `sequential: true` (the default) for keyframe dependency correctness.

### New files
* `container/format.ts` — `ContainerFormat` interface and `DecodedFrame` type
* `container/consumer.ts` — Unified `Consumer` class with sequential/non-sequential delivery
* `container/cmaf/format.ts` — `CmafFormat` implementation (delegates to `decodeDataSegment`)
* `container/types.ts` — Shared `Frame`, `BufferedRanges`, `mergeBufferedRanges`
* `container/consumer.test.ts` — 25 tests covering both formats and edge cases

### Modified files
* `container/legacy.ts` — Added `LegacyFormat`, removed old `Consumer` class (kept `Producer`)
* `container/cmaf/index.ts` — Re-export `CmafFormat`
* `container/index.ts` — Re-export `Consumer`, `ConsumerProps`, shared types
* `watch/audio/decoder.ts` — Use `Consumer` with `sequential: false` for both legacy and CMAF
* `watch/audio/mse.ts` — Use `Consumer`
* `watch/video/decoder.ts` — Use `Consumer` (default `sequential: true`)
* `watch/video/mse.ts` — Use `Consumer`

### Deleted files
* `container/legacy.test.ts` — Tests migrated to unified `consumer.test.ts`

## Design decisions

* **`sequential` flag over separate classes**: One consumer with a flag is better than two classes sharing 90% of their code. The flag controls whether `next()` gates on `#active` group completion or returns frames from any group.
* **Non-sequential uses lowest-PTS selection**: When multiple groups have buffered frames, `next()` picks the frame with the lowest PTS. This preserves temporal ordering without requiring inter-group serialization.
* **Keyframe override at index 0**: Protocol invariant — groups always start at a keyframe. Enforced regardless of what the format reports. Subsequent frames trust the format's keyframe detection.
* **Error handling drops entire group**: On any decode error, the group's frames are cleared. Matches CMAF behavior where partial groups are unrecoverable.

## Net impact

`+997 / -1,157` lines across 13 files. The unified consumer is ~290 lines; the two old consumers totaled ~450 lines combined (plus ~700 lines of duplicate tests).


kixelated:
I think we need a generic `OrderedConsumer`. The problem is that `recvGroup` (and MoQ in general) returns groups out-of-order.

The idea behind `OrderedConsumer` is that we skip groups based on the target latency, which requires timestamp information unfortunately. Otherwise we would end up skipping groups immediately (latency = 0).

ksletmoe-aws:
Ah, that makes sense. I'll do that, thanks!

kixelated:
> Ah, that makes sense. I'll do that, thanks!

On the Rust side, I made an interface to parse the timestamp out of each frame. Then OrderedConsumer can be reusable.

ksletmoe-aws:
Addressed the duplicate cleanup nitpick — `close()` now just calls `this.#signals.close()`, which triggers the constructor-registered cleanup callback. Single authoritative cleanup path, no redundancy.

ksletmoe-aws:
This PR grew a bit from the original fix — I took the opportunity to create a unified `Consumer` that mirrors the Rust `Consumer<F: Container>` pattern.

**What started it:** The CMAF decoder paths were processing groups via parallel `effect.spawn()` calls with no ordering guarantees. This worked for audio (all keyframes) but was fragile for video (keyframe dependencies). Luke's comment about needing a generic `OrderedConsumer` with timestamp-based group skipping was the nudge to do it properly.

**What it does now:**
- Single `Consumer` class that accepts a `ContainerFormat` strategy (CMAF or Legacy), replacing ~450 lines of duplicate consumer code across two implementations
- `sequential` flag (default `true`) controls whether frame delivery is serialized across groups. Video uses sequential delivery for keyframe correctness; audio uses non-sequential since every frame is independently decodable
- Wall-clock latency skipping for all paths — the old CMAF audio path had none (there was a TODO)
- Non-sequential mode picks the lowest-PTS frame across all buffered groups, which is actually better cross-group ordering than the old fire-and-forget spawns

**Testing:** I noticed frequent audio underflows and late frames during integration testing and initially suspected the Consumer changes. After A/B testing against main, the pattern is identical on both branches — it's a pre-existing issue with the postMessage audio buffer fallback when SharedArrayBuffer is unavailable (the dev server doesn't serve the required COOP/COEP headers). At ~20ms target latency the postMessage path can't keep up; bumping to 70-80ms eliminates it. Separate issue to fix. All 25 unit tests pass, both `@moq/hang` and `@moq/watch` type-check clean.

Validated all three consumer paths against main:

CMAF (`just pub bbb`) — fMP4 via ffmpeg, exercises CmafFormat through Consumer
Legacy (`just pub h264 bbb`) — H.264 Annex B, exercises LegacyFormat through Consumer
HLS/MSE (`just pub hls bbb`) — HLS fMP4 passthrough, exercises the MSE consumer paths

All three play back identically across both branches. Same audio/video underflow patterns on main vs the PR. 25 unit tests pass, @moq/hang and @moq/watch type-check clean.

Net: +997 / -1,157 lines across 13 files. Ready for review.

ksletmoe-aws:
Sorry for the churn on this one — the commit history is messier than it should be. In hindsight I should have closed the original PR and opened a fresh one once the scope expanded from a targeted fix to a full consumer unification. Lesson learned for next time.

The latest commits add two fixes found during CMSF integration testing:

esds descriptor parsing — extractDescription was returning the full esds box payload instead of just the AudioSpecificConfig. Chrome's AudioDecoder.configure() expects the raw 2-byte AAC-LC config, not the entire ES_Descriptor chain. Added proper parsing to walk the descriptor tags and extract the DecoderSpecificInfo (tag 0x05) payload.

Opus description omission — Opus in CMAF uses raw packets, not OGG-wrapped. The dOps box from the init segment is not a valid OGG Identification Header, so passing it as description to AudioDecoder.configure() causes a rejection. Now omitted for Opus across all decoder paths.

Both of these only affect the MSF/CMSF catalog path (the hang catalog path already had correct descriptions from the Rust muxer).  I squashed down to one commit. This is now actually ready to review.

kixelated:
No worries, I'll take a look at it soon.

js/hang/src/container/consumer.ts
```diff
@@ -0,0 +1,294 @@
...
+import { Effect, type Getter, Signal } from "@moq/signals";
+
+import type { ContainerFormat } from "./format";
+import type { BufferedRanges, Frame } from "./types";
+
+export interface ConsumerProps {
+	format: ContainerFormat;
+	latency?: Signal<Time.Milli> | Time.Milli;
+	/** Returns the PTS that should be rendering right now, or undefined if unknown. */
+	now?: () => Time.Milli | undefined;
```
kixelated: I'd prefer not to have this, because to use it "correctly" requires signals and sleeping. The existing consumer effectively uses `latency` as a `max_buffer_size` instead.

A separate PR would make sense if you want to change the algorithm but I'm scared. This is Consumer was really hard to write and has gone through multiple revisions. We should just make the container generic (via an interface) instead of redesigning it.

js/hang/src/container/consumer.ts
```diff
@@ -0,0 +1,294 @@
...
+	/** When false, frames are delivered from any group as soon as available (no inter-group serialization).
+	 *  Useful for audio where every frame is independently decodable. Default: true. */
+	sequential?: boolean;
```
kixelated: Can we remove this? I don't know what happens when you feed an AudioDecoder out of order and I don't want to find out.

js/hang/src/container/format.ts
```diff
@@ -0,0 +1,12 @@
+import type { Time } from "@moq/lite";
+
+export interface DecodedFrame {
```
kixelated: nit: Just call it `Frame` or something? This could be used for encoding too.

js/watch/src/audio/decoder.ts
```diff
@@ -190,16 +190,19 @@ export class Decoder {
 	#runLegacyDecoder(effect: Effect, sub: Moq.Track, config: Catalog.AudioConfig): void {
 		// Create consumer with slightly less latency than the render worklet to avoid underflowing.
 		// TODO include JITTER_UNDERHEAD
-		const consumer = new Container.Legacy.Consumer(sub, {
+		const consumer = new Container.Consumer(sub, {
+			format: new Container.Legacy.LegacyFormat(),
```
kixelated: `Legacy.LegacyFormat` should be avoided IMO.

js/hang/src/container/types.ts
```diff
@@ -0,0 +1,33 @@
+import { Time } from "@moq/lite";
+
+export interface Frame {
```
kixelated, line 3: We should reuse `Frame` and `DecodedFrame`

kixelated (commented):
Try to avoid changes to the `Consumer` logic for now. Otherwise definitely the right direction

ksletmoe-aws:
Made one last change specifically for CMSF playback -- PR should be ready to re-review when you have time. Thanks!

kixelated:
I'll have Claude fix a few nits and then merge.

PR #1362, opened 2026-04-29
Add audio encoder reconfiguration

Qizot:
For some platforms (looking at you Apple) we can't reliably get the channel count from the media track, this information becomes available only when looking at the data inside of audio worklet callback.

The following changes allow for config creation only after receiving the first batch of samples.


js/publish/src/audio/encoder.ts
```diff
@@ -190,11 +198,26 @@ export class Encoder {
 			});
 			effect.cleanup(() => encoder.close());
 
-			console.debug("encoding audio", config);
-			encoder.configure(config);
+			let config = this.#config.peek();
```
kixelated: Peek is racey. 

If we really need the config, we should break early if `effect.get(this.#config)` is undefined. This function will rerun when it's set, then we can properly configure the encoder and stuff.

js/publish/src/audio/encoder.ts
```diff
@@ -102,14 +101,29 @@ export class Encoder {
...
 				numberOfOutputs: 0,
-				channelCount: settings.channelCount ?? root.channelCount,
+				channelCount,
 			});
 
 			effect.set(this.#worklet, worklet);
 
+			// The information about channels count can be unreliable on different platforms (Apple's safari).
+			// Try to get the first audio frame and only then create the configuration.
+			worklet.port.onmessage = ({ data }: { data: Capture.AudioFrame }) => {
```
kixelated: It's a little spooky setting `onmessage` in two places.

You shuld use 

```js
effect.event(worklet.port, "message", (data) => {
```

It'll unregister the event on cleanup too.

js/publish/src/audio/encoder.ts
```diff
@@ -102,14 +101,29 @@ export class Encoder {
...
+				const channelCount = data.channels.length;
+				if (!channelCount) return;
+
+				this.#config.set(this.#createConfig(worklet, channelCount));
+				worklet.port.onmessage = null;
+			};
+			effect.cleanup(() => {
+				worklet.port.onmessage = null;
+				this.#config.set(undefined);
+			});
```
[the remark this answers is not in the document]
kixelated: Can you make sure #runCatalog uses this #config signal?

kixelated:
yolo

PR #1367, opened 2026-05-01
@moq/watch: add pull mode to video renderer

skirsten:
For me (on Chrome with 144hz+ monitors) the existing renderer for some reason causes chrome to render the page at 120fps. I checked and the logic is not drawing at that rate so nothing wrong with it.

Instead, if the requestAnimationFrame is done inside another one it apparently does different schedule logic and actually syncs correctly. So the new pull mode causes it to render at the actually video refresh rate...

| push | pull |
| --- | --- |
| image | image |

Let me know what you think and if you are seeing something similar.
Visually they both should be the same so if everything checks out we can also drop the push mode if you want.

Adds a `mode: "push" | "pull"` prop on Renderer:

- "push" (default, existing behavior): the render effect subscribes to decoder.frame and schedules a single rAF paint per decoded frame.
- "pull": runs a self-recursive rAF loop and redraws only when the decoder's current frame differs from what was last drawn.

MultiBackend's WebCodecs path now constructs the renderer with `mode: "pull"` so playback runs off the display's vsync rather than the decoder's frame cadence.

### DIFF
--- a/js/watch/src/video/renderer.ts
+++ b/js/watch/src/video/renderer.ts
@@ -2,12 +2,15 @@ import { Time } from "@moq/lite";
 import { Effect, Signal } from "@moq/signals";
 import type { Decoder } from "./decoder";
 
+export type RendererMode = "push" | "pull";
+
 export type RendererProps = {
 	canvas?: HTMLCanvasElement | Signal<HTMLCanvasElement | undefined>;
 	paused?: boolean | Signal<boolean>;
+	mode?: RendererMode | Signal<RendererMode>;
 };
 
-// An component to render a video to a canvas.
+// A component to render a video to a canvas.
 export class Renderer {
 	decoder: Decoder;
 
@@ -16,6 +19,10 @@ export class Renderer {
 
 	// Whether the video is paused.
 	paused: Signal<boolean>;
+
+	// "push" re-renders each time the decoder publishes a new frame; "pull" drives a
+	// continuous rAF loop and redraws only when the decoder's current frame differs.
+	mode: Signal<RendererMode>;
 
 	// The most recently rendered frame, updated after each rAF paint.
 	readonly frame = new Signal<VideoFrame | undefined>(undefined);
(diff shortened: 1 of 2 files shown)


js/watch/src/video/renderer.ts
```diff
@@ -141,7 +158,32 @@ export class Renderer {
...
+					old?.close();
+					return current.clone();
+				});
+				this.timestamp.set(Time.Milli.fromMicro(current.timestamp as Time.Micro));
+				lastDrawn = current;
+			}
+			rafId = requestAnimationFrame(tick);
+		};
+
+		rafId = requestAnimationFrame(tick);
```
kixelated, line 179: I'm not sure if this is worth the compute cost of busy looping `requestAnimationFrame` if there's no work to be done. Is 120fps versus 144fps that important?

skirsten: its 120fps vs the actual video fps (24) so the GPU is actually less busy with the change.

kixelated: wait what? how does that work?

With `push` we should schedule 24 rAF callbacks, while `pull` should schedule refresh rate rAF callbacks? Are you sure those screenshots are correct?

skirsten: Pretty sure, but let me do some actual power testing on my phone or laptop...

I think it has to do with v-sync, but yea, it doesn't really make sense :)

skirsten: Okay, I did some more testing and the screenshots are real and there is definitely different magic on chrome's side between them.

But there is actually very little difference in the power draw. I remember this being a real issue in the past though, but that might have been fixed in a chrome or driver update.

Also it seems to depend on the display refresh rate a lot as on a 120 fps display I was getting very different numbers for both.

Anyway, I don't think this is actually needed so we can close this or continue but keep the current `push` as default?

Btw, both of these run into a issue on 60hz display with 60hz content: It's actually rendering at ~45 fps or so because some frames get unlucky and are being scheduled too late and get overwritten by the next one.

kixelated: Yeah, it might be a good idea to schedule the next RAF within the loop, but don't schedule again if the frame didn't change.

PR #1371, opened 2026-05-02
feat: relative broadcast paths for cross-broadcast catalog renditions (incl. WHEP egress)

kixelated:
## Summary

Rewritten from scratch against `dev` (the original main-based implementation predated the moq-mux catalog machinery and the js/watch rework) and now hooked up end to end on the consumer side, including WHIP/WHEP and RTMP egress.

A rendition in a hang catalog can set an optional `broadcast` field: a path relative to the broadcast that served the catalog (e.g. `"../source"`), pointing at the broadcast that actually publishes the track. A transcoder can then publish a sidecar catalog that adds new renditions while pointing unchanged ones at the original broadcast, instead of re-publishing those bytes.

### Schema / path math

- `moq-net`: new `PathRelative` type (trims slashes, drops `.` segments, preserves `..`; serde only, never on the announce/subscribe wire) and `Path::resolve` with excess-`..` clamping. Mirrored in `@moq/net` as `Path.normalizeRelative` / `Path.resolve` so both sides agree byte-for-byte.
- `hang` / `@moq/hang`: optional `broadcast` field on `VideoConfig` and `AudioConfig` (skipped when unset, so existing catalogs are byte-identical).

### Consumer wiring

- `moq-net`: `origin::Consumer::request_broadcast` now weakly caches dynamically-served (OriginDynamic) broadcasts by path, so the catalog and every same-broadcast rendition share one upstream subscription instead of re-invoking the handler N+1 times (which previously duplicated subscriptions, or failed with `Unroutable` against a one-shot handler). Announced broadcasts were already deduplicated.
- `moq-mux`: new `Source` type built via `Source::new(origin, path)` (an `OriginConsumer` plus the catalog broadcast's path). Origin is required: a source without one could never follow a cross-broadcast reference, so there's no bare-`BroadcastConsumer` path and no `MissingOrigin` error. Both the catalog broadcast and any referenced sibling resolve through `OriginConsumer::request_broadcast`, which deduplicates announced paths (a self-reference or excess-`..` falls back to the catalog broadcast). All exporters (fmp4, mkv, ts, flv, h264, h265) now take a `Source`; `Source::broadcast()` is async.
- `moq-cli subscribe`, `moq-srt` egress, `moq-hls`, the `moq-rtc` WHEP server and WHIP client egress, and the `moq-rtmp` server play path and client push all pass their origin + path through, so every place the exporters run honors cross-broadcast renditions. `moq_rtc::Client::publish` and `moq_rtmp::Client::publish` now take `(origin, path)`; `moq_hls::Broadcaster::new` is async + fallible.
- `@moq/watch`: `Broadcast.relativeBroadcast(effect, rel)` resolves the reference against the broadcast name and consumes the resolved path on the same connection, scoped to the caller's effect (no manual cache: `conn.consume` mints an independent consumer and each rendition subscribes a distinct track, so the previous per-path cache added no real dedup). Wired into the audio + video decoder and MSE pipelines.

### Public API changes

- New: `moq_net::{PathRelative, PathRelativeOwned}`, `moq_net::Path::resolve` (plus an internal `request_broadcast` dedup, no signature change), `moq_mux::Source` (`new(origin, path)` / async `broadcast()` / `subscribe_track()`), `hang::catalog::{VideoConfig,AudioConfig}::broadcast`, `@moq/net` `Path.resolve`/`Path.normalizeRelative`, `@moq/hang` `RelativeBroadcastSchema`, `@moq/watch` `Broadcast.relativeBroadcast`.
- Breaking (hence `dev`): moq-mux exporter constructors, `moq_hls::Broadcaster::new` (now `async` + `Result`), `moq-cli`'s `Subscribe::new`, `moq_rtc::EgressSource::new`, `moq_rtc::Client::publish`, and `moq_rtmp::Client::publish` all take a `Source` / `(origin, path)` instead of a `BroadcastConsumer`. Catalog format gains an optional field.

### Not covered (follow-ups)

- `moq-ffi` (and the py/swift/kt/go wrappers) and `moq-gst` consume tracks from a bare `BroadcastConsumer` and don't build a `Source`, so they don't resolve references yet; they need origin context plumbed the same way.
- The Rust client origin only resolves referenced broadcasts that are announced (or served by a dynamic handler); there's no client-side "subscribe by arbitrary path" fallback like `conn.consume(path)` in JS.
- `@moq/net` `conn.consume(path)` does not yet deduplicate the way Rust `request_broadcast` now does (JS `close()` closes shared state, so it needs a consume-ownership rework). Tracked in #2121; the current `js/watch` path works for announced broadcasts (all the browser hits) but churns on effect rerun.

## Test plan

- [x] `PathRelative` normalization + `Path::resolve` unit tests (Rust and JS mirrors)
- [x] Catalog serde round-trip: field deserializes, `None` stays off the wire, empty string normalizes (Rust `hang` + `@moq/hang` zod)
- [x] `moq_mux::Source` unit tests: no-override, self-reference reuse, excess-`..` fallback, referenced-broadcast resolution through a live origin
- [x] Rust: `cargo test` + `clippy` (default features) across moq-mux, moq-hls, moq-rtc, moq-srt, moq-rtmp, moq-cli; JS: `tsc` + biome on the watch changes
- [ ] Manual end-to-end: publish `room/source`, publish a sidecar catalog with `broadcast: "../source"`, verify `@moq/watch` and `moq subscribe` pull the track from the source broadcast

(Written by Claude Opus 4.8)


js/watch/src/broadcast.ts
```diff
@@ -201,6 +204,58 @@ export class Broadcast {
...
+	 * If `rel` is set (a rendition's catalog `broadcast` field), treat it as a path
+	 * relative to this broadcast's name and consume the resolved broadcast on the same
+	 * connection. Otherwise return the catalog's own active broadcast.
+	 *
+	 * Override broadcasts are cached per resolved path and owned by this Broadcast's
+	 * `signals`; the caller's `effect` only subscribes to the cached signal. Many
+	 * renditions referencing the same source thus share one underlying subscription,
+	 * and the override outlives any single caller effect.
+	 */
+	trackBroadcast(effect: Effect, rel: string | undefined): Moq.broadcast.Consumer | undefined {
```
kixelated: confusing name.

`relativeBroadcast` maybe?

kixelated: Also why is `Moq.broadcast` lowercase? Please make it `Moq.Broadcast`

js/watch/src/broadcast.ts
```diff
@@ -201,6 +204,58 @@ export class Broadcast {
...
+
+		// A reference that walks back to the catalog's own broadcast (or resolves to
+		// the empty root, via excess `..`) is served by the catalog broadcast itself,
+		// avoiding a duplicate subscription on the same path.
+		if (resolved === base || resolved === Path.empty()) return effect.get(this.output.active);
+
+		return effect.get(this.#override(resolved));
+	}
+
+	#overrides = new Map<Moq.Path.Valid, Signal<Moq.broadcast.Consumer | undefined>>();
```
kixelated: Why do we need to cache this? Can `conn.consume` do the deduplication instead?

rs/moq-mux/src/source.rs
```diff
@@ -0,0 +1,244 @@
...
+/// plus optional origin context for resolving cross-broadcast rendition references.
+///
+/// Build one from a bare [`moq_net::broadcast::Consumer`] (via `From` or [`Source::new`])
+/// when every track lives in the catalog's own broadcast. Add origin context with
+/// [`Source::with_origin`] to also serve catalogs whose renditions reference sibling
+/// broadcasts; without it, such a rendition fails with [`Error::MissingOrigin`](crate::Error::MissingOrigin).
+#[derive(Clone)]
+pub struct Source {
+	broadcast: moq_net::broadcast::Consumer,
+	origin: Option<(moq_net::origin::Consumer, moq_net::PathOwned)>,
```
kixelated: Should we just make `origin` required? And take a path to the broadcast instead?

js/watch/src/broadcast.ts
kixelated: Renamed `trackBroadcast` → `relativeBroadcast`. On the casing: `@moq/net` deliberately exports its role namespaces lowercase (`broadcast`/`track`/`group`/`announce`) to mirror the Rust `broadcast::` modules, so `Moq.broadcast` is the crate-wide convention, not an accident here. Renaming just this one would make it inconsistent with `track`/`group` and break symmetry with Rust, so keeping it lowercase (confirmed with Luke).

js/watch/src/broadcast.ts
kixelated: Dropped the `#overrides` cache. `conn.consume` mints an independent consumer and each rendition subscribes a distinct track, so the cache added no real dedup; the reference now resolves inline, scoped to the caller's effect exactly like the catalog broadcast.

rs/moq-mux/src/source.rs
kixelated: Done. `Source::new(origin, path)` now — origin is required and it takes the catalog broadcast's path instead of a pre-resolved `BroadcastConsumer`. Both the catalog broadcast and any referenced sibling resolve through `origin.request_broadcast` (which dedups announced paths), so `Into<Source>`, `with_origin`, and `Error::MissingOrigin` are all gone. Threaded origin+path through every exporter caller (incl. WHIP/RTMP client egress and RTMP play).

PR #1388, opened 2026-05-07
Add Low Overhead Container (LOC) frame format support

kixelated:
Implements support for the Low Overhead Container (LOC) frame format defined in draft-ietf-moq-loc across the Rust and JavaScript stacks.

## Summary
This PR adds complete encode/decode support for LOC frames, a lightweight container format that wraps codec bitstreams with minimal overhead. Each LOC frame consists of a small property block (containing timestamp and optional per-frame timescale) followed by the raw codec payload.

## Key Changes

**Rust Implementation:**
- New `moq-loc` crate providing `encode()` and `decode()` functions for LOC frame wire format
  - Handles QUIC-style varint encoding for properties
  - Supports delta-encoded property type IDs (0x06 timestamp, 0x08 timescale)
  - Comprehensive error handling and validation
  - Full test coverage including roundtrip, per-frame timescale, and error cases
- Integration into `moq-mux` container abstraction via new `Loc` struct
- Updated catalog container enum to support LOC with configurable timescale (defaults to 1,000,000 microseconds)

**JavaScript Implementation:**
- New `@moq/loc` package with `Format` (decoder) and `Producer` (encoder) classes
  - Mirrors Rust API with TypeScript types
  - Handles timescale conversion between frame timestamps and microseconds
  - Supports both `Uint8Array` and custom `Source` interface for flexible payload handling
  - Full test coverage matching Rust test scenarios
- Integration into hang catalog and watch player for audio/video decoding

**Integration:**
- Updated `moq-mux` error handling to include LOC errors
- Modified hang catalog schema to support LOC container type with optional timescale property
- Updated watch player (audio/video decoders and MSE backends) to instantiate appropriate LOC decoder based on catalog configuration
- Added LOC preference in audio source selection (prioritized after legacy, before CMAF)

## Implementation Details
- Timescale handling: Frames can specify per-frame timescale (0x08 property) or fall back to catalog default
- Property encoding: Uses delta-encoding for type IDs to minimize wire size
- Even-typed properties carry varint values; odd-typed properties carry length-prefixed bytes
- Unknown properties are silently skipped on decode, never emitted on encode
- Timestamp conversion properly scales between different timescale units while maintaining precision

https://claude.ai/code/session_01K5qt8CV2TU1hciusTu6P5r


js/hang/src/catalog/container.ts
```diff
@@ -18,6 +23,11 @@ export const ContainerSchema = z._default(
 			// Base64-encoded init segment (ftyp+moov)
 			init: z.base64(),
 		}),
+		// Low Overhead Container with optional fallback timescale (units/sec).
+		z.object({
+			kind: z.literal("loc"),
+			timescale: z._default(u53Schema, u53(1_000_000)),
```
kixelated: Actually we should remove this timescale fallback, because it's technically not in the LOC draft.

PR #1397, opened 2026-05-10
fix(lite): update in-flight group priorities on SUBSCRIBE_UPDATE

### LINKED ISSUE #1370: fix(lite): PriorityQueue does not update in-flight groups on SUBSCRIBE_UPDATE
## Problem

When a subscriber sends `SUBSCRIBE_UPDATE` to change priority, `subscriber.update()` updates the subscription but the `PriorityQueue` items for already in-flight groups retain their old `track` value. Only newly inserted groups get the updated priority.

This violates [draft-ietf-moq-transport-13, Section 6.1](https://www.ietf.org/archive/id/draft-ietf-moq-transport-13.html#section-6.1):

> Subscriber Priority is a priority number associated with an individual request. It is specified in the SUBSCRIBE or FETCH message, and can be updated via SUBSCRIBE_UPDATE message. [...] **When subscriber priority is changed, a best effort SHOULD be made to apply the change to all objects that have not been sent**, but it is implementation dependent what happens to objects that have already been received and possibly scheduled.

## Impact

Under bandwidth constraints, switching camera focus via `SUBSCRIBE_UPDATE` takes several seconds because old groups from the previously-focused camera continue to be served at high priority, starving the newly-focused camera.

## Root Cause

[`PriorityQueue::insert()`](https://github.com/moq-dev/moq/blob/main/rs/moq-lite/src/lite/priority.rs) copies the `track` value at insertion time. There is no API to update it afterward. When [`run_track`](https://github.com/moq-dev/moq/blob/main/rs/moq-lite/src/lite/publisher.rs) receives `SUBSCRIBE_UPDATE` and calls `subscriber.update()`, the `PriorityQueue` is not notified — existing `PriorityHandle`s keep their stale position.

## Proposed Fix

1. Add `subscription_id` to `PriorityItem` and `insert()`
2. Add `PriorityQueue::update_subscription(subscription_id, new_track)` — updates all items for that subscription, re-sorts the vec, and notifies all handles via watch channels
3. Call `update_subscription()` in `run_track` when `SUBSCRIBE_UPDATE` is received

Additionally:
- **Widen quinn priority range**: The current index (0–N) maps directly to `set_priority()`, but quinn needs a wider spread to effectively differentiate streams. Mapping via `index * 64` gives meaningful separation (e.g., 0 vs 64 vs 128).
- **Priority-aware `write_all`**: Wrap `write_all` in `tokio::select!` with `priority.next()` so priority changes take effect even during blocked writes under congestion.

Related: #699 (priority tie-breaking limitations), #1363 (JS SUBSCRIBE_UPDATE API)

We have a working implementation in our fork and can submit a PR if interested.

metapox:
Closes #1370

## Problem
When subscription priority changes via SUBSCRIBE_UPDATE, only new groups get the updated priority. In-flight groups continue at the old priority, causing stale scheduling under bandwidth constraints.

## Changes
- **priority.rs**: Add `subscription_id` to PriorityItem. Add `update_subscription(id, new_priority)` to re-sort existing items. Widen quinn priority range (`index * 64`) for effective differentiation.
- **publisher.rs**: Add `run_subscribe_updates()` that listens for SubscribeUpdate messages and calls `priority.update_subscription()` to update in-flight groups immediately.

## Testing
Verified with bandwidth throttling (tc 2000kbps): focus camera maintains quality while background cameras degrade. Priority switch is instant for both new and in-flight groups.

## Context
Built for [moq-multicam](https://github.com/metapox/moq-multicam) where camera switching under bandwidth constraints requires immediate priority propagation to all groups, not just future ones.

Ref: draft-ietf-moq-transport-13 Section 6.1


rs/moq-lite/src/lite/priority.rs
```diff
@@ -77,11 +84,11 @@ struct PriorityState {
 }
 
 impl PriorityState {
-	pub fn insert(&mut self, track: u8, group: u64, myself: PriorityQueue) -> PriorityHandle {
+	pub fn insert(&mut self, track: u8, group: u64, subscription_id: u64, myself: PriorityQueue) -> PriorityHandle {
```
kixelated: Could we add a method to the returned `PriorityHandle` instead of taking in `subscription_id`?

And that way we might be able to share the current index, instead of doing the expensive O(N) search for the subscription_id in your update function.

rs/moq-net/src/lite/priority.rs
```diff
@@ -156,6 +156,70 @@ impl PriorityState {
...
+						found = Some(entry);
+					} else {
+						self.overflow.push(entry);
+					}
+				}
+				found.expect("item not found in overflow heap")
+			}
+		};
+
+		if item.track == new_track {
```
kixelated: Why? It's the same as below.

rs/moq-net/src/lite/priority.rs
kixelated: Could we make a Priority struct that contains track/group so it's more obvious what the two args are?

rs/moq-net/src/lite/priority.rs
```diff
@@ -156,6 +156,70 @@ impl PriorityState {
 		});
 	}
 
+	fn set_track(&mut self, id: usize, new_track: u8) {
+		let (location, _) = self.indexes.get(&id).expect("item not in indexes");
+
+		let mut item = match location {
```
kixelated: Can you explain what we're doing here? Is this just remove? Why recode it?

rs/moq-net/src/lite/priority.rs
```diff
@@ -156,6 +156,70 @@ impl PriorityState {
...
+			return;
+		}
+
+		item.track = new_track;
+		self.reinsert(item);
+	}
+
+	// Place an item back into the queue, choosing vec vs overflow based on priority,
+	// and updating handle indices and watch channels.
+	fn reinsert(&mut self, item: PriorityItem) {
```
kixelated: Is this just `insert`? Why recode it?

I would love to see an ACTUAL `reinsert` function that takes the old and new priority or something. We could potentially avoid shifting a bunch of elements, although it would be more difficult to code.

rs/moq-net/src/lite/publisher.rs
```diff
@@ -325,20 +325,40 @@ impl<S: web_transport_trait::Session> Publisher<S> {
...
 		stream.writer.finish()?;
 		stream.writer.closed().await
 	}
 
+	async fn run_subscribe_updates<R: web_transport_trait::RecvStream>(
+		reader: &mut crate::coding::Reader<R, Version>,
+		priority_tx: &tokio::sync::watch::Sender<u8>,
+	) -> Result<(), Error> {
+		loop {
+			match reader.decode_maybe::<lite::SubscribeUpdate>().await? {
```
kixelated: while let Some?

rs/moq-net/src/lite/priority.rs
```diff
@@ -156,6 +156,70 @@ impl PriorityState {
...
+		let mut item = match location {
+			Location::Vec(idx) => {
+				let idx = *idx;
+				let item = self.vec.remove(idx);
+				self.update_indices_from(idx);
+				item
+			}
+			Location::Overflow => {
+				let mut found = None;
+				let drained: Vec<_> = self.overflow.drain().collect();
```
kixelated: Why are we draining the entire overflow? Binary search and call remove.

rs/moq-net/src/lite/publisher.rs
```diff
@@ -392,9 +416,14 @@ impl<S: web_transport_trait::Session> Publisher<S> {
...
 				_ = stream.closed() => return Err(Error::Cancel),
 				frame = group.next_frame() => frame,
-				// Update the priority if it changes.
-				priority = priority.next() => {
-					stream.set_priority(priority);
+				new_pri = priority.next() => {
+					stream.set_priority(new_pri);
+					continue;
+				}
+				res = track_priority.changed() => {
```
kixelated: Ok(res) = 

Otherwise I think we'll infinite loop if this returns Err? When can it return Err and should we do something about it instead?

rs/moq-net/src/lite/publisher.rs
```diff
@@ -411,15 +440,36 @@ impl<S: web_transport_trait::Session> Publisher<S> {
...
 					_ = stream.closed() => return Err(Error::Cancel),
 					chunk = frame.read_chunk() => chunk,
-					// Update the priority if it changes.
-					priority = priority.next() => {
-						stream.set_priority(priority);
+					new_pri = priority.next() => {
+						stream.set_priority(new_pri);
+						continue;
+					}
+					res = track_priority.changed() => {
```
kixelated: same

rs/moq-net/src/lite/publisher.rs
```diff
@@ -411,15 +440,36 @@ impl<S: web_transport_trait::Session> Publisher<S> {
...
+						tokio::select! {
+							biased;
+							result = stream.write_all(&mut chunk) => {
+								result?;
+								break;
+							}
+							new_pri = priority.next() => {
+								stream.set_priority(new_pri);
+							}
+							res = track_priority.changed() => {
```
kixelated: same

kixelated (changes requested):

PR #1401, opened 2026-05-11
Refactor/video pacing rAF

skirsten:
## Summary

Consolidate video frame pacing into the renderer's `requestAnimationFrame` loop. Previously pacing happened twice: in the decoder via `Sync.wait()` (a setTimeout/`#update` race per in-flight frame, trying to wake near vsync) and again in the renderer via a single-shot rAF triggered by a frame signal. Two layers, two places to stall, and systematic beating when the display refresh and content fps lined up (the wait would resolve a fraction of a ms past vsync, the signal-driven rAF would schedule against the *next* vsync, so every other frame got dropped).

This PR moves all pacing to paint time:

- `DecoderTrack` enqueues decoded frames into `#queue: VideoFrame[]` immediately; the output callback no longer awaits anything.
- `Decoder.consume()` returns the newest queued frame whose PTS is `<= sync.now()`, closing older entries and transferring ownership to the caller. Late-frame drop and decode-buffer trim move here.
- The renderer's rAF calls `decoder.consume()` each vsync.
- `Sync.wait()` and the `#update` notification machinery are removed. Buffer changes take effect on the next vsync (~<=16ms), which is fine for live media.

The "first-paint preview" path and the `Decoder.frame` mirror that fed it are dropped. The preview saved 20-100ms of black canvas at first paint, but it added a special case to the render loop and held a cloned `VideoFrame` resident at all times for every active track. It's replaced with two narrower signals: `DecoderTrack.display` (set once from the first decoded frame's display dimensions, as a catalog-display fallback) and `#runBuffering` now reacting to consume-time `timestamp` instead of decode-time frame presence.

Net: one rAF loop in the system. Stalls now show up as held frames instead of dropped frames.

## Test plan

- [x] `just check` (biome + tsc + clippy + cargo check, etc.)
- [x] New `js/watch/src/video/decoder.test.ts` covers `consumeFrame()`:
  - empty queue, all-future frames, boundary (`pts === now`)
  - newest-ready selection with older frames closed
  - 60Hz vsync + 120fps content (older frame closed, newer painted)
  - 120Hz vsync + 30fps content (no frame returned between content ticks)
  - late frame still selectable when nothing newer is queued
  - 60fps x 60Hz steady state: one paint per vsync, no queue growth, no spurious closes
- [x] Manual: watch a live stream and confirm no per-frame beating on a 60Hz display

### DIFF
--- a/js/watch/src/video/decoder.ts
+++ b/js/watch/src/video/decoder.ts
@@ -129,6 +129,7 @@ export class Decoder implements Backend {
 
 			// Upgrade the pending track to active.
 			// #runActive will be in charge of it now.
+			pending.promote();
 			this.#active.set(pending);
 			pending = undefined;
 
@@ -235,6 +236,12 @@ class DecoderTrack {
 
 	#queueDrain = Promise.withResolvers<void>();
 
+	// Whether this track has been promoted to active. A pending track is never
+	// consumed, so while false the output callback drops all but the newest
+	// frame to keep decoding (and `decoded`) racing toward the live edge instead
+	// of stalling at QUEUE_CAP.
+	#promoted = false;
+
 	signals = new Effect();
 
 	constructor(props: DecoderTrackProps) {
@@ -274,6 +281,15 @@ class DecoderTrack {
 				// Queue for the renderer to pick up on its next vsync.
 				this.#queue.push(frame);
 				this.decoded.set(timestamp);
+
+				// While pending, nothing consumes the queue, so drop everything but
+				// the newest frame and release backpressure. This lets the track
+				// catch up to live before it takes over (e.g. on a rendition switch).
+				if (!this.#promoted) {
+					while (this.#queue.length > 1) this.#queue.shift()?.close();
+					this.#queueDrain.resolve();
+					this.#queueDrain = Promise.withResolvers<void>();
+				}
 			},
 			// TODO bubble up error
 			error: (error) => {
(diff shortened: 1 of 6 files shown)


js/watch/src/video/decoder.ts
```diff
@@ -240,41 +238,24 @@ class DecoderTrack {
...
-						// Late frame, don't render it.
-						return;
-					}
-
-					if (this.frame.peek() === undefined) {
-						// Render something while we wait for the sync to catch up.
-						this.frame.set(frame.clone());
-					}
-
-					const wait = this.source.sync.wait(timestamp).then(() => true);
```
kixelated, line 256: This is really important for higher latency playback.

skirsten: I did some tests on Chrome and Firefox.

It's the same peak frame count either way (latency * fps). Old code holds them in awaiting output() closures, new in the queue. Same VRAM either way, and decodeQueueSize stays at 0 so the decoder isn't backpressuring us implicitly.

Only real way to shrink this would be capping the queue and pausing decoder.decode() when it's full, so we hold encoded bytes instead of decoded frames. Container.Consumer already buffers at that layer, so would probably be better to do it there 🤔

But playback seems to be the fine even at high latency:
https://pr-31-moq-monitor.skirsten.workers.dev/?relay=https%3A%2F%2Fcdn.moq.dev%2Fdemo&watch=bbb

skirsten: Hmm, actually it depends on the type of decoder used. If chrome is using hardware decoding it only stores max 10 frames, if software there does not seem to be a limit as far as I can see...
Anyway, probably makes sense to add a limit from our side: https://github.com/skirsten/moq/pull/5/changes
Should I include this in this PR?

PR #1402, opened 2026-05-12
moq-gst: Fix MoqSink CAPS handling and per-pad EOS aggregation

SteveMcFarlin:
This fixes two correctness issues in **sink/imp.rs**: CAPS events are now passed to **event_default()** after configuring the background task, so pad state still updates normally, and EOS tracking is now per-pad instead of using a single counter. That avoids premature or missed element EOS when pads are dynamic, dropped, or emit EOS more than once.

### DIFF
--- a/rs/moq-gst/src/sink/imp.rs
+++ b/rs/moq-gst/src/sink/imp.rs
@@ -4,7 +4,7 @@
 //! requested dynamically, and each pad simply forwards buffers/events to the
 //! background worker via an unbounded channel.
 
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::sync::{LazyLock, Mutex};
 
 use anyhow::{Context, Result};
@@ -103,10 +103,12 @@ enum ControlMessage {
 		data: Bytes,
 		pts: Option<gst::ClockTime>,
 	},
+	Eos {
+		pad_name: String,
+	},
 	DropPad {
 		pad_name: String,
 	},
-	Eos,
 	Shutdown,
 }
 
@@ -280,7 +282,6 @@ impl MoqSink {
 		};
 
 		let (tx, rx) = mpsc::unbounded_channel::<ControlMessage>();
-
 		let element_weak = self.obj().downgrade();
 
 		let join = RUNTIME.spawn(async move {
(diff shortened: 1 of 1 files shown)


SteveMcFarlin:
@sidsethupathi @kixelated. Let me know if you have any questions. I tested with:

`gst-launch-1.0 -v -e videotestsrc is-live=true num-buffers=120 ! video/x-raw,width=1280,height=720,framerate=60/1,format=I420 ! x264enc ! moqsink name=mux url="https://cdn.moq.dev/anon" broadcast="videotestsrc"`

My main concern was with a possible race conditions between add/remove pad and EOS. Some of the other items are just for gst state.

rs/moq-gst/src/sink/imp.rs
```diff
@@ -347,17 +348,28 @@ impl MoqSink {
...
+				if sender
+					.send(ControlMessage::Eos {
+						pad_name: pad.name().to_string(),
+					})
+					.is_err()
+				{
 					return false;
 				}
+
 				true
```
sidsethupathi, line 373: Should `gst::Pad::event_default` be called for EOS as well?

sidsethupathi (approved):
Confirmed that multiple pads still works as expected.

```
gst-launch-1.0 -v -e videotestsrc is-live=true num-buffers=120 ! video/x-raw,width=1280,height=720,framerate=60/1,format=I420 ! x264enc ! moqsink name=mux url="https://cdn.moq.dev/anon" broadcast="videotestsrc" audiotestsrc is-live=true num-buffers=100 samplesperbuffer=882 ! avenc_aac ! mux.
```

rs/moq-gst/src/sink/imp.rs
SteveMcFarlin: No. This is a terminal SINK. It will never forward on EOS downstream. The caps need to for internal state.

SteveMcFarlin:
Hand tested a failure case pointed at /dev/null. 

```
2026-05-12T03:45:22.159469Z  WARN moq_native::client: QUIC connection failed err=failed DNS lookup
2026-05-12T03:45:22.228507Z  WARN moq_native::websocket: WebSocket connection failed err=failed to connect WebSocket
```
```
ERROR: from element /GstPipeline:pipeline0/GstVideoTestSrc:videotestsrc0: Internal data stream error.
```

I will look into this more for proper handling. For now I think this will prevent a few hard to find bugs with EOS.

kixelated:
Yep this is closer to how I expected EOS handling to work.

PR #1405, opened 2026-05-14
Declare solid-js as a peerDependency

Karolk99:
@moq/publish, @moq/watch, and @moq/ui-core import from solid-js and solid-js/web, but solid-js was only listed in devDependencies. Since common/package.ts strips devDependencies at release time, the published package.json files declared no solid-js at all, while their JS still contains bare import "solid-js" statements.

Consumers ended up with either an unresolved import or two Solid runtime instances loaded side-by-side. Because Solid's delegateEvents is a per-runtime singleton, UI components rendered by these packages registered click handlers on a different runtime than the one wired up to the host app, so buttons appeared inert.

### DIFF
--- a/js/publish/package.json
+++ b/js/publish/package.json
@@ -28,6 +28,10 @@
 		"@moq/signals": "workspace:^",
 		"@moq/ui-core": "workspace:^"
 	},
+	"peerDependencies": {
+		"solid-element": "^1.9.1",
+		"solid-js": "^1.9.10"
+	},
 	"devDependencies": {
 		"@types/audioworklet": "^0.0.77",
 		"@typescript/lib-dom": "npm:@types/web@^0.0.241",
--- a/js/ui-core/package.json
+++ b/js/ui-core/package.json
@@ -20,7 +20,8 @@
 		"release": "bun ../common/release.ts"
 	},
 	"peerDependencies": {
-		"@moq/signals": "workspace:^0.1.0"
+		"@moq/signals": "workspace:^0.1.0",
+		"solid-js": "^1.9.10"
 	},
 	"devDependencies": {
 		"@biomejs/biome": "^2.4.3",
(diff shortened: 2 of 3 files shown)


kixelated:
Okay so I'm not sure if I want solid-js as a peerDependency, but 100% bundling a second SolidJS runtime is not great.

We currently build moq/watch with vite so it inlines solid; there should be no `import "solid"` in the final NPM package. Can you double check?

Karolk99:
I didn't realise Solid was being inlined. Inlining means each package ships its own Solid runtime, so the components can't be rendered inside the consumer app's Solid tree, which is the problem affecting @moq/watch/ui, @moq/watch/ui, and @moq/ui-core because they export raw Solid components.

One option is to mark solid-js as external in the Vite config across all packages; another is to do it only for the UI-focused ones (@moq/watch, @moq/publish, @moq/ui-core) and leave the ./element web-component entries alone but then u still have to import solid-js on your own.

Worth noting: after building locally with solid-js externalised, dist/ui/index.js went from 48 KB → 26 KB — that delta is the size of the Solid runtime that was previously being carried inside each package.

What do you think @kixelated?

kixelated:
I don't really want to make solid an explicit dependency just for the UI web component. It hurts dev UX.

TBH I kind of want to revert the switch to solidjs and go back to the minimal moq/signals DOM library, since it's already a dependency. Solid feels like overkill for a simple, stock video UI.

And yeah ui-core is not supposed to be consumed by external users.

Karolk99:
Sounds good — happy to close. Just one thing for the record: on the supported Web Component path (<moq-watch-ui> + <moq-watch> in a Solid app, no direct /ui imports), the overlay buttons fire, but their visual state doesn't update (play icon stays on play after pausing). Reactive signals seem not to cross between the inlined Solid in @moq/watch/element and the one in @moq/watch/ui. Probably moot once you move off Solid, but at the moment it doesn't work properly.

PR #1410, opened 2026-05-15
fix(watch): hide buffering overlay while offline

### LINKED ISSUE #737: Offline notification
The new buffering icon completely covers the element and any controls, including when the broadcast is offline. We should instead show a message that the broadcast is not live (when `controls: true`).

YogiSotho:
## Summary

Fix the watch UI buffering overlay so it does not cover the player while the broadcast is offline.

## Changes

- hide the buffering overlay when `watchStatus()` is `offline`
- keep the spinner centered by adding explicit flex centering to the buffering container

## Why

Issue #737 reports that the buffering UI covers the element and controls even when nothing is loading because the broadcast is offline. In that state, showing a spinner is misleading and blocks the offline state from being communicated clearly.

## Testing

- verified the change is limited to the watch UI buffering indicator and styles
- attempted `bun run check` in `js/watch`, but local workspace dependencies were not fully installed, so `tsc` was unavailable in this environment

Closes #737

### DIFF
--- a/js/watch/src/backend.ts
+++ b/js/watch/src/backend.ts
@@ -166,7 +166,10 @@ export class MultiBackend implements Backend {
 			paused: this.paused,
 		});
 
-		const videoRenderer = new Video.Renderer(videoSource, { canvas: element, paused: this.paused });
+		const videoRenderer = new Video.Renderer(videoSource, {
+			canvas: element,
+			paused: this.paused,
+		});
 
 		effect.cleanup(() => {
 			videoSource.close();
--- a/js/watch/src/video/renderer.ts
+++ b/js/watch/src/video/renderer.ts
@@ -100,7 +100,7 @@ export class Renderer {
 		}
 
 		// When paused, fetch a single preview frame then disable.
-		const frame = effect.get(this.frame);
+		const frame = effect.get(this.decoder.frame);
 		this.decoder.enabled.set(!frame);
 	}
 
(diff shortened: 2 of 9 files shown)


YogiSotho:
https://youtu.be/OR1ImYQX9Po

kixelated:
Hmm I definitely think we hide the buffering icon, but we should also stop rendering the most recent frame right? A black screen is more obvious than a frozen BBB image.

YogiSotho:
Added two commits to this PR:

9ca2c7f7 — clears the canvas to black when offline instead of showing the last frozen frame
cb982070 — adds a small OFFLINE badge in the bottom right corner

https://youtu.be/mG6nsVV2jwg

js/watch/src/video/renderer.ts
```diff
@@ -17,6 +18,9 @@ export class Renderer {
 	// Whether the video is paused.
 	paused: Signal<boolean>;
 
+	// Whether the watch is offline.
+	offline: Signal<boolean>;
```
kixelated: Maybe flip it and use `enabled` instead? To match other components.

kixelated: Actually, maybe the decoder should unset the frame (on offline) instead of telling the renderer to draw black? IDK.

kixelated (commented):
I think you're very close.

Can you look into Decoder and figure out what it does when a broadcast goes offline? Maybe it's just not unsetting the current frame. I think that's a better fix than a signal that toggles a black screen.

js/watch/src/backend.ts
```diff
@@ -54,6 +54,9 @@ export interface MultiBackendProps {
 	connection?: Signal<Moq.Connection.Established | undefined>;
 
 	paused?: boolean | Signal<boolean>;
+
+	// Whether the broadcast is offline.
+	offline?: boolean | Signal<boolean>;
```
kixelated: Not used any longer

js/watch/src/video/renderer.ts
```diff
@@ -17,12 +16,6 @@ export class Renderer {
 	// Whether the video is paused.
 	paused: Signal<boolean>;
 
-	// The most recently rendered frame, updated after each rAF paint.
-	readonly frame = new Signal<VideoFrame | undefined>(undefined);
```
kixelated, line 21: I think moq-boy uses these to compute a more accurate latency. Best to keep them otherwise we have to do a semver bump.

kixelated (commented):
I think this looks a lot better!

js/watch/src/video/renderer.ts
```diff
@@ -100,43 +100,41 @@ export class Renderer {
...
 		if (!ctx) return;
 
-		const paused = effect.get(this.paused);
-
-		// Read new frames from the decoder when not paused.
-		let decoded: VideoFrame | undefined;
-		if (!paused) {
-			decoded = effect.get(this.decoder.frame);
-		}
+		effect.get(this.paused);
```
kixelated: Are we not doing anything based on this pause state?

kixelated:
yolo

kixelated:
thanks @YogiSotho

PR #1411, opened 2026-05-15
Add pixel budget controls for video rendition selection

kixelated:
## Summary

Adds a new `pixels` attribute and property to the `<moq-watch>` element that controls how video renditions are selected based on pixel constraints. This enables two modes of operation: automatic sizing based on the element's rendered dimensions, or a fixed pixel-area budget for bandwidth control.

## Changes

- **New `PixelsMode` type**: Union of `"auto"` (default) or a non-negative number representing total pixel area budget.

- **Element attribute and property**: Added `pixels` to the observed attributes list and implemented getter/setter that parse and manage the pixel mode via a Signal.

- **Auto mode with ResizeObserver**: When `pixels` is `"auto"`, the element tracks its rendered size using ResizeObserver and feeds `maxWidth`/`maxHeight` constraints to the video source, scaled by `devicePixelRatio` for high-DPI displays. This prevents downloading unnecessarily high-resolution streams for small viewport slots.

- **Numeric mode**: When `pixels` is set to a number, it pins a total pixel-area budget (`pixels` field on the target) and clears dimension constraints so the two don't conflict.

- **New `byDimensions` filter**: Added rendition filtering function that respects `maxWidth` and `maxHeight` constraints independently. Selects the largest rendition within budget, or falls back to the smallest over-budget option if nothing fits.

- **Target type expansion**: Extended the `Target` type with optional `maxWidth` and `maxHeight` fields to support dimension-based filtering alongside the existing `pixels` field.

## Implementation Details

- The pixel budget logic runs as a separate Effect scoped to the element, allowing cleanup when the element is destroyed.
- ResizeObserver is seeded with the element's current bounding rect to handle cases where the element is still 0x0 when the observer attaches.
- The `parsePixelsMode` function handles null/empty/invalid inputs by defaulting to `"auto"`.
- Dimension filtering ranks within-budget renditions by area (largest first) for best quality, matching the behavior of the existing `byPixels` filter.

https://claude.ai/code/session_013QzaZSQikeDkEmuV8ZdTJE

### DIFF
--- a/js/watch/src/element.ts
+++ b/js/watch/src/element.ts
@@ -12,33 +12,8 @@ function parseCatalogFormat(value: string | null): CatalogFormat {
 	return CATALOG_FORMATS.find((f) => f === value) ?? DEFAULT_CATALOG_FORMAT;
 }
 
-function parsePixelsMode(value: string | null): PixelsMode {
-	if (value === null || value === "" || value === "auto") return "auto";
-	const parsed = Number.parseInt(value, 10);
-	if (Number.isFinite(parsed) && parsed >= 0) return parsed;
-	return "auto";
-}
-
-const OBSERVED = [
-	"url",
-	"name",
-	"paused",
-	"volume",
-	"muted",
-	"reload",
-	"latency",
-	"jitter",
-	"catalog-format",
-	"pixels",
-] as const;
+const OBSERVED = ["url", "name", "paused", "volume", "muted", "reload", "latency", "jitter", "catalog-format"] as const;
 type Observed = (typeof OBSERVED)[number];
-
-/**
- * Pixel budget for video rendition selection.
- * - `"auto"` tracks the rendered size of the element (default).
- * - A non-negative number caps selection at that pixel count.
- */
-export type PixelsMode = number | "auto";
 
 // Close everything when this element is garbage collected.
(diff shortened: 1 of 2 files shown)


js/watch/src/video/source.ts
```diff
@@ -19,9 +19,15 @@ export type Target = {
 	// Optional manual override for the selected rendition name.
 	name?: string;
 
-	// The desired size of the video in pixels.
+	// Maximum desired pixel area (codedWidth * codedHeight).
 	pixels?: number;
 
+	// Maximum desired coded width in pixels.
+	maxWidth?: number;
```
kixelated: IMO rename to `width`. All of these are maximums.

js/watch/src/element.ts
```diff
@@ -36,6 +61,11 @@ export default class MoqWatch extends HTMLElement {
 	// Set when the element is connected to the DOM.
 	#enabled = new Signal(false);
 
+	// Controls the video selection cap. "auto" derives maxWidth/maxHeight from
+	// the element's rendered size so we don't download a 1080p stream for a
+	// 360p slot. A number pins a total-pixel-area budget instead.
+	#pixelsMode = new Signal<PixelsMode>("auto");
```
kixelated: I don't understand why you would ever want to change this from auto? It's confusing, can you delete it entirely?

PR #1429, opened 2026-05-20
feat: Unified CMSF/Hang pipeline

ksletmoe-aws:
Unify CMSF support into the existing hang/CMAF pipeline. Supersedes #1408, addressing all review feedback.

## Approach

Instead of a parallel CMSF pipeline (~2400 lines duplicating CMAF logic), this extends the existing pipeline with targeted additions (+1658 lines, no duplication). The key insight: CMSF is CMAF with a different catalog format. The relay is format-agnostic, so the difference is purely at the edges.

## Changes

**MSF catalog support (subscribe side):**
- Add SAP and jitter fields to `moq_msf::Track`
- Add `MsfConsumer` that parses MSF catalogs and converts to `hang::Catalog`
- Add `--catalog msf` flag to `moq-cli subscribe`
- Wire SAP type derivation and jitter into `to_msf()` on publish side

**Explicit group mode (multi-encoder redundancy):**
- Add `with_explicit_groups()` and `start_group(sequence)` to `Fmp4` importer
- Caller-controlled group boundaries for all tracks simultaneously

**Slim C API for CMAF passthrough:**
- `moq_track_create` with versioned `moq_track_config` struct
- Mode B: `moq_broadcast_start_group` + `moq_broadcast_write` (aligned groups)
- Mode A: `moq_group_open` + `moq_group_write` + `moq_group_close` (per-track)
- `moq_track_close` with auto-finish of open groups
- No CMAF muxing in the C API; caller provides pre-built moof+mdat

**JS decoder improvement:**
- Video/audio decoders fall back to init segment description when catalog lacks one

## How it addresses #1408 feedback

- "Support both via intermediate representation" → `hang::Catalog` is the IR, one pipeline, two catalog serializations
- "Don't copy-paste shared CMAF code" → Zero duplicated code
- "Need `--catalog msf`" → Done
- "C API should be a struct" → `moq_track_config` with `size` field for ABI compat
- "Don't want C API to be a CMAF muxer" → It's not; just track/group lifecycle
- "Don't want both description and init segment" → Only `init_data` in the config

## Testing

- 133 tests pass (18 libmoq + 9 moq-msf + 106 moq-mux)
- JS typecheck and build pass
- Integration tested: relay + publish + web player via MSF catalog


rs/libmoq/src/publish.rs
```diff
@@ -72,4 +130,146 @@ impl Publish {
...
+
+		for track in self.tracks.values_mut() {
+			if track.broadcast_id != broadcast_id {
+				continue;
+			}
+			if let Some(mut prev) = track.group.take() {
+				prev.finish()?;
+			}
+			let group = track.producer.create_group(moq_lite::Group { sequence })?;
+			track.group = Some(group);
```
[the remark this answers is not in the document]
ksletmoe-aws: This is fixed in a subsequent commit

js/watch/src/audio/decoder.ts
```diff
@@ -406,8 +410,18 @@ export class Decoder {
...
-		config.codec === "opus" ? undefined : config.description ? Util.Hex.toBytes(config.description) : undefined;
+	let description: Uint8Array | undefined;
+	if (config.codec !== "opus") {
+		if (config.description) {
+			description = Util.Hex.toBytes(config.description);
+		} else if (config.container.kind === "cmaf") {
+			try {
+				description = Container.Cmaf.decodeInitSegment(base64ToBytes(config.container.init)).description;
+			} catch {
+				// Malformed init segment; let isConfigSupported decide without description.
```
kixelated, line 421: Should warn and return return false.

rs/moq-cli/src/subscribe.rs
```diff
@@ -42,7 +56,11 @@ impl Subscribe {
 		// Fmp4 subscribes to the catalog internally, builds the merged init segment
 		// from the first catalog snapshot, then yields moof+mdat fragments in
 		// timestamp order across tracks.
-		let mut fmp4 = moq_mux::export::Fmp4::new(self.broadcast)?.with_latency(self.args.max_latency);
+		let mut fmp4 = match self.args.catalog {
+			CatalogFormat::Hang => moq_mux::export::Fmp4::new(self.broadcast)?,
+			CatalogFormat::Msf => moq_mux::export::Fmp4::new_msf(self.broadcast)?,
```
kixelated, line 61: I think the second argument should be the CatalogFormat instead.

rs/moq-mux/src/catalog/producer.rs
```diff
@@ -119,6 +119,22 @@ impl Drop for Guard<'_> {
...
+///
+/// SAP type 1: closed GOP (IDR at every SAP). Used for H.264, VP8, AV1, VP9.
+/// SAP type 3: open GOP (CRA/non-IDR RAP possible). Used for H.265 as a
+/// conservative worst-case since the codec supports both IDR and CRA.
+///
+/// Returns None for unknown codecs so the field is omitted from the catalog.
+fn video_sap_type(codec: &hang::catalog::VideoCodec) -> Option<u8> {
+	use hang::catalog::VideoCodec;
+	match codec {
+		VideoCodec::H264(_) | VideoCodec::VP8 | VideoCodec::AV1(_) | VideoCodec::VP9(_) => Some(1),
```
kixelated, line 132: Should be 2 for b-frames.

rs/moq-mux/src/catalog/producer.rs
```diff
@@ -119,6 +119,22 @@ impl Drop for Guard<'_> {
...
+		VideoCodec::H265(_) => Some(3),
```
kixelated, line 133: We don't support this at all right now... I'd rather not advertise it until then.

rs/moq-mux/src/import/fmp4.rs
```diff
@@ -75,9 +80,65 @@ impl Fmp4 {
...
+	/// Enable explicit group mode. In this mode, the importer does not create
+	/// groups on keyframes. Instead, the caller must call `start_group(sequence)`
+	/// to create groups on all tracks simultaneously.
+	///
+	/// # Audio behavior in explicit mode
+	///
+	/// Audio fragments accumulate in the current group until `start_group` is called.
+	/// This means late-joining subscribers must buffer the entire group before playback.
+	/// For latency-sensitive non-redundancy use cases, use the default auto mode instead.
+	pub fn with_explicit_groups(mut self) -> Self {
```
kixelated, line 96: I'm not sure we need this option. I'm fine with `start_group`, and we should only close the existing group if there's at least 1 frame (on keyframe).

js/watch/src/video/decoder.ts
```diff
@@ -362,7 +362,7 @@ class DecoderTrack {
 
 		const initSegment = base64ToBytes(this.config.container.init);
 		const init = Container.Cmaf.decodeInitSegment(initSegment);
-		const description = this.config.description ? Util.Hex.toBytes(this.config.description) : undefined;
+		const description = this.config.description ? Util.Hex.toBytes(this.config.description) : init.description;
```
kixelated, line 365: NOTE: I want to get rid of `this.config.description` and have each container handle description.

What you have is correct.

rs/libmoq/src/api.rs
```diff
@@ -568,3 +572,249 @@ pub extern "C" fn moq_consume_close(consume: u32) -> i32 {
...
+		bitrate,
+	})
+}
+
+/// Create a track via the slim CMAF passthrough API.
+///
+/// Updates the broadcast's catalog automatically based on `config`. The codec
+/// string determines whether the track is registered as video or audio. For
+/// codec strings that are not recognized, the implementation falls back to
+/// inspecting the `width`/`height` and `sample_rate`/`channel_count` fields
```
kixelated, line 698: I'd like to make this stuff automatic if possible. (almost) all of this information should be in CMAF already, so it should get parsed out via the CMAF import. We can only shoot ourselves in the foot if we allow users to contradict the moov.

rs/libmoq/src/api.rs
```diff
@@ -568,3 +572,249 @@ pub extern "C" fn moq_consume_close(consume: u32) -> i32 {
...
+/// to decide.
+///
+/// # Group lifecycle
+/// Each track supports one of two mutually exclusive modes:
+/// - **Mode A** (per track groups): groups are managed via [`moq_group_open`]
```
kixelated, line 703: Yeah see my later comment, I'm not sure we need to make this a hard switch. `start_group` could be optional.

kixelated (commented):
MUCH better than the previous PR.

A few API nits. The libmoq API is the most concerning to me. I want to keep it simple, and I think we can do that with the existing write API + an optional start_group call.

ksletmoe-aws:
Cool, will take a look at the comments and get a new revision up. Thanks!

kixelated:
> Cool, will take a look at the comments and get a new revision up. Thanks!

I can make the API changes too, just lemme know.

kixelated:
I merged most of this via #1444 

I'm not sure about adding an advanced splitting API to libmoq though. @ksletmoe-aws we should chat about your use-case and how we could address it.

ksletmoe-aws:
> API

@kixelated sounds good, I'll reach out on Slack this week

PR #1434, opened 2026-05-21
Split OriginConsumer into cheap read handle and announcement cursor

kixelated:
## Summary

Refactor the announcement subscription API to separate concerns: `OriginConsumer` is now a cheap, cloneable read handle over the broadcast tree, while `AnnounceConsumer` is the stateful cursor that receives announcement events. This eliminates the expensive per-clone channel allocation and allows multiple independent announcement streams from the same consumer.

## Key Changes

- **`OriginConsumer` refactored to be cheap to clone**: Removed the per-instance `mpsc::UnboundedReceiver` and `ConsumerId`. Cloning now shares the underlying tree state without allocating any per-cursor resources.

- **New `AnnounceConsumer` type**: Extracted the announcement stream logic into a separate struct that holds the channel receiver and consumer ID. Created via `OriginConsumer::announced()` or `AnnounceProducer::consume()`.

- **New `AnnounceProducer` type**: Symmetric counterpart to `AnnounceConsumer`. Provides a cheap read handle to the announcement stream for a subtree, similar to how `OriginConsumer` works for the broadcast tree.

- **API method renames**: 
  - `OriginConsumer::announced()` now returns `AnnounceConsumer` (was async method returning `Option<OriginAnnounce>`)
  - `OriginConsumer::try_announced()` → `AnnounceConsumer::try_next()`
  - `OriginConsumer::announced()` (async) → `AnnounceConsumer::next()` (async)
  - Added `AnnounceConsumer::is_closed()` to check cursor state

- **`OriginProducer::announces()` method**: New method to get an `AnnounceProducer` for the subtree, providing symmetric API to `consume()`.

- **Updated internal struct**: Renamed `OriginConsumerNotify` to `AnnounceConsumerNotify` for clarity.

- **Documentation improvements**: Updated doc comments to clarify the cheap-to-clone nature of `OriginConsumer` and the allocation semantics of `AnnounceConsumer`.

## Implementation Details

- `OriginConsumer` now derives `Clone` and implements it directly (no custom impl needed).
- `AnnounceConsumer` maintains the original per-cursor state: channel receiver, consumer ID, and root path.
- Both types support the same root path stripping and absolute path conversion utilities.
- Test helpers (`assert_next`, `assert_try_next`, etc.) moved to `AnnounceConsumer` impl block.
- All call sites updated to call `.announced()` on `OriginConsumer` to get an `AnnounceConsumer` before awaiting announcements.
- Added comprehensive test `test_consumer_clone_is_side_effect_free` to verify that cloning `OriginConsumer` does not drain announcement channels and that fresh `AnnounceConsumer` instances still receive the active backlog.

https://claude.ai/code/session_01BSuKtGPPntcBMEQiGb3Pn6

### DIFF
--- a/rs/libmoq/src/origin.rs
+++ b/rs/libmoq/src/origin.rs
@@ -40,7 +40,7 @@ impl Origin {
 
 	pub fn announced(&mut self, origin: Id, on_announce: OnStatus) -> Result<Id, Error> {
 		let origin = self.active.get_mut(origin).ok_or(Error::OriginNotFound)?;
-		let consumer = origin.consume();
+		let consumer = origin.consume().announced();
 		let channel = oneshot::channel();
 
 		let entry = TaskEntry {
@@ -64,8 +64,8 @@ impl Origin {
 		Ok(id)
 	}
 
-	async fn run_announced(task_id: Id, mut consumer: moq_net::OriginConsumer) -> Result<(), Error> {
-		while let Some((path, broadcast)) = consumer.announced().await {
+	async fn run_announced(task_id: Id, mut consumer: moq_net::AnnounceConsumer) -> Result<(), Error> {
+		while let Some((path, broadcast)) = consumer.next().await {
 			let mut state = State::lock();
 
 			// Stop if the callback was revoked by close.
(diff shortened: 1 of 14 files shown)


rs/libmoq/src/origin.rs
```diff
@@ -108,8 +108,6 @@ impl Origin {
 		let origin = self.active.get_mut(origin).ok_or(Error::OriginNotFound)?;
 		// TODO: expose an async variant backed by `announced_broadcast` so FFI callers can wait
 		// for gossip instead of racing it.
-		// Uses the deprecated direct lookup to avoid the per-call cost of OriginProducer::consume().
-		#[allow(deprecated)]
 		origin.get_broadcast(path).ok_or(Error::BroadcastNotFound)
```
kixelated: We should remove this deprecated method because it's cheap to consume().get_broadcast() now

PR #1437, opened 2026-05-21
mux: emit avcC/hvcC in catalog description for inline-SPS/PPS importers

kixelated:
## Summary

The `Avc3` and `Hev1` importers cache parameter set NALs for re-insertion before keyframes, but until now they left the catalog `description` field empty. Downstream consumers that need an out-of-band `AVCDecoderConfigurationRecord` / `HEVCDecoderConfigurationRecord` (KVS via MKV CodecPrivate; CMAF muxers; anything not willing to scrape the bitstream) had no way to get one from an `avc3`/`hev1` broadcast.

This PR:

- Builds an avcC from cached SPS+PPS the first time both are observed, sets it as the rendition's `description`, and republishes the catalog. Same flow for hvcC once VPS+SPS+PPS are all cached.
- Reorders the NAL handlers so the cache update precedes `init()` — that way the description builder sees the latest NAL, not the previous one.
- In `Hev1`, splits "codec changed" (which requires a new track) from "description appeared" (in-place rendition update) so subscribers don't re-fetch the track when only the description becomes available.

The avcC layout follows ISO/IEC 14496-15 §5.3.3.1.2 (standard, no high-profile extension fields — those are inferable from the inline SPS we ship anyway). The hvcC layout follows §8.3.3; chroma/bit-depth/temporal-id fields come from the parsed SPS.

## Test plan

- [x] `cargo test --lib` on moq-mux — 84 tests pass
- [x] Verified end-to-end pilot rebuild against a downstream consumer that requires `description` populated (KVS MKV bridge in `quartermaster/moq-pilot` worker)
- [ ] Manual: confirm the description appears in a live catalog snapshot and that consumers using `description` (not just inline NALs) can decode

### DIFF
--- a/rs/moq-mux/src/import/avc3.rs
+++ b/rs/moq-mux/src/import/avc3.rs
@@ -1,4 +1,4 @@
-use super::annexb::{NalIterator, START_CODE};
+use super::annexb::NalIterator;
 use super::jitter::MinFrameDuration;
 use super::same_codec;
 
@@ -91,7 +91,9 @@ impl Avc3 {
 				profile: sps.profile_idc,
 				constraints: constraint_flags,
 				level: sps.level_idc,
-				inline: true,
+				// We now strip inline SPS/PPS from samples and ship them via
+				// `description` (avcC) instead — that's the avc1 contract.
+				inline: false,
 			}
 			.into(),
 			description,
@@ -234,7 +236,11 @@ impl Avc3 {
 		let nal_unit_type = header & 0b11111;
 		let nal_type = NalType::try_from(nal_unit_type).ok();
 
-		match nal_type {
+		// SPS/PPS are stored in the catalog `description` (avcC) and stripped
+		// from sample data — that's the avc1 contract. The rest of the NALs
+		// are emitted length-prefixed (4-byte big-endian, matching the
+		// `lengthSizeMinusOne = 3` we write in build_avcc).
+		let emit = match nal_type {
 			Some(NalType::Sps) => {
 				self.maybe_start_frame(pts)?;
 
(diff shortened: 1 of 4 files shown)


rs/moq-mux/src/import/avc3.rs
```diff
@@ -94,8 +111,23 @@ impl Avc3 {
...
+		// downstream subscribers don't have to re-fetch on every catalog tick.
+		// The first SPS (None → Some) also reuses the eagerly-created track.
+		let needs_retrack = self.config.as_ref().is_some_and(|old| !same_codec(old, &config));
+
+		if needs_retrack {
+			let old_name = self.track.name.clone();
+			tracing::debug!(name = ?old_name, "codec changed; replacing track");
+			catalog.video.renditions.remove(&old_name);
+
+			let new_track = self.broadcast.unique_track(".avc3")?;
```
kixelated: We should make this before removing the old one right?

rs/moq-mux/src/import/hev1.rs
```diff
@@ -87,17 +94,27 @@ impl Hev1 {
...
+		// update (e.g. cached_pps just arrived) reuses the existing track.
+		let needs_retrack = self.track.is_none() || !self.config.as_ref().is_some_and(|old| same_codec(old, &config));
+
+		if needs_retrack {
+			if let Some(track) = self.track.take() {
+				tracing::debug!(name = ?track.name, "reinitializing track");
+				catalog.video.renditions.remove(&track.name);
+			}
+
+			let track = self.broadcast.unique_track(".hev1")?;
```
kixelated: We should do this before removing right?

kixelated:
Closing — review feedback led us to the right architectural conclusion: the avc3/hev1 importers shouldn't transcode their own output. If a downstream consumer needs avc1/hvc1 (length-prefixed samples + out-of-band avcC/hvcC), that's the consumer's transcode and belongs in an exporter layer. Codec-aware writers like the upcoming `export::Mkv` are the right home for that conversion, where length-prefixing and CodecPrivate emission can share code with the format itself.

Not landing the avcC/hvcC emission in the importers means consumers that today need an out-of-band record will get it from the export layer instead.

Re-open or salvage commits if useful for the MKV exporter work.

PR #1439, opened 2026-05-21
Add per-track timescale and frame timestamps to moq-lite

kixelated:
This change introduces per-track timescale negotiation and per-frame timestamps to the moq-lite protocol, enabling accurate media synchronization across different frame rates and sample rates.

## Summary

Refactors the timestamp system from a generic `Timescale<const SCALE: u64>` type alias to a concrete `Timestamp` struct that carries both a raw value and its scale. This allows timestamps to be expressed in any unit (seconds, milliseconds, microseconds, nanoseconds) and enables the protocol to negotiate per-track timescales via SUBSCRIBE_OK messages.

## Key Changes

- **Timestamp struct redesign**: Replaced `Timescale<SCALE>` type alias with a concrete `Timestamp` struct containing `value: VarInt` and `scale: u64`. The scale is now carried per-timestamp instance rather than as a compile-time constant.

- **Unspecified scale handling**: `scale == 0` denotes an unspecified timescale (used by `Timestamp::ZERO` and `Timestamp::MAX` sentinels). Unit conversions and arithmetic against unspecified scales return `TimeOverflow` to prevent divide-by-zero errors.

- **Scale conversion API**: Added `Timestamp::convert()` and `Timestamp::as_scale()` methods for explicit scale conversions. All `as_*` methods (as_secs, as_millis, etc.) now return `Result<T, TimeOverflow>` instead of infallible values.

- **Signed delta encoding**: Added `Timestamp::checked_add_delta()` and `Timestamp::checked_delta_from()` for zigzag-encoded signed deltas used by moq-lite per-frame delta decoding (for B-frames).

- **Protocol integration**: 
  - Added `timescale: u64` field to `SubscribeOk` message (Lite05+)
  - Added `timescale: u64` field to `Track` struct
  - Added `timestamp: Timestamp` field to `Frame` struct
  - Updated `Version::Lite05` to indicate timestamp support on the wire

- **VarInt zigzag support**: Added `VarInt::from_zigzag()` and `VarInt::to_zigzag()` for signed integer encoding used in delta timestamps.

- **Ordering and comparison**: Implemented `Ord` and `PartialOrd` for `Timestamp` with debug assertions for scale compatibility. Cross-scale comparisons are allowed only when one side is a scale-0 sentinel.

- **Removed PartialOrd/Ord from Timescale**: The old generic type no longer derives these traits; the new `Timestamp` implements them with scale-aware semantics.

## Implementation Details

- Scale conversions use 128-bit intermediate arithmetic to avoid overflow during multiplication before division.
- The wire format encodes only the raw value as a QUIC varint; scale is carried out-of-band via track properties or SUBSCRIBE_OK.
- `Timestamp::now()` returns microsecond-scale timestamps anchored to a jittered wall-clock reference.
- All arithmetic operations (`checked_add`, `checked_sub`) require matching scales and return `TimeOverflow` on mismatch.
- The `Debug` impl chooses the largest unit (seconds, milliseconds, microseconds, nanoseconds) where the value has no fractional part.

## Compatibility

- Lite01-Lite04: Decode timestamps with `scale == 0` (unspecified)
- Lite05+: Negotiate per-track timescale via SUBSCRIBE_OK and encode per-frame timestamps

https://claude.ai/code/session_01Wy7egQjXvfPCKovs3mWANB


rs/hang/examples/video.rs
```diff
@@ -44,6 +44,7 @@ fn create_track(broadcast: &mut moq_net::BroadcastProducer) -> anyhow::Result<mo
 	let video_track = moq_net::Track {
 		name: "video".to_string(),
 		priority: 1, // Video typically has lower priority than audio
+		timescale: 1_000_000,
```
kixelated: Can you make a `Timescale` type? `Timescale::MICRO` is a lot nicer.

rs/hang/src/catalog/root.rs
```diff
@@ -80,6 +80,7 @@ impl Catalog {
 		moq_net::Track {
 			name: Catalog::DEFAULT_NAME.to_string(),
 			priority: 100,
+			timescale: 0,
```
kixelated: `Timescale::UNKNOWN`

rs/hang/src/container/frame.rs
```diff
@@ -29,14 +32,18 @@ pub struct Frame {
...
 	/// Encode the frame to the given group as a single moq-lite frame:
-	/// VarInt timestamp prefix followed by the raw codec payload.
+	/// VarInt timestamp prefix followed by the raw codec payload. Also stamps
+	/// the moq-net [`moq_net::Frame::timestamp`] so the wire layer can
+	/// delta-encode it independently on Lite05+ (the container-level prefix
+	/// stays as a duplicate for now).
 	pub fn encode(&self, group: &mut moq_net::GroupProducer) -> Result<(), Error> {
 		let mut header = BytesMut::new();
-		self.timestamp.encode(&mut header).map_err(moq_net::Error::from)?;
+		self.timestamp.encode_value(&mut header).map_err(moq_net::Error::from)?;
```
kixelated: I'd rather do something like `self.timestamp.value.encode`  or something

rs/hang/src/container/frame.rs
```diff
@@ -29,14 +32,18 @@ pub struct Frame {
...
 
 		let size = header.len() + self.payload.len();
 
-		let mut chunked = group.create_frame(size.into())?;
+		let net_frame = moq_net::Frame::new(size as u64).with_timestamp(self.timestamp);
```
kixelated: I think you remove `Frame::new` and require using the struct constructor instead. timestamp shouldn't be optional, builder doesn't make sense.

rs/hang/src/container/frame.rs
```diff
@@ -46,7 +53,7 @@ impl Frame {
 
 	/// Decode a frame from raw bytes (VarInt timestamp prefix + payload).
 	pub fn decode(mut buf: impl Buf) -> Result<Self, Error> {
-		let timestamp = Timestamp::decode(&mut buf)?;
+		let timestamp = Timestamp::decode_value(&mut buf, TIMESCALE)?;
```
kixelated: I think we decode the value separately (via u64::decode) then call Timestamp::new

rs/libmoq/src/consume.rs
```diff
@@ -47,7 +47,7 @@ impl Consume {
 
 	pub fn catalog(&mut self, broadcast: Id, on_catalog: OnStatus) -> Result<Id, Error> {
 		let broadcast = self.broadcast.get(broadcast).ok_or(Error::BroadcastNotFound)?.clone();
-		let catalog = broadcast.subscribe_track(&hang::catalog::Catalog::default_track())?;
+		let catalog = broadcast.subscribe_track_immediate(&hang::catalog::Catalog::default_track())?;
```
kixelated: no `subscribe_track_immediate`. Figure out a way to actually make this async.

rs/hang/src/container/frame.rs
kixelated: Or in this case, Timestamp::from_micros(value)

rs/moq-ffi/src/consumer.rs
```diff
@@ -77,7 +78,7 @@ impl MoqBroadcastConsumer {
 	/// Subscribe to the catalog for this broadcast.
 	pub fn subscribe_catalog(&self) -> Result<Arc<MoqCatalogConsumer>, MoqError> {
```
kixelated: Needs to also be async unfortunately.

rs/moq-ffi/src/consumer.rs
```diff
@@ -89,7 +90,11 @@ impl MoqBroadcastConsumer {
 	/// Frames are returned as plain byte payloads with no codec or container parsing.
 	pub fn subscribe_track(&self, name: String) -> Result<Arc<MoqTrackConsumer>, MoqError> {
```
kixelated: async boy

rs/moq-ffi/src/consumer.rs
```diff
@@ -89,7 +90,11 @@ impl MoqBroadcastConsumer {
...
 		let _guard = crate::ffi::RUNTIME.enter();
-		let track = self.inner.subscribe_track(&moq_net::Track { name, priority: 0 })?;
+		let track = self.inner.subscribe_track_immediate(&moq_net::Track {
+			name,
+			priority: 0,
+			timescale: 0,
```
kixelated: We need to change `subscribe_track` so it only takes a `&str name` instead of a Track. The timescale and priority arrives in the SUBSCRIBE_OK instead. To allow subscriber priorities, we should add a `Delivery` struct in the future for overrides, but we need to start splitting track properties (set by the publisher) and track requests (set by the subscriber)

rs/moq-mux/src/container/cmaf.rs
```diff
@@ -144,7 +144,7 @@ pub(crate) fn decode(data: Bytes, timescale: u64) -> Result<Vec<Frame>, Error> {
 
 			let cts = entry.cts.unwrap_or_default() as i64;
 			let pts = dts.checked_add_signed(cts).ok_or(Error::PtsOverflow)?;
-			let timestamp = Timestamp::from_scale(pts, timescale)?;
+			let timestamp = Timestamp::from_scale(pts, timescale, crate::container::TIMESCALE)?;
```
kixelated: Kinda confusing. Maybe make a `.convert` method or something to change timescale units? `Timestamp::new(pts, timescale).convert(crate::container::TIMESCALE)`

rs/moq-mux/src/export/fmp4.rs
```diff
@@ -56,7 +56,7 @@ impl Fmp4 {
 	/// The hang catalog is subscribed internally; per-rendition tracks are (un)subscribed
 	/// as the catalog changes.
 	pub fn new(broadcast: moq_net::BroadcastConsumer) -> Result<Self, crate::Error> {
-		let catalog_track = broadcast.subscribe_track(&hang::Catalog::default_track())?;
+		let catalog_track = broadcast.subscribe_track_immediate(&hang::Catalog::default_track())?;
```
kixelated: Cant do this in the constructor any longer. please stop using subscribe_track_immediate, we need to fix the API.

rs/moq-mux/src/import/aac.rs
```diff
@@ -112,7 +112,8 @@ impl Aac {
 		mut catalog: crate::catalog::Producer,
 		config: AacConfig,
 	) -> anyhow::Result<Self> {
-		let track = broadcast.unique_track(".aac")?;
+		let mut track = broadcast.unique_track(".aac")?;
```
kixelated: Need to improve this API. What about the priority too? unique_track should probably return a String or something?

I really don't like `.set_timescale`. Publisher track properties like the name/priority/timescale should be fixed. Allowing them to be updated can only cause pain.

rs/moq-net/src/model/track.rs
```diff
@@ -601,6 +652,23 @@ impl TrackConsumer {
...
+	}
+
+	/// Poll for the negotiated timescale, without blocking.
+	pub fn poll_timescale(&self, waiter: &conducer::Waiter) -> Poll<Result<u64>> {
+		self.poll(waiter, |state| state.poll_timescale())
+	}
+
+	/// Wait until the track's timescale has been negotiated (e.g. via SUBSCRIBE_OK
+	/// for subscribers, or set immediately by publishers).
+	pub async fn timescale(&self) -> Result<u64> {
```
kixelated: Please no. Return a TrackConsumer only when we get the SUBSCRIBE_OK.

rs/moq-net/src/lite/publisher.rs
```diff
@@ -643,7 +663,28 @@ impl<S: web_transport_trait::Session> Subscription<S> {
...
+		// `[zigzag-delta timestamp][size][payload]`. With `None`, the timestamp
+		// is omitted entirely — saves a byte per frame on tracks where timing
+		// isn't meaningful (catalogs, control channels, IETF transport).
+		// Refuse a frame missing a timestamp or at the wrong scale rather than
+		// silently encoding a zero delta and corrupting decode.
+		if let Some(track_timescale) = self.timescale {
+			let ts = frame.timestamp.ok_or(Error::ProtocolViolation)?;
+			if ts.scale() != track_timescale {
+				return Err(Error::ProtocolViolation);
+			}
```
kixelated: Ideally, we should catch this earlier at the `model` level. Make a new error that is more descriptive?

rs/moq-net/src/lite/publisher.rs
```diff
@@ -643,7 +663,28 @@ impl<S: web_transport_trait::Session> Subscription<S> {
...
 		mut frame: FrameConsumer,
+		prev_ts: &mut u64,
 	) -> Result<(), Error> {
+		// Per-frame wire layout when `self.timescale` is Some:
+		// `[zigzag-delta timestamp][size][payload]`. With `None`, the timestamp
+		// is omitted entirely — saves a byte per frame on tracks where timing
+		// isn't meaningful (catalogs, control channels, IETF transport).
+		// Refuse a frame missing a timestamp or at the wrong scale rather than
+		// silently encoding a zero delta and corrupting decode.
+		if let Some(track_timescale) = self.timescale {
```
kixelated: Make sure frame.timestamp is None if timescale is None.

PR #1466, opened 2026-05-23
ci: skip unchanged languages with `just changed`

kixelated:
## Summary

- `just ci` now diffs against `$GITHUB_BASE_REF` (CI) or `origin/main` (local) and only runs the per-language `ci` recipes for whichever scopes the diff touches. Same code path runs locally and in GitHub Actions, so pre-push validation matches what the workflow does.
- Each per-language justfile (`js`, `py`, `rs`, `kt`, `swift`) owns its own `ci` recipe so adding new checks lives next to the code it gates.
- `rs ci` drops the duplicated `just check --workspace` it used to run twice. `cargo publish --dry-run` moves to `release-rs.yml` as a new `Verify packaging` step (release-plz already gates publishability there).
- Adds `concurrency:` cancellation to `check.yml` so superseded PR runs stop immediately.
- New `check-swift.yml` on `macos-latest`, gated on `swift/**` and `rs/moq-ffi/**`. Previously `just swift check` silently no-op'd on the ubuntu runner, so this is the first real Swift PR coverage; the first runs may surface latent breakage.
- `CLAUDE.md`: codify "local-first" so workflow YAML delegates to `just` by default and only reaches for plugins (`dorny/paths-filter`, etc.) when a recipe can't express the logic.

## Trade-off

This deliberately keeps everything in one ubuntu job rather than splitting per-language jobs across runners. Wins come from skipping unchanged scopes, cancelling superseded runs, and removing duplicated work. Cross-runner parallelism (matrix from `just changed --json`) is straightforward to add later if a multi-language PR's wall-clock becomes the bottleneck.

## Scope detection

`just changed` maps changed paths to scope tokens. The rules (see `justfile`):

| Scope   | Triggers |
|---------|----------|
| `js`    | `js/**`, `package.json`, `bun.lock`, `bun.lockb`, `biome.jsonc` |
| `rs`    | `rs/**`, `Cargo.toml`, `Cargo.lock` |
| `py`    | `py/**`, `pyproject.toml`, `uv.lock`, `rs/moq-ffi/**` |
| `kt`    | `kt/**`, `rs/moq-ffi/**` |
| `swift` | `swift/**`, `rs/moq-ffi/**` |
| `nix`   | `flake.nix`, `flake.lock` |
| `md`    | `**/*.md` |

Changes to `justfile`, any `*/justfile`, or `.github/workflows/check.yml` fan out to all scopes (orchestration changes can affect any language).

## Test plan

- [x] `just changed` returns `js rs py kt swift nix md` in this branch (justfile + check.yml both modified, so all scopes fan out as designed).
- [x] Spot-checked the path regex against 16 simulated file lists (js-only, `rs/moq-ffi` → 4 scopes, root manifests, justfile fan-out, doc-only → `md`, `release-rs.yml`-only → no scope, etc.) — all correct.
- [x] `actionlint` clean on `check.yml`, `check-swift.yml`, and `release-rs.yml`.
- [x] All justfiles parse and the new `ci` recipes appear in `just --list`.
- [ ] **Post-merge**: open a JS-only PR and confirm only the JS scope runs.
- [ ] **Post-merge**: open a PR touching `rs/moq-ffi` and confirm rs+py+kt run on the ubuntu job and `check-swift.yml` fires on macOS.
- [ ] **Post-merge**: push two commits to a PR in quick succession and confirm the first run is cancelled.
- [ ] **Post-merge**: confirm `Verify packaging` step in `release-rs.yml` runs on the next main push.


.github/workflows/release-rs.yml
```diff
@@ -41,6 +41,12 @@ jobs:
       - uses: DeterminateSystems/nix-installer-action@00199f951aeb9404028a6e4b95ad42546f73296a # main
       - uses: DeterminateSystems/magic-nix-cache-action@908b263ff629f4cc17666315b7fd3ec127c6244d # main
 
+      # Catch packaging issues across every workspace crate before
+      # release-plz attempts the real publish. PR CI skips this step to
+      # keep PRs fast; this is the gate that catches it on main.
+      - name: Verify packaging
+        run: nix develop --command cargo publish --dry-run --workspace
```
kixelated: Why do we need to `--dry-run` if we're about to call `just rs release` immediately afterward?

justfile
```diff
@@ -41,23 +41,92 @@ check *args:
 	just rs check {{ args }}
 	bun remark . --quiet --frail
 
-# Comprehensive CI: per-language checks + cross-cutting smoke tests.
-ci:
+# Print the language scopes touched vs $GITHUB_BASE_REF (CI) or origin/main (local).
+changed:
```
kixelated: Maybe we should delegate to each `justfile` for this? So `just js changed` is the source of truth for what files to watch? IDK if it's worth it.

Or maybe do this check as part of `just ci`? We can exit early if nothing has changed?

.github/workflows/release-rs.yml
kixelated: Good point — dropped the step in 5201ec6e. release-plz already attempts the real publish for crates being released, and the marginal benefit of catching packaging issues for crates *not* being released this run wasn't worth the step.

justfile
kixelated: Kept it central for now. The early-exit in `just ci` is already in place (prints `No language scopes changed; nothing to do.`). On delegating to per-language justfiles: I considered it but the path patterns are simple enough that one regex is easier to scan than calling 7 sub-justs and unioning their output. The downside is you edit the root justfile to add a new path scope. Happy to flip if you'd rather have it next to the language.

kixelated: Done in 7a26afe6 — each per-language justfile now has a `changed BASE` recipe that prints its scope name when touched, and root just unions stdout from all five. Bonus: this lets me drop the over-conservative fan-out rules (no longer fans out on `*/justfile` or `check*.yml` changes; only on root `justfile`, which actually controls dispatch). Also added an optional BASE arg to both `just changed` and `just ci` so you can do `just ci HEAD~3` etc.

PR #1484, opened 2026-05-24
feat: add moq-audio crate, raw-audio FFI, and rename moq-codec to moq-video

kixelated:
## Summary

- New `rs/moq-audio` crate: native Opus encode/decode of raw PCM on top of `moq-mux` + `hang`, with a rubato resampler so callers can pass any WebCodecs `AudioData.format` and sample rate.
- `moq-ffi` and `libmoq` gain raw-audio publish/subscribe APIs so Python/Swift/Kotlin and C callers can drive a microphone or speaker without bringing their own codec.
- Rename the previously empty `moq-codec` placeholder crate to `moq-video` so we squat on the right name; audio and video have different enough I/O shapes to live in separate crates.

## Why

Today, audio in moq flows as already-encoded bitstreams. `moq-mux` parses codec configuration headers (OpusHead, AudioSpecificConfig) and passes raw Opus/AAC packets through to `moq-net` unchanged. The browser path uses WebCodecs; native callers (`moq-ffi`, `libmoq`) had to bolt on a codec library themselves. This PR closes that gap for Opus.

## What's in `moq-audio`

- `AudioFormat` enum mirroring WebCodecs `AudioData.format` (U8/S16/S32/F32 ± planar), with helpers to convert any layout to/from interleaved f32 (the codec boundary).
- `AudioSamples` carrier: format + sample rate + channels + timestamp + bytes.
- Generic `Encoder` / `Decoder` traits, with an Opus implementation over the [`opus`](https://crates.io/crates/opus) crate at fixed 20 ms frames.
- `AudioProducer` / `AudioConsumer` that wire PCM through `moq_mux::container::Producer<legacy::Wire>` and register/read the `hang::catalog::AudioConfig`.
- Cargo features:
  - `opus` (default): system libopus.
  - `opus-static`: vendor libopus via `audiopus_sys/static`. Enabled by `moq-ffi` and `libmoq` so distributable artifacts (Python wheels, C staticlib) don't require libopus on the host.
  - `resample` (default): rubato wrapper with a passthrough mode when rates match.

## FFI surface

**moq-ffi** (`rs/moq-ffi/src/audio.rs`):
- `MoqAudioFormat`, `MoqRawAudio` records.
- `MoqBroadcastProducer::publish_raw_audio_opus(name, format, rate, channels, bitrate?)` returning `MoqRawAudioProducer { write, finish }`.
- `MoqBroadcastConsumer::subscribe_raw_audio_opus(name, audio_config, format, output_rate?, output_channels?)` returning `MoqRawAudioConsumer { next, cancel }`.

**libmoq** (`rs/libmoq/src/audio.rs`):
- `moq_audio_format` enum, `moq_raw_audio` struct.
- `moq_publish_raw_audio_opus` / `_write` / `_close`.
- `moq_consume_raw_audio_opus` / `_close` / `_sample` / `_sample_free`.

Both stay on the existing track-selection model (name on moq-ffi, catalog index on libmoq). ABR-style auto-selection is a TODO at the entry points.

## Size impact on shipped artifacts

Measured by building `libmoq` + `moq-ffi` in `--release` with vs. without `moq-audio`:

| Artifact | Baseline | With moq-audio (libopus bundled) | Delta |
|---|---:|---:|---:|
| `libmoq.a` | 57.5 MB | 58.4 MB | +0.92 MB |
| `libmoq_ffi.a` | 70.6 MB | 71.4 MB | +0.88 MB |
| `libmoq_ffi.dylib` (unstripped) | 12.94 MB | 13.31 MB | +0.37 MB |
| `libmoq_ffi.dylib` (stripped) | 10.50 MB | 10.83 MB | **+0.34 MB** |

The cdylib is what ships in `py/moq-rs` Python wheels, so the practical cost is ~340 KB stripped, almost all libopus.

Note: on macOS the `opus-static` Cargo feature is effectively a no-op because `audiopus_sys` always vendors libopus when there's no system copy to link against. The feature still matters on Linux distributions that ship libopus, where it forces vendoring instead of dynamic-linking.

## What's deliberately out of scope

- **AAC.** Codec module shape is generic so it can drop in later behind its own feature.
- **Video.** `moq-video` is a renamed-but-still-empty placeholder. Once the input/output story for video samples is settled it can mirror `moq-audio`'s structure.
- **ABR / auto track selection.** Catalog index (libmoq) and track name (moq-ffi) selection is preserved.
- **JS side.** Browser already encodes/decodes via WebCodecs; no changes.

## Test plan

- [x] `cargo test -p moq-audio` (14 unit tests + 2 end-to-end round-trip tests, including a 44.1 kHz S16 → Opus → 48 kHz catalog → 44.1 kHz S16 path through the resampler).
- [x] `cargo test -p moq-ffi`, `cargo test -p libmoq`, full `just test` suite (Rust + JS + Python).
- [x] `just check` (cargo check + clippy -D warnings + fmt + doc + shear + sort + remark).
- [x] `cargo check -p moq-audio --no-default-features --features opus,opus-static,resample` to confirm the static-libopus build path.
- [ ] Smoke-test the Python wheel against a real microphone / speaker once `py/moq-rs` is rebuilt.


rs/libmoq/src/audio.rs
```diff
@@ -0,0 +1,378 @@
...
+			let mut state = State::lock();
+			let Some(Some(entry)) = state.audio.consumer_tasks.get(task_id) else {
+				return Ok(());
+			};
+			let callback = entry.callback;
+			let sample_id = state.audio.samples.insert(samples)?;
+			drop(state);
+
+			callback.call(Ok(sample_id));
+		}
```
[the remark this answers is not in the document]
kixelated: Skipping — same race exists in every existing libmoq callback path (`run_catalog`, `run_track` in consume.rs): callback is snapshotted under lock, lock dropped, then invoked. Fixing it only in audio.rs would be inconsistent; a libmoq-wide pattern change is out of scope for this PR.

rs/libmoq/src/audio.rs
```diff
@@ -0,0 +1,378 @@
...
+/// Returns a non-zero handle on success or a negative error code.
+///
+/// # Safety
+/// - `name` must point to `name_len` bytes of UTF-8.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn moq_publish_raw_audio_opus(
+	broadcast: u32,
+	name: *const c_char,
+	name_len: usize,
+	_input_format: moq_audio_format,
```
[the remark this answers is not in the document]
kixelated: Done in c87a3d58 — dropped `input_format` from `publish_raw_audio_opus` in both moq-ffi and libmoq. The per-write `format` field is the source of truth, so accepting it at publish time was misleading.

rs/moq-audio/src/consumer.rs
```diff
@@ -0,0 +1,200 @@
...
+	output_rate: u32,
+	output_channels: u32,
+
+	#[cfg(feature = "resample")]
+	resampler: Option<Resampler>,
+
+	/// Leftover decoded samples (in the *output* rate/channels, interleaved f32)
+	/// from the resampler that haven't been requested yet.
+	#[cfg(feature = "resample")]
+	pending: Vec<f32>,
```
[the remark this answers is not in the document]
kixelated: Done in c87a3d58 — removed the dead `pending` / `next_timestamp_us` / `flush()` plumbing from `AudioConsumer`. The resampler emits everything it can per chunk; the small (<20ms) tail loss at end-of-track is acceptable for now and can be revisited if a use case demands it.

rs/moq-audio/src/consumer.rs
```diff
@@ -0,0 +1,200 @@
...
+			None
+		} else {
+			let chunk_frames = (decoder.sample_rate() as usize * 20) / 1000;
+			Some(Resampler::new(
+				decoder.sample_rate(),
+				output_rate,
+				decoder.channel_count(),
+				chunk_frames,
+			)?)
+		};
```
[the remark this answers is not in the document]
kixelated: Done in c87a3d58 — `AudioConsumer::subscribe` now errors with `AudioError::Unsupported` when `output_channels` differs from the decoder. Channel up/downmix stays a TODO; the error message points there.

rs/moq-audio/src/error.rs
```diff
@@ -0,0 +1,60 @@
...
+		Self::Resample(err.to_string())
+	}
+}
+
+#[cfg(feature = "resample")]
+impl From<rubato::ResampleError> for AudioError {
+	fn from(err: rubato::ResampleError) -> Self {
+		Self::Resample(err.to_string())
+	}
+}
```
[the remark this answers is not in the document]
kixelated: Done in c87a3d58 — switched to `ResamplerConstruction(#[from] rubato::ResamplerConstructionError)` and `Resample(#[from] rubato::ResampleError)` so the typed source is preserved instead of being flattened to a string.

rs/moq-audio/src/producer.rs
```diff
@@ -0,0 +1,234 @@
...
+		} else {
+			// Chunk size: 20ms of input frames — same window the codec uses.
+			let chunk_frames = (input_rate as usize * 20) / 1000;
+			Some(Resampler::new(
+				input_rate,
+				encoder.sample_rate(),
+				input_channels,
+				chunk_frames,
+			)?)
+		};
```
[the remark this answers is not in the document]
kixelated: Done in c87a3d58 — same fix as the consumer side: `AudioProducer::new` rejects `input_channels` that differ from `encoder.channel_count()` instead of silently doing nothing.

rs/moq-audio/src/resample.rs
```diff
@@ -0,0 +1,220 @@
...
+		Self::with_channels(input_rate, output_rate, channels, channels, chunk_frames)
+	}
+
+	fn with_channels(
+		input_rate: u32,
+		output_rate: u32,
+		input_channels: u32,
+		output_channels: u32,
+		chunk_frames: usize,
+	) -> Result<Self, AudioError> {
```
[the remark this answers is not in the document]
kixelated: Done in c87a3d58 — `Resampler::with_channels` now returns `AudioError::Unsupported("chunk_frames must be > 0")` before any further work, so `process()` can no longer spin on a zero-length drain loop.

rs/libmoq/src/audio.rs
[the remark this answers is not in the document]
kixelated: Thanks, no need — leaving that as a judgment call for the maintainer to track however they normally surface libmoq-wide refactors (likely a code-level TODO or a manually filed issue).

.cargo/config.toml
```diff
@@ -1,2 +1,12 @@
 [build]
 rustflags = ["--cfg", "tokio_unstable"]
+
+[env]
+# audiopus_sys 0.2.2 vendors a libopus source tree whose CMakeLists.txt asks
```
kixelated: Is this a reason to look for another Opus crate?

rs/moq-audio/src/codec/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+
+use bytes::Bytes;
+
+use crate::AudioError;
+
+#[cfg(feature = "opus")]
+pub mod opus;
+
+#[cfg(feature = "opus")]
+pub use opus::{OpusDecoder, OpusEncoder};
```
kixelated: I would remove the Opus prefix.

rs/moq-audio/src/codec/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+
+/// An audio encoder that turns interleaved `f32` PCM into per-packet
+/// codec output.
+///
+/// Implementations are stateful and may buffer input until they have
+/// enough samples to emit a packet — most codecs operate on fixed-size
+/// frames (Opus: 20 ms by default; AAC: 1024 samples).
+pub trait Encoder: Send {
```
kixelated: Do we need a trait right now? I think we remove it until a second codec is added, so we can better verify the interface works.

rs/moq-audio/src/codec/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+//! layer handles format conversion and (optionally) resampling.
+
+use bytes::Bytes;
+
+use crate::AudioError;
+
+#[cfg(feature = "opus")]
+pub mod opus;
+
+#[cfg(feature = "opus")]
```
kixelated: Do we need a feature flag for a single codec supported?

rs/moq-audio/src/codec/opus.rs
```diff
@@ -0,0 +1,259 @@
...
+//! becomes its own moq-lite group (see [`AudioProducer`](crate::AudioProducer)).
+
+use bytes::Bytes;
+use opus::{Application, Channels};
+
+use crate::AudioError;
+use crate::codec::{Decoder, Encoder};
+
+/// Frame duration in milliseconds — matches the JS publisher.
+const FRAME_DURATION_MS: u32 = 20;
```
kixelated: I guess we can make it configurable? Default to 20ms of course.

rs/moq-audio/src/codec/opus.rs
```diff
@@ -0,0 +1,259 @@
...
+	bitrate: Option<u32>,
+	output: Vec<u8>,
+}
+
+impl OpusEncoder {
+	/// Build a new encoder.
+	///
+	/// `bitrate` is the target bitrate in bits per second; `None` lets
+	/// libopus pick a sensible default (96 kbps for stereo at 48 kHz).
+	pub fn new(sample_rate: u32, channel_count: u32, bitrate: Option<u32>) -> Result<Self, AudioError> {
```
kixelated: Take a EncoderConfig struct instead.

rs/moq-audio/src/codec/opus.rs
```diff
@@ -0,0 +1,259 @@
...
+
+	fn sample_rate(&self) -> u32 {
+		self.sample_rate
+	}
+
+	fn channel_count(&self) -> u32 {
+		self.channel_count
+	}
+
+	fn encode(&mut self, pcm: &[f32]) -> Result<Bytes, AudioError> {
```
kixelated: Should we support configurable formats?

rs/moq-audio/src/codec/opus.rs
```diff
@@ -0,0 +1,259 @@
...
+	max_frame_size: usize,
+}
+
+impl OpusDecoder {
+	/// Build a new decoder.
+	///
+	/// Pass the sample rate and channel count from the hang catalog
+	/// (parsed from the OpusHead `description`); defaults to 48 kHz / 2 ch
+	/// if the description is missing.
+	pub fn new(sample_rate: u32, channel_count: u32) -> Result<Self, AudioError> {
```
kixelated: Same DecoderConfig

rs/moq-audio/src/consumer.rs
```diff
@@ -0,0 +1,151 @@
...
+
+#[cfg(feature = "opus")]
+use crate::codec::OpusDecoder;
+
+#[cfg(feature = "resample")]
+use crate::resample::Resampler;
+
+/// Subscribe to a moq-mux audio track and emit decoded PCM in the caller's chosen format.
+pub struct AudioConsumer {
+	decoder: Box<dyn Decoder>,
```
kixelated: Why not <T: Decoder> instead?

rs/moq-audio/src/consumer.rs
```diff
@@ -0,0 +1,151 @@
...
+	track: moq_mux::container::Consumer<moq_mux::container::legacy::Wire>,
+
+	output_format: AudioFormat,
+	output_rate: u32,
+	output_channels: u32,
+
+	#[cfg(feature = "resample")]
```
kixelated: IDK just always enable it. Seems silly to feature gate this, it's small and useful.

rs/moq-audio/src/consumer.rs
```diff
@@ -0,0 +1,151 @@
...
+impl AudioConsumer {
+	/// Subscribe to `name` in `broadcast` using the catalog entry to pick the codec.
+	#[cfg(feature = "opus")]
+	pub fn subscribe_opus(
+		broadcast: &moq_net::BroadcastConsumer,
+		config: &hang::catalog::AudioConfig,
+		name: impl Into<String>,
+		output_format: AudioFormat,
+		output_rate: Option<u32>,
+		output_channels: Option<u32>,
```
kixelated: Should be a struct, like DecoderConfig.

rs/moq-audio/src/format.rs
```diff
@@ -0,0 +1,259 @@
...
+	/// Whether the underlying sample type is floating-point.
+	pub fn is_float(self) -> bool {
+		matches!(self, Self::F32 | Self::F32Planar)
+	}
+
+	/// Convert a raw PCM buffer in this format to interleaved `f32` in `[-1.0, 1.0]`.
+	///
+	/// Codecs (Opus, AAC) work in floating-point internally, so this is the
+	/// universal adapter between caller buffers and codec input.
+	pub fn to_interleaved_f32(self, data: &[u8], channels: u32) -> Result<Vec<f32>, AudioError> {
```
kixelated: Ideally we can avoid allocating on a no-op.

Also is there not an existing library for this? IDK seems like something that could be SIMD accelerated or something, and maybe the resampler supports.

rs/moq-audio/src/producer.rs
```diff
@@ -0,0 +1,242 @@
...
+		Ok(())
+	}
+
+	fn ensure_catalog_registered(&mut self) -> Result<(), AudioError> {
+		if self.catalog_registered {
+			return Ok(());
+		}
+		let config = self.encoder.config();
+		self.catalog.lock().audio.insert(&self.track_name, config)?;
+		self.catalog_registered = true;
```
kixelated: Why not do this on the constructor? Do we need frames to initialize the encoder config..?

CLAUDE.md
```diff
@@ -32,7 +32,8 @@ The project contains multiple layers of protocols:
    - catalog: a JSON track containing a description of other tracks and their properties (for WebCodecs).
    - container: each frame consists of a timestamp and codec bitstream
    - watch/publish: dedicated packages for subscribing/publishing with optional UI overlays
-5. **application** - Users building on top of `moq-net` or `hang`
+5. **moq-audio** - Native Rust audio codecs (Opus today; AAC/video in future crates) on top of `moq-mux` + `hang`. Wraps libopus behind `opus`/`opus-static` features and ships a rubato-backed resampler so callers can hand it any WebCodecs `AudioData.format` and sample rate. Used by `moq-ffi`/`libmoq` to import/export raw PCM samples without bringing a codec library themselves.
```
kixelated: Pretty long, should be brief. Just say it encodes and decodes Opus (more codecs to come).

rs/moq-ffi/src/audio.rs
```diff
@@ -0,0 +1,259 @@
...
+		})
+	}
+}
+
+/// A buffer of raw PCM samples.
+///
+/// `data` layout is fully described by `format` and `channel_count`.
+/// `timestamp_us` is the presentation timestamp of the first frame.
+#[derive(uniffi::Record)]
+pub struct MoqRawAudio {
```
kixelated: MoqAudioFrame maybe?

rs/moq-ffi/src/audio.rs
```diff
@@ -0,0 +1,259 @@
...
+/// Producer for a raw-audio track.
+///
+/// Built via [`MoqBroadcastProducer::publish_raw_audio_opus`]. Each
+/// [`write`](Self::write) accepts PCM whose layout is described by the
+/// per-call [`MoqRawAudio::format`] field; the producer converts to
+/// interleaved `f32` per write, resamples to the codec's rate if
+/// needed, then encodes through libopus and publishes to the
+/// underlying moq broadcast.
+#[derive(uniffi::Object)]
+pub struct MoqRawAudioProducer {
```
kixelated: I prefer MoqAudio as the prefix. MoqAudioProducer would be best.

rs/moq-ffi/src/audio.rs
```diff
@@ -0,0 +1,259 @@
...
+impl MoqBroadcastProducer {
+	/// Open a raw-audio Opus track on this broadcast.
+	///
+	/// `sample_rate` and `channel_count` describe the PCM the caller will
+	/// feed to [`MoqRawAudioProducer::write`]; a resampler runs
+	/// internally if `sample_rate` isn't one Opus supports natively. The
+	/// per-write `MoqRawAudio.format` carries the sample layout, so no
+	/// format is needed at publish time. `bitrate` is in bits per
+	/// second; pass `None` for the libopus default.
+	pub fn publish_raw_audio_opus(
```
kixelated: remove _raw IMO

rs/moq-ffi/src/audio.rs
```diff
@@ -0,0 +1,259 @@
...
+		&self,
+		name: String,
+		sample_rate: u32,
+		channel_count: u32,
+		bitrate: Option<u32>,
```
kixelated: Feels like a struct?

rs/moq-ffi/src/audio.rs
```diff
@@ -0,0 +1,259 @@
...
+}
+
+/// A buffer of raw PCM samples.
+///
+/// `data` layout is fully described by `format` and `channel_count`.
+/// `timestamp_us` is the presentation timestamp of the first frame.
+#[derive(uniffi::Record)]
+pub struct MoqRawAudio {
+	pub format: MoqAudioFormat,
+	pub sample_rate: u32,
```
kixelated, line 64: Do we need this on each Frame? Seems like it can only encourage misuse. We should require configuring the encoder with an input sample rate and output sample rate. Same with channel count.

kixelated: Same feedback for moq_audio::AudioSamples (should be rename to moq_audio::Frame).

rs/libmoq/src/audio.rs
```diff
@@ -0,0 +1,409 @@
...
+///   callers may reuse the buffer immediately.
+/// - **Consume** ([`moq_consume_raw_audio_sample`]): the pointer
+///   remains valid until the corresponding sample ID is released with
+///   [`moq_consume_raw_audio_sample_free`]. The on-frame callback only
+///   delivers an ID; the buffer outlives the callback, and
+///   [`moq_consume_raw_audio_close`] does not free already-delivered
+///   sample IDs.
+#[repr(C)]
+#[allow(non_camel_case_types)]
+pub struct moq_raw_audio {
```
kixelated: Can we use `moq_audio_frame`?

rs/libmoq/src/audio.rs
```diff
@@ -0,0 +1,409 @@
...
+	pub format: u32,
+	pub sample_rate: u32,
```
kixelated, line 99: Same deal, this shouldn't be per frame.

rs/libmoq/src/audio.rs
```diff
@@ -0,0 +1,409 @@
...
+/// - **Consume** ([`moq_consume_raw_audio_sample`]): the pointer
+///   remains valid until the corresponding sample ID is released with
+///   [`moq_consume_raw_audio_sample_free`]. The on-frame callback only
+///   delivers an ID; the buffer outlives the callback, and
+///   [`moq_consume_raw_audio_close`] does not free already-delivered
+///   sample IDs.
+#[repr(C)]
+#[allow(non_camel_case_types)]
+pub struct moq_raw_audio {
+	pub format: u32,
```
kixelated, line 97: Same deal, this shouldn't be per frame. We need to negotiate the format/sample/channel count when initializing the encoder, not per frame.

rs/libmoq/src/audio.rs
```diff
@@ -0,0 +1,409 @@
...
+		Ok(())
+	}
+
+	pub fn publish_close(&mut self, id: Id) -> Result<(), Error> {
+		let producer = self.producers.remove(id).ok_or(Error::MediaNotFound)?;
+		producer.finish()?;
+		Ok(())
+	}
+
+	#[allow(clippy::too_many_arguments)]
```
kixelated: struct pls

rs/libmoq/src/audio.rs
```diff
@@ -0,0 +1,409 @@
...
+/// format is needed at publish time.
+///
+/// `bitrate` is bits-per-second; pass 0 for the libopus default.
+///
+/// Returns a non-zero handle on success or a negative error code.
+///
+/// # Safety
+/// - `name` must point to `name_len` bytes of UTF-8.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn moq_publish_raw_audio_opus(
```
kixelated: moq_publish_audio.

Could opus be a "codec" string? Why do we need a specific method for opus?

rs/libmoq/src/audio.rs
```diff
@@ -0,0 +1,409 @@
...
+	broadcast: u32,
+	name: *const c_char,
+	name_len: usize,
+	sample_rate: u32,
```
kixelated: struct

rs/libmoq/src/audio.rs
```diff
@@ -0,0 +1,409 @@
...
+/// # Safety
+/// - `on_samples` must be valid until [`moq_consume_raw_audio_close`] is called.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn moq_consume_raw_audio_opus(
+	catalog: u32,
+	index: u32,
+	output_format: u32,
+	output_sample_rate: u32,
+	output_channels: u32,
+	on_samples: Option<extern "C" fn(user_data: *mut c_void, samples: i32)>,
```
kixelated: samples -> frame 

Unless you strongly think it should stay Samples. I can see the argument, just I use `Frame` everwhere else and I think it makes more sense IMO.

rs/libmoq/src/audio.rs
kixelated: And if this name is taken, `publish_audio_raw` works for me too. Just make sure it's `publish_audio_raw_write`, `_close`, etc.

kixelated (commented):
And if `publish_audio

.cargo/config.toml
kixelated: Investigated — `opus`/`audiopus_sys` is the only mature option, all forks land on the same vendored libopus + CMake combo. Kept it but added `OPUS_NO_PKG=1` alongside the CMake env so pkg-config can't pick up ffmpeg's transitive libopus and lie about static linking. Done in e21a232e.

rs/moq-audio/src/codec/mod.rs
kixelated: codec/mod.rs:16 (Opus prefix): Done in e21a232e — renamed `OpusEncoder` → `Encoder` and `OpusDecoder` → `Decoder`. Single-codec for now; when AAC lands we'll add a `Codec` dispatch.

rs/moq-audio/src/codec/mod.rs
kixelated: codec/mod.rs:24 (trait): Done in e21a232e — dropped the `Encoder` / `Decoder` traits entirely. Concrete types only. Will reintroduce once a second codec validates the abstraction.

rs/moq-audio/src/codec/mod.rs
kixelated: codec/mod.rs:15 (feature flag): Done in e21a232e — dropped `opus` and `resample` Cargo features. Always on.

rs/moq-audio/src/codec/opus.rs
kixelated: opus.rs:15 (frame duration): Done in e21a232e — `EncoderConfig::frame_duration: Duration` (default 20ms). Validated against libopus's 2.5/5/10/20/40/60 ms.

rs/moq-audio/src/codec/opus.rs
kixelated: opus.rs:54 (EncoderConfig): Done in e21a232e — `Encoder::new(EncoderConfig)`.

rs/moq-audio/src/codec/opus.rs
kixelated: opus.rs:103 (configurable formats): Done in e21a232e — encoder takes `input_format` in `EncoderConfig`; producer handles the conversion to libopus's interleaved f32 internally.

rs/moq-audio/src/codec/opus.rs
kixelated: opus.rs:131 (DecoderConfig): Done in e21a232e — `Decoder::new(catalog)` reads the codec config from the catalog; `AudioConsumer` carries the `DecoderConfig` (output format/rate/channels).

rs/moq-audio/src/consumer.rs
kixelated: consumer.rs:16 (<T: Decoder>): Done implicitly in e21a232e — since the `Decoder` trait is gone, the consumer just holds a concrete `Decoder`. No generic needed.

rs/moq-audio/src/consumer.rs
kixelated: consumer.rs:23 (resample always on): Done in e21a232e — `resample` feature gate removed.

rs/moq-audio/src/consumer.rs
kixelated: consumer.rs:36 (DecoderConfig struct): Done in e21a232e — `AudioConsumer::new(broadcast, catalog, name, DecoderConfig)`.

rs/moq-audio/src/format.rs
kixelated: format.rs:52 (avoid no-op alloc): Partially done in e21a232e — `as_interleaved_f32` returns `Cow<[f32]>` so the F32-interleaved-and-aligned fast path borrows the input bytes. For other layouts we still allocate; a SIMD/dasp pass is a follow-up.

rs/moq-audio/src/producer.rs
kixelated: producer.rs:186 (catalog on constructor): Done in e21a232e — `AudioProducer::new` registers the rendition immediately. No first-write side effect.

CLAUDE.md
kixelated: CLAUDE.md:35 (shorter blurb): Done in e21a232e — collapsed to one sentence.

rs/moq-ffi/src/audio.rs
kixelated: ffi/audio.rs:71 (MoqAudioFrame): Done in e21a232e.

rs/moq-ffi/src/audio.rs
kixelated: ffi/audio.rs:116 (MoqAudioProducer): Done in e21a232e — also `MoqAudioConsumer`.

rs/moq-ffi/src/audio.rs
kixelated: ffi/audio.rs:152 (drop _raw): Done in e21a232e — `publish_audio` / `subscribe_audio`. (The `_raw` ended up on the libmoq side only because `moq_consume_audio_close` collided with the existing encoded path; UniFFI doesn't have the same conflict.)

rs/moq-ffi/src/audio.rs
kixelated: ffi/audio.rs:157 (struct): Done in e21a232e — `MoqAudioEncoderConfig` record.

rs/moq-ffi/src/audio.rs
kixelated: ffi/audio.rs:73 (per-frame format): Done in e21a232e — format / sample rate / channels live in `MoqAudioEncoderConfig`. The frame is just bytes + timestamp.

kixelated: ffi/audio.rs:73 (same for moq_audio::AudioSamples): Done in e21a232e — renamed to `moq_audio::Frame` (timestamp + bytes); layout lives in `EncoderConfig` / `DecoderConfig`.

rs/libmoq/src/audio.rs
kixelated: libmoq/audio.rs:94 (moq_audio_frame): Done in e21a232e.

rs/libmoq/src/audio.rs
kixelated: libmoq/audio.rs:96 (not per frame): Done in e21a232e — `moq_audio_encoder_config` holds layout.

rs/libmoq/src/audio.rs
kixelated: libmoq/audio.rs:95 (encoder config): Same as above — `moq_audio_encoder_config` is now the negotiation point.

rs/libmoq/src/audio.rs
kixelated: libmoq/audio.rs:150 (struct pls): Done in e21a232e.

rs/libmoq/src/audio.rs
kixelated: libmoq/audio.rs:251 (moq_publish_audio + codec string): Done in e21a232e — function is `moq_publish_audio_raw` (had to suffix `_raw` to dodge the existing `moq_consume_audio_close`), and the encoder config takes a codec UTF-8 string ("opus" today; AAC etc. drop in later).

rs/libmoq/src/audio.rs
kixelated: libmoq/audio.rs:255 (struct): Done in e21a232e — `moq_audio_decoder_config`.

rs/libmoq/src/audio.rs
kixelated: libmoq/audio.rs:336 (samples → frame): Done in e21a232e — `moq_audio_frame` everywhere, no more "samples" naming.

rs/libmoq/src/audio.rs
kixelated: libmoq/audio.rs:251 (publish_audio_raw fallback): Done in e21a232e — landed on `_raw` precisely because `_close` collided. Functions are now `moq_publish_audio_raw{,_frame,_close}` and `moq_consume_audio_raw{,_close,_frame,_frame_free}`.

rs/moq-audio/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+
+[dependencies]
+audiopus_sys = { version = "0.2", default-features = false, features = ["static"] }
+bytes = "1"
+hang = { workspace = true }
+moq-mux = { workspace = true }
+moq-net = { workspace = true }
+opus = "0.3"
+rubato = "0.16"
+thiserror = "2"
```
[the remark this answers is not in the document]
kixelated: Declining — re-adding the feature flags would undo a deliberate call (see https://github.com/moq-dev/moq/pull/1484#discussion_r3295145786 / r3295149138 / r3295145617): single codec, tiny resampler, not worth the gating complexity. Will revisit if/when AAC or a heavier codec lands.

rs/moq-audio/src/format.rs
```diff
@@ -0,0 +1,279 @@
...
+
+		out
+	}
+
+	/// Convert interleaved `f32` PCM to this format's raw byte representation.
+	///
+	/// Returns a [`Cow::Borrowed`] of the input bytes when no conversion is
+	/// needed (`F32` interleaved). Integer formats clamp out-of-range samples
+	/// rather than wrapping.
+	pub fn from_interleaved_f32(self, samples: &[f32], channels: u32) -> Result<Vec<u8>, AudioError> {
```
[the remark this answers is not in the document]
kixelated: Done in 068c4cab — dropped the stale `Cow::Borrowed` line.

rs/moq-audio/src/producer.rs
```diff
@@ -0,0 +1,152 @@
...
+		config: EncoderConfig,
+	) -> Result<Self, AudioError> {
+		let encoder = Encoder::new(config)?;
+
+		let resampler = if encoder.config().input_sample_rate == encoder.codec_rate() {
+			None
+		} else {
+			let chunk_frames = (encoder.config().input_sample_rate as usize
+				* encoder.config().frame_duration.as_millis() as usize)
+				/ 1000;
```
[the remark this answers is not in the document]
kixelated: Done in 068c4cab — `chunk_frames` now uses `frame_duration.as_micros() / 1_000_000`, so 2.5 ms frame durations no longer truncate.

rs/moq-audio/tests/roundtrip.rs
```diff
@@ -0,0 +1,170 @@
...
+		let pcm: Vec<f32> = frame
+			.data
+			.chunks_exact(4)
+			.map(|c| f32::from_le_bytes([c[0], c[1], c[2], c[3]]))
+			.collect();
+		total_frames += pcm.len() as u64 / 2;
+		for s in &pcm {
+			total_energy += (*s as f64) * (*s as f64);
+		}
+	}
```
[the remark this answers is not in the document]
kixelated: Done in 068c4cab — both loops now wrap `consumer.read()` in a 5 s `tokio::time::timeout`.

rs/libmoq/src/audio.rs
```diff
@@ -0,0 +1,394 @@
...
+
+struct AudioTaskEntry {
+	#[allow(dead_code)] // Dropping signals shutdown via channel.
+	close: oneshot::Sender<()>,
+	callback: OnStatus,
+}
+
+fn parse_codec(codec_str: &str) -> Result<moq_audio::Codec, Error> {
+	match codec_str {
+		"opus" => Ok(moq_audio::Codec::Opus),
```
kixelated: "Opus" too

rs/moq-audio/src/codec.rs
```diff
@@ -0,0 +1,431 @@
...
+#[derive(Clone, Debug)]
+pub struct EncoderConfig {
+	pub codec: Codec,
+	pub input_format: AudioFormat,
+	pub input_sample_rate: u32,
+	pub input_channels: u32,
+	/// libopus bitrate in bits per second. `None` lets libopus pick.
+	pub bitrate: Option<u32>,
+	/// Encoded frame duration. Opus accepts 2.5 / 5 / 10 / 20 / 40 / 60 ms.
+	pub frame_duration: Duration,
```
kixelated, line 89: Should these be prefixed with output_?

We should also have an optional output_sample_rate. Maybe output_channels? 

Maybe two structs instead, EncoderInput and EncoderOutput?

rs/moq-audio/src/codec.rs
```diff
@@ -0,0 +1,431 @@
...
+	}
+}
+
+/// Decoder configuration.
+///
+/// `output_*` describe how the caller wants samples delivered.
+/// `None` for rate or channels means "match the codec's native shape
+/// from the catalog".
+#[derive(Clone, Debug, Default)]
+pub struct DecoderConfig {
```
kixelated: Do we need the output_ prefix?

rs/moq-audio/src/codec.rs
```diff
@@ -0,0 +1,431 @@
...
+	pub output_channels: Option<u32>,
+	/// Maximum buffering before skipping a stalled group.
+	///
+	/// Forwarded to [`moq_mux::container::Consumer::with_latency`]: if a
+	/// group is stuck and a newer group is older-than-`max_latency` ahead,
+	/// the consumer skips. `None` (the default) keeps the moq-mux default
+	/// of zero, which skips aggressively. Set this to the playout buffer
+	/// you can tolerate (typically tens to a few hundred ms) for the
+	/// best congestion-vs-quality trade-off.
+	pub max_latency: Option<Duration>,
```
kixelated: `output_latency` maybe? I kind of like the `_max` to be clear that we won't delay unless necessary though. `latency_min` would delay.

rs/moq-audio/src/codec.rs
```diff
@@ -0,0 +1,431 @@
...
+}
+
+fn opus_error(code: i32, context: &str) -> AudioError {
+	AudioError::Unsupported(format!("libopus {context} failed (code {code})"))
+}
+
+/// Pick a libopus-supported rate close to `input_rate`.
+pub fn pick_opus_rate(input_rate: u32) -> u32 {
+	const SUPPORTED: [u32; 5] = [8_000, 12_000, 16_000, 24_000, 48_000];
+	if SUPPORTED.contains(&input_rate) {
```
kixelated: Do we need this check?

rs/moq-audio/src/codec.rs
```diff
@@ -0,0 +1,431 @@
...
+			)
+		};
+		if n < 0 {
+			return Err(opus_error(n, "opus_encode_float"));
+		}
+		Ok(Bytes::copy_from_slice(&self.scratch[..n as usize]))
+	}
+
+	/// hang catalog entry describing this encoder's output stream.
+	pub fn catalog_config(&self) -> hang::catalog::AudioConfig {
```
kixelated: nit: `catalog`

CLAUDE.md
```diff
@@ -49,6 +50,8 @@ Key architectural rule: The CDN/relay does not know anything about media. Anythi
   moq-cli/           # CLI tool for media operations (binary: moq)
   moq-clock/         # Clock synchronization example (binary: moq-clock)
   moq-mux/           # Media muxers/demuxers (fMP4, CMAF, HLS)
+  moq-audio/         # Native PCM ↔ Opus encode/decode on top of moq-mux
+  moq-video/         # Counterpart to moq-audio for video (stub today)
```
kixelated: Don't bother documenting the stub.

rs/libmoq/src/audio.rs
kixelated: Done in 3f6b06f3 — added `Codec::from_str` / `Display` / `as_str` on the moq-audio enum so the codec string is the single source of truth. libmoq now does `codec_str.parse::<moq_audio::Codec>()` instead of a hand-rolled match.

rs/moq-audio/src/codec.rs
kixelated: Done in 3f6b06f3 — split into `EncoderInput { format, sample_rate, channels }` and `EncoderOutput { codec, sample_rate: Option, channels: Option, bitrate, frame_duration }`. The optional output rate/channels lets callers explicitly encode at a different rate than the input (e.g. 48 kHz mic → 24 kHz Opus). Mirrored across moq-ffi (`MoqAudioEncoderInput`/`MoqAudioEncoderOutput`), libmoq (`moq_audio_encoder_input`/`moq_audio_encoder_output`), and the Python wrapper.

rs/moq-audio/src/codec.rs
kixelated: Done in 3f6b06f3 — renamed `DecoderConfig` → `DecoderOutput` and dropped the `output_` prefix on every field (the struct name already says it). Same treatment in moq-ffi / libmoq / py wrapper.

rs/moq-audio/src/codec.rs
kixelated: Kept `max_latency` (now on `DecoderOutput`) — the `max_` prefix makes it clear we never *add* latency, just skip when newer data is already this far ahead. Added a doc paragraph that spells that out so the name does its job.

rs/moq-audio/src/codec.rs
kixelated: Right, dropped it in 3f6b06f3. The `find(|&r| r >= input_rate)` falls through to the exact match anyway, so the `SUPPORTED.contains()` short-circuit was redundant.

rs/moq-audio/src/codec.rs
kixelated: Done in 3f6b06f3 — renamed `Encoder::catalog_config()` → `Encoder::catalog()`.

CLAUDE.md
kixelated: Done in 3f6b06f3 — dropped the moq-video line from the project structure tree.

rs/moq-audio/src/codec.rs
kixelated: Done — renamed to `latency_max` (and `latency_max_ms` in moq-ffi / libmoq) so it pairs cleanly with a future `latency_min` for the jitter-buffer floor. Only the new audio API is renamed; the existing `subscribe_media(max_latency_ms)` and `moq_consume_audio_ordered` stay as-is.

PR #1495, opened 2026-05-24
moq-mux: replace anyhow with thiserror

kixelated:
## Summary
- moq-mux is a library, so per the workspace convention (thiserror for libs, anyhow for binaries) it shouldn't be exporting anyhow types in its public API.
- Each codec, container, and catalog format gets its own `Error` enum, flattened into `crate::Error` via `#[from]`: `codec::{annexb,aac,opus,h264,h265,av1}::Error`, `container::{fmp4,mkv,hls}::Error`, `catalog::msf::Error`.
- Pure parsing functions (`Config::parse`, `Sps::parse`, `Avcc::parse`, `Avc1`/`Hvc1::transform`, etc.) return their module-local `Result`. Importer/exporter struct methods return `crate::Result` because they interleave codec errors with `moq_net::Error` and the other leaf errors that flow through track operations.

## Downstream impact
- `libmoq::Error::{InitFailed,DecodeFailed}` now wrap `Arc<moq_mux::Error>` instead of `Arc<anyhow::Error>`. Added a `BufferNotConsumed` variant for the one ad-hoc `anyhow!()` site that didn't fit either bucket.
- `moq-cli`'s `PublishDecoder` keeps `anyhow::Result` and bridges `moq_mux::Result` with explicit `?`.

## Test plan
- [x] `just check` (cargo check + clippy -D warnings + fmt --check + doc + shear + sort) passes
- [x] `cargo test -p moq-mux` — 141 tests pass
- [x] `cargo check --workspace --all-targets --exclude moq-gst` builds (moq-gst is blocked by a local gstreamer install only, not by these changes)

(Written by Claude)


kixelated:
Addressed the out-of-bounds-samples finding in 4737faca: `decode()` now returns `Error::SampleRangeOutOfBounds` instead of silently truncating.

rs/moq-mux/src/codec/h264/import.rs
```diff
@@ -452,20 +453,20 @@ impl Import {
...
 	}
 
 	/// Close the current group and open the next one at `sequence`.
-	pub fn seek(&mut self, sequence: u64) -> anyhow::Result<()> {
-		let track = self.track.as_mut().context("not initialized")?;
+	pub fn seek(&mut self, sequence: u64) -> Result<()> {
+		let track = self.track.as_mut().ok_or(Error::NotInitialized)?;
 		track.seek(sequence)?;
 		Ok(())
 	}
```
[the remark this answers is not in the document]
kixelated: Fixed in 6540b493: h264/h265/av1 `seek` now resets the in-flight `current` access unit (and h264's avc3 `State`) so pre-seek NALs/OBUs don't leak into the post-seek group.

rs/libmoq/src/error.rs
```diff
@@ -85,11 +85,15 @@ pub enum Error {
 
 	/// Media decoder initialization failed.
 	#[error("init failed: {0}")]
-	InitFailed(Arc<anyhow::Error>),
+	InitFailed(Arc<moq_mux::Error>),
```
kixelated: Could we make the moq_mux::Error Clone so we don't need to wrap it in an Arc?

And maybe combine InitFailed and DecodeFailed into a single Mux(#[from] moq_mux::Error)? Don't worry about backwards compatibility for error codes.

kixelated: Done in 4196dd59: every moq-mux Error now derives Clone (non-Clone leaves like `std::io::Error`, `mp4_atom::Error`, `reqwest::Error`, and `TagWriterError` are wrapped in `Arc`), and libmoq drops `InitFailed`/`DecodeFailed` in favor of `Mux(#[from] moq_mux::Error)` with plain `?` at the call sites.

PR #1502, opened 2026-05-25
ci(swift): decouple release manifest from dev Package.swift and gate publish on SPM resolve

kixelated:
## Summary

The `moq-dev/moq-swift` mirror at `0.2.13` is unusable from SPM: `xcodebuild -resolvePackageDependencies` against it fails with

> local binary target 'MoqFFIBinary' at '/Users/.../checkouts/moq-swift/MoqFFI.xcframework' does not contain a binary artifact.

Root cause: the mirror shipped the **dev-mode** `Package.swift` (path-based `binaryTarget`) instead of the **release-mode** one (URL + SHA-256). Trace:

- [`swift/scripts/check.sh`](https://github.com/moq-dev/moq/blob/main/swift/scripts/check.sh) rewrote the in-repo `Package.swift` to a path-based form for local testing. That rewrite eventually got committed to `main`.
- [`swift/scripts/package.sh`](https://github.com/moq-dev/moq/blob/main/swift/scripts/package.sh) then tried to `sed` `REPLACE_VERSION` / `REPLACE_CHECKSUM` and the upstream URL into that same file. None of those tokens existed in the dev-mode file, so every substitution was a silent no-op and the dev-mode manifest passed straight through into the release tarball.
- `publish.sh` rsynced the tarball into the mirror clone and pushed the tag — shipping the broken `Package.swift`.

The xcframework zip itself was uploaded correctly to the [moq-ffi-v0.2.13 release](https://github.com/moq-dev/moq/releases/tag/moq-ffi-v0.2.13); only the mirror's manifest was wrong.

## What changed

Fix (2) from the report — separate the source-of-truth from the local working file — plus a CI gate that catches this class of regression before it ships.

### Manifest split

- **New** [`swift/Package.swift.template`](swift/Package.swift.template) — the released manifest, with `REPLACE_VERSION` / `REPLACE_CHECKSUM` placeholders and a URL-based `binaryTarget`. `package.sh` now reads this.
- [`swift/Package.swift`](swift/Package.swift) is now explicitly the local-dev manifest (path-based `binaryTarget`); the header documents its scope and points at the template.
- [`swift/scripts/check.sh`](swift/scripts/check.sh) no longer overwrites `Package.swift`. The committed dev-mode file already matches the local xcframework layout `check.sh` sets up, so the rewrite was both unnecessary and the source of the bug.
- [`swift/scripts/package.sh`](swift/scripts/package.sh) reads `Package.swift.template`, then runs two fail-fast guards on the generated file:
  1. `grep` for any surviving `REPLACE_*` tokens
  2. `swift package dump-package` to confirm the manifest parses

  Both of these run on PR dry-runs too, so template syntax regressions fail the PR before any tag fires.

### CI gate

- **New** [`swift/scripts/verify.sh`](swift/scripts/verify.sh) builds a throwaway SPM consumer pointed at the staged package via `.package(path:)`, then runs `swift package resolve` (fetches `MoqFFI.xcframework.zip` and verifies its SHA-256 against the manifest checksum) and `swift build` (links the host slice).
- **New** `verify` job in [`.github/workflows/release-swift.yml`](.github/workflows/release-swift.yml) runs on `macos-latest` after the `release` job uploads the asset. `publish` now depends on `verify`, so a manifest SPM cannot resolve never reaches the mirror.

## Test plan

- [x] `bash -n` and `shellcheck` clean on all four scripts (only a pre-existing unused-`SCRIPT_DIR` warning in `publish.sh`).
- [x] `just swift check` passes locally — Rust build + uniffi bindgen + `swift test` (2 smoke tests, 0 failures). Confirms removing the heredoc rewrite did not regress local dev.
- [x] Simulated the full `package.sh` substitution on `Package.swift.template` with a stub version + checksum. Output has no surviving `REPLACE_*` tokens and `swift package dump-package` accepts it.
- [ ] Cut a fresh `moq-ffi-v*` patch tag after merge. The new `verify` job is the canary — if anything is still wrong, `publish` stays blocked and the mirror is unaffected.
- [ ] Smoke-test the resulting tag from `~/work/moq-voice/swift`: bump `from:` in `project.yml`, `xcodegen generate`, `xcodebuild -scheme MoqVoiceMac -destination 'platform=macOS' build`.

(Written by Claude)


kixelated:
Addressed in d418538: replaced the literal URL match with a `REPLACE_URL` token, and extended the grep guard to catch surviving `REPLACE_URL` too. Now `--release-url` overrides can never silently no-op.

(Written by Claude)

PR #1504, opened 2026-05-25
moq-relay: restore gossip-style cluster discovery via --cluster-mesh

### LINKED ISSUE #1499: old leaf discovery strategy gone?
Hi,

my previous pulled version had cluster leaves use cluster.node="<leaf-url>" to inform the cluster root, who would then inform all existing leaves of the new leaf, creating the full connection mapping.

I merged latest main (bit over a week ago), but the node variable is removed, and it seems i need to manually add the full list of relays in each cluster.connect in each relay (seems inefficient, especially compared to the previous discovery method).

Is there a better leaf discovery process (not manual), that I missed in the new version?

kixelated:
## Summary

Fixes #1499. Restores gossip-style peer discovery for `moq-relay` clusters by adding `--cluster-mesh <self-url>`, while keeping the new hop-routed origin and loop detection from #1322.

- New: `--cluster-mesh <self-url>` publishes a placeholder broadcast at `.internal/origins/<url>` so peers reached via `--cluster-connect` discover and dial this relay automatically. Pairs with `--cluster-connect` for the bootstrap; multiple nodes form a full mesh after a single rendezvous. A relay with `--cluster-mesh` alone (no `--cluster-connect`) runs as a passive rendezvous without a QUIC client.
- New: `.block(prefix)` view on `OriginProducer` / `OriginConsumer` (symmetric with `scope` and `with_root`) that refuses publishes and hides announces under a prefix. Applied to non-mTLS sessions in `Cluster::publisher` / `Cluster::subscriber` so JWT and anonymous clients can never see or publish into `.internal/*`, no matter how broad their declared scope or what their `token.root` is. (Block is applied to `self.origin` before `with_root`/`scope` so a JWT with `token.root = \".internal\"` cannot sidestep it; `OriginProducer::consume()` propagates `blocked` to the derived consumer.)
- Removed flags `--cluster-root` and `--cluster-node` (the latter existed both pre-`91ea43c5` and briefly in earlier commits of this branch) are kept as hidden `Option<String>` fields solely so the relay can bail at startup with a migration message pointing at `--cluster-connect` and `--cluster-mesh`. Catches old configs without surfacing them in `--help`.
- Dial-task provenance: `active: HashMap<String, DialEntry { handle, is_static }>`. Static `--cluster-connect` peers are exempt from gossip-driven aborts so a peer restart doesn't kill the reconnect loop.
- Docs: rewrote `doc/bin/relay/cluster.md` to cover both modes, the `.internal/*` access rule, and a migration table for `root → connect` and `node → mesh`. Updated `doc/bin/relay/config.md`, `rs/moq-relay/README.md` to match.
- Demos: switched `demo/relay/leaf{0,1}.toml` to use `mesh =` instead of static `connect = [root, leaf0]`, so the existing cluster-loop-detection exercise still runs but via discovery.

## Why

Without gossip, every operator had to enumerate every peer URL in every relay's config, and adding a node required editing every existing node. The hop-based loop detection added in #1322 makes the old gossip trick safe to bring back without the special-case `cluster: bool` auth flag it used to need.

## Reviewer notes

- The `.internal/` prefix matches the existing dotted-infrastructure convention (`.stats/...` from #1322). Choosing it (versus `internal/origins` from the pre-rewrite code) gives a path that's also conventionally hidden from app-level subscribers.
- Loop detection: when a relay's own registration is gossiped back through a peer, `OriginProducer::publish_broadcast` already refuses it because the hop chain contains this relay's `OriginId`. No special case needed.
- Dial dedup: the gossip discovery loop and static `--cluster-connect` share a `HashMap<String, DialEntry>` so a peer reached via both paths only opens one session. `DialEntry.is_static` keeps gossip unannounces from tearing down static reconnect loops.
- mTLS-only access: `AuthToken::unrestricted()` (mTLS) sets `internal: true` and skips the `.block(\".internal\")` wrap; JWT sessions always get the block, even with `subscribe = [\"\"]`. JWT cluster peers can still relay user traffic but cannot participate in gossip discovery.
- Passive rendezvous: a relay with only `--cluster-mesh` set runs without a QUIC client, publishes its self-registration, and parks via `std::future::pending()` so the registration stays alive.

## Test plan

- [x] `cargo test -p moq-net` — 332 pass, including 5 tests for the `block()` view (announce filter, publish refusal, get_broadcast filter, stacking, with_root interaction).
- [x] `cargo test -p moq-relay` — 87 pass, including 10 cluster tests: `.internal/*` invisible to broad-scope JWT, invisible even when `token.root = \".internal\"`, visible to mTLS, `--cluster-root` errors with migration message (CLI + TOML), `--cluster-node` errors with migration message (CLI + TOML), passive rendezvous runs without a client and keeps its registration alive, gossip unannounce preserves static peers, `cluster.mesh` config round-trips.
- [x] `cargo clippy -p moq-net -p moq-relay --all-targets --all-features -- -D warnings` — clean.
- [x] `cargo fmt --all --check` and `bun biome check` — clean.
- [x] `cargo check --workspace --exclude moq-gst` — clean. (moq-gst fails on this machine due to missing `gstreamer-1.0` system lib, unrelated.)
- [ ] Manual e2e: bring up `demo/relay/root.toml` + leaf0 + leaf1, publish through leaf0, subscribe from leaf1; verify it flows without leaf1 having leaf0 in its static `connect` list. (Not run locally.)

### DIFF
--- a/demo/relay/leaf0.toml
+++ b/demo/relay/leaf0.toml
@@ -36,12 +36,12 @@ listen = "[::]:4444"
 
 [cluster]
 # Dial the root for the initial cluster session. The root then gossips our
-# `node` URL to other leaves via the .internal/origins/<node> mesh path so
+# `mesh` URL to other leaves via the .internal/origins/<url> mesh path so
 # they discover and dial us back. leaf1 reaches us this way, forming a loop
 # (leaf1 -> root, leaf1 -> leaf0, leaf0 -> root) that exercises cluster
 # loop detection.
 connect = ["localhost:4443"]
-node = "localhost:4444"
+mesh = "localhost:4444"
 
 [auth]
 # Allow JWT tokens that are signed by this root key.
--- a/demo/relay/leaf0.toml
+++ b/demo/relay/leaf0.toml
@@ -35,9 +35,13 @@ tls.root = ["ca.pem"]
 listen = "[::]:4444"
 
 [cluster]
-# Connect only to the root. leaf1 will dial us directly, forming a loop
-# (leaf1 -> root, leaf1 -> leaf0, leaf0 -> root) to exercise cluster loop detection.
+# Dial the root for the initial cluster session. The root then gossips our
+# `node` URL to other leaves via the .internal/origins/<node> mesh path so
+# they discover and dial us back. leaf1 reaches us this way, forming a loop
+# (leaf1 -> root, leaf1 -> leaf0, leaf0 -> root) that exercises cluster
+# loop detection.
 connect = ["localhost:4443"]
+node = "localhost:4444"
 
 [auth]
 # Allow JWT tokens that are signed by this root key.
--- a/demo/relay/leaf0.toml
+++ b/demo/relay/leaf0.toml
@@ -35,13 +35,9 @@ tls.root = ["ca.pem"]
 listen = "[::]:4444"
 
 [cluster]
-# Dial the root for the initial cluster session. The root then gossips our
-# `mesh` URL to other leaves via the .internal/origins/<url> mesh path so
-# they discover and dial us back. leaf1 reaches us this way, forming a loop
-# (leaf1 -> root, leaf1 -> leaf0, leaf0 -> root) that exercises cluster
-# loop detection.
+# Connect only to the root. leaf1 will dial us directly, forming a loop
+# (leaf1 -> root, leaf1 -> leaf0, leaf0 -> root) to exercise cluster loop detection.
 connect = ["localhost:4443"]
-mesh = "localhost:4444"
 
 [auth]
 # Allow JWT tokens that are signed by this root key.
--- a/demo/relay/leaf1.toml
+++ b/demo/relay/leaf1.toml
@@ -41,7 +41,7 @@ listen = "[::]:4445"
 # exercises cluster loop detection. The mesh edge is now discovered, not
 # hardcoded (ex. imagine us-east learning about us-west via us-central).
 connect = ["localhost:4443"]
-node = "localhost:4445"
+mesh = "localhost:4445"
 
 [auth]
 # Allow JWT tokens that are signed by this root key.
--- a/demo/relay/leaf1.toml
+++ b/demo/relay/leaf1.toml
@@ -35,9 +35,13 @@ tls.root = ["ca.pem"]
 listen = "[::]:4445"
 
 [cluster]
-# Connect to both the root and leaf0, forming a loop to exercise cluster loop detection.
-# ex. imagine us-east connects to us-central AND directly to us-west.
-connect = ["localhost:4443", "localhost:4444"]
+# Dial only the root explicitly; gossip via .internal/origins/* will discover
+# leaf0 and dial it automatically. Combined with leaf0's own outbound connect
+# to the root, this still forms the leaf0 <-> leaf1 <-> root loop that
+# exercises cluster loop detection. The mesh edge is now discovered, not
+# hardcoded (ex. imagine us-east learning about us-west via us-central).
+connect = ["localhost:4443"]
+node = "localhost:4445"
 
 [auth]
 # Allow JWT tokens that are signed by this root key.
--- a/demo/relay/leaf1.toml
+++ b/demo/relay/leaf1.toml
@@ -35,13 +35,9 @@ tls.root = ["ca.pem"]
 listen = "[::]:4445"
 
 [cluster]
-# Dial only the root explicitly; gossip via .internal/origins/* will discover
-# leaf0 and dial it automatically. Combined with leaf0's own outbound connect
-# to the root, this still forms the leaf0 <-> leaf1 <-> root loop that
-# exercises cluster loop detection. The mesh edge is now discovered, not
-# hardcoded (ex. imagine us-east learning about us-west via us-central).
-connect = ["localhost:4443"]
-mesh = "localhost:4445"
+# Connect to both the root and leaf0, forming a loop to exercise cluster loop detection.
+# ex. imagine us-east connects to us-central AND directly to us-west.
+connect = ["localhost:4443", "localhost:4444"]
 
 [auth]
 # Allow JWT tokens that are signed by this root key.
--- a/doc/bin/relay/cluster.md
+++ b/doc/bin/relay/cluster.md
@@ -5,96 +5,125 @@ description: Run multiple moq-relay instances across multiple hosts/regions
 
 # Clustering
 
-Multiple relay instances can cluster for geographic distribution and improved latency.
+Multiple relay instances can join a cluster for geographic distribution and improved latency. Every cluster peer publishes into the same logical origin; loop detection and shortest-path preference come from a hop list on each broadcast, so peers can be connected in arbitrary topologies without duplicating data.
 
-## Overview
+## Two ways to form a cluster
 
-`moq-relay` uses a simple clustering scheme:
+Pick the mode that matches your operational constraints. Both can be combined in a single deployment.
 
-1. **Root node** - A single relay (can serve public traffic) that tracks cluster membership
-2. **Other nodes** - Accept internet traffic and consult the root for routing
+### Static topology
 
-When a relay publishes a broadcast, it advertises its `node` address to other relays via the root.
-
-## Configuration
+Enumerate every peer by URL. Best for small clusters (2-5 nodes) where membership rarely changes.
 
 ```toml
 [cluster]
-root = "https://root-relay.example.com"  # Root node
-node = "https://us-east.relay.example.com"  # This node's address
+connect = ["us-east.example.com:4443", "eu-west.example.com:4443"]
 ```
 
-### Cluster Arguments
+Each entry in `connect` is dialed at startup and kept alive with exponential backoff. There is no discovery: every new node requires editing every existing config.
 
-- `--cluster-root <HOST>` - Hostname/IP of the root node (omit to make this node the root)
-- `--cluster-node <HOST>` - Hostname/IP of this instance (needs valid TLS cert)
+### Gossip discovery
 
-## How It Works
+Each relay sets `node` to its own externally-reachable URL. Connecting to a single peer is enough; that peer gossips the new node's address to everyone else.
 
-1. Each relay connects to the root node on startup
-2. When a publisher connects to any relay, that relay announces the broadcast
-3. The root node tracks which relay has which broadcasts
-4. When a subscriber connects, the relay queries the root to find the broadcast
-5. Relays connect to each other to forward traffic
+```toml
+# On the rendezvous (every other relay connects here)
+[cluster]
+node = "rendezvous.example.com:4443"
 
-## Benefits
-
-- **Lower latency** - Users connect to nearest relay
-- **Higher availability** - Redundancy across regions
-- **Geographic distribution** - Serve global audiences
-
-## Example Topology
-
-```text
-                    ┌─────────────┐
-                    │  Root Node  │
-                    │   (US-C)    │
-                    └──────┬──────┘
-           ┌───────────────┼───────────────┐
-           │               │               │
-    ┌──────┴──────┐ ┌──────┴──────┐ ┌──────┴──────┐
-    │   US-East   │ │   EU-West   │ │   Asia-SE   │
-    │   Relay     │ │   Relay     │ │   Relay     │
-    └─────────────┘ └─────────────┘ └─────────────┘
+# On a leaf joining the cluster
+[cluster]
+node = "us-east.example.com:4443"
+connect = ["rendezvous.example.com:4443"]
 ```
 
-## Peer Authentication
+When a leaf with `node` set connects to `rendezvous`, it publishes a registration broadcast at `.internal/origins/<node>` on the cluster origin. Other peers reachable from `rendezvous` see the registration and dial the new leaf, building a full mesh. Removing a node unannounces its registration, which aborts the dial on every other peer.
 
-Cluster peers must authenticate to each other. Two options:
+A relay with `node` set and no `connect` entries waits passively for inbound connections. A relay with `connect` and no `node` dials peers but isn't itself advertised, so others won't discover it via gossip.
 
-### JWT token
+## How gossip works
 
-Each leaf reads a JWT from `cluster.token` (see [Authentication](/bin/relay/auth))
-and presents it to the root on connect. The token must grant cluster privileges
-and full publish/subscribe access.
+1. On startup, a relay with `cluster.node = "<url>"` publishes a placeholder broadcast at `.internal/origins/<url>` on its own origin. The broadcast carries no tracks: the path is the registration.
+2. Cluster sessions exchange their origins both ways. The registration propagates to every connected peer, accumulating a hop chain along the way.
+3. Each peer watches `.internal/origins/*` for newly announced URLs and dials any it isn't already connected to. Dials are deduplicated by URL, so a peer reached via both `connect` and gossip uses a single session.
+4. When a peer goes away, its registration is unannounced and every other relay aborts the dial it spawned in response.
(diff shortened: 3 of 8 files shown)


doc/bin/relay/cluster.md
```diff
@@ -5,96 +5,125 @@ description: Run multiple moq-relay instances across multiple hosts/regions
...
+node = "us-east.example.com:4443"
+connect = ["rendezvous.example.com:4443"]
+token = "cluster.jwt"
+```
 
-Configure the root with `tls.root` pointing at the CA that signed the leaf
-certificates. Leaves connect with a client certificate signed by that CA —
-no JWT needed. The leaf's cluster node name is taken from the first DNS SAN on
-its certificate, so identity is bound to the cert rather than self-declared.
+JWT-authenticated cluster sessions are tagged as external for stats purposes (unlike mTLS) but otherwise behave the same. Note: by default a JWT session is also filtered out of `.internal/*`. A cluster JWT must include the cluster privileges that bypass that filter; if you find yourself reaching for this path, prefer mTLS.
```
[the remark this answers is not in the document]
kixelated: Fixed in ea8678f3 — rewrote the JWT subsection to drop the bypass claim. `.internal/*` is now documented as mTLS-only; JWT cluster peers can still relay user traffic but cannot participate in gossip discovery. (Written by Claude)

rs/moq-relay/src/cluster.rs
```diff
@@ -98,24 +125,45 @@ impl Cluster {
...
 	}
 
 	/// Returns an [`OriginProducer`] scoped to this session's publish permissions.
+	///
+	/// Non-internal tokens cannot publish into `.internal/*` regardless of their
+	/// declared scope.
 	pub fn publisher(&self, token: &AuthToken) -> Option<OriginProducer> {
-		self.origin.with_root(&token.root)?.scope(&token.publish)
+		let view = self.origin.with_root(&token.root)?.scope(&token.publish)?;
+		Some(if token.internal { view } else { view.block(".internal") })
```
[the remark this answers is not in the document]
kixelated: Fixed in ea8678f3 — `Cluster::access_origin(token)` now applies `.block(".internal")` to `self.origin` *before* `with_root`/`scope`. Also fixed `OriginProducer::consume()` to propagate `blocked` to the derived consumer (the original code blocked only the producer half, which the bug exposed). Regression test: `internal_paths_invisible_when_token_root_is_internal`. (Written by Claude)

rs/moq-relay/src/cluster.rs
```diff
@@ -133,22 +181,116 @@ impl Cluster {
...
+		// Track active dial tasks by URL so static and gossip-discovered peers don't
+		// duplicate, and so the discovery side can abort a task when a peer unannounces.
+		let active: Arc<Mutex<HashMap<String, AbortHandle>>> = Arc::new(Mutex::new(HashMap::new()));
+
 		let mut tasks = tokio::task::JoinSet::new();
+
+		// Seed static peers from --cluster-connect.
 		for peer in &self.config.connect {
+			Self::spawn_dial(&mut tasks, &active, self.clone(), peer.clone(), token.clone());
+		}
```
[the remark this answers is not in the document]
kixelated: Fixed in ea8678f3 — `active` is now `HashMap<String, DialEntry>` where `DialEntry { handle, is_static }` records provenance. `Cluster::handle_gossip_unannounce` only aborts non-static entries; static peers from `--cluster-connect` keep their reconnect loop running regardless of gossip churn. Regression test: `gossip_unannounce_preserves_static_peer`. (Written by Claude)

kixelated:
@coderabbitai, fixed in 0524a101 — split `has_work` into `has_outbound = !connect.is_empty()` and a node-only check. The `with_client` requirement and the discovery spawn are now both gated behind `has_outbound`. Also fixed a related bug where the `_self_registration` broadcast would be dropped immediately in passive-rendezvous mode (no tasks to wait on → `while tasks.join_next()` exited): `run()` now parks with `std::future::pending()` when no dial tasks are spawned, keeping the registration alive. Regression test: `passive_rendezvous_runs_without_client_and_advertises_self`. (Written by Claude)

demo/relay/leaf0.toml
```diff
@@ -35,9 +35,13 @@ tls.root = ["ca.pem"]
...
 [cluster]
-# Connect only to the root. leaf1 will dial us directly, forming a loop
-# (leaf1 -> root, leaf1 -> leaf0, leaf0 -> root) to exercise cluster loop detection.
+# Dial the root for the initial cluster session. The root then gossips our
+# `mesh` URL to other leaves via the .internal/origins/<url> mesh path so
+# they discover and dial us back. leaf1 reaches us this way, forming a loop
+# (leaf1 -> root, leaf1 -> leaf0, leaf0 -> root) that exercises cluster
+# loop detection.
 connect = ["localhost:4443"]
+mesh = "localhost:4444"
```
kixelated: Can you remove this from this config? I want to test proxying, not mesh.

demo/relay/leaf1.toml
```diff
@@ -35,9 +35,13 @@ tls.root = ["ca.pem"]
...
-# Connect to both the root and leaf0, forming a loop to exercise cluster loop detection.
-# ex. imagine us-east connects to us-central AND directly to us-west.
-connect = ["localhost:4443", "localhost:4444"]
+# Dial only the root explicitly; gossip via .internal/origins/* will discover
+# leaf0 and dial it automatically. Combined with leaf0's own outbound connect
+# to the root, this still forms the leaf0 <-> leaf1 <-> root loop that
+# exercises cluster loop detection. The mesh edge is now discovered, not
+# hardcoded (ex. imagine us-east learning about us-west via us-central).
+connect = ["localhost:4443"]
+mesh = "localhost:4445"
```
kixelated: Same

doc/bin/relay/cluster.md
```diff
@@ -5,96 +5,126 @@ description: Run multiple moq-relay instances across multiple hosts/regions
 
 # Clustering
 
-Multiple relay instances can cluster for geographic distribution and improved latency.
+Multiple relay instances can join a cluster for geographic distribution and improved latency. Every cluster peer publishes into the same logical origin; loop detection and shortest-path preference come from a hop list on each broadcast, so peers can be connected in arbitrary topologies without duplicating data.
```
kixelated: > Every cluster peer publishes into the same logical origin; loop detection and shortest-path preference come from a hop list on each broadcast, so peers can be connected in arbitrary topologies without duplicating data.

This is confusing.

Just say something like relays can be joined together to proxy announcements/subscriptions.

doc/bin/relay/cluster.md
```diff
@@ -5,96 +5,126 @@ description: Run multiple moq-relay instances across multiple hosts/regions
...
 
-`moq-relay` uses a simple clustering scheme:
+Pick the mode that matches your operational constraints. Both can be combined in a single deployment.
 
-1. **Root node** - A single relay (can serve public traffic) that tracks cluster membership
-2. **Other nodes** - Accept internet traffic and consult the root for routing
+### Static topology
 
-When a relay publishes a broadcast, it advertises its `node` address to other relays via the root.
+Enumerate every peer by URL. Best for small clusters (2-5 nodes) where membership rarely changes.
```
kixelated: Do a proper example with multiple hops. This lets you control the hops, unlike a mesh where there's N * N-1 connections.

ex. us-west connects to us-east, us-east connects to eu-west. If us-west wants to pull data from eu-west, it goes to us-east instead (cached/deduplicated).

doc/bin/relay/cluster.md
```diff
@@ -5,96 +5,126 @@ description: Run multiple moq-relay instances across multiple hosts/regions
...
+# On the rendezvous (every other relay connects here)
 [cluster]
-root = "https://root-relay.example.com"  # Root node
-node = "https://us-east.relay.example.com"  # This node's address
+mesh = "rendezvous.example.com:4443"
+
+# On a leaf joining the cluster
+[cluster]
+mesh = "us-east.example.com:4443"
+connect = ["rendezvous.example.com:4443"]
```
kixelated: Note that you still need nodes to connect (or accept) connections from at least one node to discover each other.

doc/bin/relay/cluster.md
```diff
@@ -5,96 +5,126 @@ description: Run multiple moq-relay instances across multiple hosts/regions
...
+### Gossip discovery
+
+Each relay sets `mesh` to its own externally-reachable URL. Connecting to a single peer is enough; that peer gossips the new relay's address to everyone else.
 
 ```toml
+# On the rendezvous (every other relay connects here)
 [cluster]
-root = "https://root-relay.example.com"  # Root node
-node = "https://us-east.relay.example.com"  # This node's address
+mesh = "rendezvous.example.com:4443"
```
kixelated: I wouldn't have this sepreate example. Just say connect is optiona.

doc/bin/relay/cluster.md
```diff
@@ -5,96 +5,126 @@ description: Run multiple moq-relay instances across multiple hosts/regions
...
 
-### Cluster Arguments
+When a leaf with `mesh` set connects to `rendezvous`, it publishes a registration broadcast at `.internal/origins/<url>` on the cluster origin. Other peers reachable from `rendezvous` see the registration and dial the new leaf, building a full mesh. Removing a node unannounces its registration, which aborts the dial on every other peer.
 
-- `--cluster-root <HOST>` - Hostname/IP of the root node (omit to make this node the root)
-- `--cluster-node <HOST>` - Hostname/IP of this instance (needs valid TLS cert)
+A relay with `mesh` set and no `connect` entries waits passively for inbound connections (it does not need a QUIC client). A relay with `connect` and no `mesh` dials peers but isn't itself advertised, so others won't discover it via gossip.
 
-## How It Works
+## How gossip works
```
kixelated: Too detailed I think.

Just say each node creates a broadcast with its address, which other nodes will discover.

rs/moq-net/src/model/origin.rs
```diff
@@ -643,6 +643,10 @@ pub struct OriginProducer {
 
 	// The prefix that is automatically stripped from all paths.
 	root: PathOwned,
+
+	// Absolute path prefixes for which publishes are silently refused.
+	// Populated by [`OriginProducer::block`]. Empty by default.
+	blocked: Vec<PathOwned>,
```
kixelated: Actually I don't think we need this.

Our relay should make tokens that scope access instead. If customers are allowed to read/write to *... something is probably wrong.

kixelated (commented):
Remove the .block functionality, don't make .internal special. relay operators should figure it out.

doc/bin/relay/cluster.md
```diff
@@ -5,96 +5,126 @@ description: Run multiple moq-relay instances across multiple hosts/regions
...
+mesh = "rendezvous.example.com:4443"
+
+# On a leaf joining the cluster
+[cluster]
+mesh = "us-east.example.com:4443"
+connect = ["rendezvous.example.com:4443"]
 ```
 
-### Cluster Arguments
+When a leaf with `mesh` set connects to `rendezvous`, it publishes a registration broadcast at `.internal/origins/<url>` on the cluster origin. Other peers reachable from `rendezvous` see the registration and dial the new leaf, building a full mesh. Removing a node unannounces its registration, which aborts the dial on every other peer.
```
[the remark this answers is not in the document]
kixelated: Fixed in c81ce3c1 — reworded the unannounce sentence to specify it aborts dials spawned in response to the gossip, and added a clause noting that static `connect` peers are exempt so a remote restart cannot break a hand-configured edge. (Written by Claude)

demo/relay/leaf0.toml
kixelated: Removed in 1094b83f. Both leaves now use the original `connect = ["localhost:4443"]` (leaf1 also keeps `connect = ["localhost:4443", "localhost:4444"]` for the loop-detection triangle), no `mesh =` lines. (Written by Claude)

demo/relay/leaf1.toml
kixelated: Removed in 1094b83f along with the leaf0 mesh line. (Written by Claude)

doc/bin/relay/cluster.md
kixelated: Rewrote in 1094b83f. The doc now opens with: "Relays can be joined together to proxy announcements and subscriptions between each other." The "same logical origin" wording is gone. (Written by Claude)

doc/bin/relay/cluster.md
kixelated: Done in 1094b83f. Primary example is now `eu-west <- us-east <- us-west` with a paragraph explaining that `us-east` caches and dedups Atlantic fetches across many `us-west` viewers, which a full mesh would skip. (Written by Claude)

doc/bin/relay/cluster.md
kixelated: Added in 1094b83f. The auto-discovery section now states: "`connect` is optional once gossip is running, but you still need at least one connection somewhere (either you dial a peer or a peer dials you) for the advertisement to flow." (Written by Claude)

doc/bin/relay/cluster.md
kixelated: Done in 1094b83f. The separate rendezvous-vs-leaf example is gone; the auto-discovery section is now one TOML block plus a sentence noting that `connect` is optional and that `mesh` alone makes the relay a passive rendezvous. (Written by Claude)

doc/bin/relay/cluster.md
kixelated: Trimmed in 1094b83f. The "How gossip works" numbered list is gone; replaced with one sentence: "Each node with `mesh` set creates a broadcast carrying its address, which other nodes pick up." (Written by Claude)

rs/moq-net/src/model/origin.rs
kixelated: Reverted in 1094b83f. Dropped `OriginProducer::block` / `OriginConsumer::block`, the `blocked` field, `is_blocked` helpers, `new_with_blocked`, and the five block tests. `Cluster::access_origin` is gone too; `publisher`/`subscriber` collapse back to the simple `with_root(token.root).scope(token.subscribe)` form. Net diff was -463/+70. (Written by Claude)

rs/moq-relay/src/cluster.rs
```diff
@@ -133,22 +195,146 @@ impl Cluster {
...
+				tasks.spawn(async move {
+					this.run_discovery(self_url, token, active).await;
+				});
+			}
+		}
+
+		if tasks.is_empty() {
+			// Passive rendezvous: park to keep `_self_registration` alive. The
+			// process still exits via the other arms of `tokio::select!` in main.
+			std::future::pending::<()>().await
```
kixelated, line 332: You should add this task if self.config.mesh.is_some(). Probably replace .map with if let Some above.

rs/moq-relay/src/cluster.rs
```diff
@@ -133,22 +195,146 @@ impl Cluster {
...
+			return;
+		};
+
+		while let Some((relative, announced)) = consumer.announced().await {
+			let peer = relative.as_str();
+			if peer == self_url {
+				continue;
+			}
+
+			match announced {
```
kixelated: if let Some() .. else is a bit cleaner.

rs/moq-relay/src/cluster.rs
kixelated: Done in afb75426. Merged the `_self_registration` block and the discovery spawn into one `if let Some(mesh) = self.config.mesh.as_deref() { ... } else { None }`. Discovery still requires `has_outbound` inside that block (passive rendezvous has nothing to discover), but the outer mesh-is-set check happens once. (Written by Claude)

rs/moq-relay/src/cluster.rs
kixelated: Done in afb75426. Rewrote as `if announced.is_some() { ... } else { Self::handle_gossip_unannounce(...) }` since the BroadcastConsumer in the Some arm was never bound. (Written by Claude)

PR #1509, opened 2026-05-26
moq-native(jemalloc): drop runtime activation; fixes moq-boy startup crash

### LINKED ISSUE #1507: moq_native::jemalloc: profiling activation errors should warn, not be fatal
### Summary

[rs/moq-native/src/jemalloc.rs](rs/moq-native/src/jemalloc.rs) propagates errors via `?` from `run()`, so any failure to activate jemalloc heap profiling (or set up the SIGUSR1 listener) becomes fatal to the caller. Heap profiling is a debug aid; it should never take down a process that would otherwise be healthy.

### Current behavior

[rs/moq-native/src/jemalloc.rs:9](rs/moq-native/src/jemalloc.rs#L9):

```rust
pub async fn run() -> anyhow::Result<()> {
    let prof_active = b"prof.active\0";

    match unsafe { raw::read::<bool>(prof_active) } {
        Ok(true) => tracing::info!("jemalloc heap profiling is active"),
        Ok(false) => {
            tracing::info!("jemalloc profiling compiled in; activating");
            unsafe { raw::write(prof_active, true) }
                .map_err(|err| anyhow::anyhow!("failed to activate jemalloc profiling: {err}"))?;
        }
        Err(err) => {
            tracing::debug!(%err, "jemalloc profiling not available — set MALLOC_CONF=prof:true to enable");
            return Ok(());
        }
    }

    let mut sig = tokio::signal::unix::signal(tokio::signal::unix::SignalKind::user_defined1())?;
    ...
}
```

Two failure paths are fatal:

1. `raw::write(prof_active, true)` failing when `prof.active` is readable but unwritable.
2. `tokio::signal::unix::signal(...)` failing (e.g. SIGUSR1 already claimed).

When a binary `tokio::spawn`s this task with `?` or `.unwrap()` in a `try_join!`, the binary exits.

### Expected behavior

`MALLOC_CONF` profiling is a debug aid, not load-bearing. Failures should `tracing::warn!` and return `Ok(())` so the rest of the process keeps running. Either:

- Change `run()` to never return `Err` (log + return on failure), or
- Keep the signature but document that callers should ignore the result.

### Impact

Production binaries built with the jemalloc feature can crash on startup or first profile dump if the runtime jemalloc environment isn't exactly what the code expects, even when no heap dumps were actually requested.

(Written by Claude)

kixelated:
## Summary

Root cause of moq-boy crashing on first startup with `jemalloc profiler failed: name or mib specifies an unknown/invalid value`:

[rs/moq-native/src/jemalloc.rs](rs/moq-native/src/jemalloc.rs) had a branch that tried to flip on profiling at runtime:

```rust
Ok(false) => {
    unsafe { raw::write(prof_active, true) }
        .map_err(|err| anyhow::anyhow!("failed to activate jemalloc profiling: {err}"))?;
}
```

But jemalloc only initializes the profiling backend if `opt.prof=true` at init. Writing `prof.active` later returns EINVAL. The branch was dead code in the success case and a crash in the failure case.

moq-relay's nix module sets `MALLOC_CONF=prof:true,prof_active:true` at the systemd unit, so it hits the `Ok(true)` branch and never noticed. moq-boy compiles in the same jemalloc feature but is launched without any MALLOC_CONF wrapper, takes the broken `Ok(false)` path, and crashes. systemd restarts it, restart works fine if the env got applied elsewhere, masking the bug.

This drops the activation attempt entirely. If `prof.active` reads false, log a hint to set `MALLOC_CONF=prof:true,prof_active:true` at startup and return Ok(()).

Closes #1507.

## Test plan

- [x] `cargo check --features jemalloc` on `rs/moq-native` builds cleanly
- [ ] Manual: launch moq-boy without `MALLOC_CONF` (previously crashed; now logs a hint and continues without profiling)
- [ ] Manual: launch moq-relay via the nix unit (already worked, should still work — `Ok(true)` branch unchanged)
- [ ] Manual: launch a binary with `MALLOC_CONF=prof:true,prof_active:true,prof_prefix=/tmp/heap` and send SIGUSR1 — should still dump

## Follow-up (separate PR)

moq-boy probably wants a launcher / nix module mirroring the relay's `MALLOC_CONF` env so heap profiling is opt-in via the same surface as the relay, instead of relying on the caller to set it manually.

(Written by Claude)

### DIFF
--- a/rs/moq-native/src/jemalloc.rs
+++ b/rs/moq-native/src/jemalloc.rs
@@ -5,7 +5,8 @@ pub use tikv_jemallocator;
 /// Activate jemalloc heap profiling and listen for SIGUSR1 to dump profiles.
 ///
 /// The dump path is controlled by `MALLOC_CONF=prof_prefix:<path>`.
-/// Returns `Ok(())` if profiling is not available (i.e. MALLOC_CONF=prof:true was not set).
+/// Profiling is a debug aid, so any setup failure is logged and swallowed
+/// rather than propagated. Returns when profiling can't be set up.
 pub async fn run() -> anyhow::Result<()> {
 	let prof_active = b"prof.active\0";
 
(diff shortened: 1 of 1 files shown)


kixelated:
@coderabbitai the docstring nit is moot — commit 3b8c2697 (after the review snapshot at 124b97f3) rewrote the docstring to describe the new contract ("Profiling must be enabled at startup via MALLOC_CONF=prof:true ... jemalloc only initializes the profiling backend when opt.prof is set at init; toggling prof.active later returns EINVAL"), since I also dropped the runtime-activation branch entirely as the real root cause fix. (Written by Claude)

PR #1514, opened 2026-05-27
moq-net: linger upstream subscriptions across consumer churn (moq-lite)

kixelated:
## Summary

When the last viewer of a track drops, the upstream subscription used to tear down immediately. A returning viewer milliseconds later triggered a fresh `Subscribe`, and the publisher re-served the latest cached group. Under churn (page reloads, reconnects, toggles) the same group got re-fetched over and over.

This PR keeps the upstream `TrackProducer` alive for up to 5s after the last consumer leaves. On `track.unused()` the lite subscriber sends `SubscribeUpdate(priority = 0)` + FIN upstream and waits for one of:

- **Complete** — the upstream FINs back (subscription is done)
- **Cancelled** — the 5s linger timeout expires (no one came back)
- **Reused** — a returning consumer arrives during the wait; reissue with `start_group = latest + 1` so the publisher doesn't re-send already-cached groups

Late duplicates from the prior subscription round get silently dropped in `recv_group` (the data is already cached locally, so STOP_SENDING is the right move).

Lite01/02 fall back to the old immediate-cancel since they lack `SubscribeUpdate`. IETF transport is unchanged for now.

## Why the broadcast cleanup change

The auto-cleanup spawn in `BroadcastConsumer::subscribe_track` previously fired on `weak.unused()`, which raced the linger logic and yanked the lookup entry the instant consumer count hit 0 — defeating the point of linger. Re-anchored it to `weak.closed()` instead: the producer owner is now responsible for closing the producer (via `abort`/`finish`/drop) once truly done, and that close fires the cleanup eagerly. No in-memory leak of dead broadcasts.

The `requested_unused` test was rewritten to match the new contract: while the producer is alive, re-subscribing to the same name returns the existing producer's consumer (no new request); a fresh request only appears after the producer itself is closed.

## Files

- `rs/conducer/src/weak.rs` — add `Weak::closed()` (mirror of `Producer::closed`)
- `rs/moq-net/src/model/track.rs` — `TrackProducer::latest()`, `TrackWeak::closed()`
- `rs/moq-net/src/model/broadcast.rs` — cleanup spawn waits on `weak.closed()`; test rewrite
- `rs/moq-net/src/lite/subscriber.rs` — restructure `run_subscribe` into a linger loop driving a new `run_subscribe_session`; silently drop `Error::Duplicate` in `recv_group`
- `rs/moq-native/tests/broadcast.rs` — `linger_resubscribe_keeps_flowing_moq_lite_03` smoke test

## Test plan

- [x] `just check`
- [x] `cargo test --workspace --exclude moq-gst --exclude moq-ffi --exclude libmoq` — all green
- [x] Existing `model::broadcast::test::requested_unused` updated and passes against new contract
- [x] New `linger_resubscribe_keeps_flowing_moq_lite_03` end-to-end test passes

## Known caveats

Distinguishing the **Reused** vs **Complete** linger branches from a black-box client test over localhost turned out to be too timing-dependent (the publisher's FIN-back roundtrip races the test thread's resubscribe call). The included test verifies the end-to-end smoke — drop + resub keeps working through the linger window. A more targeted branch-specific test would need either a slow-publisher harness or wire-level interception. (Written by Claude)

### DIFF
--- a/rs/moq-net/src/lite/publisher.rs
+++ b/rs/moq-net/src/lite/publisher.rs
@@ -378,14 +378,10 @@ impl<S: web_transport_trait::Session> Publisher<S> {
 
 		// Track-level subscriber priority. SUBSCRIBE_UPDATE messages broadcast new values
 		// to both run_track (so future groups inherit the new priority) and serve_group
-		// tasks (so in-flight groups update via PriorityHandle::set_track).
+		// tasks (so in-flight groups update via PriorityHandle::set_track). The Sender
+		// stays in run_subscribe and gets handed to run_track so the same loop that
+		// parses SUBSCRIBE_UPDATEs also fans the new priority out.
 		let (track_priority_tx, track_priority_rx) = tokio::sync::watch::channel(track.priority);
-		// `end_group` is a serving cap, not a subscription terminator: groups with
-		// sequence > cap are held until the subscriber raises the cap (or unsets
-		// it) via SUBSCRIBE_UPDATE, then served in order. Only FIN ends the
-		// subscription. This is what relays use to pause an upstream subscription
-		// across consumer churn without tearing it down.
-		let (end_group_tx, end_group_rx) = tokio::sync::watch::channel(subscribe.end_group);
 
 		let sub = Subscription {
 			session,
@@ -397,25 +393,22 @@ impl<S: web_transport_trait::Session> Publisher<S> {
 			version,
 		};
 
-		tokio::select! {
-			res = sub.run_track(track, subscribe.start_group, end_group_rx) => res?,
-			res = Self::run_subscribe_updates(&mut stream.reader, &track_priority_tx, &end_group_tx) => res?,
-		}
+		// `end_group` is a serving cap, not a subscription terminator: groups with
+		// sequence > cap are held in the producer's cache until the subscriber raises
+		// the cap (or unsets it) via SUBSCRIBE_UPDATE, then served in order. Only a
+		// peer FIN actually ends the subscription. This is what lets relays pause an
+		// upstream subscription across consumer churn without tearing it down.
+		sub.run_track(
+			track,
+			subscribe.start_group,
+			subscribe.end_group,
+			&mut stream.reader,
+			&track_priority_tx,
+		)
+		.await?;
 
 		stream.writer.finish()?;
 		stream.writer.closed().await
-	}
-
-	async fn run_subscribe_updates<R: web_transport_trait::RecvStream>(
-		reader: &mut crate::coding::Reader<R, Version>,
-		priority_tx: &tokio::sync::watch::Sender<u8>,
-		end_group_tx: &tokio::sync::watch::Sender<Option<u64>>,
-	) -> Result<(), Error> {
-		while let Some(upd) = reader.decode_maybe::<lite::SubscribeUpdate>().await? {
-			let _ = priority_tx.send(upd.priority);
-			let _ = end_group_tx.send(upd.end_group);
-		}
-		Ok(())
 	}
 }
 
@@ -439,7 +432,9 @@ impl<S: web_transport_trait::Session> Subscription<S> {
 		mut self,
 		mut track: TrackConsumer,
 		start_group: Option<u64>,
-		mut end_group: tokio::sync::watch::Receiver<Option<u64>>,
+		initial_end_group: Option<u64>,
+		reader: &mut crate::coding::Reader<S::RecvStream, Version>,
+		track_priority_tx: &tokio::sync::watch::Sender<u8>,
 	) -> Result<(), Error> {
 		let mut tasks: FuturesUnordered<futures::future::BoxFuture<'static, ()>> = FuturesUnordered::new();
 
@@ -449,8 +444,8 @@ impl<S: web_transport_trait::Session> Subscription<S> {
 		}
 
 		// Apply the initial cap from the original Subscribe. Subsequent updates
-		// arrive via the end_group watch channel and get re-applied below.
-		track.end_at(*end_group.borrow_and_update());
+		// flow through the SubscribeUpdate select arm below.
+		track.end_at(initial_end_group);
 
 		loop {
 			tokio::select! {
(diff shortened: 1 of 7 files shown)


rs/moq-net/src/lite/subscriber.rs
```diff
@@ -318,99 +339,212 @@ impl<S: web_transport_trait::Session> Subscriber<S> {
...
+			};
+
+			match outcome {
+				SessionOutcome::Reused => {
+					// A new consumer arrived during linger. Drop the lingering subscribe
+					// id and restart with a fresh one, resuming after the latest group we
+					// already cached so the publisher doesn't re-send anything we have.
+					self.subscribes.lock().remove(&id);
+					id = self.next_id.fetch_add(1, atomic::Ordering::Relaxed);
+					start_group = track.latest().and_then(|s| s.checked_add(1));
```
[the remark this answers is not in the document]
kixelated: Good catch — keeping the old id mapped. Switched  to track every allocated id in a Vec and remove them all together at task exit, so late uni streams from a prior subscription instance still resolve to the same TrackEntry and hit the Duplicate-drop path. Fixed in a09e1e8c.

rs/moq-net/src/model/broadcast.rs
```diff
@@ -578,54 +582,48 @@ mod test {
...
+		// the existing producer (no new request) — this is what lets the relay
+		// linger upstream subscriptions across transient consumer churn.
+		let consumer3 = broadcast.consume().assert_subscribe_track(&Track::new("unknown_track"));
+		consumer3.assert_is_clone(&producer1.consume());
+		broadcast.assert_no_request();
+		drop(consumer3);
+
+		// Aborting the producer triggers cleanup; the next subscribe creates a fresh request.
+		producer1.abort(Error::Cancel).unwrap();
 		tokio::time::sleep(std::time::Duration::from_millis(1)).await;
```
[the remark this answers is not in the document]
kixelated: Done — moved the test to `#[tokio::test(start_paused = true)]` and replaced the sleep with `tokio::time::advance`. The cleanup task is not actually time-driven (it waits on `weak.closed()`), but pausing time + advancing yields the runtime in the same shape without depending on wall-clock scheduling. Fixed in a09e1e8c.

rs/moq-net/src/lite/publisher.rs
```diff
@@ -395,63 +400,113 @@ impl<S: web_transport_trait::Session> Publisher<S> {
...
+							}
+						}
+						None => track_done = true,
+					}
+				}
+
+				Ok(()) = end_group.changed() => {
+					// Loop back to drain any newly-permitted held groups.
+				}
+			}
```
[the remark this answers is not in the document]
kixelated: Good catch. Replaced the unbounded `Vec<GroupConsumer>` with an `Option<GroupConsumer>` and gated `track.recv_group()` on `pending.is_none()`, so we pull at most one group beyond the cap. Anything the publisher produces while pending is set sits in the track's own cache (bounded by its 5s eviction policy) and lands on the wire normally once the cap rises. Fixed in 7c99e409.

rs/moq-net/src/lite/publisher.rs
```diff
@@ -381,11 +380,26 @@ impl<S: web_transport_trait::Session> Publisher<S> {
 		// to both run_track (so future groups inherit the new priority) and serve_group
 		// tasks (so in-flight groups update via PriorityHandle::set_track).
 		let (track_priority_tx, track_priority_rx) = tokio::sync::watch::channel(track.priority);
-		let track_stats = std::sync::Arc::new(track_stats);
+		// `end_group` is a serving cap, not a subscription terminator: groups with
+		// sequence > cap are held until the subscriber raises the cap (or unsets
+		// it) via SUBSCRIBE_UPDATE, then served in order. Only FIN ends the
+		// subscription. This is what relays use to pause an upstream subscription
+		// across consumer churn without tearing it down.
+		let (end_group_tx, end_group_rx) = tokio::sync::watch::channel(subscribe.end_group);
```
kixelated: I think we can call `track.end_at` instead?

The TrackConsumer itself can store the largest group returned. If `track.end_at(None)` then it can go back and return anything in cache greater than the largest group returned? Or something.

Also same with `start_at`? We can store the minimum group served and allow expanding the range. 

Both of these will need unit tests because it's kind of annoying to get right, especially if start_at/end_at gets toggled.

rs/moq-net/src/lite/subscriber.rs
```diff
@@ -352,65 +377,162 @@ impl<S: web_transport_trait::Session> Subscriber<S> {
...
-		stream.reader.closed().await?;
+		// Lifecycle loop: serve → linger → resume → serve → ... → FIN.
+		loop {
+			// Phase 1 — serving. Wait for either:
+			// - the last consumer to drop (enter linger), or
+			// - the upstream to close the stream (subscription is over).
+			tokio::select! {
+				_ = track.unused() => {}
+				res = stream.reader.closed() => {
+					return match res {
```
kixelated: if let Err(err) { } else { }

I think?

rs/moq-net/src/lite/publisher.rs
kixelated: Done — moved end_at onto TrackConsumer (`end_sequence` field + `end_at(Option<u64>)` setter). poll_next_group now goes through a sequence-based scan (`State::poll_next_in_range`) so the cap can be raised / lowered / unset and previously-blocked cached groups become eligible automatically. next_group is now sequence-ordered (smallest-first) rather than arrival-ordered with a monotonic drop filter, which is strictly better at not losing data on out-of-order arrivals. Publisher.rs is back to a single track.next_group() loop with track.end_at() applied on each end_group.changed() — no more local pending buffer. Added four unit tests covering cap/raise/lower/toggle-with-out-of-order. Skipped the dynamic start_at expansion for now since we have no consumer of it yet; happy to add when one shows up. Fixed in 88c297b1.

rs/moq-net/src/lite/subscriber.rs
kixelated: Done — swapped both `match res` blocks in the linger select arms over to `if let Err(err) = res { return ... }` style. Fixed in 88c297b1.

rs/moq-net/src/model/track.rs
```diff
@@ -596,6 +659,16 @@ impl TrackConsumer {
 		self.min_sequence = sequence;
 	}
 
+	/// Cap the consumer at the specified sequence (inclusive), or remove the cap entirely.
+	///
+	/// Affects [`Self::next_group`] only: groups beyond the cap stay in the producer's
+	/// cache rather than being skipped past, so a later call to [`Self::end_at`] with a
+	/// higher value (or `None`) makes them available again. Lowering the cap below the
+	/// consumer's current cursor parks the consumer until the cap is raised.
+	pub fn end_at(&mut self, sequence: Option<u64>) {
```
kixelated: `impl Into<Option<u64>>`

kixelated: Or we could use u64::MAX I guess.

kixelated: Done — `end_at` now takes `impl Into<Option<u64>>` so callers can pass a bare `u64`, `Some(u64)`, or `None`. Test sites updated to use the bare form where it reads cleaner. Fixed in 57d180bf.

PR #1517, opened 2026-05-27
moq-net(stats): aggregate per-node into a single gzipped broadcast

kixelated:
## Summary

- Collapses the per-(level × node) stats fan-out into a single `.stats/node/{name}` broadcast per relay (or `.stats/node` when unset). The four `tier × role` tracks remain, renamed to `.json.gz`.
- Each frame is a gzipped JSON map of broadcast path to cumulative counter snapshot for every broadcast active in the last `retention_ticks` ticks. Tier / role / node are implied by the track and broadcast paths, not repeated in the payload.
- `--stats-levels` is gone; level aggregation moves to a separate downstream binary. New flags: `--stats-tick-secs` (default `1`) and `--stats-retention-ticks` (default `10`), both `Option<T>` per the TOML-merge rule.

## Why

Per-(level × node) broadcasts grow as `(N + K) × M` (broadcasts × levels × relays), and a browser dashboard subscribing to all of them hit the 100-subscription ceiling almost immediately. Per-frame payload now grows linearly in active-broadcast count instead, and gzip handles the repetitive JSON shape so the wire cost stays bounded.

## Wire format

```jsonc
// .stats/node/sjc/1, track publisher.json.gz, gunzipped
{
  \"demo/bbb\": { \"broadcasts\": 1, \"broadcasts_closed\": 0, \"subscriptions\": 5,
                 \"subscriptions_closed\": 2, \"bytes\": 12345, \"frames\": 678, \"groups\": 9 },
  \"anon/foo\": { \"broadcasts\": 1, \"broadcasts_closed\": 0, \"subscriptions\": 2,
                 \"subscriptions_closed\": 0, \"bytes\": 234,   \"frames\": 12,  \"groups\": 1 }
}
```

A broadcast appears in the frame for a given (tier, role) while it has at least one active subscription, and lingers for `retention_ticks` ticks after the last one drops. Counters are cumulative; downstream computes rates.

## Implementation notes

- `BroadcastEntry` holds four `Counters` and four `last_active_tick` atomics (one per `tier × role`). The single snapshot task ticks every `tick_secs`, builds per-slot frames, writes them gzipped, and GCs entries that have no outstanding RAII guards (`Arc::strong_count == 1`) and are past the retention window. The strong-count check is the safety hatch: a bump on an orphaned `Arc` after GC can't be silently lost because we never GC while a guard is still alive.
- The task lazily spawns on the first `broadcast()` call and exits after `2 × retention_ticks` of an empty entry map (next bump respawns it). One broadcast and four tracks per node lifetime.
- Idle-frame skipping is preserved by byte-comparing the per-slot JSON payload to the last emitted one.

## Reviewer notes

- `Stats::new` signature changed (`levels: u32` → `tick: Duration, retention_ticks: u32`). Only moq-relay constructs `Stats` so the rest of moq-net's `BroadcastStats` / guard surface is unchanged.
- `js/net` doesn't consume the stats wire format today, so no JS sync is needed per the Cross-Package Sync table; `doc/bin/relay/config.md` is updated with the new schema, flags, env vars.

## Test plan

- [x] `cargo test --workspace --all-targets` (sans `moq-gst`, which needs system libs locally) — 11 new stats tests cover per-broadcast isolation, tier independence, single-broadcast announce, gzip round-trip, retention-keeps and retention-evicts, prefix/disabled no-ops, plus existing 78 moq-relay tests including the updated TOML/CLI regression test for `tick_secs` / `retention_ticks`.
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo fmt --all --check`
- [x] `RUSTDOCFLAGS=-D warnings cargo doc --no-deps --workspace`
- [x] `cargo shear` / `cargo sort --workspace --check`
- [ ] End-to-end against a local relay: run `moq-relay --stats-enabled --stats-node sjc/1`, publish a couple broadcasts, subscribe to `.stats/node/sjc/1/publisher.json.gz` through `gunzip`, drop a publisher and confirm the entry lingers ~10 ticks then disappears.

(Written by Claude)

### DIFF
--- a/doc/bin/relay/config.md
+++ b/doc/bin/relay/config.md
@@ -186,7 +186,7 @@ Each stats broadcast carries four tracks, one per `(tier, role)` pair:
 
 Each frame is a JSON object mapping broadcast path to a cumulative
 counter snapshot. An entry surfaces whenever its snapshot differs from
-the last one emitted, and lingers for `retention_ticks` ticks past the
+the last one emitted, and lingers for `retention` intervals past the
 most recent change:
 
 ```json
--- a/doc/bin/relay/config.md
+++ b/doc/bin/relay/config.md
@@ -160,13 +160,13 @@ enabled = true
 # Top-level path under which stats broadcasts are published (defaults to ".stats")
 prefix = ".stats"
 
-# Tick interval in seconds between snapshots (defaults to 1)
-tick_secs = 1
+# Seconds between snapshot publishes (defaults to 1)
+interval = 1
 
-# Number of ticks an entry lingers in the emitted frame after its last
-# observed change (defaults to 1). A short reconnect window keeps the
-# entry visible across brief disconnects.
-retention_ticks = 1
+# Number of intervals an entry lingers in the emitted frame after its
+# last observed change (defaults to 1). A short reconnect window keeps
+# the entry visible across brief disconnects.
+retention = 1
 
 # Node identifier appended to the advertised path to disambiguate broadcasts
 # when multiple relays share a cluster origin. May be multi-segment, e.g.
@@ -240,9 +240,9 @@ byte-identical to the last emitted frame; new subscribers still pick up
 a baseline immediately via track-latest semantics.
 
 Every flag also accepts an equivalent CLI argument (`--stats-enabled`,
-`--stats-prefix`, `--stats-tick-secs`, `--stats-retention-ticks`,
+`--stats-prefix`, `--stats-interval`, `--stats-retention`,
 `--stats-node`) and environment variable (`MOQ_STATS_ENABLED`,
-`MOQ_STATS_PREFIX`, `MOQ_STATS_TICK_SECS`, `MOQ_STATS_RETENTION_TICKS`,
+`MOQ_STATS_PREFIX`, `MOQ_STATS_INTERVAL`, `MOQ_STATS_RETENTION`,
 `MOQ_STATS_NODE`).
 
 ### \[iroh]
--- a/doc/bin/relay/config.md
+++ b/doc/bin/relay/config.md
@@ -163,10 +163,10 @@ prefix = ".stats"
 # Tick interval in seconds between snapshots (defaults to 1)
 tick_secs = 1
 
-# Number of ticks an idle broadcast lingers in the emitted frame after its
-# last observed active subscription (defaults to 10). A short reconnect
-# window keeps the entry visible across brief disconnects.
-retention_ticks = 10
+# Number of ticks an entry lingers in the emitted frame after its last
+# observed change (defaults to 1). A short reconnect window keeps the
+# entry visible across brief disconnects.
+retention_ticks = 1
 
 # Node identifier appended to the advertised path to disambiguate broadcasts
 # when multiple relays share a cluster origin. May be multi-segment, e.g.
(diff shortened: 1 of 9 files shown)


rs/moq-net/src/stats.rs
```diff
@@ -577,141 +545,185 @@ impl SubscriberTrack {
...
+				.zip(frames.iter_mut())
+			{
+				let snap = counters.snapshot();
+				if counters.active() {
+					last_tick.store(current_tick, Ordering::Relaxed);
+				}
+				let last = last_tick.load(Ordering::Relaxed);
+				if last != 0 && current_tick.saturating_sub(last) < retention_ticks as u64 {
+					frame.insert(entry.path.as_str().to_string(), snap);
+				}
```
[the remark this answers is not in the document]
kixelated: Fixed in 71e38cb1 — switched the retention check to `<=` and added a separate active-branch emit so `retention_ticks=0` still surfaces live subs and lingers zero idle ticks. The GC condition uses the same comparison for consistency. Added `retention_boundary_keeps_last_idle_tick` to lock the boundary semantics in. Thanks for catching this. (Written by Claude)

rs/moq-net/src/stats.rs
```diff
@@ -577,141 +545,185 @@ impl SubscriberTrack {
...
+				Err(err) => {
+					tracing::debug!(?err, slot, "stats: failed to gzip frame");
+					continue;
+				}
+			};
+			if let Err(err) = track.write_frame(compressed) {
+				tracing::debug!(?err, slot, "stats: failed to write frame");
+			}
+			*last = json;
+		}
```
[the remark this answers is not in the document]
kixelated: Fixed in 71e38cb1 — `continue` on write_frame failure so `*last` only advances after a successful write. Failed snapshots are now retried on the next tick. (Written by Claude)

doc/bin/relay/config.md
```diff
@@ -146,6 +146,105 @@ tls.disable_verify = true
...
+# Top-level path under which stats broadcasts are published (defaults to ".stats")
+prefix = ".stats"
+
+# Tick interval in seconds between snapshots (defaults to 1)
+tick_secs = 1
+
+# Number of ticks an entry lingers in the emitted frame after its last
+# observed change (defaults to 1). A short reconnect window keeps the
+# entry visible across brief disconnects.
+retention_ticks = 1
```
kixelated: IMO just call it `retention`

doc/bin/relay/config.md
```diff
@@ -146,6 +146,105 @@ tls.disable_verify = true
...
+```toml
+[stats]
+# Master switch (defaults to false)
+enabled = true
+
+# Top-level path under which stats broadcasts are published (defaults to ".stats")
+prefix = ".stats"
+
+# Tick interval in seconds between snapshots (defaults to 1)
+tick_secs = 1
```
kixelated: IMO call it `interval`

rs/moq-net/src/stats.rs
```diff
@@ -85,56 +124,76 @@ pub struct Counters {
...
 
-impl Role {
-	fn as_str(&self) -> &'static str {
-		match self {
-			Role::Publisher => "publisher",
-			Role::Subscriber => "subscriber",
-		}
+const NUM_SLOTS: usize = 4;
+
+fn slot_index(tier: Tier, role: Role) -> usize {
```
kixelated: It feels gross doing this. The Role is always hard-coded, so we should definitely have a separate `publisher` and `subscriber` field. I'm fine having slots for internal/external.

doc/bin/relay/config.md
kixelated: Done in 555df3fa — renamed to `interval`, units in the field doc-comment ("seconds between snapshot publishes"). (Written by Claude)

doc/bin/relay/config.md
kixelated: Done in 555df3fa — renamed to `retention`, unit ("intervals") in the field doc-comment. (Written by Claude)

rs/moq-net/src/stats.rs
kixelated: Done in 555df3fa — `Role` enum is gone; `BroadcastEntry` now has named `publisher` and `subscriber` fields (each `[Counters; 2]` indexed by Tier). `EntrySnapState` mirrors that shape for the snapshot task. (Written by Claude)

PR #1530, opened 2026-05-28
REANNOUNCE via duplicate ANNOUNCE; announced() returns (path, Announced)

kixelated:
Implements REANNOUNCE (per [moq-dev/drafts#23](https://github.com/moq-dev/drafts/pull/23)). Rebased onto current `dev`.

REANNOUNCE is an atomic broadcast replacement: the broadcast at a path is swapped for a new one without an interruption in availability (relay failover, upstream restart, a shorter hop path arriving).

## Wire design: duplicate ANNOUNCE = REANNOUNCE

Rather than an explicit status byte, a reannounce **is** a duplicate ANNOUNCE — an active announcement for a path that is already announced, with no intervening unannounce. This means:

- **No wire-format change** (`lite/announce.rs` is untouched), **no new status**, and **no version gating** — every version that speaks ANNOUNCE gets it.
- **No JS change**: a duplicate ANNOUNCE is idempotent for `@moq/net`'s set-based announce consumers (reconnect republish, `@moq/boy`), so they keep working unchanged.

## API change (Rust)

`AnnounceConsumer::next()` / `try_next()` / `poll_next()` now yield `(PathOwned, Announced)` instead of `(PathOwned, Option<BroadcastConsumer>)`:

```rust
pub enum Announced {
    Active(BroadcastConsumer),
    Reannounce(BroadcastConsumer),
    Ended,
}
```

`.broadcast()` gives the legacy `Option<BroadcastConsumer>` view (Some for both Active and Reannounce, None for Ended), so callers that don't care about the distinction stay simple.

## Behavior

- The origin already coalesced atomic replacements (backup promotion, shorter hop path) into an internal reannounce; those now surface as a single `Announced::Reannounce` delivery instead of an Ended-then-Active pair. A genuine unannounce+announce is still preserved as two deliveries.
- **Lite publisher:** a `Reannounce` is sent as a duplicate `Announce::Active` (keeping the stats guard for continuity).
- **Lite subscriber:** an `Active` for an already-announced path is applied as an atomic replacement (publish the replacement so the origin demotes the old broadcast to a backup, then retire it) instead of returning a `Duplicate` error.
- **IETF moq-transport** has no reannounce, so it splits a `Reannounce` into `namespace_done` + `namespace`.

## Drive-by

Adds the missing `compress` field (`..Default::default()`) to libmoq's `Track` initializers, which fails to compile after dev's compression feature landed (this was the `error[E0063]: missing field compress` CI failure).

## Tests

- `model::origin`: reannounce after the consumer drained is delivered as a single `Reannounce`; an undrained reannounce stays `Active`; `test_duplicate` backup-promotion asserts a single reannounce.
- moq-net 327 unit tests + moq-relay 93 lib tests pass; `moq-net`/`moq-relay`/`moq-native`/`moq-ffi`/`libmoq`/`hang` build clean with no clippy `-D warnings`; `cargo fmt --all --check` clean (verified on rust 1.96, the nix-pinned toolchain).

(Written by Claude)

### DIFF
--- a/rs/hang/examples/subscribe.rs
+++ b/rs/hang/examples/subscribe.rs
@@ -2,6 +2,8 @@
 
 use std::time::Duration;
 
+use anyhow::Context;
+
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
 	// Optional: Use moq_native to configure a logger.
@@ -40,13 +42,11 @@ async fn run_session(origin: moq_net::OriginProducer) -> anyhow::Result<()> {
 // Subscribe to a broadcast and read media frames.
 async fn run_subscribe(consumer: moq_net::OriginConsumer) -> anyhow::Result<()> {
 	// Wait for a broadcast to be announced.
-	let (path, broadcast) = consumer
-		.announced()
-		.next()
-		.await
-		.ok_or_else(|| anyhow::anyhow!("origin closed"))?;
+	let (path, broadcast) = consumer.announced().next().await.context("origin closed")?;
 
-	let broadcast = broadcast.ok_or_else(|| anyhow::anyhow!("broadcast unannounced: {path}"))?;
+	let broadcast = broadcast
+		.broadcast()
+		.with_context(|| format!("broadcast unannounced: {path}"))?;
 
 	tracing::info!(%path, "broadcast announced");
 
--- a/rs/moq-ffi/src/origin.rs
+++ b/rs/moq-ffi/src/origin.rs
@@ -28,14 +28,16 @@ impl Announced {
 	async fn next(&mut self) -> Result<Option<Arc<MoqAnnouncement>>, MoqError> {
 		loop {
 			match self.inner.next().await {
-				Some((path, Some(broadcast))) => {
+				// Active and Restart both carry a broadcast; skip unannounce events.
+				Some((path, event)) => {
+					let Some(broadcast) = event.broadcast() else {
+						continue;
+					};
 					return Ok(Some(Arc::new(MoqAnnouncement {
 						path: path.to_string(),
 						broadcast: Arc::new(MoqBroadcastConsumer::new(broadcast)),
 					})));
 				}
-				// TODO moq-lite will change to not emit None (unannounce) events here.
-				Some((_path, None)) => continue,
 				None => return Ok(None),
 			}
 		}
@@ -44,11 +46,11 @@ impl Announced {
 	async fn available(&mut self) -> Result<Arc<MoqBroadcastConsumer>, MoqError> {
 		loop {
 			match self.inner.next().await {
-				Some((_path, Some(broadcast))) => {
-					return Ok(Arc::new(MoqBroadcastConsumer::new(broadcast)));
-				}
-				// TODO moq-lite will change to not emit None (unannounce) events here.
-				Some((_path, None)) => continue,
+				// Active and Restart both carry a broadcast; skip unannounce events.
+				Some((_path, event)) => match event.broadcast() {
+					Some(broadcast) => return Ok(Arc::new(MoqBroadcastConsumer::new(broadcast))),
+					None => continue,
+				},
 				None => return Err(MoqError::Closed),
 			}
 		}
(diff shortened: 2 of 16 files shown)


js/net/src/lite/announce.ts
```diff
@@ -9,22 +9,52 @@ import { Version } from "./version.ts";
...
 		this.hops = props.hops ?? [];
 		if (this.hops.length > MAX_HOPS) {
 			throw new Error(`hop count ${this.hops.length} exceeds maximum ${MAX_HOPS}`);
 		}
 	}
 
 	async #encode(w: Writer, version: Version) {
-		await w.bool(this.active);
+		if (this.status === AnnounceStatus.Reannounce && !reannounceSupported(version)) {
+			throw new Error("reannounce not supported for this version");
```
kixelated: If not supported, do an unnannounce then a regular announce.

js/net/src/announced.ts
```diff
@@ -8,7 +8,15 @@ import * as Path from "./path.js";
  */
 export interface AnnouncedEntry {
 	path: Path.Valid;
+	/** True if a broadcast is available (announce or reannounce); false if it ended. */
 	active: boolean;
```
kixelated, line 11: Shouldn't this be AnnounceStatus now?

rs/moq-ffi/src/origin.rs
```diff
@@ -28,14 +28,16 @@ impl Announced {
 	async fn next(&mut self) -> Result<Option<Arc<MoqAnnouncement>>, MoqError> {
 		loop {
 			match self.inner.next().await {
-				Some((path, Some(broadcast))) => {
+				// Active and Reannounce both carry a broadcast; skip unannounce events.
+				Some((path, event)) => {
+					let Some(broadcast) = event.into_broadcast() else {
```
kixelated: `.into_broadcast()` is pretty awkard. Just call it `broadcast()`?

rs/hang/examples/subscribe.rs
```diff
@@ -46,7 +46,9 @@ async fn run_subscribe(consumer: moq_net::OriginConsumer) -> anyhow::Result<()>
 		.await
 		.ok_or_else(|| anyhow::anyhow!("origin closed"))?;
 
-	let broadcast = broadcast.ok_or_else(|| anyhow::anyhow!("broadcast unannounced: {path}"))?;
+	let broadcast = broadcast
+		.broadcast()
+		.ok_or_else(|| anyhow::anyhow!("broadcast unannounced: {path}"))?;
```
kixelated: I'm not a fan of .ok_or_else(|| anyhow::anyhow!

Can you use .context instead for this example?

rs/moq-native/examples/clock.rs
```diff
@@ -101,13 +101,13 @@ async fn main() -> anyhow::Result<()> {
 
 			loop {
 				tokio::select! {
-					Some(announce) = origin.next() => match announce {
-						(path, Some(broadcast)) => {
+					Some((path, event)) = origin.next() => match event {
+						Announced::Active(broadcast) | Announced::Reannounce(broadcast) => {
```
kixelated: matching on `event.broadcast()` might be better?

rs/moq-net/src/ietf/publisher.rs
```diff
@@ -448,79 +448,94 @@ impl<S: web_transport_trait::Session> Publisher<S> {
...
-								.encode_message(&ietf::PublishNamespaceDone {
-									track_namespace: suffix.as_path(),
-									request_id,
-								})
-								.await;
-						}
-						_ => {}
-					}
-					stream.writer.finish().ok();
+				crate::Announced::Reannounce(_) => {
```
kixelated: Maybe rename this to Announced::Restart? Or ::Retry?

kixelated: Renamed to `Announced::Restart` (and the internal restart-coalescing path) across moq-net + consumers. Went with `Restart` over `Retry` since `restart` is the status name in the draft (and "retry" reads like a failed attempt being re-sent, whereas this is a clean replacement). Happy to switch to `Retry` if you'd prefer. (Written by Claude)

---

PR #1531, opened 2026-05-29
lite-05: negotiate per-frame compression via SUBSCRIBE_OK (Rust + JS)

kixelated:
## Summary

Opt-in frame compression for the `moq-lite-05` (WIP) draft, implemented on **both the Rust (`moq-net`) and browser (`@moq/net`) sides**. The publisher marks a track as worth compressing, the codec is negotiated in `SUBSCRIBE_OK`, and the subscriber blocks on that message before decoding any frame.

- **`Track` gains a `compress` flag** (Rust `compress: bool` / `Track::with_compress(..)`; JS `track.compress`) — a hint that the track's frames are worth compressing. The hang JSON catalog (`Catalog::default_track`) is marked compressible, since it's text and re-sent on every change.
- **`SUBSCRIBE_OK` carries a negotiated `Compression` codec**, gated to lite-05+. The publisher picks `Deflate` when the track is marked `compress` and the peer speaks lite-05; older drafts (lite-04 and below) always negotiate `None` and stream frames verbatim, so the wire stays backwards compatible. Negotiation lives in `SUBSCRIBE_OK` (not `SUBSCRIBE`) specifically so the subscriber has to block on it to learn how to decode.
- **Subscribers block on `SUBSCRIBE_OK` before decoding frames.** A group's QUIC stream can be accepted before `SUBSCRIBE_OK` lands on the subscribe stream, so the negotiated codec is fanned out to each group stream (Rust: a per-subscribe `watch` channel; JS: a per-subscribe signal) and the group read path waits on it before touching any frame payload.
- **Frames are compressed independently** so the codec never carries state across the lossy, out-of-order group boundary. Decompression caps the inflated size at 16 MiB (`MAX_FRAME_SIZE`) to reject zip bombs.
- **Codec = raw DEFLATE**, no zlib/gzip header (QUIC already guarantees integrity). Rust uses `flate2` (pure-Rust `miniz_oxide` backend, no C deps — keeps `moq-ffi` cross-compilation and Nix builds painless); the browser uses the native `CompressionStream`/`DecompressionStream` with `"deflate-raw"`, which produces the same bytes. The `Compression` enum is the extension point if we want zstd later.

## Wire format (lite-05 only)

`SUBSCRIBE_OK` appends a single `Compression (i)` varint after the existing fields (`0` = none, `1` = deflate). On a compressed track, each `FRAME`'s size prefix becomes the *compressed* length; the subscriber inflates it back, so the public `Frame`/`Group`/`Track` API is unchanged on both sides.

## Scope / known limitations (follow-ups)

- **Hop-by-hop, not end-to-end.** A relay learns the codec from its upstream `SUBSCRIBE_OK`, decompresses on ingest, and caches the plaintext. Its downstream producer track isn't marked `compress`, so it re-serves uncompressed. Compression therefore applies on a hop where the publisher's track carries the flag (origin → direct subscriber). Propagating the flag through relays is a follow-up.
- **JS publisher pull-model timing.** `@moq/net` creates the served track lazily when a subscribe arrives, so an app must set `track.compress` before the track is served for the publisher to advertise it. The browser's primary role here is the *subscriber* (decoding a Rust-published compressed catalog), which is fully wired.
- **lite-05 is unadvertised/WIP**, so this is opt-in only (`Lite05Wip` / `DRAFT_05_WIP`); nothing negotiates it by default.

## Cross-package sync

| Table row | Action |
|---|---|
| `rs/moq-net` wire/API → `js/net` | **Done** — `@moq/net` implements the same `Compression` codec, `SUBSCRIBE_OK` field, blocking, and per-frame (de)compression. |
| `rs/moq-net` wire/API → `doc/concept` | Deferred: the wire spec belongs in `moq-dev/drafts`; `doc/concept` is a stable-feature overview and shouldn't document an unadvertised draft yet. |
| `rs/hang` catalog → `js/hang` | No catalog *format* change; `compress` is a transport-layer track hint, not part of the serialized catalog JSON (`skip_serializing_if`). |

## Test plan

- [x] Rust: `Compression` codec round-trips (none + deflate), rejects garbage, caps inflated size, wire-code round-trip; `SUBSCRIBE_OK` wire round-trip (compression on lite-05, absent on lite-04). `cargo test -p moq-net --lib` (327) and `-p hang` pass; `cargo check --all-targets --workspace`, `clippy -D warnings`, `cargo doc -D warnings` clean (minus ffmpeg/gst crates needing system toolchains locally).
- [x] JS: codec round-trip (incl. max-size + garbage rejection) and `SUBSCRIBE_OK` wire format across versions; `bun test` (148 net tests) and `bun run --filter='*' check` + Biome clean.
- [ ] End-to-end session test exercising the SUBSCRIBE_OK-blocking + compress/decompress path across a live transport — needs an in-memory transport harness `moq-net` doesn't have yet; flagging as a follow-up.

(Written by Claude)


rs/moq-net/src/lite/publisher.rs
```diff
@@ -543,45 +562,87 @@ impl<S: web_transport_trait::Session> Subscription<S> {
...
-								result = stream.write_all(&mut chunk) => {
-									result?;
-									break;
-								}
-								new_pri = priority.next() => {
-									stream.set_priority(new_pri);
-								}
-								Ok(()) = self.track_priority.changed() => {
-									priority.set_track(*self.track_priority.borrow_and_update());
+			match self.compression {
```
kixelated: Any way to split this into separate functions? It's very nested.

PR #1532, opened 2026-05-29
Add duration-based skipping to the container jitter buffer

kixelated:
## Summary

The container consumer's jitter buffer advances by group sequence and skips stalled or missing groups once the configured latency budget is exceeded. This adds a third skip trigger driven by frame duration.

CMAF carries a per-sample duration. A stalled group whose most recent frame *ends* (`timestamp + duration`) at or past the next group's first timestamp has nothing left worth waiting for, so we skip it immediately, regardless of the latency budget. Containers that don't carry a duration (Legacy, LOC) leave it `None`, which disables the new check and falls back to the existing latency behavior.

## Changes

- **`Frame` gains an optional `duration`** (`Option<Timestamp>`, in the frame's own scale). The CMAF decoder fills it from the `trun` sample-duration (falling back through `trun → tfhd`), and `encode_fragment` writes it back, so **fMP4 → fMP4 round-trips it**. Other producers construct frames with `None`.
- **Producer backfill (no added latency):** when a keyframe rolls a group over, `Producer` now passes that keyframe's timestamp into the flush, so the previous group's *last* frame can borrow it as a duration boundary. That recovers the one sample whose successor wasn't visible when it arrived. The keyframe is already in hand, so nothing extra is buffered.
- **Consumer** tracks each group's furthest presentation point in wall-clock terms (so timestamp/duration scales needn't match) and skips the active group when a newer group's start is covered. The skip only fires once the active group is fully consumed, so frames the viewer hasn't seen are never dropped.
- **js/hang** is mirrored on the decode + consumer side.

## Why `dev`

Adds a public field to `Frame` (rs/moq-mux) and to `@moq/hang`'s `Frame`. Per branch-targeting, container/API changes target `dev`. Branch is rebased directly on top of `dev`.

## Cross-package sync

| Row | Status |
|---|---|
| `rs/moq-net` wire/API | n/a |
| `rs/hang` ↔ `js/hang` container | done (decode + consumer mirrored) |
| `doc/concept` | no change needed: the docs describe the jitter buffer only at a high level and stay accurate. |

Most of the file count is a one-line `duration: None` per `Frame` constructor, since adding a field makes Rust's exhaustive struct literals require it.

## Test plan

- [x] `cargo test -p moq-mux` (159 passing), incl. `consumer::tests::{duration_skip_advances_to_next_group, duration_below_gap_does_not_skip}`, `fmp4::tests::{decode_reads_trun_sample_duration, duration_round_trips_through_encode, decode_without_duration_reports_none}`, and `producer::tests::keyframe_backfills_last_frame_duration`.
- [x] `cargo clippy -p moq-mux --all-targets` + `cargo fmt` clean; dependent crates build (`moq-audio`, `hang` examples, `libmoq`, `moq-ffi`).
- [x] `bun test` in `js/hang` (26 passing); `tsc --noEmit` clean.

(Written by Claude)

### DIFF
--- a/rs/moq-mux/src/container/producer.rs
+++ b/rs/moq-mux/src/container/producer.rs
@@ -75,9 +75,10 @@ impl<C: Container> Producer<C> {
 	/// A keyframe closes any open group and starts a new one. A non-keyframe extends
 	/// the current group; if no group is open, returns a protocol violation.
 	pub fn write(&mut self, frame: Frame) -> Result<(), C::Error> {
-		// Close the current group on an explicit keyframe.
+		// Close the current group on an explicit keyframe, passing its timestamp so
+		// the previous group's last frame can borrow it as a duration boundary.
 		if frame.keyframe {
-			self.finish_group()?;
+			self.finish_group_at(Some(frame.timestamp))?;
 		}
 
 		// Start a new group if needed; the first frame of a group must be a keyframe.
@@ -108,7 +109,7 @@ impl<C: Container> Producer<C> {
 				let first = iter.next().unwrap();
 				let (min, max) = iter.fold((first, first), |(min, max), d| (min.min(d), max.max(d)));
 				if max.saturating_sub(min) >= self.latency {
-					self.flush()?;
+					self.flush(None)?;
 				}
 			}
 		}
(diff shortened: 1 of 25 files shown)


rs/moq-mux/src/container/producer.rs
```diff
@@ -120,7 +121,14 @@ impl<C: Container> Producer<C> {
...
 	/// The next [`write`](Self::write) must be a keyframe.
 	pub fn finish_group(&mut self) -> Result<(), C::Error> {
-		self.flush()?;
+		self.finish_group_at(None)
+	}
+
+	/// Like [`finish_group`](Self::finish_group), but uses `next` (the timestamp of the
+	/// keyframe that rolled the group over) as the duration boundary for the group's
+	/// last frame. See [`flush`](Self::flush).
+	fn finish_group_at(&mut self, next: impl Into<Option<moq_net::Timestamp>>) -> Result<(), C::Error> {
```
kixelated: Actually no need for Into here, since it's private

kixelated: Done — reverted `finish_group_at` to a plain `Option<moq_net::Timestamp>` and pass `Some(...)`/`None` at the two call sites. (The CLAUDE.md note I added now scopes the `impl Into<Option<T>>` convention to public APIs only.) (Written by Claude)

---

PR #1533, opened 2026-05-29
Add libmoq catalog producer + raw moq-net track API

kixelated:
## Summary

Adds two related, additive pieces to libmoq's C API, both mirroring functionality that already exists in `moq-ffi` but was missing from the C bindings. Together they let a C caller describe a broadcast's catalog **and** write/read arbitrary track contents themselves, the producer-side opposite of the existing consume API.

### 1. Catalog producer API

The producer counterpart to `moq_consume_video_config` / `moq_consume_audio_config`. Lets callers author catalog renditions directly instead of relying on `moq_publish_media_ordered` to derive them.

- `moq_publish_video_config(broadcast, const moq_video_config*)` / `moq_publish_audio_config(broadcast, const moq_audio_config*)` — insert or replace a rendition, keyed by `name`.
- `moq_publish_video_remove` / `moq_publish_audio_remove` — remove by name.

These reuse the existing `moq_video_config` / `moq_audio_config` structs as **inputs** (the same structs the consume side fills as outputs), so producer and consumer are mirror images. Each call republishes the catalog automatically via the existing moq-mux catalog `Producer` guard. `description` / `coded_width` / `coded_height` pointers may be `NULL` to omit those fields.

### 2. Raw moq-net track API

The catalog API above only writes *metadata*. To actually feed a track, libmoq previously offered only `moq_publish_media_ordered`, which requires a known `FramedFormat` and owns its own catalog/container. There was no way to publish or consume a **non-media** track (control channel, JSON metadata, or your own encoded bitstream) over the C bindings.

This adds the bare moq-net primitives, mirroring `moq-ffi`'s `publish_track` / `subscribe_track` (broadcast → track → group → frame, raw bytes, no codec/container/catalog):

**Publish:**
- `moq_publish_track(broadcast, name, name_len)` → track handle (`create_track`)
- `moq_publish_track_group(track)` → group handle (`append_group`)
- `moq_publish_track_frame(track, payload, len)` — convenience one-frame-per-group write
- `moq_publish_group_frame(group, payload, len)` / `moq_publish_group_close(group)`
- `moq_publish_track_close(track)`

**Consume:**
- `moq_consume_track(broadcast, name, name_len, on_frame, user_data)` → track handle
- `moq_consume_track_frame(frame, moq_frame*)` — fills `payload`/`payload_size`; `timestamp_us`/`keyframe` are 0/false (not meaningful for raw tracks)
- `moq_consume_track_frame_close(frame)` / `moq_consume_track_close(track)`

The consumer iterates groups in sequence order and delivers **every** frame within each group, so it sees exactly what the producer wrote regardless of grouping. Adds a `GroupNotFound` error code (`-31`).

These two pieces compose: describe a track with `moq_publish_*_config`, then push its frames with `moq_publish_track`/`_frame`.

## Cross-package sync

| Change in | Also updated |
|---|---|
| `rs/libmoq` | `CHANGELOG.md`; doc comments flow into the cbindgen-generated `moq.h` that `doc/lib/c` points at |

Confined to `rs/libmoq` and purely additive (no existing signatures change), so it does not touch `rs/moq-ffi`, the wire/catalog formats, or the language wrappers — and targets `main`.

## Test plan

- [x] `cargo test -p libmoq` — 19 tests pass, including:
  - `publish_catalog_roundtrip`: author video + audio renditions, publish, consume the catalog back (codec, coded dimensions, sample rate, channels), then remove the video rendition and confirm the republished catalog drops it.
  - `raw_track_publish_consume`: publish a raw `data` track, consume it via `moq_consume_track`, verify a one-frame-per-group write and a multi-frame group both round-trip byte-for-byte, and that `timestamp_us`/`keyframe` come back 0/false.
  - invalid-handle and null-pointer tests for both APIs.
- [x] `cargo clippy -p libmoq --all-targets` — clean.
- [x] `cargo fmt -p libmoq -- --check` — clean.
- [x] Generated `moq.h` picks up all new functions (one declaration each).

(Written by Claude)

### DIFF
--- a/rs/libmoq/CHANGELOG.md
+++ b/rs/libmoq/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Catalog producer API to author renditions directly (`moq_publish_video_config`, `moq_publish_audio_config`, `moq_publish_video_remove`, `moq_publish_audio_remove`), mirroring the consume-side config queries.
+- Raw moq-net track API for arbitrary (non-media) byte tracks, mirroring the moq-ffi primitives:
+  - Publish: `moq_publish_track`, `moq_publish_track_group`, `moq_publish_track_frame`, `moq_publish_group_frame`, `moq_publish_group_close`, `moq_publish_track_close`.
+  - Consume: `moq_consume_track`, `moq_consume_track_frame`, `moq_consume_track_frame_close`, `moq_consume_track_close`.
 
 ## [0.2.17](https://github.com/moq-dev/moq/compare/libmoq-v0.2.16...libmoq-v0.2.17) - 2026-05-24
 
--- a/rs/libmoq/CHANGELOG.md
+++ b/rs/libmoq/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Catalog producer API to author renditions directly (`moq_publish_video_config`, `moq_publish_audio_config`, `moq_publish_video_remove`, `moq_publish_audio_remove`), mirroring the consume-side config queries.
+
 ## [0.2.17](https://github.com/moq-dev/moq/compare/libmoq-v0.2.16...libmoq-v0.2.17) - 2026-05-24
 
 ### Added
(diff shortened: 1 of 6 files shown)


kixelated:
@coderabbitai review

The previous review aborted because I pushed a follow-up commit (clippy `useless_conversion` + rustfmt fixes) while the review was in flight. The branch is settled now at `eec70b9`. Thanks!

(Written by Claude)

---

PR #1536, opened 2026-05-29
moq-net: auto-create Origin on connect/accept, expose via Session

kixelated:
## Summary

Makes explicitly constructing an Origin **optional** on both client and server in `moq-net` and `moq-ffi`. The common-case setup collapses from this:

```rust
let origin = Origin::random().produce();
let client = Client::new().with_origin(origin.clone());
let session = client.connect(transport).await?;
// ... use origin to publish / consume ...
```

to this:

```rust
let session = Client::new().connect(transport).await?;
let pub_origin = session.publisher().unwrap();
let sub_origin = session.consumer().unwrap();
// ... use them ...
```

`with_publish` / `with_consume` / `with_origin` stay for callers that need a custom topology (relay, gst, cli, ffi); when any of those is set, `session.publisher()` and `session.consumer()` return `None` and the caller drives things through the origin it already holds.

Targeting `dev` per CLAUDE.md (breaking change to public APIs in `rs/moq-net`, `rs/moq-ffi`, and the language wrappers).

## Rust layer (`rs/moq-net`)

`Session` now carries the auto-created origin sides directly:

```rust
pub struct Session {
    /* existing fields */
    publisher: Option<OriginProducer>,
    consumer: Option<OriginConsumer>,
}

impl Session {
    pub fn publisher(&self) -> Option<&OriginProducer> { ... }
    pub fn consumer(&self) -> Option<&OriginConsumer> { ... }
}
```

`Client::connect` and `Server::accept` both auto-create an Origin when neither publish nor consume was set, then surface it through these accessors. One type instead of a separate `ClientSession` wrapper, symmetric across the client and server paths.

Downstream call sites in `moq-native`, `moq-relay`, `moq-cli`, `moq-gst`, `libmoq`, and `moq-ffi` thread `session.publisher()` / `session.consumer()` where they need the auto-created sides; everything else just continues to use `session` as before.

One internal moq-net test was depending on "no origin → trivial close with code 0 (Cancel)" to validate SETUP framing. With auto-origin attached the Lite01 session now closes with `RequiredExtension` (code 1) since Lite01 can't satisfy the new origin. Loosened the assertion to "any non-Version error" since the test's purpose is the framing path, not the close code.

## FFI layer (`rs/moq-ffi`)

Mirror at the FFI level: `MoqSession` gains `publisher()` and `consumer()` accessor methods (uniffi Objects, return `Option<Arc<MoqOriginProducer>>` / `Option<Arc<MoqOriginConsumer>>`). `MoqClient::connect` continues to return `Arc<MoqSession>`.

Renamed `MoqOriginProducer::publish` → `add_broadcast` so `session.publisher().unwrap().add_broadcast(path, broadcast)` reads cleanly instead of stuttering `…publisher().publish(...)`.

New test `server_client_roundtrip_auto_origin` validates the no-config path end-to-end (client doesn't call `set_publish`/`set_consume`; `session.publisher()` and `session.consumer()` are `Some` and drive a server → client broadcast).

## Per-language wrapper sync (CLAUDE.md cross-package table)

- **`py/moq-rs/moq/client.py`**: no extraction step needed, just stores the returned `MoqSession` directly.
- **`py/moq-rs/moq/origin.py`**: Python `OriginProducer.publish` wrapper kept (no Python API churn); now calls `add_broadcast` on the FFI underneath.
- **`swift/README.md`, `doc/lib/{swift,kt}/{index,moq}.md`**: show the new shape — no `MoqOriginProducer()` construction in the happy path; `session.publisher()` / `session.consumer()` return the auto-created sides.
- **`go/`**: pure binding shim, no Go-side wrapper to update; consumers get the regenerated `MoqClient.Connect()` returning `*MoqSession` with the new accessors.
- **Swift `Moq.swift` / Kotlin `Moq.kt`**: already deleted in the prior PR #1526, so no entry-point shims to update here.

## Why fold into `Session` instead of keeping `ClientSession`

The earlier version of this PR added a separate `ClientSession { session, publisher, consumer }` wrapper. Two reasons to fold:

1. **Symmetry**: the same auto-origin behavior applies to servers too. Keeping `ClientSession` would force either a parallel `ServerSession` (more types) or asymmetric APIs. Folding the fields onto `Session` lets `Server::accept` use the same path with zero new naming.
2. **Naming**: there's no obvious neutral name for the wrapper (`Endpoint` and `Connection` are overloaded), and `Session` already represents "the connected thing." Adding optional fields fits cleanly without inventing a new concept.

## Test plan

- [x] `cargo test --workspace --tests` passes (all suites, including the new `server_client_roundtrip_auto_origin`).
- [x] `cargo fmt --check` clean; `cargo clippy --workspace --all-targets` clean.
- [x] `just swift check` passes (2/2 smoke tests).
- [ ] Kotlin: gradle toolchain not available locally, so `just kt check` only ran the binding regeneration step. CI exercises the full Kotlin build.
- [ ] Python: smoke tests not run locally (no uv environment set up in this worktree). CI runs `just py check`.

(Written by Claude)

### DIFF
--- a/doc/lib/kt/index.md
+++ b/doc/lib/kt/index.md
@@ -46,8 +46,7 @@ val cs = client.connect("https://relay.example.com")
 // cs.consumer() and cs.publisher() are always populated: by whatever
 // origin you wired via setPublish / setConsume before connect, or by a
 // fresh auto-created one for any side you didn't set.
-val consumer = cs.consumer()
-val announced = consumer.announced("demos/")
+val announced = cs.consumer().announced("demos/")
 announced.announcements().collect { announcement ->
     println("got broadcast ${announcement.path()}")
 
--- a/doc/lib/kt/index.md
+++ b/doc/lib/kt/index.md
@@ -43,11 +43,10 @@ import uniffi.moq.MoqClient
 val client = MoqClient()
 val cs = client.connect("https://relay.example.com")
 
-// `cs.consumer()` returns the auto-created origin consumer for reading
-// announcements; `cs.publisher()` returns the producer side for publishing
-// broadcasts. Both are null if you set publish / consume manually before
-// connect (subscribe-only, publish-only, or custom split-origin cases).
-val consumer = cs.consumer()!!
+// cs.consumer() and cs.publisher() are always populated: by whatever
+// origin you wired via setPublish / setConsume before connect, or by a
+// fresh auto-created one for any side you didn't set.
+val consumer = cs.consumer()
 val announced = consumer.announced("demos/")
 announced.announcements().collect { announcement ->
     println("got broadcast ${announcement.path()}")
--- a/doc/lib/kt/index.md
+++ b/doc/lib/kt/index.md
@@ -57,7 +57,7 @@ announced.announcements().collect { announcement ->
     }
 }
 
-cs.session().shutdown()
+cs.shutdown()
 ```
 
 Cancelling the surrounding coroutine scope propagates through to the native consumer's `cancel()` via the wrapper's `onCompletion` hook.
--- a/doc/lib/kt/index.md
+++ b/doc/lib/kt/index.md
@@ -39,30 +39,25 @@ Published to Maven Central via [release-kt.yml](https://github.com/moq-dev/moq/b
 import dev.moq.*
 import kotlinx.coroutines.flow.collect
 import uniffi.moq.MoqClient
-import uniffi.moq.MoqOriginProducer
 
-// Wire an origin as both publish source and consume sink. Set just one
-// side for a subscribe-only or publish-only client.
-val origin = MoqOriginProducer()
 val client = MoqClient()
-client.setPublish(origin)
-client.setConsume(origin)
+val cs = client.connect("https://relay.example.com")
 
-val session = client.connect("https://relay.example.com")
+// `cs.consumer()` returns the auto-created origin consumer for reading
+// announcements; `cs.publisher()` returns the producer side for publishing
+// broadcasts. Both are null if you set publish / consume manually before
+// connect (subscribe-only, publish-only, or custom split-origin cases).
+val consumer = cs.consumer()!!
+val announced = consumer.announced("demos/")
+announced.announcements().collect { announcement ->
+    println("got broadcast ${announcement.path()}")
 
-origin.use {
-    val consumer = origin.consume()
-    val announced = consumer.announced("demos/")
-    announced.announcements().collect { announcement ->
-        println("got broadcast ${announcement.path()}")
-
-        announcement.broadcast().subscribeCatalog().updates().collect { catalog ->
-            println("catalog: $catalog")
-        }
+    announcement.broadcast().subscribeCatalog().updates().collect { catalog ->
+        println("catalog: $catalog")
     }
 }
 
-session.shutdown()
+cs.session().shutdown()
 ```
 
 Cancelling the surrounding coroutine scope propagates through to the native consumer's `cancel()` via the wrapper's `onCompletion` hook.
--- a/doc/lib/kt/moq.md
+++ b/doc/lib/kt/moq.md
@@ -30,18 +30,12 @@ Android uses JNI (`jniLibs/`), desktop JVM uses JNA (resource-classpath layout).
 
 ```kotlin
 import uniffi.moq.MoqClient
-import uniffi.moq.MoqOriginProducer
 
-// Wire an origin as both publish source and consume sink for the
-// typical full-duplex client. Set just one side for a subscribe-only
-// or publish-only client.
-val origin = MoqOriginProducer()
 val client = MoqClient()
-client.setPublish(origin)
-client.setConsume(origin)
-
-val session = client.connect("https://relay.example.com")
+val cs = client.connect("https://relay.example.com")
 ```
+
+`MoqClient.connect(url)` returns a `MoqClientSession` that bundles the session with auto-created publish and consume origin sides. The convenience accessors `cs.publisher()` and `cs.consumer()` return those origins (or `null` if you wired your own via `setPublish` / `setConsume` before connect).
 
 For development against a relay with a self-signed certificate, configure the client before connecting:
 
@@ -49,15 +43,13 @@ For development against a relay with a self-signed certificate, configure the cl
 val client = MoqClient()
 client.setTlsDisableVerify(true)
 client.setBind("127.0.0.1:0")
-client.setPublish(origin)
-client.setConsume(origin)
-val session = client.connect("https://localhost:4443")
+val cs = client.connect("https://localhost:4443")
 ```
 
 When you're done, signal graceful shutdown to the peer:
 
 ```kotlin
-session.shutdown()  // alias for cancel(0u)
+cs.session().shutdown()  // alias for cancel(0u)
 ```
 
 ## Subscribe
@@ -66,7 +58,7 @@ session.shutdown()  // alias for cancel(0u)
 import dev.moq.*
 import kotlinx.coroutines.flow.collect
 
-val consumer = origin.consume()
+val consumer = cs.consumer()!!
 val announced = consumer.announced("demos/")
 
 announced.announcements().collect { announcement ->
@@ -86,7 +78,7 @@ import uniffi.moq.MoqBroadcastProducer
 val broadcast = MoqBroadcastProducer()
 val audio = broadcast.publishMedia("opus", opusInitBytes)
 
-origin.publish("my-stream", broadcast)
+cs.publisher()!!.addBroadcast("my-stream", broadcast)
 
 audio.writeFrame(payload, timestampUs = 0u)
 audio.writeFrame(payload, timestampUs = 20_000u)
--- a/doc/lib/kt/moq.md
+++ b/doc/lib/kt/moq.md
@@ -78,7 +78,7 @@ import uniffi.moq.MoqBroadcastProducer
 val broadcast = MoqBroadcastProducer()
 val audio = broadcast.publishMedia("opus", opusInitBytes)
 
-cs.publisher().addBroadcast("my-stream", broadcast)
+cs.publisher().announce("my-stream", broadcast)
 
 audio.writeFrame(payload, timestampUs = 0u)
 audio.writeFrame(payload, timestampUs = 20_000u)
--- a/doc/lib/kt/moq.md
+++ b/doc/lib/kt/moq.md
@@ -35,7 +35,7 @@ val client = MoqClient()
 val cs = client.connect("https://relay.example.com")
 ```
 
-`MoqClient.connect(url)` returns a `MoqClientSession` that bundles the session with auto-created publish and consume origin sides. The convenience accessors `cs.publisher()` and `cs.consumer()` return those origins (or `null` if you wired your own via `setPublish` / `setConsume` before connect).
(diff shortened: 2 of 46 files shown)


doc/lib/swift/index.md
```diff
@@ -49,16 +49,13 @@ The package depends on a prebuilt `MoqFFI.xcframework` attached to the matching
...
-client.setConsume(origin: origin)
+let cs = try await client.connect(url: "https://relay.example.com")
 
-let session = try await client.connect(url: "https://relay.example.com")
-
-let consumer = origin.consume()
+// cs.consumer() and cs.publisher() are always populated: by whatever
+// origin you wired via setPublish / setConsume before connect, or by a
+// fresh auto-created one for any side you didn't set.
+let consumer = cs.consumer()
```
kixelated: I don't think we need this temporary variable.

doc/lib/swift/moq.md
```diff
@@ -33,38 +33,31 @@ Supported platforms: iOS 15+, iPadOS 15+, macOS 12+. The package ships an XCFram
...
-// Wire an origin as both publish source and consume sink for the
-// typical full-duplex client. Set just one side for a subscribe-only
-// or publish-only client.
-let origin = MoqOriginProducer()
 let client = MoqClient()
-client.setPublish(origin: origin)
-client.setConsume(origin: origin)
-
-let session = try await client.connect(url: "https://relay.example.com")
+let cs = try await client.connect(url: "https://relay.example.com")
```
kixelated: Use cdn.moq.dev as the example in general.

doc/lib/swift/moq.md
```diff
@@ -81,7 +74,7 @@ for try await announcement in announced.announcements {
 let broadcast = try MoqBroadcastProducer()
 let audio = try broadcast.publishMedia(format: "opus", init: opusInitBytes)
 
-try origin.publish(path: "my-stream", broadcast: broadcast)
+try cs.publisher().addBroadcast(path: "my-stream", broadcast: broadcast)
```
kixelated: Can you use the verb `announce` instead? addBroadcast -> announce

rs/hang/examples/video.rs
```diff
@@ -29,8 +29,9 @@ async fn run_session(origin: moq_net::OriginConsumer) -> anyhow::Result<()> {
 	let url = url::Url::parse("https://cdn.moq.dev/anon/video-example").unwrap();
 
 	// Establish a connection with automatic reconnection.
-	// with_publish() registers an OriginConsumer for outgoing data.
-	// Use with_consume() if you also want to subscribe/consume from the session.
+	// with_publish() registers an OriginProducer; the client reads from
+	// its consumer view to forward broadcasts to the remote. Pair with
+	// with_consume() if you also want to subscribe to remote announcements.
 	let reconnect = client.with_publish(origin).reconnect(url);
```
kixelated: `with_publish` should take an OriginConsumer

rs/libmoq/src/session.rs
```diff
@@ -22,7 +22,7 @@ impl Session {
 	pub fn connect(
 		&mut self,
 		url: Url,
-		publish: Option<moq_net::OriginConsumer>,
+		publish: Option<moq_net::OriginProducer>,
```
kixelated, line 25: OriginConsumer

doc/lib/swift/index.md
kixelated: Done — inlined as `let announced = try cs.consumer().announced(prefix: "demos/")`.

doc/lib/swift/moq.md
kixelated: Done — switched both connect examples to `https://cdn.moq.dev/anon/demo`.

doc/lib/swift/moq.md
kixelated: Done — renamed FFI `MoqOriginProducer::add_broadcast` → `announce`. Python's `OriginProducer.publish` wrapper kept its public name and now forwards to `inner.announce` underneath.

rs/hang/examples/video.rs
kixelated: Done — reverted `with_publish` (Client + Server + Request + moq-native + FFI + libmoq) to take `OriginConsumer`. All callers re-add the `.consume()`; relay's `cluster.subscriber()` returns `OriginConsumer` again.

rs/libmoq/src/session.rs
kixelated: Done — `Session::connect`'s `publish` parameter is back to `Option<OriginConsumer>`; `api.rs` derives the consumer at the call.

rs/moq-net/src/client.rs
```diff
@@ -221,20 +252,78 @@ impl Client {
...
+/// the wire depends on whether the user overrode that side:
+/// - Neither overridden: one shared origin powers everything (duplex).
+/// - Override only publish: the user's consumer drives the wire publish
+///   side; `Session.publisher` becomes a standalone no-op producer.
+///   `Session.consumer` and wire-consume share a fresh shared origin.
+/// - Override only consume: symmetric to above.
+/// - Both overridden: `Session.publisher` and `Session.consumer` are
+///   independent standalone defaults (no-ops); the wire uses what the
+///   user passed.
+pub(crate) fn resolve_origins(
```
kixelated: I have no idea what this is for.

rs/moq-net/src/server.rs
```diff
@@ -51,10 +58,17 @@ impl Server {
...
+	/// [`Origin`](crate::Origin) is auto-created and shared between
+	/// both. The same origin drives the wire. When the caller
+	/// overrode either side, the session's accessor for that side
+	/// becomes a standalone auto-created no-op decoupled from the wire.
 	pub async fn accept<S: web_transport_trait::Session>(&self, session: S) -> Result<Session, Error> {
-		if self.publish.is_none() && self.consume.is_none() {
-			tracing::warn!("not publishing or consuming anything");
-		}
+		let (publisher, consumer_view, publish, consume) =
+			crate::client::resolve_origins(self.publish.clone(), self.consume.clone());
```
kixelated: I'm so confused.

What about:

```rust
let publisher = self.publish.unwrap_or(OriginProducer::random().consume());
let consumer = self.consume.unwrap_or(OriginProducer::random());
```

I don't know why you're returning a 4-tuple?

doc/lib/kt/index.md
```diff
@@ -39,30 +39,24 @@ Published to Maven Central via [release-kt.yml](https://github.com/moq-dev/moq/b
...
-
-        announcement.broadcast().subscribeCatalog().updates().collect { catalog ->
-            println("catalog: $catalog")
-        }
+val cs = client.connect("https://relay.example.com")
+
+// cs.consumer() and cs.publisher() are always populated: by whatever
+// origin you wired via setPublish / setConsume before connect, or by a
+// fresh auto-created one for any side you didn't set.
+val consumer = cs.consumer()
```
kixelated: Don't need the temporary variable

doc/lib/kt/moq.md
```diff
@@ -66,7 +58,7 @@ session.shutdown()  // alias for cancel(0u)
 import dev.moq.*
 import kotlinx.coroutines.flow.collect
 
-val consumer = origin.consume()
+val consumer = cs.consumer()
```
kixelated: not needed

doc/lib/rs/env/native.md
```diff
@@ -80,13 +80,13 @@ See the full [video.rs](https://github.com/moq-dev/moq/blob/main/rs/hang/example
...
 
-To consume a broadcast, use `with_consume()` and listen for announcements:
+To consume a broadcast, use `with_consumer()` and listen for announcements:
 
 ```rust
 let origin = moq_net::Origin::new().produce();
 let mut consumer = origin.consume();
 let session = client
-    .with_consume(origin)
+    .with_consumer(origin)
```
kixelated: None of this is needed now

doc/lib/swift/moq.md
```diff
@@ -33,38 +33,31 @@ Supported platforms: iOS 15+, iPadOS 15+, macOS 12+. The package ships an XCFram
...
 ```swift
-session.shutdown()  // alias for cancel(code: 0)
+cs.shutdown()  // alias for cancel(code: 0)
 ```
 
 ## Subscribe
 
 ```swift
-let consumer = origin.consume()
+let consumer = cs.consumer()
```
kixelated: not needed

rs/moq-net/src/client.rs
```diff
@@ -221,20 +249,30 @@ impl Client {
...
 
-		Ok(Session::new(session, version, recv_bw))
+		Ok(Session::new(session, version, recv_bw, publisher, consumer_view))
+	}
+}
+
+/// Pick the publish / consume [`OriginProducer`]s for a session,
+/// defaulting any side the caller didn't set. When both are unset, a
+/// single shared fresh origin powers both (the typical duplex client).
+pub(crate) fn resolve_origins(
```
kixelated: I think you just Default publish/consume to the shared Origin::random().produce() (instead of None), then you can delete this.

rs/moq-net/src/client.rs
```diff
@@ -51,10 +60,18 @@ impl Client {
...
+	/// / [`Self::with_origin`], or a fresh auto-created [`Origin`] for
+	/// any side the caller left unset. When neither side is set, both
+	/// default to the same shared origin (the typical full-duplex client).
 	pub async fn connect<S: web_transport_trait::Session>(&self, session: S) -> Result<Session, Error> {
-		if self.publish.is_none() && self.consume.is_none() {
-			tracing::warn!("not publishing or consuming anything");
-		}
+		let (publisher, consumer) = resolve_origins(self.publish.clone(), self.consume.clone());
+		let publish = Some(publisher.consume());
+		let consume = Some(consumer.clone());
```
kixelated: You should make the inner libraries no longer take Option.

doc/lib/kt/index.md
kixelated: Done.

doc/lib/kt/moq.md
kixelated: Done — inlined as `val announced = cs.consumer().announced("demos/")`.

doc/lib/rs/env/native.md
kixelated: Done — rewrote both Publishing and Subscribing sections in native.md to use `session.publisher()` / `session.consumer()` directly. No more manual `Origin::new().produce()` + `with_publisher()` setup.

doc/lib/swift/moq.md
kixelated: Done — inlined to `let announced = try cs.consumer().announced(prefix: "demos/")`.

rs/moq-net/src/client.rs
kixelated: Done — Client/Server fields now `OriginProducer` (non-Option) and `Default::default()` initializes both sides to a shared `Origin::random().produce()`. `resolve_origins` deleted; `connect`/`accept` just clone the fields.

rs/moq-net/src/client.rs
kixelated: Done — `lite::start`, `ietf::start`, `PublisherConfig`, `SubscriberConfig`, and the ietf `Subscriber` field all take/store non-Option now. No-op fallbacks (`Origin::random().produce().consume()`, `has_origin()` guards, the `.as_ref().expect(...)` chains) all gone. Callers that still hold an Option (FFI, libmoq, relay) move the None-handling to their own call site.

PR #1540, opened 2026-05-29
moq-net: make subscribe_track async, blocking on SUBSCRIBE_OK

kixelated:
## Summary

Reshapes `subscribe_track` so a subscription resolves once the publisher confirms it, and separates subscriber-controlled knobs from publisher-owned track properties. (Picks up the idea started in #1439, implemented cleanly on top of `dev`.)

- **`Subscription` (new)** holds the subscriber wire params: `priority`, `ordered`, `max_latency`, `start_group`, `end_group`.
- **`Track`** now holds only immutable publisher properties — just `name` today; timescale can be added later.
- **`BroadcastConsumer::subscribe_track(name: &str, subscription: Subscription)` is now `async`** and returns once the publisher accepts. Concurrent subscribers to the same name coalesce onto one request.
- **`BroadcastDynamic::requested_track()` returns a `TrackRequest`.** The publisher calls `accept(Track) -> TrackProducer` (creating the producer and unblocking the waiting subscriber) or `deny(err)`. Dropping the request denies with `Cancel`.
- **lite subscriber:** `accept`s right after SUBSCRIBE_OK, so a downstream `subscribe_track` resolves exactly when the upstream confirms — woven into the existing linger lifecycle (#1514).
- **ietf subscriber:** accepts immediately, since group data can arrive before SUBSCRIBE_OK.
- Per-consumer preferences are an `Arc<Mutex<Subscription>>` updatable via `TrackConsumer::update_subscription`; `TrackProducer::subscription()` exposes the aggregate (most-demanding) across all live subscribers, which the relay forwards upstream.
- `conducer::Waiter` gained a `waker()` accessor so poll-based callers (moq-mux export) can drive the async subscribe inside the existing poll model.

### Cross-package
- **FFI (`moq-ffi` / `libmoq`)**: keeps its **synchronous** surface by driving the async subscribe on the runtime (`block_on` / inside spawned tasks), so the py/swift/kt/go bindings need **no regeneration**.
- **`js/net`**: no change. It already exposes `subscribe(name, priority)` with a reactive `Track` (`updatePriority`); this change brings the Rust API closer to it, and the **wire format is unchanged**.
- **docs**: updated the Rust example in `doc/lib/rs/env/native.md`. The Python doc examples are unchanged (FFI API unchanged).

## Test plan

- [x] `cargo check --workspace --all-targets` clean (no warnings)
- [x] `cargo clippy --workspace --all-targets` clean
- [x] `cargo test` for `moq-net`, `moq-mux`, `moq-audio`, `moq-native`, `moq-relay` — all pass, including the moq-native/moq-relay integration tests that exercise the real SUBSCRIBE_OK round-trip
- [ ] `just check` full gate (JS/bun, etc.) — not run locally

(Written by Claude)


rs/hang/src/catalog/root.rs
```diff
@@ -77,12 +77,14 @@ impl Catalog {
...
-			compress: true,
-		}
+		// The catalog is JSON and re-sent on every change, so it pays to compress.
+		moq_net::Track::new(Catalog::DEFAULT_NAME).with_compress(true)
+	}
+
+	/// The subscription preferences used for the catalog track (high priority so
+	/// it preempts media tracks).
+	pub fn default_subscription() -> moq_net::Subscription {
+		moq_net::Subscription::new(100)
```
kixelated: I think this is confusing. We should remove `::new` and require `priority` as a key, either by constructing the struct or using a builder.

rs/moq-ffi/src/audio.rs
```diff
@@ -250,7 +248,9 @@ impl MoqBroadcastConsumer {
 		cfg.description = catalog_audio.description.map(Into::into);
 		cfg.container = catalog_audio.container.into();
 
-		let consumer = moq_audio::AudioConsumer::new(
+		// `subscribe_track` now blocks on SUBSCRIBE_OK; drive it on the runtime so
+		// this synchronous FFI method keeps its signature.
+		let consumer = crate::ffi::RUNTIME.block_on(moq_audio::AudioConsumer::new(
```
kixelated: Never use block_on. Change the method to be async.

rs/moq-ffi/src/consumer.rs
```diff
@@ -82,8 +82,12 @@ impl Media {
 impl MoqBroadcastConsumer {
 	/// Subscribe to the catalog for this broadcast.
 	pub fn subscribe_catalog(&self) -> Result<Arc<MoqCatalogConsumer>, MoqError> {
-		let _guard = crate::ffi::RUNTIME.enter();
-		let track = self.inner.subscribe_track(&hang::catalog::Catalog::default_track())?;
+		// `subscribe_track` now blocks on SUBSCRIBE_OK; drive it on the runtime so
+		// this synchronous FFI method keeps its signature.
+		let track = crate::ffi::RUNTIME.block_on(self.inner.subscribe_track(
```
kixelated: nope

rs/moq-mux/src/container/source.rs
```diff
@@ -46,18 +48,40 @@ impl VideoTransform {
...
 	transform: Option<VideoTransform>,
 	/// Resolved codec configuration record (avcC / hvcC / AudioSpecificConfig /
 	/// OpusHead). Some once the codec config is available — from the catalog
 	/// `description`, or synthesized by the transform.
 	description: Option<Bytes>,
 }
 
+/// Build the (deferred) subscribe future so `ExportSource` constructors stay
+/// synchronous and the subscription resolves inside the existing poll loop.
+fn subscribe_future(
```
kixelated: This is super gross, how can we improve it?

You are allowed to change methods and interfaces.

rs/moq-net/src/lite/subscriber.rs
```diff
@@ -479,62 +468,93 @@ impl<S: web_transport_trait::Session> Subscriber<S> {
...
+		// they still read it through a watch receiver (a group's QUIC stream can
+		// otherwise race ahead of SUBSCRIBE_OK).
+		let mut track = match request.accept(Track::new(name)) {
+			Ok(track) => track,
+			Err(err) => {
+				stream.writer.abort(&err);
+				return SessionOutcome::Error(err);
+			}
+		};
+		let (_compression_tx, compression_rx) = tokio::sync::watch::channel(Some(info.compression));
```
kixelated: This shouldn't be compression specific any longer. We could even save the SubscribeOK itself.

And ideally we don't use tokio. Can you use kio Producer/Consumer? The Producer would be in this task (writes the Option<SubscribeOk>) while the Consumer goes in the subscribes map.

rs/moq-net/src/model/broadcast.rs
```diff
@@ -32,13 +34,32 @@ impl Broadcast {
 	}
 }
 
-#[derive(Default, Clone)]
+/// One subscriber awaiting a pending request: its preferences and the channel
+/// that delivers the resolved [`TrackConsumer`] (or an error).
+type Resolver = (Subscription, oneshot::Sender<Result<TrackConsumer, Error>>);
```
kixelated: Bonus points if you can use kio instead.

rs/moq-net/src/model/broadcast.rs
```diff
@@ -352,66 +499,62 @@ impl Deref for BroadcastConsumer {
...
+	/// returns immediately. Otherwise queues a dynamic request and waits for the
+	/// broadcast's producer to serve it via [`BroadcastDynamic::requested_track`]
+	/// and [`TrackRequest::accept`] (for the wire this happens on SUBSCRIBE_OK).
+	/// Returns [`Error::NotFound`] if no dynamic producer exists to handle the
+	/// request.
+	///
+	/// The returned [`TrackConsumer`] carries the publisher's [`Track`]; the given
+	/// [`Subscription`] is this subscriber's preferences and feeds the producer's
+	/// [`TrackProducer::subscription`] aggregate.
+	pub async fn subscribe_track(&self, name: &str, subscription: Subscription) -> Result<TrackConsumer, Error> {
```
kixelated: Maybe mock up if we can return a `TrackRequest` (or another name if that's used) instead? Then this function wouldn't be async, but `TrackRequest::ok()` (or something) would be? The only difference is that we could implement `poll_ok`.

rs/moq-net/src/model/track.rs
```diff
@@ -63,6 +66,49 @@ impl Track {
...
+/// Each subscriber holds its own [`Subscription`]; the publisher observes an
+/// aggregate across all live subscribers via [`TrackProducer::subscription`].
+/// A subscriber can change its preferences after the fact with
+/// [`TrackConsumer::update_subscription`].
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Subscription {
+	/// Delivery priority. Higher values preempt lower ones when bandwidth is constrained.
+	pub priority: u8,
+	/// Whether groups should be delivered in sequence order.
+	pub ordered: bool,
```
kixelated, line 92: defaults to false, so we skip during congestion

rs/moq-net/src/model/track.rs
```diff
@@ -63,6 +66,49 @@ impl Track {
...
+	/// Maximum delay before a group is considered stale. `Duration::ZERO` means
+	/// no deadline (wait forever).
+	pub max_latency: Duration,
```
kixelated: This is the duration before we consider a group stale if it's not the most recent. ZERO means immediate; group 8 arriving meaning group 7 should be skipped.

Can you rename this to `stale`  across the board?

rs/moq-net/src/model/track.rs
```diff
@@ -63,6 +66,49 @@ impl Track {
...
+	/// First group to deliver, or `None` to start at the latest group.
+	pub start_group: Option<u64>,
```
kixelated: Do you like `start_group` and `end_group`, or `group_start` and `group_end`? I kinda like the latter more.

rs/moq-net/src/model/track.rs
```diff
@@ -267,6 +317,60 @@ impl State {
...
 		}
 	}
+
+	/// Aggregate every live subscriber's preferences into a single [`Subscription`]
+	/// describing the most demanding request, pruning entries for dropped consumers.
+	///
+	/// The result is what an upstream relay should forward: highest priority,
+	/// ordered if anyone wants it, the tightest latency deadline, the earliest
+	/// start group, and the latest end group (unbounded wins).
+	fn aggregate_subscription(&mut self) -> Subscription {
```
kixelated: return Option instead of default

rs/moq-net/src/model/track.rs
```diff
@@ -392,18 +496,40 @@ impl TrackProducer {
 		Ok(())
 	}
 
-	/// Create a new consumer for the track, starting at the beginning.
-	pub fn consume(&self) -> TrackConsumer {
+	/// Create a new consumer for the track with the given subscriber preferences.
+	///
+	/// The preferences feed the producer's [`Self::subscription`] aggregate and
+	/// can be changed later via [`TrackConsumer::update_subscription`].
+	pub fn consume_with(&self, subscription: Subscription) -> TrackConsumer {
```
kixelated: Rename this to `subscribe`

rs/moq-net/src/model/track.rs
```diff
@@ -392,18 +496,40 @@ impl TrackProducer {
...
+			subscription,
 			index: 0,
 			min_sequence: 0,
 			next_sequence: 0,
 			end_sequence: None,
 		}
 	}
 
+	/// Create a new consumer for the track with default ([`Subscription::default`]) preferences.
+	pub fn consume(&self) -> TrackConsumer {
```
kixelated: `subscribe_default`?

rs/moq-net/src/model/track.rs
```diff
@@ -683,6 +821,16 @@ impl TrackConsumer {
 		self.end_sequence = sequence.into();
 	}
 
+	/// This subscriber's current preferences.
+	pub fn subscription(&self) -> Subscription {
+		self.subscription.lock().unwrap().clone()
+	}
+
+	/// Replace this subscriber's preferences, updating the producer's aggregate.
+	pub fn update_subscription(&self, subscription: Subscription) {
```
kixelated: `update`

rs/moq-net/src/stats.rs
```diff
@@ -807,11 +807,7 @@ async fn run_publisher(weak: Weak<StatsShared>, advertised: PathOwned, interval:
 	let mut broadcast = Broadcast::new().produce();
 	let mut tracks: Vec<TrackProducer> = Vec::with_capacity(NUM_SLOTS);
 	for name in TRACK_ORDER {
-		match broadcast.create_track(Track {
-			name: name.into(),
-			priority: 0,
-			..Default::default()
-		}) {
+		match broadcast.create_track(Track::new(name)) {
```
kixelated, line 810: Can create_track take `impl Into<Track>`, so Track::new is optional?

kixelated:
Addressed the review in `8b77d931`:

**`subscribe_track` returns a pollable handle (no forced await).** It's now sync and returns a `TrackPending`; call `poll_ok(waiter)` (kio-style) or `ok().await`. `moq-mux`'s `ExportSource` uses `poll_ok` directly now — the `Box<dyn Future>` + `Context::from_waker` hack and the unused `kio::Waiter::waker` are gone.

**No more `block_on`.** The `moq-ffi` `subscribe_catalog`/`subscribe_track`/`subscribe_media`/`subscribe_audio` methods are async (uniffi async export). The only `block_on` left in ffi/libmoq are the runtime-keeper thread and a test helper.

**More kio, less tokio.**
- Broadcast resolver channels: tokio `oneshot` → kio `Producer`/`Consumer`.
- lite subscriber: the tokio `watch<Option<Compression>>` is now a kio `Consumer<Option<SubscribeOk>>` — the whole SUBSCRIBE_OK is carried, not just compression. The producer writes it in the subscribe task; the consumer lives in the `subscribes` map.

**Renames / API:**
- `Subscription`: `max_latency` → `stale` (with the reorder-tolerance doc you described), `start_group`/`end_group` → `group_start`/`group_end`, removed `Subscription::new` (construct the struct).
- `TrackProducer`: `consume_with` → `subscribe`, `consume` → `subscribe_default`, `subscription()` now returns `Option`.
- `TrackConsumer::update_subscription` → `update`.
- `create_track` takes `impl Into<Track>` (+ `From<&str>`/`From<String>`), so `create_track("video")` works.

One subtlety worth flagging: the consumer is now built at `accept` time (delivered through the channel) rather than lazily on poll, so it counts toward the producer immediately. Building it lazily caused a no-linger cancel race on lite-01/02 (the publisher saw `unused()` before the subscriber materialized its consumer).

Bindings: py + swift wrappers updated to `await` the now-async subscribe methods; Kotlin's `suspend` wrapper is transparent. `js/net` unchanged (wire format untouched).

`cargo fmt`/`clippy` clean; moq-net/mux/native/relay/libmoq/rtc/audio/kio tests all pass.

(Written by Claude)

PR #1541, opened 2026-05-29
moq-ffi: route Android logs to logcat

Qizot:
While on iOS the logs can be inspected in xcode the same thing can't be said about android. For this particular usecase, this PR adds support for a android locat tracing subscriber so the logs are properly routed and can be inspected using ADB or Android Studio (both using logcat).

### DIFF
--- a/rs/moq-ffi/src/log.rs
+++ b/rs/moq-ffi/src/log.rs
@@ -1,6 +1,7 @@
 use std::str::FromStr;
 
 use crate::error::MoqError;
+use tracing::Level;
 
 /// Initialize logging with a level string: "error", "warn", "info", "debug", "trace", or "".
 ///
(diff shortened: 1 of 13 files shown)


rs/moq-ffi/src/log.rs
```diff
@@ -1,27 +1,62 @@
...
 	}
 
-	log.init();
+	init_logging(level)?;
 
 	Ok(())
 }
+
+#[cfg(all(target_os = "android", feature = "android-logcat"))]
+fn init_logging(level: Level) -> Result<(), MoqError> {
```
kixelated: Should this be in moq-native instead?

rs/moq-ffi/src/log.rs
```diff
@@ -1,27 +1,62 @@
...
+		.add_directive("tungstenite=info".parse().unwrap())
+		.add_directive("rustls=info".parse().unwrap())
+		.add_directive("tracing::span=off".parse().unwrap())
+		.add_directive("tracing::span::active=off".parse().unwrap())
+		.add_directive("tokio=info".parse().unwrap())
+		.add_directive("runtime=info".parse().unwrap());
+
+	let logcat_layer = tracing_android::layer("MoQNative")
+		.map_err(|err| MoqError::Log(format!("failed to initialize Android logcat layer: {err}")))?
+		.with_filter(filter);
```
kixelated: If we move this to moq-native, this is the #[cfg] gate, while the rest is shared.

PR #1544, opened 2026-05-29
libmoq: auto-reconnect sessions; conducer-based Reconnect notifications

kixelated:
## Summary

`moq_session_connect` established a single session that died permanently when the connection dropped. It now drives `moq_native`'s `Reconnect` helper (the same one used by `moq-cli`, `moq-boy`, and the `hang`/`moq-native` examples) so libmoq sessions reconnect with exponential backoff. Origins outlive the connection, so published broadcasts are re-announced and consumers re-subscribed automatically on each reconnect.

## `Reconnect` rebuilt on `kio`

`Reconnect` stores its state (current status + terminal error) in a `kio` channel instead of `tokio::sync::watch`, matching the rest of moq-net. It exposes:

- `status(&mut self) -> Result<Status>` / `poll_status` — returns the current `Status` (`Connected`/`Disconnected`) whenever it differs from what this handle last reported. Level-triggered: it tracks the *current* state, so a flip-and-flip-back between polls collapses to a single report. `Err` is the terminal signal (give-up error, or generic once the handle is dropped).
- `closed() -> Result<()>` / `poll_closed` — for callers that only want to await termination (unchanged for existing consumers).

`kio` is now re-exported from moq-net (`pub use kio`). Its public poll_* API already exposes `kio::Waiter`, so downstream callers need to name the type; moq-native uses `moq_net::kio` rather than a direct dep. (Making `Reconnect` cloneable for multiple handles is deferred to a follow-up PR.)

## libmoq callback contract

`moq_session_connect`'s `on_status` reports connection state via the sign of the code:

- **positive** — connected; the value is the connection epoch (1 = first connect, 2+ = reconnect).
- **negative** — reconnection permanently gave up (backoff timeout exceeded). Terminal.

Transient disconnects are not reported (a separate change reserves 0 for "closed"). The status loop lives in `report()`, which returns `anyhow::Result<()>` so `?` flows; `connect_run` maps the terminal error to `Connect` once. The callback is copied out before the global lock is released, so the C callback never runs while the lock is held. The function signature is unchanged (no wire/ABI break), so this targets `main`.

## Follow-up (not in this PR)

`rs/moq-ffi/src/session.rs` has the identical single-shot `.connect()` gap for the Python/Swift/Kotlin/Go bindings. The same `.connect()` -> `.reconnect()` swap applies there.

## Test plan

- [x] `cargo clippy --all-targets --workspace -- -D warnings` clean
- [x] `cargo test -p libmoq -p moq-native -p moq-net` pass
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --workspace` clean
- [x] `moq-cli`/`moq-boy`/`hang`/`moq-native` consumers + examples still build
- [ ] Manual: drop/restore a relay mid-session and confirm the callback reports a reconnect (epoch++) and the publisher re-announces / subscriber resumes

(Written by Claude)


rs/libmoq/src/session.rs
```diff
@@ -57,22 +62,25 @@ impl Session {
...
-			.connect(url)
-			.await
-			.map_err(|err| Error::Connect(Arc::new(err)))?;
+			.with_consume(consume);
 
-		// "Connected" callback — copy from slab if not revoked.
-		if let Some(Some(entry)) = State::lock().session.task.get(task_id) {
-			entry.callback.call(());
-		}
+		let reconnect = client.reconnect(url);
```
kixelated: Chain this above.

rs/libmoq/src/session.rs
```diff
@@ -57,22 +62,25 @@ impl Session {
...
-		session.closed().await?;
-		Ok(())
+		loop {
+			tokio::select! {
+				res = reconnect.closed() => return res.map_err(|err| Error::Connect(Arc::new(err))),
+				next = reconnect.connected(epoch) => {
+					epoch = next;
+					// "Connected" callback: copy from slab if not revoked.
+					if let Some(Some(entry)) = State::lock().session.task.get(task_id) {
+						entry.callback.call(());
```
kixelated: We should fire the callback with the epoch?

kixelated: Or only fire this once? IDK if it would be helpful to know there was a reconnect.

rs/moq-native/src/reconnect.rs
```diff
@@ -66,24 +66,32 @@ impl Default for Backoff {
 pub struct Reconnect {
 	abort: tokio::task::AbortHandle,
 	closed_rx: tokio::sync::watch::Receiver<Option<Arc<anyhow::Error>>>,
+	connected_rx: tokio::sync::watch::Receiver<u64>,
 }
 
 impl Reconnect {
 	pub(crate) fn new(client: Client, url: Url, backoff: Backoff) -> Self {
 		let (closed_tx, closed_rx) = tokio::sync::watch::channel(None::<Arc<anyhow::Error>>);
+		let (connected_tx, connected_rx) = tokio::sync::watch::channel(0u64);
```
kixelated: Can you use conducer instead? And make a poll_closed method?

rs/moq-native/src/reconnect.rs
```diff
@@ -117,6 +126,24 @@ impl Reconnect {
...
 
+	/// Wait for the next successful (re)connect after the given epoch.
+	///
+	/// The epoch counts successful connects: it is 1 after the first connect,
+	/// 2 after the first reconnect, and so on. Pass the previously returned
+	/// value to wait for a newer connection, or 0 to wait for the first.
+	///
+	/// Never resolves once the reconnect loop has stopped, so pair it with
+	/// [`closed`](Self::closed) in a `select!` to observe termination.
+	pub async fn connected(&self, since: u64) -> u64 {
```
kixelated: Can you also make a `disconnected` method that is similar? Closed would still be the only fatal signal.

rs/libmoq/src/session.rs
kixelated: Done — chained `.reconnect(url)` directly onto the `ClientConfig::default().init()?.with_publish(..).with_consume(..)` builder and dropped the intermediate `client` binding. (Written by Claude)

rs/libmoq/src/session.rs
kixelated: Done — the callback now fires the connection epoch as a positive code: 1 on the first connect, 2+ on each reconnect. So a caller can both treat "any positive" as connected and tell whether it was a reconnect (epoch > 1); if it only cares about the first connect it can ignore the rest. Transient disconnects fire 0, and a terminal give-up fires a negative code. (Written by Claude)

rs/moq-native/src/reconnect.rs
kixelated: Done — `Reconnect` is now built on `conducer` instead of `tokio::sync::watch`. I re-exported `conducer` from moq-net (`pub use conducer`), since its public poll_* API already exposes `conducer::Waiter` and callers couldn't name it otherwise; moq-native now uses `moq_net::conducer` rather than a direct dep. Added `poll_closed` plus `poll_connected`/`poll_disconnected`, each with an async wrapper via `conducer::wait`, following the poll_* convention in group.rs/track.rs. (Written by Claude)

rs/moq-native/src/reconnect.rs
kixelated: Done — added `Reconnect::disconnected(since)` (and `poll_disconnected`), analogous to `connected`. It counts session drops and is non-fatal: the loop keeps reconnecting. `closed()` remains the only terminal signal (Err on backoff-timeout give-up, Ok on close/drop). (Written by Claude)

rs/libmoq/src/session.rs
```diff
@@ -49,30 +49,60 @@ impl Session {
...
-		if let Some(Some(entry)) = State::lock().session.task.get(task_id) {
-			entry.callback.call(());
+		loop {
+			tokio::select! {
+				res = reconnect.closed() => return res.map_err(|err| Error::Connect(Arc::new(err))),
+				epoch = reconnect.connected(connects) => {
+					connects = epoch;
+					// Positive status carries the connection epoch, so callers can tell a
+					// reconnect (>1) from the first connect (1).
+					Self::notify(task_id, i32::try_from(epoch).unwrap_or(i32::MAX));
```
kixelated: IDK throw an error if this happens.

rs/moq-native/src/reconnect.rs
```diff
@@ -59,31 +61,52 @@ impl Default for Backoff {
...
+/// Shared reconnect state, observed by consumers through a [`conducer`] channel.
+///
+/// The channel closing (all producers dropped) is the terminal signal; `error`
+/// distinguishes a permanent give-up from a graceful close.
+#[derive(Default)]
+struct State {
+	/// Incremented on each successful connect: 1 after the first, 2 after the first reconnect, etc.
+	connects: u64,
+	/// Incremented each time an established session drops.
+	disconnects: u64,
```
kixelated: Is there any way we could use the same epoch? I would want connected and disconnected to be aligned.

ex. 
connected(0) waits until the first connection is established
disconnected(0) waits until the first connection is lost
connected(1) waits until the second connection is established
disconnected(1) waits until the second connection is lost
...etc

I'm not sure how this would work though.

Maybe there should be a single method that returns an enum instead, and keeps track of the last epoch/result returned?

kixelated: Done — added `Reconnect::status(&mut self) -> Result<Status>` exactly as you sketched: a single method with an internal cursor that returns `Status::Connected`/`Status::Disconnected` in strict alternating order (Err = terminal give-up). Internally it's one transition counter (odd = connected, even = disconnected). This replaces the separate connected()/disconnected(), so they're inherently aligned and can't be reordered. libmoq just loops on it. (Written by Claude)

rs/libmoq/src/session.rs
kixelated: Done — on the (practically unreachable) i32 epoch overflow, connect_run now returns a terminal `Error::Connect` instead of clamping to i32::MAX. (Written by Claude)

rs/moq-native/src/reconnect.rs
```diff
@@ -117,19 +146,84 @@ impl Reconnect {
...
+	/// [`poll_closed`](Self::poll_closed).
+	pub fn poll_connected(&self, waiter: &conducer::Waiter, since: u64) -> Poll<u64> {
+		match self.state.poll(waiter, |state| match state.connects {
+			connects if connects > since => Poll::Ready(connects),
+			_ => Poll::Pending,
+		}) {
+			Poll::Ready(Ok(connects)) => Poll::Ready(connects),
+			// Loop stopped: let `closed` deliver the outcome instead.
+			Poll::Ready(Err(_)) | Poll::Pending => Poll::Pending,
+		}
```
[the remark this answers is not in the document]
kixelated: The old poll_connected is gone (replaced by status()). It wasn't actually dropping epochs: conducer's poll evaluates the predicate before the closed check under the same lock, and a write() bump wakes the waiter, so a recorded transition is always delivered via the ready path; the closed branch only fires when epoch <= since (nothing newer). status() relies on the same property. (Written by Claude)

rs/libmoq/src/session.rs
```diff
@@ -49,30 +49,60 @@ impl Session {
...
+				}
+				epoch = reconnect.disconnected(disconnects) => {
+					disconnects = epoch;
+					// Status 0: transiently disconnected, reconnect in progress.
+					Self::notify(task_id, 0);
+				}
+			}
```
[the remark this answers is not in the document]
kixelated: Fixed — replaced the two-branch select! with a single ordered status() that returns transitions from an internal cursor, so connect/disconnect notifications can no longer be reordered. (Written by Claude)

rs/libmoq/src/session.rs
```diff
@@ -40,39 +40,73 @@ impl Session {
...
+				moq_native::Status::Connected => {
+					connects += 1;
+					// Positive status carries the connection epoch, so callers can tell a
+					// reconnect (>1) from the first connect (1).
+					let code = i32::try_from(connects)
+						.map_err(|_| Error::Connect(Arc::new(anyhow::anyhow!("connection epoch exceeded i32::MAX"))))?;
+					Self::notify(task_id, code);
+				}
+				// Status 0: transiently disconnected, reconnect in progress.
+				moq_native::Status::Disconnected => Self::notify(task_id, 0),
```
kixelated: I think we avoid notifying, because another PR is making 0 mean `closed`. Just ignore disconnected for now.

rs/moq-native/src/reconnect.rs
```diff
@@ -59,31 +61,63 @@ impl Default for Backoff {
...
+/// Shared reconnect state, observed by consumers through a [`conducer`] channel.
+///
+/// The channel closing (all producers dropped) is the terminal signal; `error`
+/// distinguishes a permanent give-up from a graceful close.
+#[derive(Default)]
+struct State {
+	/// Number of connect/disconnect transitions so far. Connects and disconnects strictly
+	/// alternate starting with a connect, so odd values are [`Status::Connected`] and even
+	/// values are [`Status::Disconnected`].
+	epoch: u64,
```
kixelated: IDK just store the current status, and last reported status, as Option<Status>. When they're != then return from async status()

kixelated: Done — State now holds the current status as `Option<Status>` and each handle tracks its `last_reported: Option<Status>`. async `status()` returns once they differ, then updates last_reported. Also made `Reconnect` `#[derive(Clone)]` (abort wrapped in `Arc<AbortOnDrop>`, so the loop stops on last-handle drop). (Written by Claude)

rs/libmoq/src/session.rs
kixelated: Done — libmoq no longer notifies on disconnect; it still consumes the Disconnected status internally (to detect the next reconnect) but doesn't call the callback. So only positive (connected epoch) and negative (terminal) are reported, leaving 0 free for "closed". (Written by Claude)

rs/moq-native/src/reconnect.rs
```diff
@@ -117,19 +159,75 @@ impl Reconnect {
...
+		})
+		.await;
+
+		if !advanced {
+			// Channel closed: surface the terminal error (or a generic one for a graceful stop).
+			return Err(self
+				.outcome()
+				.err()
+				.unwrap_or_else(|| anyhow::anyhow!("reconnect stopped")));
+		}
```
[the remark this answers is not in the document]
kixelated: The epoch design this referenced is gone. status() now tracks the current status vs the handle's last-reported one. conducer's poll evaluates the predicate before the closed check under the same lock, so if the status differs from last-reported it's returned (Ok) even as the channel closes; the terminal Err only fires when there's no pending change. The model is intentionally level-triggered now (a flip-and-flip-back collapses), per maintainer guidance. (Written by Claude)

rs/libmoq/src/session.rs
```diff
@@ -40,39 +40,73 @@ impl Session {
...
 
-		// "Connected" callback — copy from slab if not revoked.
-		if let Some(Some(entry)) = State::lock().session.task.get(task_id) {
-			entry.callback.call(());
+		// status() reports connection changes and Err on terminal give-up, so the ? both ends the
+		// loop and propagates the negative status code.
+		let mut connects: u64 = 0;
+		loop {
+			match reconnect.status().await.map_err(|err| Error::Connect(Arc::new(err)))? {
+				moq_native::Status::Connected => {
```
kixelated:
```
while let moq_native::Status::Connected = reconnect.status().await? {
  connects += 1;
  ...
}
```

I think you move this to another function that returns Result so we can use ? easier.

rs/moq-native/src/reconnect.rs
```diff
@@ -59,31 +61,72 @@ impl Default for Backoff {
...
+/// Spawns a tokio task that connects, waits for session close, then reconnects with exponential
+/// backoff. [`status`](Self::status) reports connection changes; [`closed`](Self::closed) waits for
+/// the loop to stop. Cheaply cloneable for multiple handles; the loop stops once the last handle
+/// is dropped.
+#[derive(Clone)]
 pub struct Reconnect {
-	abort: tokio::task::AbortHandle,
-	closed_rx: tokio::sync::watch::Receiver<Option<Arc<anyhow::Error>>>,
+	#[allow(dead_code)] // Held for its Drop: aborts the task when the last handle goes.
+	abort: Arc<AbortOnDrop>,
```
kixelated: Let's actually move the Clone and AbortOnDrop stuff to another PR. `Drop` for now.

rs/moq-native/src/reconnect.rs
```diff
@@ -117,31 +166,72 @@ impl Reconnect {
...
-			Ok(v) => {
-				let err = Arc::clone(v.as_ref().expect("predicate matched Some"));
-				Err(anyhow::anyhow!("{err:#}"))
+	/// Returns the current [`Status`] (`Connected` or `Disconnected`). Since the loop alternates
+	/// between the two, successive calls alternate as well, but a status that flips and flips back
+	/// before the caller polls is reported once: this tracks the *current* state, not every edge.
+	///
+	/// Returns `Err` once the loop has stopped: the give-up error if it exhausted its backoff
+	/// timeout, or a generic "stopped" error after the last handle is dropped.
+	pub async fn status(&mut self) -> anyhow::Result<Status> {
```
kixelated: make `poll_status` please.

rs/moq-native/src/reconnect.rs
```diff
@@ -117,31 +166,72 @@ impl Reconnect {
...
+		let consumer = self.state.clone();
+		let current = conducer::wait(|waiter| {
+			match consumer.poll(waiter, |state| match state.status {
+				Some(status) if Some(status) != last => Poll::Ready(status),
+				_ => Poll::Pending,
+			}) {
+				Poll::Ready(Ok(status)) => Poll::Ready(Some(status)),
+				Poll::Ready(Err(_)) => Poll::Ready(None),
+				Poll::Pending => Poll::Pending,
+			}
```
kixelated: Move this to poll_status.

You can use `?` and `ready!` to simplify this logic, see moq-net.

rs/moq-native/src/reconnect.rs
```diff
@@ -117,31 +166,72 @@ impl Reconnect {
...
+			Some(status) => {
+				self.last_reported = Some(status);
+				Ok(status)
 			}
-			Err(_) => Ok(()),
+			// Channel closed: surface the terminal error (or a generic one for a graceful stop).
+			None => Err(self
+				.outcome()
+				.err()
+				.unwrap_or_else(|| anyhow::anyhow!("reconnect stopped"))),
```
kixelated: .context

rs/moq-native/src/reconnect.rs
```diff
@@ -117,31 +166,72 @@ impl Reconnect {
...
 
-	/// Stop the background reconnect loop.
-	pub fn close(self) {
-		self.abort.abort();
+	/// Poll whether the reconnect loop has stopped.
+	///
+	/// Ready with `Err` if it permanently gave up (reconnect timeout exceeded),
+	/// or `Ok(())` if stopped by dropping the last handle.
+	pub fn poll_closed(&self, waiter: &conducer::Waiter) -> Poll<anyhow::Result<()>> {
+		match self.state.poll_closed(waiter) {
```
kixelated: Can this be written simpler?

rs/moq-native/src/reconnect.rs
```diff
@@ -117,31 +166,72 @@ impl Reconnect {
...
+	/// Returns `Err` with the most recent connection error if the reconnect
+	/// timeout was exceeded.
+	pub async fn closed(&self) -> anyhow::Result<()> {
+		conducer::wait(|waiter| self.poll_closed(waiter)).await
+	}
+
+	/// Read the terminal outcome; only meaningful once the channel has closed.
+	fn outcome(&self) -> anyhow::Result<()> {
+		match self.state.read().error.clone() {
+			Some(err) => Err(anyhow::anyhow!("{err:#}")),
```
kixelated: why ????

rs/libmoq/src/session.rs
kixelated: Done — moved the loop into `report(task_id, reconnect) -> anyhow::Result<()>`, so `status().await?` and `.context()?` flow directly; connect_run maps the terminal error to Connect once. I kept a `loop { if let Connected = ...? }` rather than `while let`, since while-let would break on the first Disconnected and stop reconnecting. (Written by Claude)

rs/moq-native/src/reconnect.rs
kixelated: Done — reverted to a single-handle Reconnect with a plain `Drop` that aborts the task. Clone/AbortOnDrop deferred to a follow-up PR. (Written by Claude)

rs/moq-native/src/reconnect.rs
kixelated: Done — added `poll_status(&mut self, waiter) -> Poll<Result<Status>>`; `status()` is now just `kio::wait(|w| self.poll_status(w))`. (Written by Claude)

rs/moq-native/src/reconnect.rs
kixelated: Done — the change-detection logic now lives in poll_status, using `ready!` and an early `return` on the closed Err (mirroring moq-net's poll_* style). (Written by Claude)

rs/moq-native/src/reconnect.rs
kixelated: Addressed — the `.err().unwrap_or_else(|| anyhow!(...))` is gone (status() now returns the terminal error straight from poll_status). I used `.context()` where it now fits: the i32 epoch-overflow in libmoq's report(). (Written by Claude)

rs/moq-native/src/reconnect.rs
kixelated: Done — poll_closed is now `ready!(self.state.poll_closed(waiter)); Poll::Ready(match &self.state.read().error { ... })`. (Written by Claude)

rs/moq-native/src/reconnect.rs
kixelated: Dropped the Arc — the error is stored as `Option<anyhow::Error>` directly now. The `{err:#}` format remains because anyhow::Error isn't Clone and we only hold a shared ref to it (multiple handles may read the terminal error), so we surface a flattened copy. (Written by Claude)

PR #1571, opened 2026-05-31
moq-relay: add --cluster-connect-api and split cluster identity from gossip

kixelated:
## Summary

Adds `--cluster-connect-api` so a relay can fetch its cluster peer list from an external source and reconcile dials at runtime, without a redeploy. The relay stays topology-agnostic: it dials whatever hostnames the source returns, so all routing decisions (which node connects where) live in whatever service answers the endpoint.

Also splits cluster identity from the gossip toggle, which `--cluster-mesh` previously conflated.

### `--cluster-connect-api`
- Value is an `http(s)` URL or a local file (a path or `file://` URL).
- Returns a bare JSON array of peer hostnames: `["a.pop.example", "b.pop.example"]`.
- Reconciles against live dials: new peers dialed, dropped peers aborted. Composes with static `--cluster-connect` (never reconciled away) and gossip.
- HTTP: polled, with the re-poll cadence following `Cache-Control` (`max-age` + `stale-while-revalidate`). Conditional revalidation (ETag / Last-Modified) and stale-if-error come from the shared HTTP cache client.
- File: re-read when its mtime changes.
- Fails static: a failed fetch keeps the last good list, so a control-plane blip never tears down the cluster.
- The relay's `--cluster-node` value is sent as `?node=`, and the cluster mTLS client cert identifies the caller, so the endpoint can return per-node peer lists.

### Identity vs gossip
- `--cluster-node <url>` (re-added): this relay's own URL / identity. Used for `?node=` and as the address gossip advertises.
- `--cluster-mesh` is now a boolean gossip toggle (it previously took a URL). Enabling it without `--cluster-node` errors at startup. This is a breaking change to a recently-added flag (`#1504`).

### Other
- New `http_client` module factored out of `auth.rs`, reused for the peer-list poller.
- Docs: `doc/bin/relay/cluster.md`, `doc/bin/relay/config.md`.

## Follow-ups (not in this PR)
- The `/cluster/connect` endpoint itself (proprietary routing over the node inventory) lives in moq-pro.
- True non-blocking stale-while-revalidate: `http-cache` does not parse the SWR directive, so the cadence honors it but a stale fetch still revalidates synchronously on the poll task. Tracked as a follow-up.

## Test plan
- [x] `cargo test -p moq-relay` (97 pass; new tests cover DialMap API reconcile, config round-trip, SWR cadence parsing, gossip-requires-node)
- [x] `cargo clippy -p moq-relay --all-targets` clean
- [ ] Manual: `--cluster-connect-api file:///tmp/peers.json`, edit the file, confirm dials reconcile; HTTP endpoint with `Cache-Control` + a 5xx to confirm fail-static

(Written by Claude)

### DIFF
--- a/doc/bin/relay/cluster.md
+++ b/doc/bin/relay/cluster.md
@@ -35,17 +35,41 @@ Pick the shape that matches your traffic. Linear chains are great for fanout; sm
 
 ## Auto-discovery
 
-Listing every peer by hand can get tedious in larger clusters. Set `cluster.mesh` to this relay's own URL and connected peers will discover and dial it back automatically:
+Listing every peer by hand can get tedious in larger clusters. Tell the relay its own URL with `cluster.node`, then enable gossip with `cluster.mesh`; connected peers will discover and dial it back automatically:
 
 ```toml
 [cluster]
 connect = ["us-east.example.com:4443"]
-mesh    = "us-west.example.com:4443"
+node    = "us-west.example.com:4443"
+mesh    = true
 ```
 
-Each node with `mesh` set creates a broadcast carrying its address, which other nodes pick up. `connect` is optional once gossip is running, but you still need at least one connection somewhere (either you dial a peer or a peer dials you) for the advertisement to flow.
+`node` is this relay's identity (its externally-reachable URL); `mesh` is a boolean that turns gossip on. Each gossiping node creates a broadcast carrying its `node` address, which other nodes pick up. `connect` is optional once gossip is running, but you still need at least one connection somewhere (either you dial a peer or a peer dials you) for the advertisement to flow. Enabling `mesh` without `node` is an error, since there'd be no address to advertise.
 
-A relay with `mesh` set and no `connect` is a passive rendezvous: it sits and waits for inbound connections, then helps everyone else find each other.
+A relay with `node` + `mesh` and no `connect` is a passive rendezvous: it sits and waits for inbound connections, then helps everyone else find each other.
+
+## Dynamic peer lists
+
+`cluster.connect` is fixed at startup, so adding or removing a node means editing every affected config and restarting. When you'd rather keep the topology somewhere external and change it without a redeploy, point `cluster.connect_api` at an HTTP(S) endpoint or a local file:
+
+```toml
+[cluster]
+connect_api = "https://api.example.com/cluster/connect"
+mesh        = "us-west.example.com:4443"
+```
+
+The source returns a bare JSON array of peer hostnames:
+
+```json
+["eu-west.example.com:4443", "us-east.example.com:4443"]
+```
+
+The relay reconciles that list against its live dials: new entries are dialed, entries that disappear are dropped. It composes with `connect` (static seeds that are never reconciled away) and `mesh` (gossip). The relay's own `node` value, when set, is sent as a `?node=` query parameter so the endpoint can return the peers for that specific node; for mTLS-gated endpoints the cluster client certificate identifies the caller as well.
+
+- **HTTP(S) URL**: polled, with the re-poll cadence taken from the response's `Cache-Control` reuse window (`max-age` plus any `stale-while-revalidate`; default 30s, floored at 5s). Conditional revalidation (`ETag` / `Last-Modified`) and stale-if-error (serving the last cached body when a revalidation request fails) are handled by the underlying HTTP cache, so transient endpoint blips don't churn the dial set.
+- **Local file** (a path or `file://` URL): re-read whenever its mtime changes.
+
+If a fetch fails or returns garbage, the relay logs and keeps the last good list rather than tearing the cluster down. This keeps the moq-relay binary generic: all routing decisions (which node connects where) live in whatever service answers the endpoint.
 
 ## Authentication
 
@@ -58,13 +82,13 @@ See [Authentication](/bin/relay/auth) for the full setup.
 
 ## Migration from older configs
 
-`cluster.root` and `cluster.node` were both removed. If a config still sets either flag, the relay errors at startup with a message pointing at the replacements:
+`cluster.root` was removed; setting it errors at startup with a message pointing at the replacement. `cluster.mesh` is now a boolean gossip toggle (it used to take this relay's URL); the URL moved to `cluster.node`.
 
 | Old | New |
 |---|---|
-| `root = "rendezvous:4443"` + `node = "us-east:4443"` | `connect = ["rendezvous:4443"]` + `mesh = "us-east:4443"` |
-| `root = "rendezvous:4443"` only | `mesh = "rendezvous:4443"` (passive rendezvous) |
-| `node = "us-east:4443"` | `mesh = "us-east:4443"` |
+| `root = "rendezvous:4443"` + `node = "us-east:4443"` | `connect = ["rendezvous:4443"]` + `node = "us-east:4443"` + `mesh = true` |
+| `root = "rendezvous:4443"` only | `node = "rendezvous:4443"` + `mesh = true` (passive rendezvous) |
+| `mesh = "us-east:4443"` | `node = "us-east:4443"` + `mesh = true` |
 
 ## Next steps
 
--- a/doc/bin/relay/cluster.md
+++ b/doc/bin/relay/cluster.md
@@ -84,7 +84,7 @@ See [Authentication](/bin/relay/auth) for the full setup.
 
 ## Migration from older configs
 
-`cluster.root` was removed; setting it errors at startup with a message pointing at the replacement. `cluster.mesh` is now a boolean gossip toggle (it used to take this relay's URL); the URL moved to `cluster.node`.
+`cluster.root` was removed; setting it errors at startup with a message pointing at the replacement. `cluster.mesh` is now a boolean gossip toggle (it used to take this relay's URL); the URL moved to `cluster.node`. The old `mesh = "<url>"` form still works for backwards compatibility: it enables gossip and is treated as `cluster.node`, with a deprecation warning (or an error if it conflicts with an explicit `cluster.node`).
 
 | Old | New |
 |---|---|
--- a/doc/bin/relay/cluster.md
+++ b/doc/bin/relay/cluster.md
@@ -67,7 +67,7 @@ The source returns a bare JSON array of peer hostnames:
 The relay reconciles that list against its live dials: new entries are dialed, entries that disappear are dropped. It composes with `connect` (static seeds that are never reconciled away) and `mesh` (gossip). The relay's own `node` value, when set, is sent as a `?node=` query parameter so the endpoint can return the peers for that specific node; for mTLS-gated endpoints the cluster client certificate identifies the caller as well.
 
 - **HTTP(S) URL**: polled, with the re-poll cadence taken from the response's `Cache-Control` reuse window (`max-age` plus any `stale-while-revalidate`; default 30s, floored at 5s). Conditional revalidation (`ETag` / `Last-Modified`) and stale-if-error (serving the last cached body when a revalidation request fails) are handled by the underlying HTTP cache, so transient endpoint blips don't churn the dial set.
-- **Local file** (a path or `file://` URL): re-read whenever its mtime changes.
+- **Local file** (a path or `file://` URL): watched via OS filesystem notifications (inotify / FSEvents / kqueue), with a periodic re-check as a safety net.
 
 If a fetch fails or returns garbage, the relay logs and keeps the last good list rather than tearing the cluster down. This keeps the moq-relay binary generic: all routing decisions (which node connects where) live in whatever service answers the endpoint.
 
--- a/rs/moq-relay/src/cluster.rs
+++ b/rs/moq-relay/src/cluster.rs
@@ -1,5 +1,5 @@
 use std::{
-	collections::HashMap,
+	collections::{HashMap, HashSet},
 	path::PathBuf,
 	sync::{Arc, Mutex},
 	time::{Duration, Instant},
@@ -7,6 +7,7 @@ use std::{
 
 use anyhow::Context;
 use moq_net::{BroadcastProducer, Origin, OriginConsumer, OriginProducer, Path, Stats, Tier};
+use reqwest_middleware::ClientWithMiddleware;
 use tokio::task::AbortHandle;
 use url::Url;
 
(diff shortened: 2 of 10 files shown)


rs/moq-relay/src/cluster.rs
```diff
@@ -24,24 +25,37 @@ const SWEEP_INTERVAL: Duration = Duration::from_secs(30);
...
-	peer > self_url
+/// Re-poll cadence for `--cluster-connect-api` when the response carries no
+/// `Cache-Control` reuse window (`max-age` / `stale-while-revalidate`).
+const CONNECT_API_DEFAULT_INTERVAL: Duration = Duration::from_secs(30);
+
+/// Floor on the `--cluster-connect-api` poll cadence, so a tiny (or zero) reuse
+/// window can't spin the relay into a tight fetch loop.
+const CONNECT_API_MIN_INTERVAL: Duration = Duration::from_secs(5);
+
+/// How often a local `--cluster-connect-api` file is stat'd for mtime changes.
```
kixelated: Is there some library that automatically does this for us, using OS functionality (ex. inotify?)

rs/moq-relay/src/cluster.rs
```diff
@@ -24,24 +25,37 @@ const SWEEP_INTERVAL: Duration = Duration::from_secs(30);
...
-/// registration goes away (operator intent says "always dial"). Gossip-discovered
-/// peers carry an `unannounced_at` timestamp that the periodic sweep uses to
-/// decide when a peer has truly left vs. is just flapping between paths.
+/// One entry in [`DialMap`]. Gossip-discovered peers carry an `unannounced_at`
+/// timestamp that the periodic sweep uses to decide when a peer has truly left
+/// vs. is just flapping between paths.
 struct DialEntry {
 	handle: AbortHandle,
-	is_static: bool,
+	source: DialSource,
```
kixelated: Does this need to support Gossip and API? I think both are possible if you use --cluster-connect-api and --cluster-mesh? Should be two booleans right?

rs/moq-relay/src/cluster.rs
```diff
@@ -307,25 +382,42 @@ impl Cluster {
 					tracing::warn!(%err, peer = %peer_for_task, "cluster peer connection ended");
 				}
 			});
-			dialed.insert(peer, handle, true);
+			dialed.insert(peer, handle, DialSource::Static);
+		}
+
+		if let Some(source) = self.config.connect_api.clone() {
+			let tls = self
```
kixelated: This is only needed if HTTPS URLs are used.

rs/moq-relay/src/cluster.rs
```diff
@@ -307,25 +382,42 @@ impl Cluster {
...
+		let _self_registration: Option<BroadcastProducer> = if gossip {
+			// Checked above: gossip requires `node`.
+			let node = self.config.node.as_deref().expect("gossip requires --cluster-node");
+			let path = Path::new(MESH_PREFIX).join(node);
 			let broadcast = self
 				.origin
 				.create_broadcast(&path)
 				.expect(".internal/origins is within the relay origin's root");
-			tracing::info!(url = %mesh, %path, "advertising cluster mesh URL");
+			tracing::info!(url = %node, %path, "advertising cluster node URL");
```
kixelated: Remove the `url = `

rs/moq-relay/src/cluster.rs
kixelated: Yes — switched to the `notify` crate (inotify on Linux, FSEvents on macOS, kqueue on BSD). It watches the parent directory rather than the file itself, since editors and `mv` replace the file by atomic rename (new inode), which would drop a watch set on the file. A periodic re-check stays as a safety net for missed events / network filesystems, and as the sole mechanism if a watcher can't be created. (Written by Claude)

rs/moq-relay/src/cluster.rs
kixelated: Good catch — a peer can be wanted by both gossip and the API list at once (e.g. `--cluster-mesh` + `--cluster-connect-api`). Replaced the single `source` enum with a `DialSources` set (static/gossip/api booleans). A dial is kept until the *last* source releases it: the gossip stale-sweep and the API reconcile each clear only their own bit and abort only when nothing else wants the peer. `insert` now merges onto an existing entry and aborts the redundant handle, so two sources discovering the same peer concurrently can't leak a dial. Added tests for the multi-source survive/teardown paths. (Written by Claude)

rs/moq-relay/src/cluster.rs
kixelated: Fixed — the TLS requirement and HTTP cache-client build now happen inside the http(s) branch of `run_connect_api`. A file/`file://` source no longer needs (or builds) a client TLS config. `run` only asserts client TLS is present when the source parses as an http(s) URL (`connect_api_is_http`). (Written by Claude)

rs/moq-relay/src/cluster.rs
kixelated: Done. (Written by Claude)

doc/bin/relay/cluster.md
```diff
@@ -35,17 +35,41 @@ Pick the shape that matches your traffic. Linear chains are great for fanout; sm
...
+
+## Dynamic peer lists
+
+`cluster.connect` is fixed at startup, so adding or removing a node means editing every affected config and restarting. When you'd rather keep the topology somewhere external and change it without a redeploy, point `cluster.connect_api` at an HTTP(S) endpoint or a local file:
+
+```toml
+[cluster]
+connect_api = "https://api.example.com/cluster/connect"
+mesh        = "us-west.example.com:4443"
+```
```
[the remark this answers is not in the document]
kixelated: Already fixed in the merge commit (db2ad6a): the example now uses `node = "us-west.example.com:4443"` instead of the legacy `mesh = "..."`. (Written by Claude)

rs/moq-relay/src/cluster.rs
```diff
@@ -297,25 +382,42 @@ impl Cluster {
...
-			dialed.insert(peer, handle, true);
+			dialed.insert(peer, handle, DialSource::Static);
+		}
+
+		if let Some(source) = self.config.connect_api.clone() {
+			let tls = self
+				.client_tls
+				.clone()
+				.context("cluster.connect_api set but no client TLS attached (call Cluster::with_client_tls)")?;
+			let http = crate::http_client::build(&tls)?;
```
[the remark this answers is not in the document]
kixelated: Fixed in 947b239: the `client_tls` requirement and HTTP cache-client build moved into the http(s) branch of `run_connect_api`. `run` only asserts client TLS when the source parses as an http(s) URL (`connect_api_is_http`); a local file source needs no TLS. (Written by Claude)

rs/moq-relay/src/cluster.rs
```diff
@@ -394,6 +496,147 @@ impl Cluster {
...
+					},
+					Err(err) => {
+						tracing::warn!(%err, ?path, "failed to read cluster.connect_api file; keeping current peers")
+					}
+				},
+				Ok(_) => {} // unchanged
+				Err(err) => tracing::warn!(%err, ?path, "cluster.connect_api file unavailable; keeping current peers"),
+			}
+
+			tokio::time::sleep(CONNECT_API_FILE_INTERVAL).await;
```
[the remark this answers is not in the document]
kixelated: Good catch — the 'addressed' badge was a false positive; the notify rewrite kept the same bug. Fixed in 72509b4a: `reload_connect_api_file` now advances `last_seen_mtime` as soon as a new mtime is observed (not only on successful parse), so a malformed file at a constant mtime is attempted once per real edit instead of every tick. The last-good list is already retained in the dial map. (Written by Claude)

rs/moq-relay/src/cluster.rs
```diff
@@ -34,14 +48,59 @@ fn should_dial(self_url: &str, peer: &str) -> bool {
...
+	Gossip,
+	/// Supplied by `--cluster-connect-api`. Released when a fetched peer list no
+	/// longer contains the peer.
+	Api,
+}
+
+/// The set of [`DialSource`]s currently keeping a dial alive.
+#[derive(Clone, Copy, Default)]
+struct DialSources {
+	static_: bool,
```
kixelated: Maybe find another word that isn't a keyword

rs/moq-relay/src/cluster.rs
```diff
@@ -59,66 +118,123 @@ impl DialMap {
...
-		);
+	/// Record a freshly-spawned dial for `peer` under `source`. If `peer` is
+	/// already dialed, add `source` to its set and abort the redundant `handle`
+	/// (the existing dial stands, since dialing dedupes by URL). Always spawn the
+	/// task first, then call this: it resolves the "two sources discover the same
+	/// peer at once" race without leaking a task.
+	fn insert(&self, peer: String, handle: AbortHandle, source: DialSource) {
+		let mut map = self.inner.lock().expect("dial map poisoned");
+		match map.get_mut(&peer) {
+			Some(entry) => {
```
kixelated: `if let` is nicer IMO

rs/moq-relay/src/cluster.rs
```diff
@@ -140,21 +256,50 @@ pub struct ClusterConfig {
...
+	#[arg(
+		id = "cluster-mesh",
+		long = "cluster-mesh",
+		env = "MOQ_CLUSTER_MESH",
+		default_missing_value = "true",
+		num_args = 0..=1,
+		require_equals = true,
+		value_parser = clap::value_parser!(bool),
+	)]
+	pub mesh: Option<bool>,
```
kixelated: Maybe we make this a String for backwards compatibility? `true` or `false` are valid, while anything else emits a warning (or error if it clashes with cluster-node

rs/moq-relay/src/cluster.rs
```diff
@@ -408,6 +586,206 @@ impl Cluster {
...
+
+	/// Watch a local peer-list file, reconciling whenever it changes. Prefers OS
+	/// filesystem notifications (inotify / FSEvents / kqueue via `notify`), with a
+	/// periodic re-check as a safety net for missed events and network filesystems.
+	/// Fails static: a missing or malformed file keeps the current dials.
+	async fn run_connect_api_file(&self, path: PathBuf, node: Option<String>, token: String, dialed: DialMap) {
+		let mut last_seen_mtime = None;
+		self.reload_connect_api_file(&path, &mut last_seen_mtime, &node, &token, &dialed);
+
+		let mut tick = tokio::time::interval(CONNECT_API_FILE_INTERVAL);
```
kixelated: Do we really need this? IDK I think we can just reload on the `recommended_watcher` callback instead of all this async/timer nonsense. It would be crazy if `notify` didn't default to a poll mode or something.

rs/moq-relay/src/cluster.rs
```diff
@@ -408,6 +586,206 @@ impl Cluster {
...
+		token: &str,
+		dialed: &DialMap,
+	) {
+		let mtime = match std::fs::metadata(path).and_then(|m| m.modified()) {
+			Ok(mtime) => mtime,
+			Err(err) => {
+				tracing::warn!(%err, ?path, "cluster.connect_api file unavailable; keeping current peers");
+				return;
+			}
+		};
```
kixelated: let Ok(mtime) = ... else { return };

Add a note to claude.md that `Ok(v) => v` or `Some(v) => v` generally means there's a cleaner alternative using `if let` and friends. I think `match` on an Option is always cleaner as an `if let` instead?

rs/moq-relay/src/cluster.rs
```diff
@@ -59,66 +118,123 @@ impl DialMap {
...
+				tracing::debug!(%peer, "peer dropped from cluster-connect-api; still wanted by another source");
 				true
+			} else {
+				tracing::info!(%peer, "peer dropped from cluster-connect-api; abandoning dial");
+				entry.handle.abort();
+				false
 			}
 		});
+
+		desired
```
kixelated, line 223: I think you could do this above with `retain_mut`?

rs/moq-relay/src/cluster.rs
kixelated: Renamed the field to `seeded` (the variant stays `DialSource::Static`, which is a valid identifier; only the lowercase struct field needed the trailing-underscore escape). (Written by Claude)

rs/moq-relay/src/cluster.rs
kixelated: Done — `insert` now uses `if let Some(entry) = map.get_mut(&peer) { ... } else { ... }`. (Written by Claude)

rs/moq-relay/src/cluster.rs
kixelated: Done. `mesh` is `Option<String>` again with a bool-or-string deserializer (so TOML `mesh = true` and `mesh = "..."` both work). `resolve_mesh()` maps `true`/`false` to the toggle; any other value is the legacy 'advertise this URL' form — it enables gossip and is used as `--cluster-node` with a deprecation warning, or errors if it conflicts with an explicit `--cluster-node`. Added tests for both the legacy-enable and the conflict paths. (Written by Claude)

rs/moq-relay/src/cluster.rs
kixelated: Agreed. Dropped the `select!`/interval machinery: the happy path is now just `while let Some(event) = events.recv().await { reload }`. (`notify`'s callback runs on its own thread off the tokio runtime, and reconciling spawns dial tasks, so a channel still bridges the callback into this async task — but no timer.) `recommended_watcher` picks the native backend and doesn't auto-poll, so I kept a simple poll loop as the fallback *only* when a watcher can't be created. (Written by Claude)

rs/moq-relay/src/cluster.rs
kixelated: Switched to `let Ok(mtime) = std::fs::metadata(path).and_then(|m| m.modified()) else { ... };`. Note: `let-else` drops the `Err` binding, so that branch logs without the error detail (the path is still logged). On the CLAUDE.md note: I tried to add it under Rust Conventions but the edit was blocked by the harness's self-modification guard (it won't let me edit CLAUDE.md unprompted). Could you add it, or approve the edit? Proposed text: 'Prefer `if let` / `let ... else` over a `match` whose only job is to unwrap (`Ok(v) => v` / `Some(v) => v`); keep `match` when both arms do real work or you need the Err/None payload.' (Written by Claude)

rs/moq-relay/src/cluster.rs
kixelated: Done — folded the API-bit handling into the single `map.retain(...)` pass (it already gives `&mut entry`): set the bit on listed peers, clear it on the rest (aborting only when nothing else wants them). The follow-up pass is now just `desired.iter().filter(|p| !map.contains_key(*p))` to collect peers needing a fresh dial. (Written by Claude)

rs/moq-relay/src/cluster.rs
kixelated: Added the CLAUDE.md note in bf17e2d9 (under Rust Conventions). (Written by Claude)

PR #1573, opened 2026-05-31
moq-lite-05: add AnnounceOk message (responder origin + initial active count)

kixelated:
## Summary

Adds an `AnnounceOk` message to the moq-lite-05 announce stream, sent once by the publisher right after it reads `AnnounceInterest` and before any `Announce`. It does two things:

1. **Reports the responder's origin id once** instead of stamping it onto the trailing hop of every `Announce`. That id is identical for every announce on a session, so re-encoding it per-message is pure waste. In Lite05 a node no longer stamps its *own* origin onto a chain; the **receiver** stamps the *remote* sender's origin (from `AnnounceOk`) on receipt. The stored hop chain is byte-identical to Lite04, so loop detection / shortest-path selection are unchanged.
2. **Reports `active: N`**, the count of currently-active broadcasts, followed by exactly `N` initial `Announce::Active`. This gives the announce stream a discrete initial-set boundary (the successor to `AnnounceInit`).

The count lets `connect()` **block until the initial set has landed**, closing the startup race where a synchronous `get_broadcast()` right after connect could miss broadcasts that were live but not yet gossiped. A `SyncLatch` fires once every announce-prefix stream has its initial set, generalized across both boundaries — `AnnounceInit` (Lite01/02) and `AnnounceOk + N` (Lite05); Lite03/04 have no boundary and resolve immediately. `OriginConsumer::announced_broadcast()` is **kept** — it still serves the distinct "wait for a broadcast that comes online *after* connect" case.

This is opt-in: `Lite05Wip` is not advertised over ALPN or in the default version set, so nothing negotiates it by default.

## Why

The trailing self-hop is redundant on every announce, and Lite03+ had no way to know when the initial set was complete. `AnnounceOk` fixes both with one message.

## JS mirror

`@moq/net` mirrors the wire: `AnnounceOk` message, no-self-hop send, and the subscriber re-appending the responder origin (so the `ignoreSelf` filter still sees the full chain). It also wires `ALPN_05_WIP` into `connect`/`accept`, which previously had **no** Lite05 negotiation branch at all.

JS does **not** add connect-blocking: its `Connection` is pull-based (`announced()` opens the announce stream lazily and `consume(path)` subscribes directly without consulting announcements), so the synchronous `get_broadcast()` race doesn't exist there.

## Notes for reviewers

- Targets `dev` per the wire-protocol branch convention. Rebased onto `dev`, which had substantially reworked these files (`AnnounceConsumer` split, `Session::new` gaining publisher/consumer args, restart/REANNOUNCE semantics, its own lite-05 additions). Integrated with all of those:
  - The publisher's self-stamp now lives in `prepare_active_hops`, made version-aware so Lite05 skips it.
  - Both `start_announce` and `restart_announce` take the responder origin and append it (with MAX_HOPS handling) on the receive side.
- The IETF draft / spec is being updated separately.

## Test plan

- [x] `AnnounceOk` round-trip / old-version reject / zero-active / zero-origin reject (rs)
- [x] `SyncLatch` unit tests: arm/complete, arm(0), guard-fires-on-drop, idempotent completion (rs)
- [x] Existing restart/REANNOUNCE tests still pass (336 moq-net tests, 93 moq-relay incl. smoke)
- [x] New `integration: lite draft-05-wip` end-to-end test exercises `AnnounceOk` over the wire (js)
- [x] `just rs check` (clippy + fmt + tests) and `just js check` (biome + tsc + tests) green

(Written by Claude)

### DIFF
--- a/js/net/src/lite/publisher.ts
+++ b/js/net/src/lite/publisher.ts
@@ -6,7 +6,7 @@ import * as Path from "../path.ts";
 import { type Stream, Writer } from "../stream.ts";
 import type { Track } from "../track.ts";
 import { error } from "../util/error.ts";
-import { Announce, AnnounceInit, type AnnounceInterest } from "./announce.ts";
+import { Announce, AnnounceInit, type AnnounceInterest, AnnounceOk } from "./announce.ts";
 import { Group as GroupMessage } from "./group.ts";
 import type { Origin } from "./origin.ts";
 import { Probe } from "./probe.ts";
@@ -106,6 +106,10 @@ export class Publisher {
 			active.add(suffix);
 		}
 
+		// Lite05+ reports our origin once via AnnounceOk; the subscriber stamps it
+		// onto each announce, so we no longer put it in the per-announce hop list.
+		const selfHops = this.version === Version.DRAFT_05_WIP ? [] : [this.origin];
+
 		switch (this.version) {
 			case Version.DRAFT_01:
 			case Version.DRAFT_02: {
@@ -113,10 +117,20 @@ export class Publisher {
 				await init.encode(stream.writer, this.version);
 				break;
 			}
-			default:
-				// Draft03+: send individual Announce messages for initial state.
+			case Version.DRAFT_05_WIP: {
+				// Report our origin id and the count of initial announces that follow.
+				const ok = new AnnounceOk(this.origin, active.size);
+				await ok.encode(stream.writer, this.version);
 				for (const suffix of active) {
-					const wire = new Announce({ suffix, active: true, hops: [this.origin] });
+					const wire = new Announce({ suffix, active: true, hops: selfHops });
+					await wire.encode(stream.writer, this.version);
+				}
+				break;
+			}
+			default:
+				// Draft03/04: send individual Announce messages for initial state.
+				for (const suffix of active) {
+					const wire = new Announce({ suffix, active: true, hops: selfHops });
 					await wire.encode(stream.writer, this.version);
 				}
 				break;
@@ -147,7 +161,7 @@ export class Publisher {
 			// Announce any new broadcasts.
 			for (const added of newActive.difference(active)) {
 				console.debug(`announce: broadcast=${added} active=true`);
-				const wire = new Announce({ suffix: added, active: true, hops: [this.origin] });
+				const wire = new Announce({ suffix: added, active: true, hops: selfHops });
 				await wire.encode(stream.writer, this.version);
 			}
 
--- a/js/net/src/lite/publisher.ts
+++ b/js/net/src/lite/publisher.ts
@@ -106,10 +106,6 @@ export class Publisher {
 			active.add(suffix);
 		}
 
-		// Lite05+ reports our origin once via AnnounceOk; the subscriber stamps it
-		// onto each announce, so we no longer put it in the per-announce hop list.
-		const selfHops = this.version === Version.DRAFT_05_WIP ? [] : [this.origin];
-
 		switch (this.version) {
 			case Version.DRAFT_01:
 			case Version.DRAFT_02: {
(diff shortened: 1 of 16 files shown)


js/net/src/lite/publisher.ts
```diff
@@ -106,17 +106,31 @@ export class Publisher {
 			active.add(suffix);
 		}
 
+		// Lite05+ reports our origin once via AnnounceOk; the subscriber stamps it
+		// onto each announce, so we no longer put it in the per-announce hop list.
+		const selfHops = this.version === Version.DRAFT_05_WIP ? [] : [this.origin];
```
kixelated: Is this needed? Just inline it.

rs/moq-net/src/lite/session.rs
```diff
@@ -1,9 +1,19 @@
...
+use super::{Publisher, PublisherConfig, Subscriber, SubscriberConfig, SyncLatch, Version};
+
+/// Start a lite session.
+///
+/// Returns the receive-bandwidth consumer (if any) and a `synced` receiver that
+/// resolves once the initial announce set has been inserted into the subscribe
+/// origin, letting `connect()` block past the startup race. The receiver fires
+/// immediately when there is nothing to wait on (no subscribe origin, or a version
+/// without an initial-set boundary).
 pub fn start<S: web_transport_trait::Session>(
```
kixelated, line 14: Can you make this async instead of returning a oneshot?

rs/moq-net/src/lite/subscriber.rs
```diff
@@ -17,13 +17,83 @@ use crate::{
...
 
 /// Keep an upstream subscription alive briefly after the last consumer leaves,
 /// so a returning subscriber reuses the same TrackProducer instead of forcing a
 /// fresh fetch (and the publisher re-serving the latest cached group).
 const LINGER_TIMEOUT: Duration = Duration::from_secs(5);
 
+/// One-shot latch that fires once every announce-prefix stream has received its
+/// initial set (AnnounceInit for Lite01/02, AnnounceOk + N for Lite05). Lets
+/// `connect()` block until the initial broadcasts are available locally.
+pub(super) struct SyncLatch {
```
kixelated: Maybe this should be more generic. Like call it Setup or something so we can use it to track when extensions are also done being negotiated in the future.

And IDK maybe try to use `kio`. It wouldn't be that hard to have a Producer/Consumer pair and a poll_ready method that checks when remaining <= 0.

kixelated: Also while you're here, add a note to try to use `kio` instead of `tokio` primitives for any new async stuff. The goal is to fully support both an async and synchronous (poll-based) API.

rs/moq-net/src/lite/session.rs
kixelated: Done. Replaced the oneshot with a `Setup` handle (kio-backed) returned from `lite::start`; `connect()` now does `setup.ready().await`. It can't be a plain `async fn start` because the server calls `start` too and must not block on the initial set, so readiness is an awaitable handle rather than folded into `start` itself. (Written by Claude)

rs/moq-net/src/lite/subscriber.rs
kixelated: Done. Renamed to `Setup` (in `lite/setup.rs`) and documented it as the generic place to track session-setup completion, with extension negotiation called out as a future step. Reworked it onto a `kio` Producer/Consumer pair with both `poll_ready(&Waiter)` and async `ready()` (the latter is implemented via the former). (Written by Claude)

kixelated: Added a note to the `Setup` module doc-comment to prefer `kio` over `tokio` primitives for new async state, so we keep both an async and a poll-based API available. Happy to also drop a line in CLAUDE.md if you'd like it somewhere more discoverable. (Written by Claude)

js/net/src/lite/publisher.ts
kixelated: Inlined. (Written by Claude)

rs/moq-net/src/lite/subscriber.rs
kixelated: Renamed it from `Setup` to `Connecting` per your follow-up — also avoids confusion with the IETF SETUP message module (`setup::Client`/`setup::Server`). Now in `rs/moq-net/src/lite/connecting.rs`. (Written by Claude)

PR #1576, opened 2026-06-01
moq-net: split TrackConsumer into a track handle + TrackSubscriber

kixelated:
## Summary

Picks up the next step after #1540. Separates *a handle to a track you can subscribe to* from *a live subscription you read from*, and moves the SUBSCRIBE_OK await off the broadcast lookup and onto the subscription itself.

- **Rename** the old read+demand handle `TrackConsumer` → **`TrackSubscriber`**. It keeps everything (`recv_group`/`next_group`/`read_frame`, `update`, `subscription`) and still carries the per-subscriber `Subscription` that feeds the producer's aggregate.
- **Add** a new lightweight **`TrackConsumer { broadcast, name }`**: a cheap, cloneable handle that names a track but sends nothing until you call it.
  - `async fn subscribe(Subscription) -> Result<TrackSubscriber>` blocks on SUBSCRIBE_OK.
  - `subscribe_pending(Subscription) -> TrackPending` for poll-based callers (moq-mux's container `source.rs`).
  - `fetch` is reserved (`TODO`) for one-shot range retrieval; not added yet.
- **`BroadcastConsumer::subscribe_track(name, sub) -> TrackPending`** becomes **`consume_track(name) -> TrackConsumer`**; the await now lives on `.subscribe()`:

```rust
let track: TrackConsumer = broadcast.consume_track("video");
let sub: TrackSubscriber = track.subscribe(Subscription { priority: 5, ..Default::default() }).await?;
sub.recv_group().await?;
```

All ~30 call sites across the workspace migrated.

## Cross-package

- **FFI (`moq-ffi` / `libmoq`)**: unchanged surface. `MoqBroadcastConsumer::subscribe_track` keeps its name and async signature; only its internal `self.inner.consume_track(...)` calls were renamed. No py/swift/kt/go regeneration needed.
- **`js/net`**: no change. Wire format is unchanged, and it has its own `subscribe(name, priority)` API.
- **docs**: updated the Rust example in `doc/lib/rs/env/native.md`. Python doc examples are unchanged (FFI API unchanged).

## Test plan

- [x] `cargo check --workspace --all-targets` clean
- [x] `cargo clippy --workspace --all-targets` clean (via pinned nix toolchain)
- [x] `cargo test` for `moq-net` (328), `moq-native`, `moq-relay`, `moq-mux` — all pass
- [x] `cargo fmt` via nix

## Follow-ups (not in this PR)

- The `stale` field in `aggregate_subscription` looks inverted (forwards the *soonest* skip deadline upstream rather than the *most patient* subscriber's). Flagged for a separate change once confirmed.
- SUBSCRIBE_UPDATE aggregate plumbing: the relay still doesn't re-aggregate downstream preferences and push an upstream SUBSCRIBE_UPDATE when they change.

(Written by Claude)

### DIFF
--- a/doc/lib/rs/env/native.md
+++ b/doc/lib/rs/env/native.md
@@ -125,7 +125,7 @@ Subscribe to a media track and read frames using [`OrderedConsumer`](https://doc
 ```rust
 let track_consumer = broadcast
     .consume_track(&track_name)
-    .subscribe(moq_net::Subscription::new(1))
+    .subscribe(moq_net::Subscription::default().with_priority(1))
     .await?;
 let mut ordered = hang::container::OrderedConsumer::new(
     track_consumer,
--- a/doc/lib/rs/env/native.md
+++ b/doc/lib/rs/env/native.md
@@ -107,7 +107,8 @@ Subscribe to it using [`CatalogConsumer`](https://docs.rs/hang/latest/hang/catal
 
 ```rust
 let catalog_track = broadcast
-    .subscribe_track(hang::Catalog::DEFAULT_NAME, hang::Catalog::default_subscription())
+    .consume_track(hang::Catalog::DEFAULT_NAME)
+    .subscribe(hang::Catalog::default_subscription())
     .await?;
 let mut catalog = hang::CatalogConsumer::new(catalog_track);
 let info = catalog.next().await?.expect("no catalog");
@@ -123,7 +124,8 @@ Subscribe to a media track and read frames using [`OrderedConsumer`](https://doc
 
 ```rust
 let track_consumer = broadcast
-    .subscribe_track(&track_name, moq_net::Subscription::new(1))
+    .consume_track(&track_name)
+    .subscribe(moq_net::Subscription::new(1))
     .await?;
 let mut ordered = hang::container::OrderedConsumer::new(
     track_consumer,
--- a/rs/moq-net/src/model/broadcast.rs
+++ b/rs/moq-net/src/model/broadcast.rs
@@ -4,7 +4,7 @@ use std::{
 	task::{Poll, ready},
 };
 
-use crate::{Error, Subscription, TrackConsumer, TrackProducer, model::track::TrackWeak};
+use crate::{Error, Subscription, TrackProducer, TrackSubscriber, model::track::TrackWeak};
 
 use super::{OriginList, Track};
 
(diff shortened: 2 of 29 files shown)


doc/lib/rs/env/native.md
```diff
@@ -123,7 +124,8 @@ Subscribe to a media track and read frames using [`OrderedConsumer`](https://doc
 
 ```rust
 let track_consumer = broadcast
-    .subscribe_track(&track_name, moq_net::Subscription::new(1))
+    .consume_track(&track_name)
+    .subscribe(moq_net::Subscription::new(1))
```
kixelated: Can you remove `::new`? It's not obvious this is priority. 

Force struct construction and/or make a builder. `moq_net::Subscription::default().with_priority(1)`

rs/moq-net/src/model/broadcast.rs
```diff
@@ -654,6 +662,43 @@ impl BroadcastConsumer {
...
+	pub async fn subscribe(&self, subscription: Subscription) -> Result<TrackSubscriber, Error> {
+		self.subscribe_pending(subscription).ok().await
+	}
+
+	/// Like [`Self::subscribe`], but returns the unresolved [`TrackPending`] so a
+	/// poll-based caller can drive it via [`TrackPending::poll_ok`].
+	pub fn subscribe_pending(&self, subscription: Subscription) -> TrackPending {
+		self.broadcast.request_subscribe(&self.name, subscription)
+	}
+}
```
kixelated: I think we just combine these. TrackPending could impl Future

doc/lib/rs/env/native.md
kixelated: Done in f6e7b717d423e845997845a04f00a64901a0a1e9. Replaced `Subscription::new(1)` (which didn't actually exist) with builder methods: `Subscription::default().with_priority(1)`. Added `with_priority`/`with_ordered`/`with_stale`/`with_group_start`/`with_group_end` and updated `hang::Catalog::default_subscription` too.

(Written by Claude)

rs/moq-net/src/model/broadcast.rs
kixelated: Done in f6e7b717d423e845997845a04f00a64901a0a1e9. `TrackPending` now impls `Future`, so `TrackConsumer::subscribe(sub)` returns it directly: `.await` it for the `TrackSubscriber`, or drive it with `poll_ok` from a poll loop (moq-mux's container source). Dropped `subscribe_pending` and the redundant `ok()`.

(Written by Claude)

PR #1581, opened 2026-06-01
relay: unified --auth-api (one call returns key + public + alias)

kixelated:
## What & why

Adds **`--auth-api`**: a single endpoint that resolves everything the relay needs to authorize a connection in **one HTTP call**, instead of separate fetches for the verifying key, public access, and path alias. This supports the MoQ CDN dashboard ([moq-dev/moq-pro#47](https://github.com/moq-dev/moq-pro/pull/47)) giving each project a stable id plus an editable vanity path that all resolve to one broadcast tree.

> This PR was reworked from an earlier `--auth-alias-api` approach (separate alias endpoint + client-side segment rewriting) into the unified one-call design below.

## The call

Per connection: `GET <base>/<connection-path>?kid=<kid>` (kid sent only when a JWT is present), over the existing cached, mTLS-gated client. Response — **all fields optional**:

```json
{ "alias": "x7k2qp/room/cam",
  "public": { "subscribe": ["..."], "publish": ["..."] },
  "key": "<verifying JWK string>" }
```

- `alias` — the **full** canonical root (first segment resolved to the project's stable id, rest preserved; e.g. `demo/room/cam` → `x7k2qp/room/cam`). Used verbatim, so the server owns the whole mapping. Absent → request path unchanged.
- `public` — anonymous subscribe/publish prefixes, relative to the root. Absent → none.
- `key` — verifying JWK for the `kid`. Absent → reject the JWT.

With the API resolving `demo` → `x7k2qp`, both `cdn.moq.dev/demo/foo` and `cdn.moq.dev/x7k2qp/foo` scope to `/x7k2qp/foo`.

## Behavior

- Kept alongside the existing flags (`--auth-key`, `--auth-key-dir`, `--auth-public`, `--auth-public-api`, `--auth-domain`); `--auth-api` is **mutually exclusive** with the key/public sources (startup error otherwise).
- **Fails closed**: the verifying key comes from this call, so any network/non-2xx/parse error rejects the connection (the cached `Cache-Control` softens blips). This is deliberately unlike the removed fail-open alias step.
- `verify()` short-circuits to `verify_via_api` when configured; the overlap+scope reduction is factored into `Auth::finalize` and shared with the standalone path (all existing tests unchanged). mTLS peers resolve only the alias via `Auth::resolve_root`; root connections (`/`) skip the call. `web.rs`/`websocket.rs` route through the same paths.
- Removes `--auth-alias-api`, `resolve_path`, `fetch_alias`, `rewrite_first_segment`, `AliasResponse`.

## Tests
wiremock coverage for the unified endpoint: JWT scopes-to-alias, full-root passthrough (`/demo/room/cam` → `/x7k2qp/room/cam`), anonymous-via-public, unknown-project (absent alias → path), missing-key → reject, server-error → fail-closed, mTLS `resolve_root`, root-skips-API, and `--auth-api`+`--auth-key-dir` config-exclusivity. Plus the config-merge regression test.

`cargo fmt`, `cargo clippy -p moq-relay --all-targets` (clean), `cargo test -p moq-relay` (111 passed), `cargo build` all pass.

## Pairs with
[moq-dev/moq-pro#47](https://github.com/moq-dev/moq-pro/pull/47), which serves `GET /cluster/auth/<path>?kid=<kid>` with this exact contract and points the relay deploy at `--auth-api "<api>/cluster/auth"`.


rs/moq-relay/src/auth.rs
```diff
@@ -98,6 +98,32 @@ fn match_domain(host: Option<&str>, domains: &[String]) -> Option<String> {
...
 }
 
+/// Rewrite the first path segment of `path` to `canonical`, preserving the rest
+/// of the path and the leading-`/` shape.
+///
+/// Pure (no HTTP) so the rewrite logic can be unit-tested directly. Callers
+/// guarantee `canonical` is non-empty and that `path` has a non-empty first
+/// segment; a root path (`/` or empty first segment) is returned unchanged as
+/// a defensive no-op.
+fn rewrite_first_segment(path: &str, canonical: &str) -> String {
```
kixelated: Also can you get rid of rewrite_first_segment? The API should return the full root to use, so if the path is /demo/room/cam  then the alias/root should be /x7k2qp/room/cam . This gives us more flexibility in the future

rs/moq-relay/src/auth.rs
```diff
@@ -306,6 +332,35 @@ pub struct AuthConfig {
...
+	/// JSON parse error, or URL-join error the path is left unchanged and a
+	/// warning is logged. The canonical id is the SDK default, so id-based
+	/// connections must keep working during an API outage; the response cache
+	/// (`Cache-Control` from the endpoint) softens transient failures. This is
+	/// deliberately the opposite of the public-API path, which fails closed.
+	///
+	/// Example: `https://api.moq.dev/cluster/alias`.
+	#[arg(long = "auth-alias-api", env = "MOQ_AUTH_ALIAS_API")]
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub alias_api: Option<String>,
```
kixelated: Could we combine --auth-key-dir --auth-public-api and --auth-alias-api   into a new --auth-api field? Then we can return key/public/alias as object K/V values, instead of requiring multiple HTTP calls.

rs/moq-relay/src/auth.rs
```diff
@@ -420,15 +455,31 @@ impl Serialize for PublicConfig {
...
 	publish: Vec<String>,
 }
 
+/// Response from the unified `--auth-api` endpoint. Every field is optional; the
+/// relay defaults anything absent (see [`AuthConfig::auth_api`]).
+#[derive(Debug, Default, Deserialize)]
+struct AuthApiResponse {
+	/// Canonical full root to scope to; absent -> use the request path as-is.
+	#[serde(default)]
+	alias: Option<String>,
```
kixelated, line 488: PathOwned?

rs/moq-relay/src/auth.rs
```diff
@@ -420,15 +455,31 @@ impl Serialize for PublicConfig {
...
+	/// Anonymous access prefixes; absent -> no public access.
+	#[serde(default)]
+	public: Option<PublicResponse>,
+	/// Verifying JWK (JSON string) for the requested kid; absent -> not found.
+	#[serde(default)]
+	key: Option<String>,
```
kixelated: Does moq-token implement serde? Could we avoid the String?

rs/moq-relay/src/connection.rs
```diff
@@ -112,21 +112,21 @@ impl Connection {
...
 			None => AuthParams::default(),
 		};
+
+		if self.request.has_peer_certificate() {
+			tracing::debug!("mTLS peer authenticated");
+			// Scope the grant to the canonical root. An mTLS publisher dialing a
+			// vanity alias lands on the same tree a JWT would; cluster peers dial
+			// "/", which resolves to an empty (unscoped) root. mTLS peers skip the
+			// JWT/public parts of --auth-api, so only the alias is resolved.
+			let root = self.auth.resolve_root(&params.path).await;
```
kixelated: We shouldn't do this for mTLS? 

I mean maybe we should always hit the API? And indicate if the connection was mTLS verified? Then we can actually return the tier instead of hard-coding it as internal?

kixelated:
@coderabbitai review

PR #1591, opened 2026-06-02
js/signals: add Computed derived signals

kixelated:
## Summary

Adds `Computed` to `@moq/signals`: a read-only signal derived from other signals. The compute function receives an `Effect` and reads dependencies with `effect.get(...)`, exactly like a normal effect, but the result is cached and recomputed when a dependency changes.

```ts
const first = new Signal("Ada");
const last = new Signal("Lovelace");
const full = new Computed((effect) => `${effect.get(first)} ${effect.get(last)}`);
full.peek(); // "Ada Lovelace" — never undefined
```

### Why it wasn't here before

The open question was whether a computed's value should be `T | undefined`: effects run asynchronously (deliberately, to catch code that assumes synchronous reruns), so a naive computed would be `undefined` between construction and its first async run. The fix is to compute **lazily on first read**, synchronously, so the value is always a real `T`. Subsequent reads recompute synchronously when a dependency has changed.

Staleness matches the rest of the library: a `.set()` does not synchronously rerun effects, and likewise does not synchronously recompute a computed. A bare `peek()` immediately after a dependency `set()` is **stale-but-defined** (one-microtask lag, like every `Signal` subscriber). Inside the reactive flow it's always fresh, because the propagate → recompute → flush chain completes before a downstream effect reads.

### How it works

- **Reuses `Effect` for tracking** via a new derived mode: instead of rerunning itself asynchronously when a dependency changes, the `Effect` calls an `onInvalidate` callback. `Computed` drives a synchronous recompute through a new `Effect.track()` that runs the body and returns its value.
- **Equality-filtered, coalesced downstream notification** comes for free by routing the cached value through an internal `Signal`.
- **Cycle detection**: a per-instance `evaluating` flag turns a self-referential or transitive cycle (`a` → `b` → `a`) into a thrown `Computed cycle detected` error instead of a stack overflow.
- **`effect.computed(fn)`** ties a derived signal's lifetime to the parent effect (closed on rerun/close). Standalone `new Computed(fn)` has an explicit `close()`.

### Migration proof point

Migrated the `catalog` derivation in watch's audio and video `Source` from the hand-rolled `effect.set(target, derived)` pattern to `effect.computed`. The public `catalog` field stays typed as `Getter`, so the API is unchanged.

### Tests

First test suite for `@moq/signals` (`bun test`, wired like `net`/`watch`/`token`): 23 tests covering `Signal` (coalescing, equality filtering), `Effect` (rerun, cleanup ordering), and `Computed` (lazy/never-undefined, stale-then-fresh, equality filtering, coalescing, nesting, `close()`, cycle detection, `effect.computed` lifetime).

## Test plan

- [x] `just js check` (typecheck + biome) clean across all packages
- [x] `just js test` passes (signals 23, watch 64, net 137, token 61, loc 6)
- [x] `bun run build` for signals succeeds (publint passes)

(Written by Claude)


js/signals/src/index.ts
```diff
@@ -625,6 +672,105 @@ export class Effect {
...
+		try {
+			const value = this.#effect.track(this.#fn);
+			if (this.#signal) {
+				this.#signal.set(value);
+			} else {
+				this.#signal = new Signal(value);
+			}
+		} finally {
+			this.#evaluating = false;
+		}
```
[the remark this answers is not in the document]
kixelated: Fixed in c2f38fb9: `#recompute` now clears `#dirty` only after a successful compute, so a throwing compute (cycle or compute error) stays dirty and resurfaces on the next read instead of returning the stale cached value. The background recompute microtask catches the throw so it does not escape uncaught. Added a regression test covering succeed → self-reference → recover. (Written by Claude)

js/signals/src/index.ts
```diff
@@ -207,19 +207,27 @@ export class Effect {
...
 	#stopped: PromiseWithResolvers<void>;
 	#closed: PromiseWithResolvers<void>;
 
 	#abort: AbortController = new AbortController();
 
 	// If a function is provided, it will be run with the effect as an argument.
-	constructor(fn?: (effect: Effect) => void) {
+	// onInvalidate switches the effect into derived mode (see #onInvalidate); it's
+	// internal, used by Computed.
+	constructor(fn?: (effect: Effect) => void, onInvalidate?: () => void) {
```
kixelated: I don't like this. Shouldn't it be possible to use the existing fn instead of making a new (internal) callback?

js/signals/src/index.ts
```diff
@@ -625,6 +672,115 @@ export class Effect {
...
+
+	#dirty = true;
+	#scheduled = false;
+	#closed = false;
+
+	// True while the compute function is running, so a self-referential read can be
+	// caught instead of recursing forever. Also catches transitive cycles (a -> b -> a).
+	#evaluating = false;
+
+	constructor(fn: (effect: Effect) => T) {
```
kixelated, line 650: Do we need this? Why can't we just create our own `new Effect()`

js/signals/src/index.ts
kixelated: Removed in 415325fd. Effect is reverted to stock (no `onInvalidate`, no `track()`). Computed now subscribes to its dependencies directly via `signal.changed(...)` and marks itself dirty, so the internal callback lives in Computed where it belongs. (Written by Claude)

js/signals/src/index.ts
kixelated: Computed no longer uses Effect at all. I tried a stock `new Effect(body)` and it loops forever: the constructor always schedules an initial async run, and `Effect.#run` clears `#unwatch` before running the body, so the deps read during `peek()` get wiped and the body re-subscribes to nothing (which then re-notifies, rebuilds, and repeats). Computed instead tracks its own deps; it keeps an internal Signal only for value storage / equality / async notification, which runs nothing on its own. (Written by Claude)

js/signals/src/index.ts
```diff
@@ -625,6 +633,139 @@ export class Effect {
...
+				// The compute threw (e.g. a cycle). We stay dirty, so the error
+				// resurfaces synchronously on the next read rather than escaping here.
+			}
+		});
+	}
+
+	peek(): T {
+		this.#assertNotEvaluating();
+		if (!this.#signal || this.#dirty) this.#recompute();
+		return (this.#signal as Signal<T>).peek();
```
kixelated: maybe have this.#recompute return this.#signal so typescript doesn't need a cast.

js/signals/src/index.ts
```diff
@@ -625,6 +633,139 @@ export class Effect {
...
+	}
+
+	changed(fn: Subscriber<T>): Dispose {
+		this.#assertNotEvaluating();
+		if (!this.#signal) this.#recompute();
+		return (this.#signal as Signal<T>).changed(fn);
+	}
+
+	subscribe(fn: Subscriber<T>): Dispose {
+		this.#assertNotEvaluating();
```
kixelated: maybe you should just make a method that ensures this.#signal. Lots of duplcation.

js/signals/src/index.ts
```diff
@@ -625,6 +633,139 @@ export class Effect {
...
+	}
+
+	peek(): T {
+		this.#assertNotEvaluating();
+		if (!this.#signal || this.#dirty) this.#recompute();
+		return (this.#signal as Signal<T>).peek();
+	}
+
+	// Alias for peek(), matching Signal.get().
+	get(): T {
```
kixelated: IMO remove it. 

Also does Signal.get actually exist? I didn't want it to on purpose.

js/signals/src/index.ts
```diff
@@ -625,6 +633,139 @@ export class Effect {
...
+		for (const unwatch of this.#unwatch) unwatch();
+		this.#unwatch.length = 0;
+
+		const reader: Computing = {
+			get: <V>(signal: Getter<V>): V => {
+				// changed() is one-shot, like Effect.get: re-subscribed on each recompute.
+				this.#unwatch.push(signal.changed(() => this.#invalidate()));
+				return signal.peek();
+			},
+		};
```
kixelated: Can we reuse this instead of making a new one every time?

js/signals/src/index.ts
kixelated: Resolved in 431c5151: the whole `#recompute`/`peek` path is gone. Computed is now a stock Effect writing an internal Signal, so `peek()` is just `this.#signal.peek()` with no cast. (Written by Claude)

js/signals/src/index.ts
kixelated: Resolved in 431c5151: no more duplicated `#signal` guard. The internal Signal exists from construction, so peek/changed/subscribe are one-line delegations. (Written by Claude)

js/signals/src/index.ts
kixelated: Resolved in 431c5151: there is no per-run reader anymore. The compute function receives the real Effect (the internal one), so tracking is the stock `effect.get`. (Written by Claude)

js/signals/src/index.ts
kixelated: Removed the `get()` alias in 431c5151. And yes, `Signal.get` does exist (index.ts:54) alongside `peek()`, with a `// TODO rename to get once we've ported everything` note. It predates this PR, so I left it alone; happy to remove it (or finish the peek->get rename) in a separate change if you want it gone. (Written by Claude)

kixelated:
Switched `Computed` to a fully async `T | undefined` model (431c5151).

Reviewing the actual use cases settled the question: the watch audio/video catalog computeds both already return `| undefined`, and their consumers already guard it (`effect.get(this.catalog)?.renditions ?? {}`) because the catalog was an optional `Signal` before the migration. `| undefined` is the prevailing type across `watch`, so a synchronous never-undefined computed was the outlier, not the norm.

So `Computed` is now just a stock `Effect` writing an internal `Signal`:

```ts
new Effect((effect) => internalSignal.set(fn(effect)))
```

Value is `undefined` until the first async run (and after `close()`); recomputes coalesce + equality-filter via the internal Signal. This deletes ~90 lines of dirty/peek/recompute/cycle machinery (down to ~20), drops the custom reader, and resolves the four review nits by removing the code that had them. The compute fn now gets the real `Effect`.

Tradeoffs accepted: no synchronous cycle detection (a self-referential computed becomes an async loop, like any effect that writes a signal it reads), and creating+observing a computed in the same rerunning effect body loops — so `effect.computed` is for the create-once-on-a-container pattern that `watch` uses.

(Written by Claude)

PR #1592, opened 2026-06-02
js: make @moq/watch component signals explicit inputs vs outputs

kixelated:
## Why

A long-standing complaint about `@moq/watch` (and `@moq/publish`) is that you can't tell, from a component's public surface, which signals you're meant to **drive** (inputs) and which you're meant to **read** (outputs). Everything was a bare `Signal<T>`, so nothing stopped a caller from `.set()`-ing a value the component owns, and wiring one component's output into another's input gave no protection against clobbering it.

This reshapes every `@moq/watch` component around an explicit `input` / `output` split. `@moq/publish` is intentionally left for a follow-up.

## The pattern

Three small helpers added to `@moq/signals`:

- `getter(value)` — read-only counterpart to `Signal.from`. Reuses a passed `Signal`/`Getter` (so one component's output wires straight into another's input) or wraps a raw value.
- `readonlys(map)` — identity-at-runtime cast exposing a record of `Signal`s as read-only `Getter`s.
- `InputProps<I>` — derives a component's `{ [k]?: value | Signal }` constructor props from its input map, so the props contract lives in one place.

Each component now looks like:

```ts
type FooInput = { enabled: Getter<boolean> };          // the component only READS these
type FooOutput = { status: Signal<Status> };           // the component WRITES these
export type FooProps = InputProps<FooInput>;

class Foo {
  readonly input: Readonlys<FooInput>;                 // read-only to callers
  readonly #output: FooOutput = { status: new Signal("offline") };
  readonly output = readonlys(this.#output);           // read-only to callers
}
```

- **Inputs are read-only `Getter`s.** Whoever owns the backing `Signal` (the caller, or another component whose output is wired in) does the writing. This is what stops `renderer.input.frame.set(...)` from clobbering the `source.output.frame` it's wired to.
- **Outputs are read-only `Getter`s** over private `Signal`s the component owns.
- TypeScript enforces both: calling `.set()` on an input/output, or reassigning one, is a compile error.

## Owner-at-top

Because inputs are read-only, the mutable source of truth for user controls (`volume`, `muted`, `paused`, `latency`, the video `target`, plus `name`/`reload`/`catalogFormat`/`catalog`) now lives at the **top** — the `<moq-watch>` element owns those writable `Signal`s and wires read-only views down into `broadcast`/`backend`. The UI and the element's attribute/property accessors read and write the element's `controls` group. `MultiBackend` and everything below only ever see read-only inputs.

This required relocating a few feedback policies to their owners (these are the **behavioral** parts — see Testing):

- **Source↔Sync cycle broken**: `Source` no longer holds `Sync`; it produces `output.jitter`, which `MultiBackend` wires into `Sync`'s `video`/`audio` inputs. Sources are constructed before `Sync`.
- **Decoders/MSE take `sync` as a constructor arg** instead of reaching through `source.sync`.
- **Video enable policy** (`paused ? !frame : visible`) moved from `Renderer` (which now just exposes `output.visible`) to `MultiBackend`, which owns the decoder's `enabled` input.
- **Audio enable** is now the emitter's `output.enabled`, proxied into the audio decoder.
- **`volume`↔`muted` coupling** (mute zeroes/stashes volume; volume 0 reports muted) moved from the audio `Emitter` to the element owner. Note: it now runs in **both** backends; previously it only ran in the WebCodecs path (the emitter didn't exist under MSE).
- **Native `<video>` `volumechange` → volume** moved from the audio MSE backend (which now only applies volume one-way) to the element.
- Per the design discussion, this means a **standalone `MultiBackend`** no longer bundles the mute-coupling policy; the owner provides it.

## Other changes

- `@moq/hang` `Container.Consumer`'s `latency` widened from `Signal<Time.Milli>` to `Getter<Time.Milli>` (it only ever `.peek()`s it) so a component's read-only `output.buffer` can be passed directly. Backward compatible (a `Signal` is a `Getter`).
- Fixed a latent bug surfaced by explicit ownership: `Video.Mse.close()` used to close the shared `#videoSource` (owned by `MultiBackend`), causing a double-close and a use-after-close if the `<canvas>`/`<video>` element was swapped. It no longer does, matching `Audio.Mse`.
- Updated the two external low-level consumers: `demo/web` and `@moq/boy` (which builds its own pipeline from `Watch.Sync`/`Decoder`/`Renderer`/`Source`) — new constructors + read-only inputs (it now owns its own `canvas`/`target` signals).

## Testing

- `bun run --filter='*' check` (tsc + unit tests) passes across all 11 JS packages.
- `biome check` clean.
- **Not** automatically verified: the real-time media pipeline (decode/sync/buffer/MSE) has no automated tests, so the behavioral relocations above warrant a manual smoke test — WebCodecs and MSE playback, mute/unmute + volume slider, pause/resume, quality switching, and the moq-boy grid (expand/collapse, audio on hover).

(Written by Claude)

### DIFF
--- a/js/hang/src/container/consumer.ts
+++ b/js/hang/src/container/consumer.ts
@@ -1,6 +1,6 @@
 import type { Time } from "@moq/net";
 import * as Moq from "@moq/net";
-import { Effect, type Getter, getter, Signal } from "@moq/signals";
+import { Effect, type Getter, getter, type GetterInit, Signal } from "@moq/signals";
 
 import type { Format } from "./format";
 import type { BufferedRanges, Frame } from "./types";
@@ -9,7 +9,7 @@ export interface ConsumerProps {
 	format: Format;
 	// Target latency in milliseconds (default: 0). Read-only: a Getter (e.g. another
 	// component's output) is accepted directly.
-	latency?: Getter<Time.Milli> | Time.Milli;
+	latency?: GetterInit<Time.Milli>;
 }
 
 interface Group {
--- a/js/hang/src/container/consumer.ts
+++ b/js/hang/src/container/consumer.ts
@@ -1,14 +1,15 @@
 import type { Time } from "@moq/net";
 import * as Moq from "@moq/net";
-import { Effect, type Getter, Signal } from "@moq/signals";
+import { Effect, type Getter, getter, Signal } from "@moq/signals";
 
 import type { Format } from "./format";
 import type { BufferedRanges, Frame } from "./types";
 
 export interface ConsumerProps {
 	format: Format;
-	// Target latency in milliseconds (default: 0)
-	latency?: Signal<Time.Milli> | Time.Milli;
+	// Target latency in milliseconds (default: 0). Read-only: a Getter (e.g. another
+	// component's output) is accepted directly.
+	latency?: Getter<Time.Milli> | Time.Milli;
 }
 
 interface Group {
@@ -22,7 +23,7 @@ interface Group {
 export class Consumer {
 	#track: Moq.Track;
 	#format: Format;
-	#latency: Signal<Time.Milli>;
+	#latency: Getter<Time.Milli>;
 	#groups: Group[] = [];
 	#active?: number; // the active group sequence number
 
@@ -37,7 +38,7 @@ export class Consumer {
 	constructor(track: Moq.Track, props: ConsumerProps) {
 		this.#track = track;
 		this.#format = props.format;
-		this.#latency = Signal.from(props.latency ?? Moq.Time.Milli.zero);
+		this.#latency = getter(props.latency ?? Moq.Time.Milli.zero);
 
 		this.#signals.spawn(this.#run.bind(this));
 		this.#signals.cleanup(() => {
--- a/js/hang/src/container/consumer.ts
+++ b/js/hang/src/container/consumer.ts
@@ -1,6 +1,6 @@
 import type { Time } from "@moq/net";
 import * as Moq from "@moq/net";
-import { Effect, type Getter, getter, type GetterInit, Signal } from "@moq/signals";
+import { Effect, type Getter, type GetterInit, getter, Signal } from "@moq/signals";
 
 import type { Format } from "./format";
 import type { BufferedRanges, Frame } from "./types";
--- a/js/moq-boy/src/index.ts
+++ b/js/moq-boy/src/index.ts
@@ -117,8 +117,8 @@ export class Game {
 		});
 		this.#signals.cleanup(() => this.broadcast.close());
 
-		// Sources no longer depend on Sync; they produce the per-rendition jitter that
-		// Sync consumes, so they're created first to avoid a construction cycle.
+		// Sources produce the per-rendition jitter that Sync reads, so they're created
+		// before Sync to avoid a construction cycle.
 		this.videoSource = new Watch.Video.Source({ broadcast: this.broadcast, target: this.#target });
 		this.#signals.cleanup(() => this.videoSource.close());
 
--- a/js/moq-boy/src/index.ts
+++ b/js/moq-boy/src/index.ts
@@ -75,6 +75,13 @@ export class Game {
 	readonly status = new Moq.Signals.Signal<GameStatus | undefined>(undefined);
 	readonly viewerId = new Moq.Signals.Signal<string | undefined>(undefined);
 
+	// The canvas to render into, owned here and wired into the renderer (read-only there).
+	// The UI sets this once the <canvas> is mounted.
+	readonly canvas = new Moq.Signals.Signal<HTMLCanvasElement | undefined>(undefined);
+
+	// The video rendition target, owned here and wired into the video source as an input.
+	readonly #target = new Moq.Signals.Signal<Watch.Video.Target | undefined>(undefined);
+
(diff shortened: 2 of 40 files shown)


js/hang/src/container/consumer.ts
```diff
@@ -1,14 +1,15 @@
...
 import type { Format } from "./format";
 import type { BufferedRanges, Frame } from "./types";
 
 export interface ConsumerProps {
 	format: Format;
-	// Target latency in milliseconds (default: 0)
-	latency?: Signal<Time.Milli> | Time.Milli;
+	// Target latency in milliseconds (default: 0). Read-only: a Getter (e.g. another
+	// component's output) is accepted directly.
+	latency?: Getter<Time.Milli> | Time.Milli;
```
kixelated: Can you make a type for this? GetterInit maybe?

js/watch/src/audio/decoder.ts
```diff
@@ -3,68 +3,79 @@ import * as Container from "@moq/hang/container";
...
 	// Whether the audio buffer is stalled (waiting to fill)
-	#stalled = new Signal<boolean>(true);
-	readonly stalled: Getter<boolean> = this.#stalled;
+	stalled: Signal<boolean>;
+
+	// Combined buffered ranges (network jitter + decode buffer)
+	buffered: Signal<BufferedRanges>;
+};
+
+export type DecoderProps = InputProps<DecoderInput>;
```
kixelated: Can we rename `Props` to `Input` everywhere? DecoderInput and DecoderOutput are nice types.

js/moq-boy/src/index.ts
```diff
@@ -110,30 +117,37 @@ export class Game {
 		});
 		this.#signals.cleanup(() => this.broadcast.close());
 
-		this.sync = new Watch.Sync({ latency: this.latency, connection: connection.established });
-		this.#signals.cleanup(() => this.sync.close());
-
-		this.videoSource = new Watch.Video.Source(this.sync, { broadcast: this.broadcast });
+		// Sources no longer depend on Sync; they produce the per-rendition jitter that
```
kixelated: No legacy comments pls.

js/signals/src/index.ts
```diff
@@ -186,7 +186,44 @@ export class Signal<T> implements Getter<T>, Setter<T> {
...
+	if (typeof value === "object" && value !== null && SIGNAL_BRAND in value) {
+		return value as Getter<T>;
+	}
+	return new Signal(value as T);
+}
+
+// Derive a component's constructor props from its input map: every input becomes
+// optional and accepts a raw value, a Signal, or another component's output Getter
+// (the `getter()` contract). Removes the hand-written, drift-prone props interface.
+export type InputProps<I extends SignalMap> = { [K in keyof I]?: GetterType<I[K]> | I[K] };
```
kixelated: `Inputs`?

js/watch/src/audio/decoder.ts
```diff
@@ -409,7 +420,7 @@ export class Decoder {
 	}
 }
 
-async function supported(config: Catalog.AudioConfig): Promise<boolean> {
+export async function decoderSupported(config: Catalog.AudioConfig): Promise<boolean> {
```
kixelated: I'm confused, why export these? it seems very out of place.

js/watch/src/backend.ts
```diff
@@ -1,10 +1,14 @@
 import * as Moq from "@moq/net";
-import { Effect, Signal } from "@moq/signals";
+import { Effect, type Getter, getter, type InputProps, type Readonlys, readonlys, Signal } from "@moq/signals";
 import * as Audio from "./audio";
+import { decoderSupported as audioDecoderSupported } from "./audio/decoder";
```
kixelated: not a fan. `Audio.decoderSupported` at least.

js/watch/src/video/decoder.ts
```diff
@@ -494,7 +511,7 @@ class DecoderTrack {
 	}
 }
 
-async function supported(config: Catalog.VideoConfig): Promise<boolean> {
+export async function decoderSupported(config: Catalog.VideoConfig): Promise<boolean> {
```
kixelated: keep it `supported` IMO.

kixelated:
Thanks for the review — addressed in d4ecc8c4:

- **`GetterInit` type** (consumer.ts): added `export type GetterInit<T> = T | Getter<T>` to `@moq/signals`; `Container.Consumer.latency` and `getter()` now use it.
- **`InputProps` → `Inputs`** (signals.ts): renamed.
- **Rename `Props` to `Input` everywhere** / `DecoderInput`+`DecoderOutput`: dropped the per-component `*Props` aliases entirely; constructors now take `props?: Inputs<XxxInput>` inline. The `XxxInput`/`XxxOutput` map types stay.
- **`supported` exports felt out of place** / `keep it supported` / `Audio.decoderSupported at least`: made each probe a `static` member named `supported`, so it's reached as `Video.Decoder.supported` / `Video.Mse.supported` / `Audio.Decoder.supported` / `Audio.Mse.supported`. No more module-level exports, no deep imports in `backend.ts`.
- **No legacy comments** (moq-boy): reworded that comment (and the few other "relocated here / no longer" ones) to describe current state only.

All packages still pass `tsc` + tests + biome.

(Written by Claude)

Issue #1614, opened 2026-06-03
Add a flush/cancel primitive to AudioProducer (Python) for real-time interruption use cases (ai voice agents)


vipyne:
## tl;dr:
claude wrote this (and spoiler- it wrote the accompanying PR too). 
I'm open to other solutions, but it does seem like `AudioProducer.flush()` would be clutch to interrupt voice agents.

## Summary
     
`moq.AudioProducer.write()` (Python bindings, `moq-rs 0.2.17`) is fire-and-forget — there's no way to drop already-written audio from the encoder/wire/jitter buffer without closing the whole track. For real-time interactive use cases (voice agents, where the user interrupts the bot mid-utterance), this means we can stop generating audio but the already-buffered audio finishes playing on the consumer side.  
 
## Repro / context

Voice-agent pipeline:

```python     
broadcast = moq.BroadcastProducer()
audio = broadcast.publish_audio(
    "bot-audio",
    moq.AudioEncoderInput(format=moq.AudioFormat.S16, sample_rate=24000, channels=1),
    moq.AudioEncoderOutput(codec=moq.AudioCodec.OPUS, frame_duration_ms=20, ...),
)  
# Bot starts speaking; many `write()`s queue up in the encoder/wire/browser. 
for chunk in tts_stream(): 
    audio.write(moq.AudioFrame(timestamp_us=0, data=chunk))  
   
# User interrupts. We stop generating new audio. But there's no API to  
# drop the in-flight buffer — the user keeps hearing the bot for ~hundreds
# of ms more. 
```
   
Attempted workaround: `audio.finish()` followed by `broadcast.publish_audio("bot-audio", ...)` raises `Error processing frame: duplicate` — the broadcast won't accept republishing under the same track name.  
   
Today we work around this by pacing `write()` calls against a wall-clock virtual timer in Python so the in-flight buffer never exceeds ~20 ms. It works but it's a backpressure mechanism reinvented in userland.
   
## Requested API 

Add **`AudioProducer.flush()`** / **`AudioProducer.cancel()`** — drop any unencoded + unsent frames without closing the track. Consumer-side observable as a brief skip in playback. 
 
## Why this matters
 
[...]

PR #1615, opened 2026-06-03
moq-ffi, moq-mux: add flush/cancel

vipyne:
addresses #1614

### DIFF
--- a/rs/moq-mux/src/codec/opus/import.rs
+++ b/rs/moq-mux/src/codec/opus/import.rs
@@ -57,6 +57,17 @@ impl Import {
 		Ok(())
 	}
 
+	/// Close the current group early without finishing the track.
+	///
+	/// The next group opens at the natural next sequence (current + 1). Use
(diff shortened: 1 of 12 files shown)


rs/moq-mux/src/codec/opus/import.rs
kixelated, line 92: We should already flush each audio group immediately. 

Video is a bit different but yeah there's still no harm flushing at the next keyframe. It just means the QUIC stream is held open an extra ~30ms and the FIN is sent at the same time as the keyframe.

kixelated (commented):
I think what you actually want is to make timestamps go backwards?

Like if group 7 has a smaller timestamp than group 4-6, it would effectively invalidate that data in the jitter buffer.

This should already work on the JS player side but it will overwrite, not drop. So you could get some of the old audio instead of silence. Relatively easy to fix provided we're okay with timestamps going backwards and invalidating previous groups?

PR #1617, opened 2026-06-04
moq-mux: ingest SCTE-35 from MPEG-TS, and tolerate mid-stream joins

arielmol:
## Summary

Two related changes to the MPEG-TS importer in `moq-mux`:

1. **SCTE-35 ingest** — detect a SCTE-35 elementary stream and carry its `splice_info_section`s into the broadcast verbatim.
2. **Mid-stream join hardening** — let the importer start decoding a transport stream it joined already in flight. This is general TS-importer robustness; the SCTE work surfaced it, but it is not SCTE-specific.

## SCTE-35 ingest

A SCTE-35 PID is announced by a program-level registration descriptor with format_identifier `CUEI`, and its elementary stream uses `stream_type 0x86`. `0x86` alone is ambiguous (mpeg2ts maps it to a DTS audio variant), so detection keys off the `CUEI` descriptor, not the stream type.

SCTE-35 rides in private sections (table_id `0xFC`), not PES. The mpeg2ts reader PES-parses every PMT elementary PID and would abort on a private section, so the importer intercepts SCTE PIDs ahead of the reader and reassembles their sections (pointer_field alignment, sections split across packets, continuity gaps, adaptation-field discontinuities). Each whole `splice_info_section` is published verbatim on a `.scte35` track through a catalog extension, not hang core, so an application routes cues into its own catalog section.

## Mid-stream join hardening

A live capture joins a stream already in flight: PES arrive before the first PAT/PMT, and the first decodable access unit can be a delta rather than a keyframe. Two assumptions aborted the decode:

- **Unrouted PID** — the mpeg2ts reader aborts on a PES whose PID it has not learned yet. The importer now gates what reaches the reader: only the PAT, the PMT PIDs it announces, and the elementary PIDs a PMT registers. Everything else drops until the layout is learned (PSI repeats), then normal demux resumes.
- **Delta before keyframe** — the container `Producer` rejects a non-keyframe when no group is open. Rather than relax that globally, an opt-in `lenient_start` mode drops leading non-keyframes until a keyframe anchors a group. The default stays strict; the H.264 and H.265 importers opt in. This touches `container/producer.rs` and the codec importers, outside the `ts/` tree.

## Evidence

- **Synthetic** — `survives_midstream_join` carves a mid-GOP join from `bbb.ts` and exercises both paths in one test.
- **Real encoder** — `kyrion_dirtystart_extracts_real_cues` runs a real Ateme Kyrion broadcast captured mid-stream: the importer survives the join and extracts all six SCTE-35 splice_inserts, which TSDuck decodes CRC32 OK (golden checked in beside the fixture). The fixture is our own encoder's output, a black feed with no broadcast content, byte-exact payload.

The Kyrion carries MP2 audio, not yet supported as an audio track, so it is dropped without blocking SCTE-35 extraction. The fixture keeps it on purpose: it is a real industry configuration and exercises dropping an unsupported elementary stream while the cues still come through.

## Review focus

`lenient_start` lives on the shared `container::Producer`, outside the `ts/` tree. The strict default is unchanged (`first_frame_must_be_keyframe` pins it); only the video importers enable it. That is the one change worth a careful look.

## Test plan

- [x] `cargo test -p moq-mux` (all pass)
- [x] SCTE reassembly edge cases (pointer_field, cross-packet split, continuity gap, discontinuity, stuffing)
- [x] mid-stream join: synthetic (`survives_midstream_join`) + real Kyrion (`kyrion_dirtystart_extracts_real_cues`)


js/hang/src/catalog/root.ts
```diff
@@ -5,13 +5,15 @@ import { AudioSchema } from "./audio";
...
 import { LocationSchema } from "./location";
+import { Scte35Schema } from "./scte35";
 import { TrackSchema } from "./track";
 import { UserSchema } from "./user";
 import { VideoSchema } from "./video";
 
 export const RootSchema = z.object({
 	video: z.optional(VideoSchema),
 	audio: z.optional(AudioSchema),
+	scte35: z.optional(Scte35Schema),
```
kixelated: How about we go back to your `meta` section instead of scte35 at the root?

js/hang/src/catalog/scte35.ts
```diff
@@ -0,0 +1,20 @@
+import * as z from "zod/mini";
+import { ContainerSchema } from "./container";
+
+// A SCTE-35 cue track: an MPEG-TS SCTE-35 PID carried as the verbatim
+// splice_info_section bytes (table_id 0xFC). The catalog only says how the bytes
+// were carried; it does not parse the splice. The stream_type (always 0x86) and
+// CUEI signaling are implicit to SCTE-35.
+export const Scte35ConfigSchema = z.object({
+	// The container format, used to decode the timestamp.
+	container: ContainerSchema,
```
kixelated: ...do you really want to support CMAF? I would just hard-code legacy for now, we can always negotiate a different container later.

js/hang/src/catalog/scte35.ts
```diff
@@ -0,0 +1,20 @@
...
+});
+
+export const Scte35Schema = z.object({
```
kixelated: I think nest scte35 an extra level.

```
export const MetaSchema = z.object({
	scte35: z.optional(z.record(z.string(), Scte35Schema)),
});
```

kixelated: Or maybe we just make it:

export const MetaSchema = z.object({});

And it's up to external applications to determine how to signal "scte35" tracks. hang feels like a weird spot for scte35 especially since TS is explicitly not supported as a container.

kixelated:
Also a claude review:

Solid PR overall — the byte-level reassembler is carefully written and the test coverage for the reassembler edge cases is genuinely good. The catalog typing is clean and the serde round-trip test pins the wire format properly. My findings cluster around detection robustness and timestamping, plus a few cleanups.

Correctness
1. [High] SCTE-35 detection only checks the program-level CUEI descriptor, missing the common ES-level placement — import.rs:164-167

let scte = pmt.program_info.iter().any(|d| d.tag == 0x05 && ... == b"CUEI");
The CUEI registration_descriptor is frequently carried in the per-elementary-stream descriptor loop of the splice PID, not the program loop. The mpeg2ts crate already exposes this as es.descriptors ([pmt.rs:160](https://github.com/)). As written, a stream that signals CUEI only on the 0x86 PID's own ES loop falls through to ensure_stream → the other => arm → "unsupported, dropping", and the cue track is silently never created. Consider OR-ing in a per-ES check: es.descriptors.iter().any(|d| d.tag == 0x05 && d.data.starts_with(b"CUEI")). This also fixes [#5](https://github.com/moq-dev/moq/issues/5) below for free (per-ES signaling disambiguates a real 0x86 audio ES from a splice PID).

2. [Med-High] SCTE sections are stamped Timestamp::ZERO whenever no video PTS has been seen yet — import.rs:135, 252-261

last_pts is only advanced in the H264 | H265 branch of handle_pes_start. For an audio-only TS with cues, or for any cue arriving before the first video PES, last_pts is None, so every such section is emitted at timestamp 0 and they collapse onto the same timeline point. Worth falling back to audio PTS (or wall clock) when video is absent.

3. [Med] seek() doesn't reset reassembler or media-clock state — import.rs:407-410, 314-322

ScteStream::seek only seeks the track; ScteReassembler (acc, last_cc) and media_unwrap/last_pts are untouched. On a loop/restart, a partial section left in acc from before the seek can be completed with post-seek bytes and emitted as a corrupt frame (it can still pass the 0xFC + length checks). Clear the reassembler on seek.

4. [Med] The byte-wise resync comment overstates recovery — import.rs:124-147

The comment says scanning for 0x47 one byte at a time "recovers instead of desyncing permanently," but once it locks onto a 0x47 it feeds those 188 bytes straight to read_ts_packet()?. I confirmed in the mpeg2ts source that an unknown PID (or "unexpected remaining data") returns Err, which ? propagates and aborts the entire decode(). So a false-sync hit on garbage — or any elementary/SCTE packet that legally precedes the first PMT — kills ingest rather than recovering. (This abort-on-unknown-PID is pre-existing reader strictness, but the new comment claims a robustness the code doesn't deliver.) Worth swallowing the per-packet reader error instead of ?-propagating.

5. [Low-Med] The scte flag is program-wide, so a genuine 0x86 DTS-HD lossless audio ES gets hijacked — import.rs:169

If a CUEI program also carries real DTS-HD lossless audio (stream_type 0x86), that audio PID is also routed to ensure_scte and fed into the section reassembler, dropping the audio track. Rare, but the per-ES check from #1 is the correct fix.

Cleanup / efficiency
6. [Efficiency] Per-packet reader feed is a hot-path regression — import.rs:140-149

The old path locked the Feed mutex once and drained the whole buffer; the rewrite, for every non-SCTE packet, does feed.lock() + data.clear() + extend_from_slice(&pkt) + a one-iteration read_ts_packet loop. For a 64 KB chunk that's ~340 locks/clears and a second full copy of every media byte (scratch → feed → reader). You could feed contiguous runs of non-SCTE packets in one drain, breaking the run only to divert a SCTE PID.

7. [Simplification] Duplicate PTS-unwrap state — import.rs:64-66, 259

media_unwrap runs the same video PTS sequence through a second PtsUnwrap that the Stream::H264/H265 variant already unwraps. Two state machines tracking the same wrap can drift. Consider having the video stream expose its latest unwrapped PTS and reading the media clock from there.

8. [Low] Redundant .map under an is_some() guard — import.rs:258-259

if pes.header.pts.is_some() { ... pes.header.pts.map(|t| t.as_u64()) ... } then unwrap_pts re-matches the Option. Cleaner as if let Some(pts) = pes.header.pts { self.last_pts = unwrap_pts(&mut self.media_unwrap, Some(pts.as_u64()))?; }.

9. [Low] ScteStream::new bypasses the Scte35::insert dedup helper — import.rs:373

It inserts into the raw BTreeMap (silent overwrite) rather than using the Scte35::insert helper that returns Error::Duplicate. Safe today thanks to unique_track, but it skips the invariant the helper exists to enforce.

Note (pre-existing, but amplified by this feature)
The mpeg2ts reader's StreamType::from_u8 returns Err for any stream_type not in its table ([stream_type.rs:104](https://github.com/)), which fails the whole PMT parse and aborts the import. SCTE-35 streams commonly coexist with vendor-specific PIDs, so this feature increases the odds of hitting it. Not introduced here, but a match/skip on the PMT parse would harden TS ingest meaningfully.

js/hang/src/catalog/scte35.ts
arielmol: > export const MetaSchema = z.object({});

(from Discord)

i went typed "scte35"  bcause of your earlier pushback on a generic data bucket but `export const MetaSchema = z.object({});` and  
> it's up to external applications to determine how to signal "scte35" tracks." 

signals the other way I'm happy either way, I just don't want to guess wrong. Which is the target? Typed meta.scte35 or empty metaSchema?

arielmol:
## Heads up!
This PR was force-pushed to a clean two-commit history, rebased onto current main (post #1658 / #1667). The earlier discussion and the CodeRabbit walkthrough describe the original approach, a typed scte35 section inside hang plus JS/zod changes **which has been dropped**. Please disregard the inline comments on that code; they're historical but kept on the same PR to follow the story :)

arielmol:
@coderabbitai review

arielmol:
@coderabbitai help

arielmol:
I see, rabbit. 1) The media clock is **video-only by design:** last_pts advances solely on a video PES, and for

**thread at import.rs:207 (CUEI per-ES / 0x86 disambiguation)**, ES-level-only CUEI is real and additive. Every encoder in scope, and all the checked-in fixtures (ffmpeg, GStreamer, TSDuck, and the real Ateme CM5000 "Kyrion" capture), signals CUEI at the program level, which is what detection keys off today.
A genuine 0x86 (DTS-HD MA) ES misrouted to SCTE can't be fixed by "require ES-level CUEI," because real SCTE streams carry CUEI only at the program level, so that rule would break every fixture.

The last one is real, ScteStream::seek seeks the track but leaves a half-buffered section in the reassembler, so a partial could splice onto post-seek bytes and emit a corrupt section. The fix is small: a reset() that clears `acc/last_cc/last_pkt` 

I'll fold that last one it in with the rest of @kixelated reviews rather than force-push the branch again now :)

PR #1626, opened 2026-06-04
Add moq-hls: HLS / LL-HLS gateway for MoQ

kixelated:
## Summary

Adds **`moq-hls`**, an HLS / LL-HLS gateway for MoQ, sibling to `moq-rtc`. It bridges HLS (and Low-Latency HLS) with MoQ broadcasts in both directions:

- **`export`** — subscribe to a MoQ broadcast and serve HLS + LL-HLS over HTTP. Path-based (one server, many broadcasts), per-rendition media playlists with `EXT-X-PART` parts, blocking reload (`_HLS_msn`/`_HLS_part`), and preload hints. TLS comes from a flattened `moq_native::ServerTlsConfig` (cert/key files, `--tls-generate` self-signed, `--server-tls-root` mTLS), built into a rustls config via the new `ServerTlsConfig::server_config()`.
- **`import`** — pull a remote HLS master/media playlist and publish it into MoQ (the HLS importer relocated out of `moq-mux`).

All CMAF byte handling stays in `moq-mux`; `moq-hls` owns the HLS manifest generation, segment/part windowing, and HTTP surface. `--part-target` and `--window` are humantime durations; the export store evicts segments by retained window duration rather than a fixed count.

## Design: reuse `fmp4::Export`, no bespoke segmenter

Rather than a new segmenter, export runs **one `moq-mux` `fmp4::Export` per rendition** (catalog narrowed to a single track via `catalog::Filter`), with `with_fragment_duration(part_target)` producing the LL-HLS parts. The only addition `Export` needed was segment metadata:

- New public `Fragment { data, init, independent, duration }` and `next_fragment()` / `poll_next_fragment()`.
- `independent` marks a fragment that opens on a keyframe (a GOP boundary → an HLS segment start); `duration` is the fragment's presentation seconds.
- `next()` / `poll_next()` still return `Bytes` (they delegate), so the existing `moq-cli` exporter caller is unaffected.

A bounded per-rendition store applies the segment policy (video rolls on `independent`; audio accumulates to a duration target) and a `watch` channel drives blocking reload. Playlists are hand-written because `m3u8-rs` can't emit the LL-HLS tags; the import side parses with `m3u8-rs`.

## Breaking changes (why this targets `dev`)

- `moq-mux`: the HLS importer moves to `moq-hls`. Removes `container::hls`, the `Hls` error variant, and the `m3u8-rs` + `reqwest` dependencies.
- `moq-cli`'s HLS publish format now uses `moq-hls`'s `import` module (depends on `moq-hls` with `default-features = false`, so the HTTP export server isn't pulled in).
- `moq-native`: adds `ServerTlsConfig::server_config(alpn)` (additive, `#[non_exhaustive]` struct).

## Out of scope

DASH is deferred. The segment store and `Fragment` are format-agnostic, so an MPD generator can be added over the same parts later.

## Test plan

- [x] `just rs check` (clippy `-D warnings`, fmt, `cargo shear`, `cargo sort`) clean.
- [x] `cargo test -p moq-mux -p moq-hls -p moq-native` — moq-hls 6 (playlist + master golden tests, import, `next_fragment` metadata) all green.
- [x] All three feature paths build: `moq-cli` (server-off `moq-hls`), `moq-hls` default (binary), `moq-hls --no-default-features` (import-only lib).
- [ ] Live end-to-end (needs a running relay + media): `moq-hls export` + `ffplay master.m3u8`, and an `import` round-trip. Not run in CI; commands documented in `doc/bin/hls.md`.

(Written by Claude)


rs/moq-hls/src/ingest.rs
```diff
@@ -1,7 +1,9 @@
-//! HLS (HTTP Live Streaming) ingest built on top of fMP4.
+//! HLS ingest: pull an HLS master/media playlist and publish it into MoQ.
```
kixelated: Can you call it `import/export` instead of `ingress/egress`?

rs/moq-hls/bin/moq-hls.rs
```diff
@@ -0,0 +1,173 @@
...
+	relay: Url,
+
+	#[command(subcommand)]
+	command: Command,
+}
+
+#[derive(Subcommand, Clone)]
+enum Command {
+	/// Serve HLS / LL-HLS over HTTP from MoQ broadcasts (path-based, multi-broadcast).
+	Serve {
```
kixelated: Export

rs/moq-hls/bin/moq-hls.rs
```diff
@@ -0,0 +1,173 @@
...
+		/// LL-HLS part target duration, in milliseconds.
+		#[arg(long, env = "MOQ_HLS_PART_TARGET_MS", default_value = "500")]
+		part_target_ms: u64,
+
+		/// Number of segments kept in each rendition's sliding window.
+		#[arg(long, env = "MOQ_HLS_WINDOW", default_value = "8")]
+		window: usize,
+	},
+	/// Pull a remote HLS master/media playlist and publish it into MoQ.
+	Ingest {
```
kixelated: Import

rs/moq-hls/bin/moq-hls.rs
```diff
@@ -0,0 +1,173 @@
...
+		/// HTTP listener for the HLS endpoints.
+		#[arg(long, env = "MOQ_HLS_LISTEN", default_value = "[::]:8089")]
+		listen: SocketAddr,
+
+		/// Optional TLS cert (PEM). Requires `--tls-key`.
+		#[arg(long, env = "MOQ_HLS_TLS_CERT", requires = "tls_key")]
+		tls_cert: Option<PathBuf>,
+
+		#[arg(long, env = "MOQ_HLS_TLS_KEY", requires = "tls_cert")]
+		tls_key: Option<PathBuf>,
```
kixelated: Can you make a proper TLS struct? There's other options we need (maybe use moq-native?)

rs/moq-hls/bin/moq-hls.rs
```diff
@@ -0,0 +1,173 @@
...
+
+		/// LL-HLS part target duration, in milliseconds.
+		#[arg(long, env = "MOQ_HLS_PART_TARGET_MS", default_value = "500")]
+		part_target_ms: u64,
```
kixelated: std::time::Duration

humantime maybe? 

Remve _MS

rs/moq-hls/bin/moq-hls.rs
```diff
@@ -0,0 +1,173 @@
...
+
+		/// Number of segments kept in each rendition's sliding window.
+		#[arg(long, env = "MOQ_HLS_WINDOW", default_value = "8")]
+		window: usize,
```
kixelated: Should this be a duration instead of a count?

rs/moq-hls/src/export/playlist.rs
```diff
@@ -0,0 +1,147 @@
...
+		.map(|s| s.duration)
+		.fold(0.0_f64, f64::max)
+		.max(snapshot.part_target);
+	let target_duration = max_segment.ceil().max(1.0) as u64;
+
+	// PART-HOLD-BACK must be at least 3x the part target (HLS spec).
+	let part_hold_back = snapshot.part_target * 3.0;
+
+	let mut out = String::new();
+	let _ = writeln!(out, "#EXTM3U");
```
kixelated, line 33: Any HLS crate we should use instead? It's not a big deal, but might be worth it.

rs/moq-hls/src/ingest.rs
kixelated: Done in cad100fc — renamed the `ingest` module to `import` and `egress` to `export` (and the `Serve`/`Ingest` subcommands to `Export`/`Import`). moq-cli and the docs are updated to match.

(Written by Claude)

rs/moq-hls/bin/moq-hls.rs
kixelated: Renamed `Serve` → `Export` in cad100fc.

(Written by Claude)

rs/moq-hls/bin/moq-hls.rs
kixelated: Renamed `Ingest` → `Import` in cad100fc.

(Written by Claude)

rs/moq-hls/bin/moq-hls.rs
kixelated: Done in cad100fc. The `export` subcommand now flattens `moq_native::ServerTlsConfig`, so it picks up `--tls-cert`/`--tls-key`, `--tls-generate` (self-signed), and `--server-tls-root` (mTLS) for free. I added `ServerTlsConfig::server_config(alpn)` to moq-native that builds a rustls `ServerConfig` for the axum listener, reusing the QUIC backend's cert resolver so both paths share cert loading.

(Written by Claude)

rs/moq-hls/bin/moq-hls.rs
kixelated: Done in cad100fc — it's now `--part-target` taking a humantime `Duration` (default `500ms`), matching the `humantime::parse_duration` pattern used elsewhere in the workspace. Dropped the `_ms` suffix.

(Written by Claude)

rs/moq-hls/bin/moq-hls.rs
kixelated: Changed in cad100fc — `--window` is now a humantime `Duration` (default `16s`) instead of a segment count. The export store evicts the oldest segment only while the remaining ones still cover the window span.

(Written by Claude)

rs/moq-hls/src/export/playlist.rs
kixelated: We already split it along those lines: the **import** side decodes with `m3u8-rs` (`parse_master_playlist` / `parse_media_playlist_res`), but the **export** side is hand-written because `m3u8-rs` v6 can't emit the LL-HLS tags we need — `EXT-X-PART`, `EXT-X-PART-INF`, `EXT-X-SERVER-CONTROL`, `EXT-X-PRELOAD-HINT`. So the plan is to keep the hand-rolled encoder and lean on the library for parsing. If m3u8-rs grows LL-HLS emit support we can revisit and drop the hand-written half.

(Written by Claude)

PR #1627, opened 2026-06-04
fix(moq-gst): follow catalog updates dynamically in moqsrc

kixelated:
## Problem

The interop smoke matrix (moq-dev/smoke) is red on every channel: `js-vite -> gst`, `js-esbuild -> gst`, and `js-jsdelivr -> gst` all fail, while `rust -> gst` and `python -> gst` pass and every *other* subscriber reads the browser broadcasts fine. The `gst` cells fail as a silent ~30s timeout (no bytes), not a fast error.

## Root cause

`moqsrc` (`rs/moq-gst/src/source/imp.rs`) read the hang catalog **exactly once** and built all its pads from that single snapshot, then never looked at the catalog again:

```rust
let catalog = catalog.next().await?.context("catalog missing")?.clone();
for (track_name, config) in catalog.video.renditions { … }
```

Reactive publishers — the browser via `@moq/hang` — announce an **initial catalog before the WebCodecs encoder has configured** (camera/display dimensions unresolved), so the first catalog frame carries **zero renditions**; the populated `video/hd` + `video/sd` catalog arrives a beat later as a new group. Every other consumer follows catalog updates and recovers; `moqsrc` latched the empty snapshot, created **no pads**, and produced no output → the pipeline sits with no data until it times out.

Matches all the evidence: `rust`/`python` emit a fully-populated catalog up front (pass); browser publishers race and lose (fail); ~30s timeout not fast-fail (the catalog *read* succeeds, there are just no renditions); identical across apt/brew/nix/cargo (platform-independent logic bug). Independent of the `moq-gst 0.2.4` release, which was the zigbuild/glib packaging fix that finally ships the plugin — the `moqsrc` logic was unchanged.

## Fix

Rather than reading the catalog once, `moqsrc` now **follows the catalog for the whole session and reconciles its pads against every update**. This fixes the reported race and also makes the source handle catalogs changing in general:

- **Reconcile loop** diffs the catalog's renditions against the live pumps on each update: spawn pads for newly announced renditions, tear down vanished ones, and recreate any whose **caps or container changed**. An empty catalog simply yields no pads (and we hold off `no_more_pads` until the first real one), so the browser's empty-first-frame is handled naturally.
- **Unique pad ids**: pads are named `video_<id>`/`audio_<id>` from a process-unique counter (matching the `%u` templates) instead of after the track name, so a recreated rendition never collides with the pad still being torn down. The `Drop` handler also evicts its map entry so long sessions don't leak handles.
- **Per-track teardown**: each pump gets its own cancel signal and reports completion over a channel correlated by pad id, so a pump can be stopped individually and a stale completion from a replaced rendition can't evict its successor.
- **Honor the per-track container** (legacy/cmaf/loc) from the catalog via the existing `moq_mux` `Container::try_from`, instead of hardcoding `Legacy`.

## Test

- `cargo check -p moq-gst` ✅
- `cargo clippy -p moq-gst` ✅ (clean)
- `cargo fmt -p moq-gst --check` ✅
- End-to-end confirmation is the smoke matrix's `js-* -> gst` cells (needs system GStreamer + the headless-Chromium publisher); not run locally.

### DIFF
--- a/rs/moq-gst/src/source/imp.rs
+++ b/rs/moq-gst/src/source/imp.rs
@@ -93,7 +93,6 @@ enum ControlMessage {
 		caps: gst::Caps,
 		reply: oneshot::Sender<PadEndpoint>,
 	},
-	NoMorePads,
 	ReportError(anyhow::Error),
 }
 
@@ -332,9 +331,6 @@ impl MoqSrc {
 					gst::error!(CAT, obj = self.obj(), "failed to create pad: {err:?}");
 				}
 			}
-			ControlMessage::NoMorePads => {
-				self.obj().no_more_pads();
-			}
 			ControlMessage::ReportError(err) => {
 				gst::element_error!(self.obj(), gst::CoreError::Failed, ("session error"), ["{err:?}"]);
 			}
@@ -454,7 +450,6 @@ async fn run_session(
 	let mut active: HashMap<String, ActiveTrack> = HashMap::new();
 	let (done_tx, mut done_rx) = mpsc::unbounded_channel::<TrackDone>();
 	let mut next_pad_id: u64 = 0;
-	let mut announced_no_more_pads = false;
 	// A detached clone for the pumps, kept separate so the `shutdown.changed()`
 	// branch below can borrow `shutdown` mutably without conflicting with reconcile.
 	let pump_shutdown = shutdown.clone();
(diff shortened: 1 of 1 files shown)


kixelated:
### Verified end-to-end: this fixes `browser → gst` ✅

Ran the actual `js-* → gst` cells locally (the ones the PR description couldn't run), using the in-tree smoke harness from #1639 with `MOQ_GST_PLUGIN_DIR` pointed at a `moqsrc` built from this branch. Same machine, same nix GStreamer 1.26.11, same published `@moq/publish` browser packages, same relay/cli — the **only** variable is `moqsrc`.

**Baseline (`main`, no fix):**
```
rust -> gst       PASS
js-vite -> gst     FAIL   (silent ~30s timeout, no bytes)
```

**This branch (`d399080f`):**
```
rust -> gst        PASS
js-vite -> gst     PASS
js-esbuild -> gst  PASS
js-jsdelivr -> gst PASS
```

Matches the diagnosis exactly: the reconcile loop picks up the populated catalog the browser announces a beat after its empty first frame, so `moqsrc` now builds pads and produces output instead of latching the empty snapshot. The `rust -> gst` control still passes, so nothing regressed for publishers that emit a full catalog up front.

Once this merges and a `moq-gst` release ships, `browser -> gst` goes green in the nightly matrix (#1639), and that cell can come back into the PR-smoke set.

(Written by Claude)

PR #1628, opened 2026-06-04
feat(moq-relay): graceful shutdown via GOAWAY drain

kixelated:
## Summary

Adds graceful shutdown to `moq-relay`. The first shutdown signal drains connections; a second forces termination.

- **First signal** (Ctrl+C / `SIGINT`, or `SIGTERM` from `systemctl stop`): stop accepting new connections, send a `GOAWAY` to every active session, and wait for them all to close on their own. The relay keeps serving so in-flight groups can finish. On systemd it also reports `STOPPING=1`.
- **Second signal**: force shutdown immediately, dropping every connection.

This lets an operator roll a relay node without dropping in-flight media: clients that honor `GOAWAY` migrate to another relay (or re-resolve DNS behind a load balancer) before the session closes.

## What changed

**`rs/moq-net` (the GOAWAY send path).** The GOAWAY message types already existed but were only ever *received*. This adds sending:

- `session.drain()` returns a public `Drain` handle: `start(uri)` sends the GOAWAY (without closing the session), `complete().await` waits for the peer to leave. `start` takes `impl Into<Option<&str>>` (`None` to just drain). Backed by `kio::Producer`/`Consumer`, not `tokio::sync::watch`.
- **moq-lite (04+)** opens a dedicated control stream and writes `GOAWAY`. Older lite drafts simply never send it (drain still works by waiting for the peer to leave, or a forced shutdown).
- **IETF moq-transport** sends `GOAWAY` on the shared control stream for draft-14-16 (via the control-stream adapter) and on the SETUP uni stream for draft-17+.

The signal is plumbed into `lite::start` / `ietf::start`; the trigger lives on the returned `Session`.

**`rs/moq-native`.** `Server::accept`'s built-in Ctrl+C handler (which hard-closes the QUIC endpoints) is now opt-out via `Server::with_signal_handler(false)`. Default behavior is unchanged for existing consumers (`moq-cli`, `moq-ffi`); the relay disables it and drives signals itself.

**`rs/moq-relay`.** Two-stage signal handling in `main`, plus connection tracking in `serve` (an mpsc-sender refcount) so we can wait for every in-flight connection to drain. Each connection task watches the drain signal and drains its session (`session.drain()`) when it fires.

## Test plan

- [x] New end-to-end test `goaway_drains_peer_moq_transport_14` (`rs/moq-native/tests/broadcast.rs`): server fires `session.drain().start(None)`, both sides observe the session close. Uses draft-14 because receiving a `GOAWAY` there closes the session, making the drain observable.
- [x] Lite `Goaway` encode/decode roundtrip unit tests (`rs/moq-net/src/lite/goaway.rs`).
- [x] All 57 `broadcast` integration tests pass (every lite + transport draft) — no handshake regression from the added plumbing.
- [x] `cargo clippy` clean and `cargo fmt --check` clean (pinned nix toolchain) for the three crates.

## Cross-package sync / out of scope

- `doc/bin/relay/index.md` documents the new shutdown behavior.
- **`js/net`**: not touched. This PR is the *server-side send* path. Browser clients honoring an incoming `GOAWAY` (migrating away) is a separate follow-up.
- **Cluster outbound** connections (relay dialing other relays as a client) are not drained; only inbound/accepted sessions receive `GOAWAY`. Cluster peers detect departure on their own.

Targeting `dev` per the branch-targeting rules (changes under `rs/moq-net`).

(Written by Claude)


rs/moq-native/src/server.rs
```diff
@@ -476,7 +494,7 @@ impl Server {
 						Err(err) => tracing::debug!(%err, "failed to accept session"),
 					}
 				}
-				_ = tokio::signal::ctrl_c() => {
+				_ = tokio::signal::ctrl_c(), if handle_ctrl_c => {
```
kixelated: I think this is wrong, because the condition is executed only after Poll::Ready.

rs/moq-net/src/session.rs
```diff
@@ -102,6 +106,16 @@ impl Session {
 		&self.consumer
 	}
 
+	/// Ask the peer to migrate away by sending a GOAWAY, without closing the session.
+	///
+	/// The session stays open so in-flight groups can finish; await
+	/// [`closed`](Self::closed) to learn when the peer actually leaves. `uri` is an
+	/// optional redirect target (pass `""` to just drain). Calling more than once,
+	/// or on a protocol version that predates GOAWAY, is harmless.
+	pub fn goaway(&self, uri: &str) {
```
kixelated: `impl Into Option<&str>`?

rs/moq-net/src/session.rs
```diff
@@ -126,6 +140,57 @@ impl Drop for Session {
...
+/// frame when fired.
+pub(crate) fn goaway_channel() -> (GoawayTrigger, GoawaySignal) {
+	let (tx, rx) = watch::channel(None);
+	(GoawayTrigger { tx }, GoawaySignal { rx })
+}
+
+/// Sender half of a session's GOAWAY signal, held by [`Session`].
+#[derive(Clone)]
+pub(crate) struct GoawayTrigger {
+	tx: watch::Sender<Option<Arc<str>>>,
```
kixelated: Can we combine Goaway  and Drain?

And use kio instead of watch. IDK I really hate the tokio ecosystem.

rs/moq-relay/src/main.rs
```diff
@@ -21,8 +21,10 @@ async fn main() -> anyhow::Result<()> {
 
 	let mtls_enabled = !config.server.tls.root.is_empty();
 
+	// We drive shutdown ourselves (GOAWAY drain on the first Ctrl+C, force on the
+	// second), so opt out of moq-native's built-in Ctrl+C-closes-everything handler.
 	#[allow(unused_mut)]
-	let mut server = config.server.init()?;
+	let mut server = config.server.init()?.with_signal_handler(false);
```
kixelated: This could use a better name.

rs/moq-net/src/session.rs
kixelated: And maybe `Drain` is part of the public API? So you would call `session.drain().start(url)` and `session.drain().complete().await` or something?

rs/moq-native/src/server.rs
kixelated: Switched to an explicit `ctrl_c`-or-`pending()` future instead of the `select!` `if` guard, so it no longer depends on guard-evaluation timing. When the handler is disabled the branch just never resolves.

(For the record the guard was evaluated before the future is polled, not after Poll::Ready — but the explicit future is clearer regardless.) (Written by Claude)

rs/moq-net/src/session.rs
kixelated: Done — `Drain::start` now takes `impl Into<Option<&str>>` (`None` to just drain, `Some(uri)` to redirect). (Written by Claude)

rs/moq-net/src/session.rs
kixelated: Done both. The trigger is now `kio::Producer`/`Consumer` instead of `tokio::sync::watch`, and Goaway + the drain wait are combined into one `Drain` type (see the public-API thread below). (Written by Claude)

rs/moq-relay/src/main.rs
kixelated: Renamed to `with_ctrl_c_handler`. (Written by Claude)

rs/moq-net/src/session.rs
kixelated: Implemented roughly as you sketched: `session.drain()` returns a public `Drain` with `start(uri)` (sends GOAWAY) and `complete().await` (waits for the peer to leave). The relay now does `let d = session.drain(); d.start(None); d.complete().await;`.

One boundary I kept: the relay still has its own process-level drain broadcast (one signal fanned out to all connections) separate from the per-session `Drain` — that broadcast is inherent since the signal originates at the process level, not per session. Each connection then calls `session.drain()`. Happy to fold those together differently if you had something else in mind. (Written by Claude)

PR #1630, opened 2026-06-05
feat(moq-relay): reload TLS certs on filesystem change instead of SIGUSR1

kixelated:
## Summary

Replace the `SIGUSR1`-driven certificate reloaders with a filesystem watcher, so cert-manager, Kubernetes secret mounts, and `mv`-into-place rotate certs with **no extra signalling** — the relay reacts to the file changing on disk.

Three reload sites move off the signal:
- **QUIC server certs** (`moq-native/src/tls.rs`)
- **HTTPS web cert/key/root** (`moq-relay/src/web.rs`)
- **Cluster connect-api peer file** (`moq-relay/src/cluster.rs`) — already used `notify`; now shares the same `FileWatcher` shape.

## `FileWatcher` helper

A small internal watcher used by each reload site:
- Watches each file's **parent directory**, not the file — atomic-rename / symlink-swap updates (editors, cert-manager, the K8s `..data` symlink) change the inode and would otherwise drop a watch set directly on the path.
- Coalesces duplicate and sibling-file events by **mtime**; `changed().await` only resolves when a watched file actually advances.
- **Ignores unreadable files** (the brief mid-rename gap), so a half-written cert/key set never triggers a load. A caller that fails to load keeps its current state and the next real change fires again.
- Falls back to 5s polling when no OS watcher can be created.

It's a generic filesystem utility, not MoQ API, so it's **kept internal to each crate** (`pub(crate)`) rather than exported. `moq-relay` can't reach a `pub(crate)` item in `moq-native` across the crate boundary, so each crate carries its own small private copy: `moq-native`'s serves the cert reload, `moq-relay`'s serves the web cert + cluster peer-file reloads.

## Why not just keep the signal?

The only thing the signal bought over a watcher was multi-file atomicity (cert + key + root must reload as a consistent set). That's solved in the watcher by reloading the whole set and only swapping it in if it parses and the key matches the cert — the same fail-static discipline the cluster file already used. K8s also swaps the entire secret mount atomically via the `..data` symlink, so watching the directory is consistent even with separate `tls.crt` / `tls.key` files (which is how K8s/cert-manager shape TLS secrets — you can't force a single PEM there).

## Notable behavior change

`SIGUSR1` **no longer reloads certs**. It was also wired to jemalloc heap-profile dumps (`moq-native/src/jemalloc.rs`), so previously one signal did both; jemalloc is now its sole consumer. Anyone who scripted `kill -USR1` to reload certs should know it's now automatic-on-write. No docs referenced the signal, so there's nothing to update there.

## Other changes

- Dropping the `#[cfg(unix)]` gates means cert reload now works on **Windows** too. notify 8.2 has a `RecommendedWatcher` for every build target (iOS/kqueue, Android+Linux/inotify, macOS/FSEvents, Windows), so the dep is safe for the mobile `moq-ffi` builds.
- In `moq-native` the watcher and its `notify` dep are gated on a QUIC backend (`noq`/`quinn`), matching the cert-reload path that uses them, so quiche/iroh-only builds don't pull in notify.
- Cluster refactor deletes the bespoke notify+channel plumbing, the `last_seen_mtime` threading, and the now-unused `CONNECT_API_FILE_INTERVAL` constant.

## Branch

Targeting `main`: no public API change (the watcher is crate-internal) and no wire/format change — just the `moq-relay` reload-trigger behavior change. Happy to retarget `dev` if a reviewer prefers the behavior change settle there first.

## Test plan

- [x] `cargo clippy -p moq-native -p moq-relay --all-targets` (clean; also checked `--no-default-features --features quiche,aws-lc-rs` — no unused watcher/notify)
- [x] `cargo fmt --check` (clean)
- [x] `cargo test -p moq-relay -p moq-native` — 125 unit + 2 smoke tests pass
- [ ] Manual: rotate cert/key on disk under a running relay, confirm reload without a signal

(Written by Claude)


rs/moq-native/src/watch.rs
```diff
@@ -0,0 +1,107 @@
...
+		let mtimes = paths.iter().map(|p| mtime(p)).collect();
+		let events = watcher.is_some().then_some(rx);
+
+		Self {
+			paths,
+			mtimes,
+			_watcher: watcher,
+			events,
+		}
+	}
```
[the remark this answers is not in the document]
kixelated: Added a deterministic unit test in `watch.rs` for `advanced()`'s mtime comparison (new-file detection + missing-file skip), the core both the event and polling paths gate on. The OS-event/polling paths themselves are timing-dependent so I left them untested by design. Note the watcher is now `pub(crate)`, not public API. (Written by Claude)

rs/moq-native/src/watch.rs
```diff
@@ -0,0 +1,151 @@
...
+	/// Resolve once at least one watched file's mtime advances.
+	///
+	/// Coalesces the duplicate and sibling-file events the OS delivers: a wakeup
+	/// that didn't actually change any watched file is swallowed and we keep
+	/// waiting.
+	pub(crate) async fn changed(&mut self) {
+		loop {
+			match &mut self.events {
+				// `recv` only yields `None` once the watcher is dropped, which
+				// can't happen while we hold it; park rather than busy-loop.
```
kixelated: ?? Return an error or panic then? An infinite sleep is gross.

rs/moq-native/src/watch.rs
```diff
@@ -0,0 +1,151 @@
...
+	/// Begin watching `paths`. Records their current mtimes as the baseline so
+	/// [`changed`](Self::changed) only resolves on a subsequent change.
+	pub(crate) fn new(paths: Vec<PathBuf>) -> Self {
+		// A capacity-1 channel of unit wakeups coalesces the flood of raw events
+		// notify emits (duplicates per change, plus unrelated sibling churn in the
+		// watched directory): when the buffer is already full there's a pending
+		// wakeup, so extra sends are simply dropped. The payload is irrelevant;
+		// `changed` rescans mtimes to decide what actually moved.
+		let (tx, rx) = mpsc::channel(1);
+		let watcher = notify::recommended_watcher(move |_event| {
```
kixelated: When can this return an Error? I see `poll` in the watcher code, why are we making our own poll based backup? Just return an error instead and delete our manual backup?

rs/moq-native/src/watch.rs
```diff
@@ -0,0 +1,151 @@
...
+			if self.advanced() {
+				return;
+			}
+		}
+	}
+
+	/// Refresh stored mtimes, returning true if any readable file is newer than
+	/// its baseline. Unreadable files are skipped so a mid-rename gap doesn't
+	/// fire a reload.
+	fn advanced(&mut self) -> bool {
```
kixelated: ??? What is this used for?

rs/moq-native/src/watch.rs
kixelated: Good call. `recommended_watcher` only errors on real OS failures (inotify instance/watch limits); on platforms with no native backend notify already aliases `RecommendedWatcher` to its own `PollWatcher`, so the hand-rolled poll loop was pure redundancy. `new` now returns a Result and callers log + disable hot reload on error. Manual backup deleted. (Written by Claude)

rs/moq-native/src/watch.rs
kixelated: Gone with the poll branch. The sender lives inside the watcher we hold across `&mut self`, so the channel can't close there; `changed` now just asserts that invariant via expect instead of parking forever. (Written by Claude)

rs/moq-native/src/watch.rs
kixelated: It was mtime bookkeeping to dedupe notify's per-change event bursts and filter sibling churn. Deleted it entirely: reloads are idempotent and K8s `..data` symlink swaps fire directory events anyway, so reacting to any watched-dir activity is correct and far simpler (worst case: an occasional redundant reload). (Written by Claude)

PR #1633, opened 2026-06-05
refactor(moq-gst): pump moqsrc pads directly instead of bridging to glib

kixelated:
> **Stacked on #1627** (base = `fix/moqsrc-catalog-wait-for-renditions`). Review/merge #1627 first; this diff is just the refactor. Once #1627 lands, GitHub retargets this to `main`.

## Motivation

`moqsrc`'s "async spaghetti": every buffer crossed **two** channels and a glib bridge.

The session task (tokio) couldn't touch GStreamer pads, so it marshalled everything onto the element's GLib main context:

- a `ControlMessage` channel to ask the element to create a pad,
- a per-pad `PadMessage` channel to forward each buffer / EOS / drop,
- `spawn_main_context_forwarder` tasks draining both onto `MainContext::default()`.

That bridge also carried a hidden requirement: **the application had to run a GLib main loop** or the forwarders never ran and no buffers flowed. Fine under `gst-launch`; a trap for an embedded host.

## Change

A source pad's `push` *is* its streaming thread, so there's nothing to marshal back. Each rendition now runs one **pump task that owns its pad end to end**: it creates the pad, pushes buffers/events directly, and removes the pad on exit (the sink already calls `add_pad`/`remove_pad` from a tokio task via a `WeakRef`, so this is an established pattern in the crate).

Deleted: `ControlMessage`, `PadMessage`, `PadEndpoint`, `PadHandle`, `request_pad`, `create_pad`/`dispatch_pad_message`/`handle_control_message`, `spawn_main_context_forwarder`, and the three `Mutex` fields behind them.

The session loop simplified to match:

- pumps are reaped via `JoinHandle::is_finished()` (woken by a unit `done` channel) instead of a name+id-correlated completion channel,
- a single per-pump `cancel` watch replaces the old shutdown+cancel pair.

**Net -219 lines** (`imp.rs` 848 → 628). No consumer-facing behavior change.

## Test

Built the plugin and exercised it at runtime (system GStreamer 1.26 via Nix + a local relay + a synthetic publisher):

- **Regression** — real H264 from `cdn.moq.dev/demo` decodes: 40 NV12 720p frames, exit 0.
- **Empty-first catalog race + unsupported skip + EOS** — publisher announces an empty catalog, then `video/hd` (H264) + `video/vp8` (unsupported): VP8 is skipped with a warning, exactly 120 buffers flow on `video/hd` (4s × 30fps), clean EOS on publisher finish.
- **Abrupt teardown** — SIGTERM mid-stream (`PausedToReady`): prompt clean shutdown, no hang/panic (the case the new pump-owns-its-pad model most needed to prove).
- `cargo clippy -p moq-gst` clean, `cargo fmt --check` clean.

Sizable rewrite of `moqsrc`'s internals, so a careful read on the threading is welcome even though all runtime tests pass.

(Written by Claude)

### DIFF
--- a/rs/moq-gst/src/source/imp.rs
+++ b/rs/moq-gst/src/source/imp.rs
@@ -205,7 +205,15 @@ impl ElementImpl for MoqSrc {
 					gst::error!(CAT, obj = self.obj(), "failed to start session: {err:?}");
 					return Err(gst::StateChangeError);
 				}
-				let success = self.parent_change_state(transition)?;
+				// Roll back the session we just started if the parent transition fails,
+				// otherwise it would keep running while the element stays in READY.
+				let success = match self.parent_change_state(transition) {
+					Ok(success) => success,
+					Err(err) => {
+						self.stop_session();
+						return Err(err);
+					}
+				};
 				// A live source never prerolls.
 				Ok(match success {
 					gst::StateChangeSuccess::Async => gst::StateChangeSuccess::Async,
@@ -296,11 +304,19 @@ async fn run_session(
 		}
 
 		tokio::select! {
-			// Full session shutdown: cancel every pump and wait for it to drop its pad.
+			// Full session shutdown: cancel every pump, then wait for them all to drop
+			// their pads. Cancel all up front (pumps only exit on their own `cancel`), or
+			// awaiting one at a time would let the not-yet-cancelled pumps keep streaming.
 			_ = shutdown.changed() => {
-				for (_, track) in active.drain() {
-					let _ = track.cancel.send(true);
-					let _ = track.handle.await;
+				let handles: Vec<_> = active
+					.drain()
+					.map(|(_, track)| {
+						let _ = track.cancel.send(true);
+						track.handle
+					})
+					.collect();
+				for handle in handles {
+					let _ = handle.await;
 				}
 				break;
 			}
--- a/rs/moq-gst/src/source/imp.rs
+++ b/rs/moq-gst/src/source/imp.rs
@@ -6,7 +6,7 @@ use anyhow::{bail, Context, Result};
 use gst::glib;
 use gst::prelude::*;
 use gst::subclass::prelude::*;
-use tokio::sync::{mpsc, watch};
+use tokio::sync::watch;
 
 use hang::moq_net;
 
@@ -207,12 +207,9 @@ impl ElementImpl for MoqSrc {
 				}
 				// Roll back the session we just started if the parent transition fails,
 				// otherwise it would keep running while the element stays in READY.
-				let success = match self.parent_change_state(transition) {
-					Ok(success) => success,
-					Err(err) => {
-						self.stop_session();
-						return Err(err);
-					}
+				let Ok(success) = self.parent_change_state(transition) else {
+					self.stop_session();
+					return Err(gst::StateChangeError);
 				};
 				// A live source never prerolls.
 				Ok(match success {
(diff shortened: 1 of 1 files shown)


rs/moq-gst/src/source/imp.rs
```diff
@@ -440,72 +276,44 @@ async fn run_session(
...
 
 		tokio::select! {
-			// Full session shutdown: cancel any pumps still running. (They watch this
-			// same signal too, so this is belt-and-suspenders for a prompt teardown.)
+			// Full session shutdown: cancel every pump and wait for it to drop its pad.
 			_ = shutdown.changed() => {
 				for (_, track) in active.drain() {
 					let _ = track.cancel.send(true);
+					let _ = track.handle.await;
 				}
```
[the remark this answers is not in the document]
kixelated: Fixed in dc97cd6e: shutdown now broadcasts `cancel` to every pump first, then awaits all handles, so no pump keeps streaming while we tear down. (Written by Claude)

rs/moq-gst/src/source/imp.rs
```diff
@@ -257,12 +205,20 @@ impl ElementImpl for MoqSrc {
 					gst::error!(CAT, obj = self.obj(), "failed to start session: {err:?}");
 					return Err(gst::StateChangeError);
 				}
-				let success = self.parent_change_state(transition)?;
-				let result = match success {
+				// Roll back the session we just started if the parent transition fails,
+				// otherwise it would keep running while the element stays in READY.
+				let success = match self.parent_change_state(transition) {
```
kixelated: let Ok(success) = ... else {

rs/moq-gst/src/source/imp.rs
```diff
@@ -440,72 +284,52 @@ async fn run_session(
...
+			// Full session shutdown: cancel every pump, then wait for them all to drop
+			// their pads. Cancel all up front (pumps only exit on their own `cancel`), or
+			// awaiting one at a time would let the not-yet-cancelled pumps keep streaming.
 			_ = shutdown.changed() => {
-				for (_, track) in active.drain() {
-					let _ = track.cancel.send(true);
+				let handles: Vec<_> = active
+					.drain()
+					.map(|(_, track)| {
+						let _ = track.cancel.send(true);
```
kixelated: Do we even need this? If the only sender is dropped, the receiver will unblock.

rs/moq-gst/src/source/imp.rs
```diff
@@ -440,72 +284,52 @@ async fn run_session(
...
+						track.handle
+					})
+					.collect();
+				for handle in handles {
+					let _ = handle.await;
```
kixelated: Is this a tokio::JoinSet or something? Does using that remove the need to reap handles?

rs/moq-gst/src/source/imp.rs
kixelated: Done in 86abee08 — `let Ok(success) = self.parent_change_state(transition) else { self.stop_session(); return Err(gst::StateChangeError); };`. `StateChangeError` is payload-free, so nothing is lost. (Written by Claude)

rs/moq-gst/src/source/imp.rs
kixelated: Removed in 86abee08 — pumps now live in a `tokio::JoinSet`, so `join_next()` is the wakeup and the `done` channel is gone entirely. (Written by Claude)

rs/moq-gst/src/source/imp.rs
kixelated: Yes — switched to `tokio::JoinSet` in 86abee08. It owns the tasks (no JoinHandle storage) and `join_next()` reaps them; I keep an `AbortHandle` per track only for `is_finished()` to prune the metadata map reconcile diffs against. Shutdown drains via `while pumps.join_next().await.is_some() {}`. (Written by Claude)

PR #1637, opened 2026-06-06
feat(moq-json): JSON Merge Patch snapshot/delta helper; route hang catalog through it

kixelated:
## Summary

Adds a generic JSON snapshot/delta helper (RFC 7396 JSON Merge Patch) in both Rust (`moq-json`) and TypeScript (`@moq/json`), and routes the hang catalog through it with **deltas disabled** so today's wire format is byte-identical. The goal is to land the infrastructure and get consumers onto the new reader now, so JSON Merge Patch deltas can be enabled later without touching every consumer.

A track carries a JSON value as a series of groups; each group is self-contained (frame 0 = full snapshot, frames 1.. = merge patches applied in order). A consumer jumps to the newest group, reads the snapshot, and applies deltas, so a late joiner never needs older groups. Deltas are opt-in via `max_delta_ratio` (`None`/`undefined` = off).

## What's here

**New `rs/moq-json` crate**
- `diff.rs`: hand-rolled RFC 7396 diff returning `{ patch, forced_snapshot }`. Arrays are replaced wholesale in a delta (still typically smaller than a full snapshot); only a literal-null value forces a snapshot, since merge patch reads null as a key deletion. Apply uses the `json-patch` crate's `merge`.
- `lib.rs`: `JsonProducer<T>` / `JsonConsumer<T>` owning a `TrackProducer` / `TrackSubscriber`. The producer picks snapshot vs delta from `max_delta_ratio`, with a 256-frame-per-group cap so late joiners can always read frame 0 before eviction. The consumer jumps to the newest group and reconstructs, modeled on the existing catalog consumer.

**New `js/json` (`@moq/json`) package** — a faithful TS port mirroring the Rust semantics.

**Catalog migration (deltas off, behavior-preserving)**
- Rust: `catalog/hang/producer.rs` holds the stateful `JsonProducer` behind an `Arc<Mutex>` so `Producer` stays `Clone`; `consumer.rs` wraps `JsonConsumer<Catalog>`, keeping `Consumer::next()` / `From<TrackSubscriber>`. Snapshots serialize `&T` directly, so bytes match `Catalog::to_string()` exactly.
- JS: `publish/broadcast.ts` (`#serveCatalog` → a per-request `Json.Producer`) and `watch/broadcast.ts` (hang fetch → `Json.Consumer`).

## Why dev

Targets `dev` because it makes breaking public-API changes in `moq-mux`: the catalog `Consumer` loses its (unused) `pub track` field and `Producer::finish()` now returns `crate::Result`. The new `moq-json` crate is also adapted to dev's `TrackSubscriber` / `.subscribe()` moq-net API.

## Test plan

- [x] Rust: `moq-json` unit + golden-vector tests; `moq-mux` catalog tests (the behavior-preservation gate) all pass.
- [x] JS: `@moq/json` 28 tests; `js/json`, `js/publish`, `js/watch` typecheck.
- [x] Cross-impl parity locked by a shared fixture (`rs/moq-json/tests/vectors.json`) asserted by both the Rust and TS suites.
- [x] `cargo check --workspace`, clippy, `cargo fmt --check`, `cargo sort` clean; biome clean.

## Cross-package sync

- `rs/moq-net` wire/API: unchanged (no wire change; deltas off ⇒ byte-identical).
- `rs/hang` / `js/hang` catalog format: unchanged.
- `doc/concept`: no change needed now (wire format preserved); worth a docs pass when deltas actually ship.
- No `moq-ffi`/wrapper changes (internal helper; `moq-ffi` only calls `catalog.finish()`, which converts the new `moq_mux::Error` transparently).

## Follow-ups

- Enable deltas by setting `max_delta_ratio` where the catalog producers are constructed — no consumer changes needed.
- A moq-net-free "pure transform" core (`Group(Bytes)`/`Frame(Bytes)` instructions) was deliberately deferred.

### DIFF
--- a/js/json/src/consumer.ts
+++ b/js/json/src/consumer.ts
@@ -1,0 +1,91 @@
+import type * as Moq from "@moq/net";
+import { Signal } from "@moq/signals";
+import type * as z from "zod/mini";
+
+import { merge } from "./diff.ts";
+
+/**
+ * Consumes a JSON value from a track, reconstructing it from snapshots and deltas.
+ *
+ * Always jumps to the newest group, reads its snapshot, and applies deltas in order, yielding
+ * the reconstructed value after each frame. A late joiner never needs older groups.
+ */
+export class Consumer<T> {
+	#track: Moq.Track;
+	#schema?: z.ZodMiniType<T>;
+
+	#group?: Moq.Group;
+	#current?: unknown;
+	#framesRead = 0;
+
+	constructor(track: Moq.Track, schema?: z.ZodMiniType<T>) {
+		this.#track = track;
+		this.#schema = schema;
+	}
+
+	/** Get the next reconstructed value, or `undefined` once the track ends. */
+	async next(): Promise<T | undefined> {
+		for (;;) {
+			// Jump to the newest group, discarding any older ones, and reset reconstruction.
+			const groups = this.#track.state.groups.peek();
+			if (groups.length > 0 && groups.at(-1) !== this.#group) {
+				while (groups.length > 1) groups.shift()?.close();
+				this.#group = groups[0];
+				this.#current = undefined;
+				this.#framesRead = 0;
+			}
+
+			const group = this.#group;
+			if (!group) {
+				const closed = this.#track.state.closed.peek();
+				if (closed instanceof Error) throw closed;
+				if (closed) return undefined;
+
+				await Signal.race(this.#track.state.groups, this.#track.state.closed);
+				continue;
+			}
+
+			// Frame 0 of a group is a snapshot, the rest are merge patches.
+			const frame = group.state.frames.peek().shift();
+			if (frame) return this.#apply(frame);
+
+			// The current group has no pending frame.
+			const groupClosed = group.state.closed.peek();
+			if (groupClosed) {
+				if (groupClosed instanceof Error) throw groupClosed;
+
+				// The group is exhausted; wait for a newer one.
+				if (this.#group === group) this.#group = undefined;
+				const closed = this.#track.state.closed.peek();
+				if (closed instanceof Error) throw closed;
+				if (closed) return undefined;
+
+				await Signal.race(this.#track.state.groups, this.#track.state.closed);
+				continue;
+			}
+
+			// Group open but no frame yet: wait for a frame, a newer group, or track close.
+			await Signal.race(group.state.frames, this.#track.state.groups, this.#track.state.closed);
+		}
+	}
+
+	async *[Symbol.asyncIterator](): AsyncIterator<T> {
+		for (;;) {
+			const value = await this.next();
+			if (value === undefined) return;
+			yield value;
+		}
+	}
+
+	#apply(frame: Uint8Array): T {
+		const parsed = JSON.parse(new TextDecoder().decode(frame));
+		if (this.#framesRead === 0) {
+			this.#current = parsed;
+		} else {
+			this.#current = merge(this.#current, parsed);
+		}
+		this.#framesRead += 1;
+
+		return this.#schema ? this.#schema.parse(this.#current) : (this.#current as T);
+	}
+}
--- a/js/json/src/consumer.ts
+++ b/js/json/src/consumer.ts
@@ -1,14 +1,13 @@
 import type * as Moq from "@moq/net";
-import { Signal } from "@moq/signals";
 import type * as z from "zod/mini";
-
 import { merge } from "./diff.ts";
+import type { Config } from "./producer.ts";
 
 /**
  * Consumes a JSON value from a track, reconstructing it from snapshots and deltas.
  *
- * Always jumps to the newest group, reads its snapshot, and applies deltas in order, yielding
- * the reconstructed value after each frame. A late joiner never needs older groups.
+ * Reads each group's snapshot (frame 0) and applies the following frames as merge patches,
+ * yielding the reconstructed value after each one.
  */
 export class Consumer<T> {
 	#track: Moq.Track;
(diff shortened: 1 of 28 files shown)


js/json/src/consumer.ts
```diff
@@ -0,0 +1,91 @@
...
+	constructor(track: Moq.Track, schema?: z.ZodMiniType<T>) {
+		this.#track = track;
+		this.#schema = schema;
+	}
+
+	/** Get the next reconstructed value, or `undefined` once the track ends. */
+	async next(): Promise<T | undefined> {
+		for (;;) {
+			// Jump to the newest group, discarding any older ones, and reset reconstruction.
+			const groups = this.#track.state.groups.peek();
```
kixelated: Please absolutely do not access internal state. Wtf why is this public, make it `#state`. 

`this.#track.nextGroup()` should return a group with a higher sequence number (might be rust only). `this.#track.recvGroup()` can return any group, but we only want the latest.

js/json/src/producer.ts
```diff
@@ -0,0 +1,105 @@
...
+
+export interface Config {
+	// Controls whether the producer emits deltas (merge patches) instead of full snapshots.
+	//
+	// `undefined` disables deltas: every change is published as a new snapshot group.
+	//
+	// A number enables deltas: a delta is appended to the current group as long as the group's
+	// total size stays within `maxDeltaRatio` times the size of a fresh snapshot; otherwise a
+	// new snapshot group is started.
+	maxDeltaRatio?: number;
```
kixelated: deltaRatio

js/json/src/producer.ts
```diff
@@ -0,0 +1,105 @@
...
+		// Serialize once; parse it back to a normalized JSON value for diffing and comparison
+		// (dropping `undefined` fields, matching what lands on the wire).
+		const text = JSON.stringify(valid);
+		const json = JSON.parse(text);
+		if (this.#last !== undefined && deepEqual(this.#last, json)) return;
+
+		const snapshot = new TextEncoder().encode(text);
+		const delta = this.#delta(json, snapshot.length);
+		if (delta) {
+			// biome-ignore lint/style/noNonNullAssertion: a delta is only produced with an open group.
```
kixelated: Just throw `this.#group` in the above conditional instead of explaining

js/json/src/producer.ts
```diff
@@ -0,0 +1,105 @@
...
+	#track: Moq.Track;
+	#config: Config;
+	#schema?: z.ZodMiniType<T>;
+
+	#group?: Moq.Group;
+	#last?: unknown;
+	#groupBytes = 0;
+	#groupFrames = 0;
+
+	constructor(track: Moq.Track, config: Config = {}, schema?: z.ZodMiniType<T>) {
```
kixelated: Can you put the schema in the config?

js/json/src/consumer.ts
```diff
@@ -0,0 +1,91 @@
...
+ */
+export class Consumer<T> {
+	#track: Moq.Track;
+	#schema?: z.ZodMiniType<T>;
+
+	#group?: Moq.Group;
+	#current?: unknown;
+	#framesRead = 0;
+
+	constructor(track: Moq.Track, schema?: z.ZodMiniType<T>) {
```
kixelated: schema in a config.

js/publish/src/broadcast.ts
```diff
@@ -79,12 +80,20 @@ export class Broadcast {
...
+					? new Json.Producer<Catalog.Root>(request.track)
+					: undefined;
+
 			effect.run((effect) => {
 				if (effect.get(request.track.state.closed)) return;
 
 				switch (request.track.name) {
 					case Broadcast.CATALOG_TRACK:
-						this.#serveCatalog(request.track, effect);
+						// biome-ignore lint/style/noNonNullAssertion: created above for this track name.
```
kixelated: Created the catalogProducer within the effect.run to avoid this.

js/watch/src/broadcast.ts
```diff
@@ -157,9 +158,15 @@ export class Broadcast {
...
 
+		// The hang catalog is reconstructed from snapshots (and future deltas) via @moq/json;
+		// MSF stays on its own one-blob-per-group fetch.
+		const catalogConsumer =
+			format === "hang" ? new Json.Consumer<Catalog.Root>(track, Catalog.RootSchema) : undefined;
+
 		const fetchNext =
 			format === "hang"
-				? async () => Catalog.fetch(track)
+				? // biome-ignore lint/style/noNonNullAssertion: defined whenever format is "hang".
```
kixelated: Please fix this. `if (format === hang)  do both of these things at once.

rs/moq-mux/src/catalog/hang/consumer.rs
```diff
@@ -6,47 +6,25 @@ use crate::Result;
...
-			}
-		}
-
-		if track_finished {
-			Poll::Ready(Ok(None))
-		} else {
-			Poll::Pending
+		match self.inner.poll_next(waiter) {
+			Poll::Ready(result) => Poll::Ready(result.map_err(Into::into)),
+			Poll::Pending => Poll::Pending,
```
kixelated: ready!

rs/moq-mux/src/catalog/hang/producer.rs
```diff
@@ -8,9 +8,13 @@ use base64::Engine;
...
 /// You'll have to call [`lock`](Self::lock) to update and publish the catalog.
 /// Both the hang (`catalog.json`) and MSF (`catalog`) tracks are published on drop of the guard.
+///
+/// The hang track is published through [`moq_json`], which currently emits one snapshot per
+/// group (deltas disabled). This routes catalog publishing through the JSON merge-patch helper
+/// so deltas can be enabled later without changing the wire format used today.
 #[derive(Clone)]
 pub struct Producer {
-	hang_track: moq_net::TrackProducer,
+	hang: Arc<Mutex<moq_json::JsonProducer<hang::Catalog>>>,
```
kixelated: Maybe JsonProducer should have an internal Arc<Mutex to match how other MoQ Producers work?

js/json/src/consumer.ts
kixelated: Rewrote the consumer to use only the public Track API: `nextGroupOrdered()` for higher-sequence groups + `group.readFrame()`. No more `.state` access. Frame 0 of each group is the snapshot, the rest are merge patches. (Strict jump-to-latest isn't exposed in the JS Track API, so it processes groups in order and converges to the latest, matching the old `Catalog.fetch`/`readFrame` behavior.) (Written by Claude)

js/json/src/producer.ts
kixelated: Renamed `maxDeltaRatio` → `deltaRatio` (and the Rust field `max_delta_ratio` → `delta_ratio`). (Written by Claude)

js/json/src/producer.ts
kixelated: Done — the condition is now `if (delta && this.#group)`, which narrows the type so the non-null assertion (and the comment) are gone. (Written by Claude)

js/json/src/producer.ts
kixelated: Moved `schema` into a `Config<T>`: `new Producer(track, { deltaRatio, schema })`. (Written by Claude)

js/json/src/consumer.ts
kixelated: The `Consumer` now takes the same `Config<T>` so the schema lives there too: `new Consumer(track, { schema })`. (Written by Claude)

js/publish/src/broadcast.ts
kixelated: Moved the `Json.Producer` creation into the `effect.run` switch case, dropping the pre-creation and the non-null assertion. (Written by Claude)

js/watch/src/broadcast.ts
kixelated: Consolidated into a single `if (format === "hang") { ... } else { ... }` block that builds `fetchNext` once. (Written by Claude)

rs/moq-mux/src/catalog/hang/consumer.rs
kixelated: Now `let result = ready!(self.inner.poll_next(waiter)); Poll::Ready(result.map_err(Into::into))`. (Written by Claude)

rs/moq-mux/src/catalog/hang/producer.rs
kixelated: Done — `moq_json::Producer` now holds an internal `Arc<Mutex<Inner>>` and is `Clone` like other MoQ producers, so the catalog `Producer` holds it directly (no outer `Arc<Mutex>`). Also renamed the types so they read as `moq_json::Producer` / `Consumer` / `Config`. (Written by Claude)

js/json/src/consumer.ts
kixelated: Follow-up: also renamed `Track.nextGroupOrdered()` → `nextGroup()` in `@moq/net` (and dropped the deprecated `nextGroup` alias, since `recvGroup` covers arrival order), so the consumer now calls the clean `nextGroup()`. Matches `Track::next_group` on the Rust side. (Written by Claude)

PR #1640, opened 2026-06-06
moq-net: RAII broadcast announcements via OriginPublish; remove broadcast abort/close

kixelated:
## Summary

Makes the origin announcement lifecycle sans-io and reframes what a broadcast is.

Previously `publish_broadcast` spawned a task per broadcast (`web_async::spawn`) that awaited `broadcast.closed()` and then removed it from the origin tree, which required a runtime. Now publishing returns an **`OriginPublish` guard** whose `Drop` does the removal synchronously, so the origin no longer watches closure and no longer spawns. This is the first concrete step toward driving `moq-net` from `poll_xxx`/wakers instead of `tokio`.

Paired conceptual change: **a broadcast is just a collection of tracks, so it can no longer be "closed" with a code.** `BroadcastProducer::abort` / `BroadcastDynamic::abort` are removed. A broadcast ends only when every producer is dropped; pending dynamic track requests are rejected when the last dynamic handler goes away. There is no wire code for broadcast closure, so nothing is lost on the wire. You can now **unannounce without closing** (drop the guard while the producer keeps serving tracks).

## API changes (moq-net, breaking → targeting `dev`)

- `publish_broadcast` returns `Result<OriginPublish, Error>` (was `bool`). Hold the guard to stay announced; drop it (or call `OriginPublish::unannounce`) to remove it.
- `create_broadcast` returns `Result<BroadcastPublish, Error>`; `BroadcastPublish` derefs to `BroadcastProducer` and unannounces on drop.
- New `Error::Loop` (broadcast's hop chain already contains this origin) and reused `Error::Unauthorized` (path outside allowed prefixes) distinguish the two reject reasons that were previously a bare `false`/`None`.
- Removed `BroadcastProducer::abort` and `BroadcastDynamic::abort`.

The `Lock`-based origin **tree is unchanged** — connections stay scoped to their place in the tree so they don't contend on a global lock.

## Callers

- **lite/ietf subscribers**: store the guard in the per-broadcast map so Ended / restart / reflected-loop drop it. Side benefit: unannounce-on-Ended is now prompt instead of waiting for the broadcast to fully close.
- **stats / relay / cli / gst / rtc / boy / examples**: hold the guard for the broadcast's lifetime.
- **FFI edges (libmoq, moq-ffi)**: keep fire-and-forget by spawning a small boundary watcher that drops the guard when the broadcast closes. Public signatures are unchanged, so the py/swift/kt/go wrappers and `doc/lib` need no changes.

## Cross-package sync

Skipped `js/net` and `doc/concept`: this is a Rust-internal lifecycle refactor (RAII guard + removing `abort`); the **wire protocol is unchanged**, and the JS `@moq/net` API has its own idioms. Worth a follow-up if we want JS to mirror the guard/unannounce model.

## Test plan

- [x] `cargo test -p moq-net` — 366 pass (origin tests now exercise `OriginPublish::drop` via a `#[cfg(test)]` helper that reproduces the old fire-and-forget behavior)
- [x] `cargo test -p moq-ffi` — 18 pass (exercises the relocated `announce` watcher)
- [x] `cargo check --all-targets` across every changed crate — clean
- [x] `cargo clippy -p moq-net --all-targets` (via nix) — clean
- [x] `cargo fmt` (via nix) applied
- [ ] full `just check` — run once any concurrent local `moq-cli` demo (shared target dir) has exited

### DIFF
--- a/rs/libmoq/src/origin.rs
+++ b/rs/libmoq/src/origin.rs
@@ -192,7 +192,20 @@ impl Origin {
 		broadcast: moq_net::BroadcastConsumer,
 	) -> Result<(), Error> {
 		let origin = self.active.get_mut(origin).ok_or(Error::OriginNotFound)?;
-		origin.publish_broadcast(path, broadcast);
+
+		// Errors only if the origin rejects the publish (loop / out of scope). libmoq origins
+		// are full-scope and broadcasts carry empty hop chains, so this is unreachable; ignore
+		// it to match the prior fire-and-forget contract.
+		if let Ok(publish) = origin.publish_broadcast(path, broadcast.clone()) {
+			// Auto-unannounce when the broadcast closes (all producers dropped). This keeps the
+			// previous behavior now that the origin no longer watches closure itself; the spawn
+			// lives at the FFI boundary, which is inherently runtime-bound.
+			tokio::spawn(async move {
+				broadcast.closed().await;
+				drop(publish);
+			});
+		}
+
 		Ok(())
 	}
 
--- a/rs/libmoq/src/origin.rs
+++ b/rs/libmoq/src/origin.rs
@@ -25,6 +25,10 @@ pub struct Origin {
 	/// Active origin producers for publishing and consuming broadcasts.
 	active: NonZeroSlab<moq_net::OriginProducer>,
 
+	/// Announcement guards from `publish`. Removing an entry (via `unpublish`) drops the
+	/// guard, which unannounces the broadcast.
+	published: NonZeroSlab<moq_net::OriginPublish>,
+
 	/// Broadcast announcement information (path, active status).
 	announced: NonZeroSlab<(String, bool)>,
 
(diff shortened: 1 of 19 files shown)


rs/libmoq/src/origin.rs
```diff
@@ -192,7 +192,20 @@ impl Origin {
 		broadcast: moq_net::BroadcastConsumer,
 	) -> Result<(), Error> {
 		let origin = self.active.get_mut(origin).ok_or(Error::OriginNotFound)?;
-		origin.publish_broadcast(path, broadcast);
+
+		// Errors only if the origin rejects the publish (loop / out of scope). libmoq origins
+		// are full-scope and broadcasts carry empty hop chains, so this is unreachable; ignore
+		// it to match the prior fire-and-forget contract.
+		if let Ok(publish) = origin.publish_broadcast(path, broadcast.clone()) {
```
kixelated: I think we need to store it in the slab? And add an `unpublish` method.

Also please return an error instead of assuming it will always be impossible.

rs/moq-cli/src/client.rs
```diff
@@ -6,7 +6,9 @@ use url::Url;
 pub async fn run_client(client: moq_native::Client, url: Url, name: String, publish: Publish) -> anyhow::Result<()> {
 	// Create an origin producer to publish to the broadcast.
 	let origin = moq_net::Origin::random().produce();
-	origin.publish_broadcast(&name, publish.consume());
+	let _publish = origin
+		.publish_broadcast(&name, publish.consume())
+		.expect("origin should allow publishing");
```
kixelated: .context instead of .expect for all of these that use anyhow.

rs/moq-gst/src/sink/imp.rs
```diff
@@ -393,11 +393,10 @@ async fn run_session(
...
 
-	anyhow::ensure!(
-		origin.publish_broadcast(&settings.broadcast, broadcast_consumer),
-		"failed to publish broadcast {}",
-		settings.broadcast
-	);
+	// Held for the lifetime of this task; dropping it (on return) unannounces the broadcast.
+	let _publish = origin
+		.publish_broadcast(&settings.broadcast, broadcast_consumer)
+		.map_err(|err| anyhow::anyhow!("failed to publish broadcast {}: {}", settings.broadcast, err))?;
```
kixelated: Please just use .context, I don't need to know the name.

rs/libmoq/src/origin.rs
kixelated: Done. `publish` now stores the `OriginPublish` guard in a new `published` slab and returns a publish handle; added `moq_origin_unpublish` (C: `moq_origin_unpublish(handle)`) which removes the entry and drops the guard to unannounce. It now propagates the origin error via `?` instead of swallowing it. Note this drops the old auto-unannounce-on-close behavior: the announcement lives until `unpublish`, independent of the broadcast closing. (Written by Claude)

rs/moq-cli/src/client.rs
kixelated: Switched all the anyhow call sites (moq-cli client/server, clock/chat/video examples, moq-boy, moq-rtc) to `.context("failed to publish broadcast")?`. (Written by Claude)

rs/moq-gst/src/sink/imp.rs
kixelated: Done, now `.context("failed to publish broadcast")?` with no name. (Written by Claude)

PR #1646, opened 2026-06-07
fix(moq-gst): stop moqsrc panicking on backwards timestamps; globally unique pad ids

kixelated:
Follow-up to #1633 (review findings #1 and #2).

## 1. PTS-underflow panic (the important one) 🟠

`build_buffer` pins its PTS reference to the **first decoded** frame and computes `frame.timestamp - reference`. Frames arrive in **decode order**, so a B-frame's presentation timestamp can fall *before* the reference. `Timestamp`'s `Sub` is `checked_sub(..).expect("time overflow")` — it **panics on underflow** ([`moq-net/src/model/time.rs:281`](https://github.com/moq-dev/moq/blob/main/rs/moq-net/src/model/time.rs)). That crashes the pump task and, because the panic happens before the pad-removal cleanup, **leaks the pad** and kills the rendition. Open-GOP / leading-B-frame content triggers it.

Fix: `checked_sub` and clamp to `ClockTime::ZERO` instead of crashing.

## 2. Pad-id counter was per-session 🟡

`next_pad_id` reset to `0` on every `run_session`, despite the doc claiming a "process-unique counter". A rapid `Paused -> Ready -> Paused` restart could mint `video_0` while the previous session's cancelled pump was still removing *its* `video_0` (two same-named pads on the element). Promote it to a process-wide `AtomicU64` so it actually matches the doc and the collision window is gone.

## Test

Added a throwaway publisher that sends a keyframe at `ts=2000ms` then frames at `ts=1000ms+` (the backwards-timestamp case):

| build | result |
|---|---|
| before | `panicked at time.rs:281: time overflow` after **1 buffer**; pad leaked |
| after | streams all **201 buffers**, clean EOS, no panic |

Real H264 decode (`cdn.moq.dev/demo` `bbb.hang`) still decodes (20/20 frames). `clippy` + `fmt --check` clean.

Deliberately **not** included (noted in the #1633 review as lower priority): blocking `pad.push()` on tokio workers (situational), redundant CMAF init parsing per catalog update (minor perf), and reconcile unit tests.

(Written by Claude)

### DIFF
--- a/rs/moq-gst/Cargo.toml
+++ b/rs/moq-gst/Cargo.toml
@@ -13,7 +13,6 @@ publish = true
 [lib]
 name = "gstmoq"
 crate-type = ["cdylib", "rlib"]
-test = false
 doctest = false
 
 [dependencies]
--- a/rs/moq-gst/src/source/imp.rs
+++ b/rs/moq-gst/src/source/imp.rs
@@ -1,4 +1,5 @@
 use std::collections::HashMap;
+use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::{LazyLock, Mutex};
 use std::time::Duration;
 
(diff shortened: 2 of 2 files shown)


kixelated:
Addressed the nitpick in 720c83d5: extracted `relative_pts` and added an in-file unit test for the backward-PTS clamp (asserts a pre-reference timestamp yields `ClockTime::ZERO` instead of panicking, and that a forward delta is preserved). The lib had `test = false`, which kept `#[cfg(test)]` from compiling, so I dropped it; the test runs via `cargo test -p moq-gst` (moq-gst is excluded from default-members, so the normal `cargo test` is unaffected). Easy to revert that one line if you’d rather keep the crate test-free. (Written by Claude)

PR #1649, opened 2026-06-08
fix(native): surface terminal auth connect errors

Qizot:
Preserve unauthorized and forbidden connection failures as typed errors through native reconnect, libmoq, and UniFFI so clients can distinguish auth failures from retryable transport failures.

Previously an auth error would not be recognized since the client would fallback to connecting to e.g. websocket. If both transports failed we were left with generic error that completely stripped the unauthorized information.


rs/libmoq/src/session.rs
```diff
@@ -111,3 +109,46 @@ impl Session {
 		Ok(())
 	}
 }
+
+fn map_connect_error(err: anyhow::Error) -> Error {
+	if let Some(err) = err.downcast_ref::<moq_native::ConnectError>() {
```
kixelated: I think we need to replace anyhow if you want to do this.

In general, libraries use thiserror and apps use anyhow for this exact reason.

kixelated:
Note, this disable automatic retries on auth failures, which I guess is the right thing to do?

Qizot:
> Note, this disable automatic retries on auth failures, which I guess is the right thing to do?

I would say so? Given the token doesn't change I wouldn't expect the relay to somehow start accepting it.

PR #1651, opened 2026-06-08
refactor(moq-native): migrate from anyhow to thiserror

kixelated:
## Summary

Migrates `rs/moq-native` from `anyhow` to `thiserror`, following the library-crate convention in CLAUDE.md (libraries use `thiserror` with `#[from]`; binaries keep `anyhow`).

Errors are split per concern so each type fully describes its own failures:

- `moq_native::tls::Error` — certificate/key loading, shared by the client TLS config and the quinn/noq servers.
- `moq_native::quinn::Error`, `noq::Error`, `quiche::Error`, `iroh::Error`, `websocket::Error` — one self-contained error type per backend, each returning its own `Result`.
- `moq_native::Error` — the top-level type, aggregating all of the above via `#[from]` and holding the cross-cutting variants (log init, reconnect, backend selection, status code).

Each variant uses `#[from]` for upstream errors so `?` keeps working, with descriptive `#[source]` variants preserving the previous `.context()` messages. `anyhow` moves to `[dev-dependencies]` (still used by examples and tests).

Because each backend now owns its proto-level error types, the quinn/noq `ConnectionError` `#[from]` collision (which needed a `cfg` workaround when both features were enabled) is gone.

Downstream:
- `libmoq` gains a `#[from] moq_native::Error` (Arc-wrapped to keep `Error: Clone`), so the `.init()` calls use `?` instead of stringifying through anyhow.
- Examples and CLIs use `Ok(reconnect.closed().await?)` at the reconnect boundary.

## Compatibility

Technically breaking: public functions return `moq_native`'s error types instead of `anyhow::Error`. `?` propagation and the `anyhow::Context` trait (`.context()` on a `Result`) still compile for callers; only direct/typed returns into an `anyhow::Error` slot, or `anyhow::Error`-inherent methods (`.downcast()`, `.chain()`, …), need changes. `cargo-semver-checks` will detect the return-type change and bump the minor version (`0.16.3 → 0.17.0`).

## Cross-package sync

No paired packages apply: this is an internal error-handling refactor of one crate with no wire, public-API-shape, or catalog change. The crate still exposes `std::error::Error` types that `anyhow`-based consumers absorb via `?`.

## Test plan

- [x] `just rs check --workspace` (default features): clippy `-D warnings`, fmt, doc (`-D warnings`), cargo shear (no unused deps), cargo sort — clean
- [x] `cargo clippy --workspace --all-features --all-targets -- -D warnings` (quinn + noq + quiche + iroh + jemalloc + websocket together) — clean
- [x] `cargo check --workspace --no-default-features` — builds
- [x] `cargo test -p moq-native` — 53 passed, 0 failed
- [x] Downstream `libmoq` / `moq-relay` / `moq-cli` / `moq-boy` / examples build

(Written by Claude)

### DIFF
--- a/rs/hang/examples/video.rs
+++ b/rs/hang/examples/video.rs
@@ -34,7 +34,7 @@ async fn run_session(origin: moq_net::OriginConsumer) -> anyhow::Result<()> {
 	let reconnect = client.with_publish(origin).reconnect(url);
 
 	// Wait until the reconnect loop stops (e.g. timeout exceeded).
-	reconnect.closed().await
+	reconnect.closed().await.map_err(Into::into)
 }
 
 // Create a video track with a catalog that describes it.
--- a/rs/hang/examples/video.rs
+++ b/rs/hang/examples/video.rs
@@ -34,7 +34,7 @@ async fn run_session(origin: moq_net::OriginConsumer) -> anyhow::Result<()> {
 	let reconnect = client.with_publish(origin).reconnect(url);
 
 	// Wait until the reconnect loop stops (e.g. timeout exceeded).
-	reconnect.closed().await.map_err(Into::into)
+	Ok(reconnect.closed().await?)
 }
 
 // Create a video track with a catalog that describes it.
--- a/rs/libmoq/src/api.rs
+++ b/rs/libmoq/src/api.rs
@@ -98,7 +98,7 @@ pub unsafe extern "C" fn moq_log_level(level: *const c_char, level_len: usize) -
 			level => moq_native::Log::new(Level::from_str(level)?),
 		}
 		.init()
-		.map_err(|err| Error::InitFailed(std::sync::Arc::new(err)))?;
+		.map_err(|err| Error::InitFailed(std::sync::Arc::new(err.into())))?;
 
 		Ok(())
 	})
--- a/rs/libmoq/src/api.rs
+++ b/rs/libmoq/src/api.rs
@@ -97,8 +97,7 @@ pub unsafe extern "C" fn moq_log_level(level: *const c_char, level_len: usize) -
 			"" => moq_native::Log::default(),
 			level => moq_native::Log::new(Level::from_str(level)?),
 		}
-		.init()
-		.map_err(|err| Error::InitFailed(std::sync::Arc::new(err.into())))?;
+		.init()?;
 
 		Ok(())
 	})
--- a/rs/libmoq/src/session.rs
+++ b/rs/libmoq/src/session.rs
@@ -69,8 +69,7 @@ impl Session {
 		consume: Option<moq_net::OriginProducer>,
 	) -> Result<(), Error> {
 		let reconnect = moq_native::ClientConfig::default()
-			.init()
-			.map_err(|err| Error::Connect(Arc::new(err.into())))?
+			.init()?
 			.with_publish(publish)
 			.with_consume(consume)
 			.reconnect(url);
--- a/rs/libmoq/src/session.rs
+++ b/rs/libmoq/src/session.rs
@@ -70,7 +70,7 @@ impl Session {
 	) -> Result<(), Error> {
 		let reconnect = moq_native::ClientConfig::default()
 			.init()
-			.map_err(|err| Error::Connect(Arc::new(err)))?
+			.map_err(|err| Error::Connect(Arc::new(err.into())))?
 			.with_publish(publish)
 			.with_consume(consume)
 			.reconnect(url);
--- a/rs/moq-native/src/client.rs
+++ b/rs/moq-native/src/client.rs
@@ -111,7 +111,7 @@ impl ClientTls {
 	/// Loads the configured roots (or the platform's native roots if none),
 	/// optionally attaches a client identity for mTLS, and disables server
 	/// certificate verification when `disable_verify` is set.
-	pub fn build(&self) -> crate::Result<rustls::ClientConfig> {
+	pub fn build(&self) -> std::result::Result<rustls::ClientConfig, crate::tls::Error> {
 		use rustls::pki_types::CertificateDer;
 		use rustls::pki_types::PrivateKeyDer;
 		use rustls::pki_types::pem::PemObject;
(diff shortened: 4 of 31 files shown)


rs/libmoq/src/api.rs
```diff
@@ -98,7 +98,7 @@ pub unsafe extern "C" fn moq_log_level(level: *const c_char, level_len: usize) -
 			level => moq_native::Log::new(Level::from_str(level)?),
 		}
 		.init()
-		.map_err(|err| Error::InitFailed(std::sync::Arc::new(err)))?;
```
kixelated, line 101: Can you make a #[from] moq_native::Error instead?

rs/hang/examples/video.rs
```diff
@@ -34,7 +34,7 @@ async fn run_session(origin: moq_net::OriginConsumer) -> anyhow::Result<()> {
 	let reconnect = client.with_publish(origin).reconnect(url);
 
 	// Wait until the reconnect loop stops (e.g. timeout exceeded).
-	reconnect.closed().await
+	reconnect.closed().await.map_err(Into::into)
```
kixelated: Ok(reconnect.closed().await?)

rs/libmoq/src/session.rs
```diff
@@ -70,7 +70,7 @@ impl Session {
 	) -> Result<(), Error> {
 		let reconnect = moq_native::ClientConfig::default()
 			.init()
-			.map_err(|err| Error::Connect(Arc::new(err)))?
+			.map_err(|err| Error::Connect(Arc::new(err.into())))?
```
kixelated: Same, we shouldn't need string errors any longer.

rs/moq-native/src/error.rs
```diff
@@ -0,0 +1,372 @@
...
+	#[cfg(any(feature = "quinn", feature = "noq"))]
+	#[error("fingerprint request failed")]
+	FingerprintStatus(#[source] reqwest::Error),
+
+	#[cfg(any(feature = "quinn", feature = "noq"))]
+	#[error("failed to read fingerprint")]
+	ReadFingerprint(#[source] reqwest::Error),
+
+	// ── quinn backend ───────────────────────────────────────────────────
+	#[cfg(feature = "quinn")]
```
kixelated: Can you make a separate `quinn::Error` type? Same for every backend.

rs/libmoq/src/api.rs
kixelated: Done — libmoq now has a `#[from] moq_native::Error` variant (Arc-wrapped so `Error: Clone` is preserved, matching the other non-Clone variants), so this is just `.init()?`. (Written by Claude)

rs/libmoq/src/session.rs
kixelated: Done — same `#[from] moq_native::Error` path; no more stringifying through anyhow here. (Written by Claude)

rs/hang/examples/video.rs
kixelated: Done — switched to `Ok(reconnect.closed().await?)` here, in subscribe.rs, and the `tokio::select!` reconnect arms in clock/moq-cli/moq-boy. (Written by Claude)

rs/moq-native/src/error.rs
kixelated: Done — added `moq_native::tls::Error` (shared cert/key loading) plus self-contained `quinn`/`noq`/`quiche`/`iroh`/`websocket` `Error` types, each returning its own `Result`. The top-level `Error` now just aggregates them via `#[from]`. Nice side effect: the quinn/noq `ConnectionError` `#[from]` collision workaround is gone, since each backend owns its proto errors. (Written by Claude)

rs/moq-native/src/client.rs
```diff
@@ -112,7 +111,7 @@ impl ClientTls {
 	/// Loads the configured roots (or the platform's native roots if none),
 	/// optionally attaches a client identity for mTLS, and disables server
 	/// certificate verification when `disable_verify` is set.
-	pub fn build(&self) -> anyhow::Result<rustls::ClientConfig> {
+	pub fn build(&self) -> std::result::Result<rustls::ClientConfig, crate::tls::Error> {
```
kixelated: crate::tls::Result?

rs/moq-native/src/client.rs
```diff
@@ -126,43 +125,48 @@ impl ClientTls {
...
 				for cert in certs {
-					roots.add(cert).context("failed to add root cert")?;
+					roots.add(cert).map_err(crate::tls::Error::AddRoot)?;
 				}
 			}
 		}
 
 		// Allow TLS 1.2 in addition to 1.3 for WebSocket compatibility.
 		// QUIC always negotiates TLS 1.3 regardless of this setting.
 		let builder = rustls::ClientConfig::builder_with_provider(provider.clone())
```
kixelated: Is any other this duplicated? Could it go into crate::tls instead?

kixelated (commented):
IMO don't `use Result` if it means you have to fully qualify `std::result::Result`. I'd rather have the opposite; fully-qualify the type alias.

rs/moq-native/src/client.rs
kixelated: Done — added a public `crate::tls::Result<T>` alias; `build`, `load_roots`, and the quiche cert helpers now use it. (Written by Claude)

rs/moq-native/src/client.rs
kixelated: Yes, it was — the PEM root loading was duplicated between `ClientTls::build` and `ServerTlsConfig::load_roots`. Moved the whole `rustls::ClientConfig` construction (and `NoCertificateVerification`) into `crate::tls::client_config`, and extracted the shared file reading into `crate::tls::read_certs`, now used by the client, server `load_roots`, and `ServeCerts::load`. `ClientTls::build` is a one-line delegate. (Written by Claude)

rs/moq-native/src/error.rs
```diff
@@ -0,0 +1,125 @@
...
+#[non_exhaustive]
+pub enum Error {
+	#[error(transparent)]
+	Io(Arc<std::io::Error>),
+
+	#[error(transparent)]
+	MoqNet(#[from] moq_net::Error),
+
+	#[error("invalid log directive")]
+	Directive(Arc<tracing_subscriber::filter::ParseError>),
```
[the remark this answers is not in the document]
kixelated: Good catch — added `#[source]` to `Directive` so the `ParseError` stays in the chain. The other `Arc`-wrapped variants are unaffected: `Io`/`Tls`/the backend ones are `#[error(transparent)]` (which delegates `source()`), and `SetSubscriber`/`Logcat` already carry `#[source]`. (Written by Claude)

PR #1657, opened 2026-06-08
feat(publish): configurable audio codec settings via a codec config

kixelated:
## Summary

The audio encoder ([`js/publish/src/audio/encoder.ts`](js/publish/src/audio/encoder.ts)) hardcoded its Opus settings: bitrate (`channelCount * 32kbps`) and frame duration (20ms), with no way to tune the rest of the Opus knobs. Worse, the frame duration only fed the catalog `jitter` field. It was never set on the actual WebCodecs `AudioEncoderConfig`, so the encoder always used its default.

This PR makes the encoder settings configurable through a single, codec-specific `codec` prop on `Audio.EncoderProps`, stored as a signal so callers can change it at runtime (e.g. `broadcast.audio.codec.set(...)`).

## API shape

```ts
type Codec = OpusConfig; // future: | AacConfig

type OpusConfig = "opus" | OpusOptions;

type OpusOptions = {
  name: "opus";
  bitrate?: number;        // bits/sec, default channelCount * 32kbps
  frameDuration?: number;  // ms (2.5-60), default 20
  complexity?: number;     // 0-10
  packetlossperc?: number; // 0-100
  useinbandfec?: boolean;  // in-band FEC
  usedtx?: boolean;        // discontinuous transmission
};
```

`codec` defaults to `"opus"`. Pass the bare string for all defaults, or an object to tune. Grouping the settings under `codec` keeps codec-specific tuning in one place and gives a clean extension point: AAC slots in as `| AacConfig` on the union plus a branch in `normalizeCodec`/`#createConfig`, instead of piling loosely-related knobs onto `EncoderProps`.

## Behavior

- **Catalog-shaping settings** (`bitrate`, `frameDuration`) flow through `#config`/`#createConfig` so decoders see them. `frameDuration` is converted to µs and set on `opus.frameDuration` in `toEncoderConfig` (previously it only affected catalog `jitter`, never the encoder), and is still reflected in catalog `jitter`.
- **Encode-only knobs** (`complexity`, `packetlossperc`, `useinbandfec`, `usedtx`) map directly to [`OpusEncoderConfig`](https://developer.mozilla.org/en-US/docs/Web/API/AudioEncoder/configure#opus) and stay out of the catalog (they don't affect decoding). Collected by `#opusOptions`, omitting unset values so the browser keeps its defaults.
- A single `#runConfig` effect reconciles the catalog-shaping settings into `#config` when `codec` changes, with an equality guard so it can't loop on its own write. Both `#runConfig` and the `serve` reconfigure effect read the `codec` signal, so any runtime change reconfigures the encoder.
- Renamed `OPUS_FRAME_DURATION` → `OPUS_FRAME_DURATION_MS` to make the unit explicit.

## Deliberately not exposed

- **`format`** — the MoQ container relies on raw `"opus"` packets; `"ogg"` would break it.
- **`application` / `signal`** — already driven by the source `kind` (voice→voip/voice, other→audio/music).

The active `lib.dom.d.ts` already types `complexity`/`packetlossperc`/`useinbandfec`/`usedtx`/`frameDuration`/`format`; only `application`/`signal` are missing, which is why `OpusEncoderConfigExt` exists.

## Notes for reviewers

- The original task referenced `publish/src/main.ts` and a UI note about a fixed bitrate, but neither exists in the current tree. This PR is the encoder-side support only; UI controls can be wired to `broadcast.audio.codec` in a follow-up.
- Cross-Package Sync: a `js/publish` API addition with no Rust/`hang` counterpart, so no paired updates required.

## Test plan

- [x] `tsc --noEmit` passes (no encoder-related type errors)
- [x] `biome check` clean
- [ ] Manual: publish audio with `codec: { name: "opus", bitrate, frameDuration, complexity, useinbandfec, usedtx }` and confirm the encoder reconfigures and catalog jitter reflects the frame duration

(Written by Claude)

### DIFF
--- a/js/publish/src/audio/encoder.ts
+++ b/js/publish/src/audio/encoder.ts
@@ -10,7 +10,7 @@ import { type Kind, normalizeSource, type Source } from "./types";
 const GAIN_MIN = 0.001;
 const FADE_TIME = 0.2;
 const OPUS_BITRATE_PER_CHANNEL = 32_000;
-const OPUS_FRAME_DURATION = 20;
+const OPUS_FRAME_DURATION_MS = 20;
 
 // Compiled and inlined as a blob URL via vite-plugin-worklet.
 import CaptureWorklet from "./capture-worklet.ts?worklet";
@@ -25,6 +25,12 @@ export type EncoderProps = {
 	sampleRate?: number | Signal<number | undefined>;
 	channelCount?: number | Signal<number | undefined>;
 
+	// Opus bitrate in bits/sec. Defaults to channelCount * 32kbps when unset.
+	bitrate?: number | Signal<number | undefined>;
+
+	// Opus frame duration in milliseconds. Opus supports 2.5-60ms; defaults to 20ms (the real-time default).
+	frameDuration?: number | Signal<number | undefined>;
+
 	container?: Catalog.Container;
 };
 
@@ -38,6 +44,8 @@ export class Encoder {
 	volume: Signal<number>;
 	sampleRate: Signal<number | undefined>;
 	channelCount: Signal<number | undefined>;
+	bitrate: Signal<number | undefined>;
+	frameDuration: Signal<number | undefined>;
 
 	source: Signal<Source | undefined>;
 
@@ -63,9 +71,12 @@ export class Encoder {
 		this.volume = Signal.from(props?.volume ?? 1);
 		this.sampleRate = Signal.from<number | undefined>(props?.sampleRate);
 		this.channelCount = Signal.from<number | undefined>(props?.channelCount);
+		this.bitrate = Signal.from<number | undefined>(props?.bitrate);
+		this.frameDuration = Signal.from<number | undefined>(props?.frameDuration);
 
 		this.#signals.run(this.#runSource.bind(this));
 		this.#signals.run(this.#runGain.bind(this));
+		this.#signals.run(this.#runConfig.bind(this));
 		this.#signals.run(this.#runCatalog.bind(this));
 	}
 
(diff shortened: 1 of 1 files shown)


kixelated:
@coderabbitai review

js/publish/src/audio/encoder.ts
```diff
@@ -282,29 +338,45 @@ function requestedChannelCount(track: MediaStreamTrack): number | undefined {
 	return constraint.exact ?? constraint.ideal ?? constraint.max ?? constraint.min;
 }
 
+// Resolve the bare "opus" shorthand to the full config object so callers can read fields uniformly.
+function normalizeCodec(codec: Codec): OpusConfig {
+	if (codec === "opus") return { mime: "opus" };
+	return codec;
+}
```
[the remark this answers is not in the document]
kixelated: Declining this one. Three reasons: (1) `AudioEncoder.configure()` already range-validates these (complexity 0-10, packetlossperc 0-100, valid Opus frame durations) and throws on bad input, which the encoder's `error` callback surfaces via `track.close(err)`. (2) `normalizeCodec` runs inside reactive effects (`#runConfig` and the `serve` reconfigure effect), not at the `codec.set()` call site, so throwing here would fail inside the effect graph (same place `configure()` already fails) rather than deterministically at the API boundary. (3) The sibling video encoder doesn't range-validate its tuning knobs either (it clamps bitrate), so adding a throw-based validator only here would be inconsistent. Happy to revisit if you'd prefer fail-fast validation across both encoders. (Written by Claude)

js/publish/src/audio/encoder.ts
```diff
@@ -148,18 +175,47 @@ export class Encoder {
...
-			jitter: Catalog.u53(OPUS_FRAME_DURATION),
+			// jitter doubles as the Opus frame duration; toEncoderConfig converts it to µs for WebCodecs.
+			jitter: Catalog.u53(opus.frameDuration ?? OPUS_FRAME_DURATION_MS),
 		};
 	}
 
+	// Reconcile the catalog-shaping codec settings (bitrate, frame duration) when the codec changes,
+	// without waiting for a channel-count change. The equality guard stops this from looping on its
+	// own #config write.
+	#runConfig(effect: Effect): void {
```
kixelated, line 199: This is scary, can you instead do this when `#config` is set? Or make another signal, I don't want a signal that is derived from itself.

kixelated: Fixed in 58ec3d54. Removed the self-referential signal: added a `#captured` signal holding the worklet-observed format (sample rate + channel count), and `#config` is now a pure derivation of `#captured` + `codec` (mirroring how `#catalog` already derives from `#config`). The worklet handlers only write `#captured` now, never read-modify-write `#config`. (Written by Claude)

PR #1658, opened 2026-06-08
feat(hang,json,moq-mux): generic catalog with application extensions

kixelated:
Makes the catalog generic: the base carries only the media sections (`video`/`audio`), and applications add their own root sections (e.g. `scte35`) without modifying hang. Built on the `@moq/json` / `moq_json` merge-patch helper (#1655).

## Extension model
- **TS**: `Catalog.RootSchema` is a `z.looseObject({video, audio})` (unknown sections pass through). Extend with `z.extend(Catalog.RootSchema, { scte35 })` and publish/consume via `@moq/json`. Edit with `producer.mutate((c) => { c.scte35 = ... })`.
- **Rust**: `moq_mux::catalog::hang::Catalog<E: CatalogExt = ()>` carries `video`/`audio` as fields plus a flattened extension `E`; it derefs to `E` so `catalog.video` and `catalog.scte35` are both flat. Define an extension with an empty `impl CatalogExt`. Edit with the `Deref`/`DerefMut` guard from `producer.lock()`.

## Producer / consumer
- `moq_mux::catalog::Producer<E>` is the single writer — it emits **both** the hang `catalog.json` and the MSF `catalog` track (MSF derived from the base media sections). Format-specific readers stay as `catalog::hang::Consumer<E>` / `catalog::msf::Consumer`.
- `@moq/publish` exposes a stable `CatalogProducer` that exists independent of subscriptions and fans out to each subscriber (seeding new/reconnecting subscribers with the current catalog).

## moq_json
- New shared editing API: Rust `Producer::lock()` (RAII `Guard`, holds the mutex, publishes on drop); JS `Producer.mutate(fn)`.

## Removed (app sections → hang.live; recoverable from git history)
- hang: `user`/`chat`/`preview` (Rust) and `chat`/`user`/`preview`/`location`/`capabilities` (JS) dropped from the base catalog.
- `@moq/watch` / `@moq/publish`: `Chat`/`Location`/`User`/`Preview` exports removed.

## Breaking changes (no external consumers yet)
- `moq_mux` catalog `Producer`/`Consumer`/`Guard` are generic over the extension `E` and live at `catalog::Producer` (was `catalog::hang::Producer`); consumers yield `Catalog<E>`.
- `@moq/hang` `RootSchema` shape change + removed section schemas; `@moq/watch`/`@moq/publish` export removals.
- `hang::Catalog` is now `{video, audio}`.

## Test plan
- [x] `just check` equivalents: workspace clippy + `cargo doc -D warnings` + cargo sort clean
- [x] Rust: hang, moq-json, moq-mux tests (incl. `extension_roundtrip` through the real Producer→Consumer) + doctest
- [x] JS: `@moq/json`, `@moq/hang` (scte35 export/import + removal), `@moq/publish` (CatalogProducer fan-out/reconnect); tsc + biome clean

(Written by Claude)


js/publish/src/broadcast.ts
```diff
@@ -31,10 +23,12 @@ export class Broadcast {
...
 
-	location: Location.Root;
-	chat: Chat.Root;
-	preview: Preview;
-	user: User.Info;
+	// The live catalog producer, set while the `catalog.json` track is being served and cleared
+	// when it tears down. Applications can `mutate` it to add their own root sections (e.g.
+	// `scte35`) alongside the base `video`/`audio`. Because every owner mutates the same shared
+	// document, their sections compose instead of clobbering one another.
+	#catalog = new Signal<Json.Producer<Catalog.Root> | undefined>(undefined);
```
kixelated: I don't think this makes sense as a signal? It makes it difficult to use.

js/json/src/producer.ts
```diff
@@ -59,6 +59,21 @@ export class Producer<T> {
...
+	 * Mutate the current value and publish the result.
+	 *
+	 * The callback receives a mutable deep clone of the last-published value (or `{}` if nothing
+	 * has been published yet) and edits it in place. Independent owners can share a single Producer
+	 * and each touch their own keys: every mutate starts from the latest value, so sections compose
+	 * instead of clobbering one another. No-op if unchanged. Use {@link update} to replace the
+	 * whole value instead.
+	 */
+	mutate(fn: (value: T) => void): void {
+		const current = (this.#last === undefined ? {} : structuredClone(this.#last)) as T;
```
kixelated: I think value should be undefined if there's no last published catalog. Or make the constructor take the default value instead of allowing #last to be undefined.

js/hang/src/catalog/scte35.ts
```diff
@@ -0,0 +1,38 @@
...
+// A single splice event (ad break) signalled via SCTE-35.
+export const SpliceSchema = z.object({
+	// splice_event_id from the SCTE-35 message.
+	id: z.number(),
+
+	// Presentation time of the splice point, in seconds. Omitted for an immediate splice.
+	startTime: z.optional(z.number()),
+
+	// Duration of the break in seconds, when known. Omitted for a cancel or open-ended break.
+	duration: z.optional(z.number()),
```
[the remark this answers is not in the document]
kixelated: Done — `id` is now `z.number().check(z.int(), z.nonnegative())` and `startTime`/`duration` are non-negative, matching the repo's integer-schema idiom. (Written by Claude)

js/json/src/producer.ts
kixelated: Fixed: `lock()` now starts from the last-published value, falling back to a `Config.initial` value, and throws if there's neither rather than inventing `{}`. The catalog producer passes `initial: {}`. (Written by Claude)

js/publish/src/broadcast.ts
kixelated: Reworked: dropped the producer-as-signal. `broadcast.catalog` is now a stable `CatalogProducer` you can `lock()` any time, independent of subscriptions. It holds the canonical catalog and fans out to each on-demand subscription track, seeding new/reconnecting subscribers with the current snapshot. (A raw Track-level proxy couldn't do this since the JS Track is a destructive single-consumer buffer, so a reconnecting relay would miss the current catalog.) (Written by Claude)

rs/moq-json/src/lib.rs
```diff
@@ -116,12 +117,84 @@ impl<T: Serialize> Producer<T> {
...
 		self.inner.lock().unwrap().finish()
 	}
 }
 
+/// An RAII editing guard returned by [`Producer::lock`].
+///
+/// Derefs to the current value. Mutating it through [`DerefMut`] marks it dirty, and dropping a
+/// dirty guard publishes the edited value.
+pub struct Guard<T: Serialize> {
+	inner: Arc<Mutex<Inner>>,
```
kixelated: Doesn't this have to be a MutexGuard?

kixelated: Good catch — the guard now holds the `MutexGuard` for its lifetime (field is `MutexGuard<'a, Inner>`), so concurrent owners serialize through the mutex and the drop publishes through the held guard instead of re-locking. No more reconstruct-from-same-snapshot clobber window. (Written by Claude)

rs/moq-mux/src/catalog/hang/ext.rs
```diff
@@ -0,0 +1,131 @@
...
+}
+
+impl<E: CatalogExt> DerefMut for Catalog<E> {
+	fn deref_mut(&mut self) -> &mut E {
+		&mut self.ext
+	}
+}
+
+// Lets the producer derive the MSF track from the media sections.
+impl<E: CatalogExt> Media for Catalog<E> {
```
kixelated: Do we need a separate trait for this? Can't we define it on Catalog<E> instead?

rs/moq-mux/src/catalog/hang/ext.rs
```diff
@@ -0,0 +1,131 @@
...
+///     splice_id: u32,
+/// }
+///
+/// impl moq_mux::catalog::hang::CatalogExt for Scte35Ext {}
+/// ```
+pub trait CatalogExt: Serialize + DeserializeOwned + Default + Clone + Send + 'static {}
+
+/// The empty extension: a [`Catalog<NoExt>`] is just the base media catalog.
+#[derive(Serialize, Deserialize, Clone, Default, Debug, PartialEq, Eq)]
+pub struct NoExt {}
```
kixelated: Could something like `()` work instead of NoExt?

js/json/src/producer.ts
```diff
@@ -59,6 +68,38 @@ export class Producer<T> {
...
+		// don't invent an empty object: locking with nothing to start from is a usage error.
+		const base = this.#last ?? this.#config.initial;
+		if (base === undefined) {
+			throw new Error("lock() requires a prior update() or `initial` in the config");
+		}
+
+		const value = structuredClone(base) as T;
+		return {
+			value,
+			[Symbol.dispose]: () => this.update(value),
```
kixelated: Don't use `lock()` in Javascript, I meant Rust only.

In Javascript you should have a `mutate((T) => void)` function or something.

rs/moq-mux/src/catalog/hang/producer.rs
```diff
@@ -1,36 +1,60 @@
...
 /// You'll have to call [`lock`](Self::lock) to update and publish the catalog.
 /// Both the hang (`catalog.json`) and MSF (`catalog`) tracks are published on drop of the guard.
 ///
 /// The hang track is published through [`moq_json`], which currently emits one snapshot per
 /// group (deltas disabled). This routes catalog publishing through the JSON merge-patch helper
 /// so deltas can be enabled later without changing the wire format used today.
-#[derive(Clone)]
-pub struct Producer {
-	hang: moq_json::Producer<hang::Catalog>,
+pub struct Producer<T = hang::Catalog> {
```
kixelated: Shouldn't this take a Catalog<T> instead of hang::Catalog?

rs/moq-mux/src/catalog/hang/ext.rs
kixelated: Dropped the trait — `media()` is now an inherent method on `Catalog<E>` and the producer works directly with `Catalog<E>` (see below), so there's nothing generic to abstract over. (Written by Claude)

rs/moq-mux/src/catalog/hang/ext.rs
kixelated: Yes! `()` flattens to nothing with serde (verified), so I replaced `NoExt` with `impl CatalogExt for ()`. `Catalog<()>` is the base catalog. (Written by Claude)

js/json/src/producer.ts
kixelated: Done — JS now uses `mutate((value) => void)` (both `@moq/json` Producer and the publish CatalogProducer); the lock guard is Rust-only. (Written by Claude)

rs/moq-mux/src/catalog/hang/producer.rs
kixelated: Reworked: it's now `Producer<E: CatalogExt = ()>` operating on `Catalog<E>` (and `Consumer<E>`/`Guard<E>` likewise). hang::Catalog is no longer the producer's payload type. (Written by Claude)

rs/moq-mux/src/container/source.rs
```diff
@@ -52,7 +52,14 @@ impl CatalogSource {
 
 	pub(crate) fn poll_next(&mut self, waiter: &kio::Waiter) -> Poll<anyhow::Result<Option<hang::Catalog>>> {
 		match self {
-			Self::Hang(c) => c.poll_next(waiter).map_err(Into::into),
+			// The hang consumer yields a `Catalog<()>`; the import path only needs the base media
+			// sections, so reduce it to a plain `hang::Catalog`.
+			Self::Hang(c) => match c.poll_next(waiter) {
+				Poll::Ready(result) => {
```
kixelated: ready!

rs/moq-mux/src/container/source.rs
```diff
@@ -52,7 +52,14 @@ impl CatalogSource {
...
+					Poll::Ready(result.map(|opt| opt.map(|catalog| catalog.media())).map_err(Into::into))
+				}
+				Poll::Pending => Poll::Pending,
```
kixelated: Add a note to claude.md that `Poll::Pending => Poll::Pending` usually means that `ready!` can simplify things. Also map_err can usually be improved via `Ok(err?)`

kixelated (commented):
And can you remove the scte35 code from `hang` now that we've validated it's possible to extend? Feel free to leave it in moq-mux (maybe commented out?) because another developer will use it to extend the catalog.

rs/moq-mux/src/container/source.rs
kixelated: Done — collapsed to `let catalog = ready!(c.poll_next(waiter))?; Poll::Ready(Ok(catalog.map(|c| c.media())))`. (Written by Claude)

rs/moq-mux/src/container/source.rs
kixelated: Added a bullet under Rust Conventions in CLAUDE.md covering both: `Poll::Pending => Poll::Pending` → `ready!`, and `.map_err(Into::into)` → `Ok(x?)` (with the combined `ready!(...)?` idiom). (Written by Claude)

PR #1666, opened 2026-06-09
feat(watch): keep hidden canvas video subscribed

Qizot:
Add a no-suspend-when-hidden option for WebCodecs canvas playback so hidden tabs can keep the video track subscription active. The default still suspends hidden-tab video, and offscreen canvas suspension is preserved.

### DIFF
--- a/demo/web/src/index.html
+++ b/demo/web/src/index.html
@@ -24,7 +24,7 @@
 				- `name`: The broadcast name, otherwise we use the last part of the URL.
 				- `paused`: Video and audio are paused.
 				- `muted`: Audio is muted.
-				- `no-suspend-when-hidden`: Keep canvas video active when the tab is hidden.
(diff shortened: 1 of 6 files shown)


demo/web/src/index.html
```diff
@@ -24,6 +24,7 @@
 				- `name`: The broadcast name, otherwise we use the last part of the URL.
 				- `paused`: Video and audio are paused.
 				- `muted`: Audio is muted.
+				- `no-suspend-when-hidden`: Keep canvas video active when the tab is hidden.
```
kixelated: Can we come up with a better/shorter name? IDK like `visible` to force it to always be visible?

Qizot: `visible` is probably too generic and doesn't reflect what is actually happening, what about `background-playback` in the newest commit?

kixelated:
merged in as #1744

PR #1669, opened 2026-06-09
feat(moq-video,moq-cli): webcam capture and publish

kixelated:
## Summary

Implements webcam capture + publish in `moq-video` (previously a published 0.0.2 stub) and wires it into `moq-cli` as a `webcam` subcommand. This makes "capture and publish a webcam" a single command, and it **encodes on demand** — the camera only runs while someone is watching.

## Stack: ffmpeg-next

`moq-video` is built on `ffmpeg-next` (8.x, `features = ["device"]`). One system dependency covers both libavdevice **capture** and hardware **H.264 encoding**.

For context: iroh-live (the reference we looked at) uses `openh264` + native capture and only shells out to the `ffmpeg` binary for file import. We chose ffmpeg-next anyway specifically to get the platform hardware encoders (`h264_videotoolbox` / `h264_nvenc` / `h264_vaapi`).

## What's here

`rs/moq-video`:

- **`Camera`** (`capture.rs`) — webcam capture via libavdevice (avfoundation / v4l2 / dshow). Uses an FFI `av_find_input_format` since the safe `format::list()` helper is compiled out on ffmpeg ≥ 5.
- **`Encoder`** (`encoder.rs`) — tries a hardware-encoder name list, falls back to `libx264`. Lazy swscale to YUV420P, emits Annex-B. `EncoderKind` = Auto / Hardware / Software / Named.
- **`VideoProducer` / `publish_camera`** (`producer.rs`) — publish through `moq_mux::codec::h264::Import` with `Mode::Avc3` (the sink `moq-boy` uses), which handles catalog registration and framing. Presentation timestamps track real elapsed time (`Instant`).

`moq-cli`:

- New `webcam` publish subcommand behind a **non-default `webcam` feature**, so ffmpeg stays off normal `cargo install moq-cli` builds. Works under both `serve` and `publish`.

```bash
moq-cli publish --url https://relay.example.com --broadcast webcam.hang \
    webcam --device 0 --width 1280 --height 720 --fps 30
```

## On-demand encoding

Modeled on `moq-boy`, which pauses its emulator on `TrackProducer::used()` / `unused()`. The avc3 track and catalog are advertised immediately, but the camera stays closed (LED off, zero CPU) until a subscriber appears, and is released again when the last viewer leaves.

`publish_camera` is async: it spawns the blocking capture/encode loop and concurrently runs a demand monitor that toggles a `Gate` (Mutex + Condvar) off `used()` / `unused()`. There is no synchronous "is it used" accessor on `TrackProducer`, hence the async→blocking bridge. The blocking loop opens the camera + encoder lazily on the first watched frame and tears them down when idle; real-elapsed-time timestamps keep the timeline continuous across idle gaps.

## Testing

- `moq-video` and `moq-cli` (with and without `webcam`) build under nix.
- `cargo fmt` + `cargo clippy` clean in both feature configs.
- 3 `moq-video` unit tests pass, including encoding a synthetic frame and asserting valid Annex-B output (no camera required).

## Not done / reviewer notes

- **Live capture was not run** in this environment (macOS camera-permission/TCC would block a headless process). Only the software-encode path is unit-tested; the capture path and the on-demand open/release transitions compile and follow standard idioms but need a real device + subscriber to validate end-to-end.
- **Decoder/consumer** side (the mirror of `AudioConsumer`) is intentionally a follow-up; native subscribers can use `moq_mux` directly for now.
- Open question: keep `webcam` opt-in (current) vs. add to moq-cli `default` features. Opt-in avoids breaking `cargo install moq-cli` on machines without ffmpeg. Worth considering adding `--features webcam` to the CI clippy matrix so the moq-cli cfg block doesn't rot.

## Cross-package sync

Per the `rs/moq-cli` → `doc/bin/cli.md` row, the CLI docs are updated with the native `webcam` command. `moq-video` is a new crate not in the sync table.

Targeting `main`: purely additive, no wire-protocol or public-API changes to existing crates.

(Written by Claude)


rs/moq-video/src/capture.rs
```diff
@@ -0,0 +1,218 @@
...
+			};
+
+			match packet {
+				Some(packet) => self.decoder.send_packet(&packet)?,
+				None => {
+					self.decoder.send_eof()?;
+					return match self.decoder.receive_frame(&mut frame) {
+						Ok(()) => Ok(Some(frame)),
+						_ => Ok(None),
+					};
```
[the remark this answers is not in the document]
kixelated: Fixed in 8c9c8444 — after `send_eof()` only `Eof`/`EAGAIN` map to `Ok(None)`; any other decoder error now propagates via `Err(e.into())`. (Written by Claude)

rs/moq-video/src/encode/producer.rs
```diff
@@ -0,0 +1,112 @@
...
+/// Presentation timestamps are derived from a monotonic frame counter at
+/// the configured framerate, so the on-wire timeline stays smooth and
+/// gap-free even when camera delivery jitters.
+pub fn publish_camera(
+	broadcast: moq_net::BroadcastProducer,
+	catalog: moq_mux::catalog::Producer,
+	config: CameraPublishConfig,
+) -> Result<(), Error> {
+	let framerate = config.camera.framerate.unwrap_or(DEFAULT_FRAMERATE);
+
```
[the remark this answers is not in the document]
kixelated: Fixed in 8c9c8444 — added `Error::InvalidFramerate(u32)` and `publish_camera` now fails fast when framerate is 0. (Note: the on-demand refactor already switched timestamps to real elapsed time via `Instant`, so there's no longer a divide; but a 0 fps still yields a degenerate codec time base, so the guard stands.) (Written by Claude)

kixelated:
Addressed the outside-diff note on `Gate::close()` in 4b74491b: `close()` now also clears `active`, so when the broadcast drops while a viewer is still subscribed the capture worker takes the inactive path, releases the camera, and exits cleanly instead of running until the next publish fails. (Written by Claude)

doc/bin/cli.md
```diff
@@ -62,6 +62,28 @@ ffmpeg -i input.mp4 -f mpegts - | moq-cli publish - https://relay.example.com/an
 
 ### Publish a Webcam
 
+The `webcam` subcommand captures and encodes a camera directly, no external
+FFmpeg process required. It is gated behind the `webcam` feature, which pulls
+in a system FFmpeg (libav\*) build dependency:
+
+```bash
+cargo build --release -p moq-cli --features webcam
```
kixelated: or `cargo run`.

Make sure we enable this feature when building moq-cli for apt/brew/docker/etc.

rs/moq-cli/src/publish.rs
```diff
@@ -14,6 +14,44 @@ pub enum PublishFormat {
...
+	#[arg(long)]
+	pub width: Option<u32>,
+
+	/// Requested capture height.
+	#[arg(long)]
+	pub height: Option<u32>,
+
+	/// Capture/encode framerate.
+	#[arg(long, default_value_t = 30)]
+	pub fps: u32,
```
kixelated: Should this also be an Option?

rs/moq-cli/src/publish.rs
```diff
@@ -35,8 +73,20 @@ impl PublishDecoder {
...
 
+enum Source {
+	/// Decode a container read from stdin (or an HLS playlist).
+	Stream(PublishDecoder),
+	/// Capture a webcam. The producer is built on the blocking capture thread,
+	/// so we just carry the catalog and config here.
+	#[cfg(feature = "webcam")]
+	Webcam {
+		catalog: moq_mux::catalog::Producer,
+		config: moq_video::CameraPublishConfig,
```
kixelated: Can you make modules for `produce`/`consume` (or `import`/`export`).
or `encode/decode`.

Or make a `camera` module, that's probably better for feature support? IDK it's probably not needed.

Maybe `CameraPublishConfig` is just fine. But I would probably go with `moq_video::encode::CameraConfig`.

rs/moq-cli/src/publish.rs
kixelated: Done in 9fced449 — `fps` is now `Option<u32>` (no `default_value_t`); when omitted, moq-video's `DEFAULT_FRAMERATE` (30) applies, so the default lives in one place rather than being duplicated in the CLI. (Written by Claude)

rs/moq-cli/src/publish.rs
kixelated: Done in 9fced449. Split into `moq_video::camera` (capture) and `moq_video::encode` (encode + publish), leaving room for a sibling `decode` module later. `CameraConfig` (capture) → `camera::Config`; the high-level `CameraPublishConfig` → `encode::CameraConfig`; `Encoder`/`EncoderConfig`/`EncoderKind`/`VideoProducer`/`publish_camera` now live under `encode`. So the call site is `moq_video::encode::CameraConfig { camera: moq_video::camera::Config { .. }, .. }`, as you suggested. (Written by Claude)

doc/bin/cli.md
kixelated: Added the `cargo run --features webcam` form to the docs in 9fced449. The "enable it in the packaged binaries" part is bigger than it looks, so I want to check direction before touching the release pipeline (a break there only surfaces on a release tag, not in this PR's CI).

The four channels split cleanly in two:

**Easy — Docker + `nix build`/`nix run`:** both go through the nix overlay (`moqCliArgs`). I can add `--features webcam` plus `ffmpeg` + `pkg-config` build inputs and verify locally with `nix build .#moq-cli`. One wrinkle: the `moq-cli-x86_64-apple-darwin` *nix* package uses `crossX86Darwin`, which the overlay comments say only works for "pure-Rust artifacts with no cross buildInputs" — so that one cross package would need to either skip webcam or move to `pkgsCross` like moq-gst does.

**Hard — Homebrew tarball + `.deb`/`.rpm`:** these don't use nix. They're `cargo zigbuild --target <triple>[.2.34]`, cross-compiling to pinned glibc / both Apple arches. Enabling webcam there means (a) providing cross ffmpeg dev libs to `pkg-config` for each target under zig, and (b) declaring the runtime deps the dynamically-linked binary now needs — `libav*` in `packaging/moq-cli/nfpm.yaml` for `.deb`/`.rpm`, and `depends_on "ffmpeg"` in the brew formula. That's real cross-compilation + a packaging-policy call (making ffmpeg a hard runtime dep of the `moq` package), and I can't validate any of it without running the actual release workflow.

How would you like to proceed? Options I see:

1. I wire the nix/Docker channel now (testable) and open a follow-up issue for the zigbuild `.deb`/`.rpm`/brew work.
2. Make `webcam` a default feature of `moq-cli` instead — simplest "on everywhere", but then `cargo install moq-cli` and the existing zigbuild release jobs need ffmpeg present or they break, so it still requires the packaging work above.
3. I attempt all four channels in this PR (higher risk; release-pipeline breakage wouldn't show until a tag).

My recommendation is **1** + a tracked follow-up. (Written by Claude)

PR #1673, opened 2026-06-10
moq-wasm: browser/WASM bindings for moq-net (@moq/wasm)

kixelated:
Experiment: compile the real `moq-net` Rust implementation to WebAssembly and expose it to JavaScript via **wasm-bindgen**, driving the browser's WebTransport from inside WASM. The aim is to eventually replace the hand-written TypeScript moq-lite wire implementation in `@moq/net` (~10k LOC) with the canonical Rust one, so the protocol lives in one place.

This is not the uniffi path (`moq-ffi` targets the C ABI: Kotlin/Swift/Python/Go). Browsers need wasm-bindgen, so this is a separate sibling crate.

## New: `rs/moq-wasm` + `js/wasm` (`@moq/wasm`)

- `transport.rs` adapts [`web-transport-wasm`](https://crates.io/crates/web-transport-wasm) (browser WebTransport) to the `web-transport-trait` abstraction `moq-net` is generic over. Newtypes (orphan rule); shapes line up ~1:1.
- `lib.rs` wraps the consume path: `MoqSession.connect → consume → MoqBroadcast.subscribe → MoqTrack.recvGroup → MoqGroup.readFrame`. The generated `.d.ts` is clean, idiomatic TS (`Promise`-returning methods, `Uint8Array`, `bigint` sequences).
- `just wasm` builds wasm, runs `wasm-bindgen` (web target) into `js/wasm/dist`, and shrinks with `wasm-opt`. The generated `dist/` (incl. the `.wasm` binary) is **git-ignored**, built not committed. Tooling (`wasm-bindgen-cli`, `binaryen`, the `wasm32` target) comes from the Nix dev shell; cfg flags live in `.cargo/config.toml`.
- The crate is gated `cfg(target_arch = "wasm32")` with wasm-only deps, so it is empty on native and `cargo check --workspace` stays green.

## `moq-net` portability (preserves wire format and public API)

1. tokio `test-util` moved to dev-deps (test-only, unsupported on wasm).
2. `Send`/`Sync` relaxed to `MaybeSend`/`MaybeSync` on the private `SessionInner` and two `.boxed()` sites. A cfg-gated `MaybeSendBox` / `.maybe_boxed()` helper (`util.rs`) picks `BoxFuture`+`boxed()` on native, `LocalBoxFuture`+`boxed_local()` on wasm. Native is unchanged (`MaybeSend`/`MaybeSync` **are** `Send`/`Sync` there).
3. Timers and `Instant` routed through `web_async::time` (web-async 0.1.4) instead of `tokio::time`, which panics on wasm. web-async re-exports `tokio::time` on native and `wasmtimer` (`performance.now()` + `setTimeout`) on wasm. `tokio::time::pause`/`advance` test clocks still work.

`model/time.rs` keeps an unused wall-clock helper (`Timestamp::now` / `From<Instant>`) that isn't wasm-portable but has no callers, so it never runs; left as-is.

## Not yet wasm-ready (follow-ups)

- **moq-mux** (media muxing): `hang` + `moq-mux` enable tokio's `fs` feature (native filesystem). Gating `fs` behind a native-only cfg is a prerequisite. Dependency left out for now.
- **Publish path**: same shape as consume, not yet wrapped.

## Branch targeting

Targets **dev**: the `moq-wasm` consume wrappers call dev-only moq-net APIs from the moq-lite-05 reshape (`TrackConsumer`/`TrackSubscriber` split, `announced_broadcast`) that don't exist on main, and the branch is built on top of `dev`. So this can only land on dev.

## Test plan

- [x] `cargo check --workspace` (native) green; `moq-wasm` empty on native.
- [x] 367 moq-net tests pass; clippy `-D warnings` + fmt clean.
- [x] Builds for `wasm32-unknown-unknown`; `@moq/wasm` generated and bundles under esbuild; strict-mode TS consumer type-checks.
- [ ] End-to-end browser run against a relay (consume path) — not yet exercised.


js/wasm/README.md
```diff
@@ -0,0 +1,42 @@
...
+That compiles `rs/moq-wasm` for `wasm32-unknown-unknown`, runs `wasm-bindgen`
+(bundler target) into `dist/`, and shrinks the binary with `wasm-opt`. The
+required toolchain (wasm target, `wasm-bindgen-cli`, `binaryen`) is provided by
+the Nix dev shell.
+
+## Status
+
+This compiles and produces a typed, importable package today. It does **not yet
+run** in a browser: `moq-net` calls `tokio::time` directly, which panics on
+wasm (no clock / time driver). See [`rs/moq-wasm/README.md`](../../rs/moq-wasm/README.md)
```
kixelated: outdated

rs/moq-wasm/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+/// with wasm-bindgen's default `init` export, which loads the module itself.)
+#[wasm_bindgen]
+pub fn setup() {
+	console_error_panic_hook::set_once();
+	let _ = tracing_wasm::try_set_as_global_default();
+}
+
+/// A connected MoQ session.
+#[wasm_bindgen]
+pub struct MoqSession {
```
kixelated: Can you remove the `Moq` prefix? IDK it's annoying, and I'd rather do `Moq.Session`.

justfile
```diff
@@ -121,6 +121,16 @@ build:
...
+    if command -v wasm-bindgen &> /dev/null; then just wasm; fi
+
+# Build the browser/WASM bindings (rs/moq-wasm) into the @moq/wasm package.
+# wasm-bindgen + binaryen come from the Nix dev shell; the wasm32 target and
+# cfg flags (getrandom, web-sys unstable) are wired in .cargo/config.toml.
+wasm:
+    cargo build -p moq-wasm --target wasm32-unknown-unknown --release
+    wasm-bindgen --target web --out-name moq \
+    	--out-dir js/wasm/dist target/wasm32-unknown-unknown/release/moq_wasm.wasm
+    wasm-opt -Oz -o js/wasm/dist/moq_bg.wasm js/wasm/dist/moq_bg.wasm
```
kixelated: should this be in the rs/justfile? IDK

PR #1674, opened 2026-06-10
feat(moq-ffi): expose dynamic track requests

Qizot:
Add a UniFFI BroadcastDynamic wrapper so publishers can observe subscriber requests for tracks that do not exist yet. Wrap the API for Python and add Swift/Kotlin stream helpers plus language docs.

### DIFF
--- a/doc/lib/kt/moq.md
+++ b/doc/lib/kt/moq.md
@@ -111,6 +111,8 @@ dynamic.requestedTracks().collect { track ->
     if (track.name() == "alerts") {
         track.writeFrame("ready".encodeToByteArray())
         track.finish()
+    } else {
+        track.abort(404)
     }
 }
 ```
--- a/doc/lib/kt/moq.md
+++ b/doc/lib/kt/moq.md
@@ -94,6 +94,27 @@ audio.finish()
 broadcast.finish()
 ```
 
+### On-demand raw tracks
+
+Use a dynamic broadcast when subscribers should be able to request raw tracks that are not published yet:
+
+```kotlin
+import dev.moq.*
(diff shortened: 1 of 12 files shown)


doc/lib/kt/moq.md
```diff
@@ -94,6 +94,27 @@ audio.finish()
...
+val broadcast = MoqBroadcastProducer()
+val dynamic = broadcast.dynamic()
+
+origin.publish("events", broadcast)
+
+dynamic.requestedTracks().collect { track ->
+    if (track.name() == "alerts") {
+        track.writeFrame("ready".encodeToByteArray())
+        track.finish()
+    }
```
kixelated, line 116: IMO `else { track.abort(404) }`

rs/moq-ffi/src/producer.rs
```diff
@@ -30,6 +31,27 @@ pub struct MoqBroadcastProducer {
...
+	inner: moq_net::BroadcastDynamic,
+}
+
+impl DynamicProducer {
+	async fn requested_track(&mut self) -> Result<Option<Arc<MoqTrackProducer>>, MoqError> {
+		match self.inner.requested_track().await {
+			Ok(track) => Ok(Some(Arc::new(MoqTrackProducer {
+				inner: std::sync::Mutex::new(Some(track)),
+			}))),
+			Err(moq_net::Error::Closed | moq_net::Error::Dropped) => Ok(None),
```
kixelated: Why return None instead of Err? We'll only get an error if we abort the broadcast? IMO just copy the Rust API, remove the Option.

Issue #690, opened 2025-11-15
ABR


kixelated:
Good news, this will be easy to implement thanks to simulcast. The video track is [selected based on the resolution](https://github.com/kixelated/moq/blob/aea140d94c744c8319c21fead60ef5f243598c01/js/hang/src/watch/video/source.ts#L289) 
With (or potentially without) #689 we would also select the track based on our maximum bitrate as well.

For low latency targets, I've had the most success rounding down the estimated to select the rendition when BBR is being used. Obviously people write academic papers on this stuff so it can always be over-engineered, but let's start simple.

The only catch is that ABR transitions won't be smooth for higher latency targets because the jitter buffer is currently per-track. That's a problem for future us.

kixelated:
Closing because adaptive bitrate selection has landed: watch uses receive bandwidth estimates with a safety margin and bitrate-based rendition filtering/selection. (Written by GPT-5)

PR #1679, opened 2026-06-11
feat(publish): simulcast attribute on <moq-publish>, demoed in the web demo

### LINKED ISSUE #691: Simulcast Demo
Simulcast publishing is implemented (in Javascript only) via publishing a [hd and/or sd](https://github.com/kixelated/moq/blob/aea140d94c744c8319c21fead60ef5f243598c01/js/hang/src/publish/video/index.ts#L13) track. Obviously this is quite simple and should be extended to an arbitrary number of tracks, but it's good enough for now. The javascript consumer [selects the track](https://github.com/kixelated/moq/blob/aea140d94c744c8319c21fead60ef5f243598c01/js/hang/src/watch/video/source.ts#L126) currently based on pixel count only.

We should have some way of demonstrating and actually testing this functionality. Modifying the [demos](https://github.com/kixelated/moq/tree/main/js/hang-demo/src) to use simulcast would be a good first step. The web component would want to monitor the parent container's size or something to determine which resolution to request.

YogiSotho:
Closes #691.

The consumer half of #691 already landed in #1411: `<moq-watch>` monitors its own rendered size via ResizeObserver and feeds devicePixelRatio scaled dimensions into the rendition selector, and the watch UI has a quality dropdown. What was missing is the producer half: nothing ever enabled the sd encoder that has existed since #602, so the demos never exercised multi rendition selection.

## Changes

- **`<moq-publish>` gets a boolean `simulcast` attribute** that enables the existing sd encoder, gated on `invisible` the same way as hd.
- **The sd encoder defaults to a quarter of the source pixel count.** A new generic `maxScale` option on the encoder caps the encoded resolution to a fraction of the source (`maxScale: 0.25` = a quarter of the pixels = half the width and height). `Video.Root` applies it to the sd encoder when no config is provided. Scaling relative to the source keeps simulcast generic: it makes no assumption about the hd resolution or a fixed baseline. Without a cap the sd encoder would run at the full source resolution, duplicating hd.
- **`Source.Camera` defaults to 720p** via a static `DEFAULT_CONSTRAINTS` (`width`/`height` `ideal: 1280x720`), merged under any caller-supplied constraints so per-field overrides still win. The browser default of 640x480 is otherwise low quality.
- **Watch demo** wraps the player in a CSS resizable container; dragging the handle switches the Auto selection between `video/hd` and `video/sd` seamlessly. New prose on both pages explains the flow.
- **Docs**: the `<moq-publish>` attribute lists in `doc/lib/js/@moq/publish.md` and `doc/lib/js/env/web.md` documented attributes that do not exist (`device`, `audio`, `video`, `controls`). They now list the real set plus `simulcast`.

## Behavior change

JS API users who already enable sd explicitly (`video: { sd: { enabled: true } }`) previously got a full resolution duplicate of hd; they now get a quarter-resolution rendition. Passing any explicit config (`config: {}` included) opts out of the default entirely. Camera capture now requests 720p by default; pass explicit `width`/`height` constraints to override.

## Known limitations (intentionally out of scope)

- hd and sd each cap independently to ~90% of estimated send bandwidth, so combined output can exceed the uplink. A shared budget is #690 (ABR) territory.
- An arbitrary number of renditions and a `<moq-publish-ui>` toggle remain future work.

## Test plan

- [x] Unit tests for the Root sd default and the `maxScale` / opt-out branches (`js/publish/src/video/index.test.ts`, `bun test`)
- [x] `tsc --noEmit` clean for `@moq/publish` and `@moq/demo`; biome clean on changed files
- [ ] Manual (procedure, not yet run): `just demo`, publish from `publish.html`, watch via `index.html?broadcast=me.hang`; quality dropdown lists both renditions; resizing switches Auto between them; manual selection pins; toggling video off removes both renditions from the catalog

(Written by Claude)


demo/web/src/publish.ts
```diff
@@ -11,6 +13,14 @@ const publish = document.querySelector("moq-publish") as MoqPublish;
 const watch = document.getElementById("watch") as HTMLAnchorElement;
 const watchName = document.getElementById("watch-name") as HTMLSpanElement;
 
+// Ask the camera for 720p. Most browsers default to 640x480, which is already below
+// the sd rendition's 480p cap, so hd and sd would otherwise be nearly identical.
+publish.video.watch((video) => {
+	if (video instanceof Source.Camera) {
+		video.constraints.set({ width: { ideal: 1280 }, height: { ideal: 720 } });
```
kixelated: Maybe just make these the default contraints within `Source.Camera`. The browser selects some low quality garbage by default.

js/publish/src/video/index.ts
```diff
@@ -21,6 +21,12 @@ export class Root {
 	static readonly TRACK_SD = "video/sd";
 	static readonly PRIORITY = Catalog.PRIORITY.video;
 
+	// The default cap for the sd rendition: 480p, the baseline of the encoder's bitrate table.
+	// Without a cap the sd encoder would run at the full source resolution, duplicating hd.
+	// Only applied when no sd config is provided; passing any config (even an empty object
+	// or a Signal) takes full ownership, so `config: {}` removes the cap entirely.
+	static readonly SD_DEFAULT_CONFIG: EncoderConfig = { maxPixels: 854 * 480 };
```
kixelated: I wouldn't hard-code this.

I think the simulcast rendition should be at least 1/4th of the pixel count (or something) or the source. There's too many assumptions being backed in here about what the HD config will be.

Here's the hang.live configuration btw: https://github.com/moq-dev/hang.live/blob/1a8d7b5b761f4f08548510f874246a56b18a0e36/app/src/room/local.ts#L87

^ That one is allowed to be specific because we know it's a webcam. But your generic simulcast needs to be more generic. Also 480p is a bit high in general for a simulcast rendition.

demo/web/src/publish.ts
kixelated: Done in d2984670. Moved the 720p request into `Source.Camera` as a static `DEFAULT_CONSTRAINTS` (merged under any caller-supplied constraints, so per-field overrides still win). The demo no longer sets constraints. (Written by Claude)

js/publish/src/video/index.ts
kixelated: Done in d2984670. Dropped the hard-coded `maxPixels: 854 * 480` and added a generic `maxScale` option to the encoder (fraction of the source pixel count). The sd default is now `maxScale: 0.25` (a quarter of the pixels = half width/height), so it scales with the source and makes no assumption about the hd config. (Written by Claude)

js/publish/src/video/encoder.ts
```diff
@@ -267,8 +273,15 @@ export class Encoder {
...
-		const ratio = Math.min(Math.sqrt(maxPixels / (frame.codedWidth * frame.codedHeight)), 1);
+		const sourcePixels = frame.codedWidth * frame.codedHeight;
+
+		// maxPixels caps absolutely; maxScale caps relative to the source. The smaller cap wins.
+		let maxPixels = user?.maxPixels ?? sourcePixels;
+		if (user?.maxScale !== undefined) {
+			maxPixels = Math.min(maxPixels, sourcePixels * user.maxScale);
+		}
+
+		const ratio = Math.min(Math.sqrt(maxPixels / sourcePixels), 1);
```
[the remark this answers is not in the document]
kixelated: Done in 5d613a41. Added a finite-and-positive guard on `maxScale` that throws before the dimension math, so 0/negative/NaN can't propagate to the ratio. (Written by Claude)

PR #1684, opened 2026-06-11
Mux import with existing track

Qizot:
The regular flow when creating mux import is to create a new track. This is fine most of the time but breaks when used together with dynamic tracks (produced by `requested_track`), we can't gracefully add the dynamic track to the catalog due to API limitations.

This PR adds support to moq-mux for importing from an already existing track.

There are no breaking changes, there is additional API and also the functionality has been added to moq-ffi.


kixelated:
I think this is fine for now to unblock you, but I want to revamp the API on `dev`.

I think each Import needs to have two states: Uninitialized (no TrackProducer) and Initialized (yes TrackProducer).

PR #1685, opened 2026-06-11
moq-mux: export SCTE-35 sections back to MPEG-TS

arielmol:
Follow-up to #1617 (SCTE-35 ingest, merged). That PR carried SCTE-35 cues from
MPEG-TS into a MoQ catalog extension; this one mirrors it on the way out, writing
the cue tracks back to MPEG-TS. Together they form a byte-exact MoQ <-> MPEG-TS
round-trip for splice information.

## Export

A `.scte35` cue track is remuxed back to the transport stream as private sections:

- The PMT gains the program-level CUEI registration descriptor and an elementary
  stream at stream_type 0x86, so a re-import re-detects and reassembles the cue
  (the exact signal #1617 keys off).
- Each section rides as a `Section` payload (pointer_field, unit-start bit) plus
  `Raw` continuations, emitted opaquely: no CRC, no reframing. The
  `splice_info_section` round-trips byte-for-byte.
- A frame that isn't a complete section (table_id 0xFC, length matching
  `section_length`) is dropped with a warning rather than emitted malformed. This
  is transport policy, not payload interpretation; the verbatim bytes are never
  touched.

The bridge stays media-agnostic at the MoQ layer: cues are carried as opaque
frames and never parsed semantically.

## A cue program requires a video track

The bridge clocks each cue from the video PTS at ingest; a `splice_info_section`
carries no PTS of its own. With no video track the export has no clock the cues
belong to (an audio-only ingest stamps them at zero), so a cue program without
video is rejected. Programs without cues keep the existing video-or-audio PCR
fallback.

This is a scope choice for broadcast plus the current clock model, not a SCTE-35
conformance claim: SCTE-35 is program-level, component splicing can target audio,
and audio dynamic ad insertion uses SCTE-35 without video. This rule can be
revisited if the bridge grows a clock source independent of video PTS.

## API

`Export<E: scte35::Catalog = ()>`. The plain constructors (`new`,
`with_catalog_format`) live on `impl Export<()>` and stay media-only; moq-cli
does not reference the SCTE types, so the default `Export<()>` path is unchanged.
A concrete `impl Export<scte35::Ext>` adds `Export::with_scte35`, which adds the
cue path; pinning the extension on the impl means callers write
`Export::with_scte35(..)` with no turbofish (same way `new` pins `Export<()>`).
The fmp4/mkv exporters keep their media-only view via `.media()`.

## Tests

A round-trip corpus runs the full TS -> MoQ -> TS seam and asserts every
`splice_info_section` survives byte-for-byte, pinning each fixture's total and
distinct cue counts so a clip that lost variety to duplicate sections fails:

- ffmpeg, GStreamer, and BigBuckBunny clips with injected SCTE-35 cues, each
  covering a muxer/codec/audio axis the others don't (H.264 video-only; H.264
  480i + AAC; H.265 + Opus).
- TSDuck-authored splices: the only section-repetition coverage (five distinct
  cues sent twice).
- The real Ateme CM5000 ("Kyrion") mid-stream capture from #1617, reused as a
  real-encoder case.

TSDuck decodes every checked-in section CRC32-OK; the decode is checked in beside
each clip as `<fixture>_tsduck.txt`. `examples/scte35_inject.rs` generates the
three injected fixtures.

The three commits are bisectable (export, force-video, corpus), each green alone.

## Test plan

- [x] `cargo test -p moq-mux` green (203 tests; 201/202/203 per commit)
- [x] clippy + fmt clean on the rebased branch
- [x] round-trip byte-exact across the corpus, including the real Kyrion capture
- [x] moq-cli does not reference the renamed SCTE types (default `Export<()>` path unchanged)

No JS or doc sync: carriage only, no wire or catalog-format change.


arielmol:
ok rabbit, last comments find a minor bug accepted, it lives in the examples not in production code, but will fix alonside @kixelated reviews/comments.

PR #1689, opened 2026-06-12
feat(publish): canvas preview + encoded "real" preview mode

kixelated:
## Summary

`<moq-publish>` previously only previewed the raw capture stream via a `<video>` element (`srcObject`). This adds a `<canvas>` preview path and a new `preview` attribute to toggle a "real" preview that decodes the actually-encoded video, so you can see exactly what a viewer receives over the network.

| Preview element | `preview="source"` (default) | `preview="encoded"` |
|---|---|---|
| `<video>` | raw source via `srcObject` (unchanged) | not possible without MSE → warns once, falls back to source |
| `<canvas>` | draws raw captured frames | **re-encodes + decodes → draws the wire output** |

```html
<moq-publish url="..." name="me.hang" preview="encoded">
    <canvas></canvas>
</moq-publish>
```

## Why

A `<video srcObject>` preview shows the pristine camera/screen source, which hides what's actually being transmitted. The `encoded` mode surfaces real compression artifacts, keyframe cadence, and frames the encoder dropped, useful for monitoring transmitted quality.

## How it works

The `encoded` path ([`preview.ts`](https://github.com/moq-dev/moq/blob/claude/quirky-napier-fdfbcc/js/publish/src/preview.ts)) runs a `VideoEncoder → VideoDecoder` chain fed by the same captured-frame signal the real encoder uses, configured with the **HD rendition's resolved config** (same codec, dimensions, and bandwidth-capped bitrate). That config resolves even with no remote subscriber, so the preview always shows something rather than going blank when nobody is watching. The renderer mirrors `@moq/watch`'s canvas renderer (flip + resize handling).

Tradeoff: this is a full extra encode + decode pass, incurred only while an `encoded`-mode canvas is active. (An exact-bytes loopback would need a local `consume()` on the Broadcast producer so publish could embed a `@moq/watch` element pointed at itself; that's a larger `js/net` change left as a follow-up.)

## Changes

- `js/publish/src/video/encoder.ts` — expose `resolved: Getter<VideoEncoderConfig>`.
- `js/publish/src/preview.ts` (new) — `Renderer` (canvas draw) + `Transcode` (encode→decode). Exported as `Publish.Preview`.
- `js/publish/src/element.ts` — discover `video, canvas`; add `preview` attribute + `state.preview` signal + JS property; branch the preview effect; warn-once on `<video>` + `encoded`.
- `js/publish/src/index.ts` — export `Preview`.
- `demo/web/src/publish.html`, `doc/lib/js/@moq/publish.md` — document the option.

## Branch targeting

Targeting `main`: additive only. New optional attribute defaulting to existing behavior, new module, no wire/protocol or catalog/container changes, and no breaking changes to existing public APIs (`resolved` and `Preview` are new surface).

## Test plan

- [x] `just check` (JS): `@moq/publish` and `@moq/demo` typecheck/lint/biome pass.
- [ ] Manual: `<canvas>` + `preview="source"` draws the live capture.
- [ ] Manual: `<canvas>` + `preview="encoded"` draws decoded output; toggling `invisible` blanks it.
- [ ] Manual: `<video>` + `preview="encoded"` logs the one-time warning and still shows the raw source.

(Written by Claude)

### DIFF
--- a/demo/web/src/publish.html
+++ b/demo/web/src/publish.html
@@ -24,6 +24,13 @@
 	-->
 	<moq-publish-ui>
 		<moq-publish url="%VITE_RELAY_URL%" name="me.hang">
+			<!--
+				The preview can be a <video> (raw camera/screen source, shown here) or a <canvas>.
+				A <canvas> with `preview="encoded"` on <moq-publish> shows a decoded copy of the encoded
+				video, i.e. what a viewer actually receives (codec artifacts and all), at the cost of an
+				extra encode + decode. Swap the line below to try it:
+				<canvas style="width: 100%; height: auto; border-radius: 4px; margin: 0 auto;"></canvas>
+			-->
 			<video style="width: 100%; height: auto; border-radius: 4px; margin: 0 auto;" muted autoplay></video>
 		</moq-publish>
 	</moq-publish-ui>
--- a/demo/web/src/publish.html
+++ b/demo/web/src/publish.html
@@ -25,11 +25,12 @@
 	<moq-publish-ui>
 		<moq-publish url="%VITE_RELAY_URL%" name="me.hang">
 			<!--
-				The preview can be a <video> (raw camera/screen source, shown here) or a <canvas>.
-				A <canvas> with `preview="encoded"` on <moq-publish> shows a decoded copy of the encoded
-				video, i.e. what a viewer actually receives (codec artifacts and all), at the cost of an
-				extra encode + decode. Swap the line below to try it:
+				The preview element can be a <video> (raw camera/screen source, shown here) or a <canvas>.
+				A <canvas> draws the frames directly. To preview what a viewer actually receives over the
+				network (a decoded copy of the encoded video, codec artifacts and all), do BOTH: add
+				preview="encoded" to the <moq-publish> element above, and replace the <video> below with:
 				<canvas style="width: 100%; height: auto; border-radius: 4px; margin: 0 auto;"></canvas>
+				The encoded preview costs an extra encode + decode pass.
 			-->
 			<video style="width: 100%; height: auto; border-radius: 4px; margin: 0 auto;" muted autoplay></video>
 		</moq-publish>
--- a/demo/web/src/publish.html
+++ b/demo/web/src/publish.html
@@ -24,15 +24,15 @@
 	-->
 	<moq-publish-ui>
 		<moq-publish url="%VITE_RELAY_URL%" name="me.hang">
-			<!--
-				The preview element can be a <video> (raw camera/screen source, shown here) or a <canvas>.
-				A <canvas> draws the frames directly. To preview what a viewer actually receives over the
-				network (a decoded copy of the encoded video, codec artifacts and all), do BOTH: add
-				preview="encoded" to the <moq-publish> element above, and replace the <video> below with:
-				<canvas style="width: 100%; height: auto; border-radius: 4px; margin: 0 auto;"></canvas>
-				The encoded preview costs an extra encode + decode pass.
-			-->
 			<video style="width: 100%; height: auto; border-radius: 4px; margin: 0 auto;" muted autoplay></video>
+
+			<!--
+				Alternatively, preview a decoded copy of the encoded video, i.e. what a viewer actually
+				receives over the network (codec artifacts and all), at the cost of an extra encode +
+				decode. Add preview="encoded" to <moq-publish> above and use this <canvas> instead of the
+				<video>:
+			-->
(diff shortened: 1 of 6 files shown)


js/publish/src/preview.ts
```diff
@@ -0,0 +1,192 @@
...
+
+		const frame = effect.get(this.#frame);
+		const display = effect.get(this.#video.display);
+		const flip = effect.get(this.#video.flip);
+
+		// Setting width/height clears the canvas, so only resize when the dimensions actually change.
+		if (display && (ctx.canvas.width !== display.width || ctx.canvas.height !== display.height)) {
+			ctx.canvas.width = display.width;
+			ctx.canvas.height = display.height;
+		}
```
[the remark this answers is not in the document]
kixelated: Fixed in dadae592: the canvas is now sized from `frame.displayWidth/Height` (the frame being drawn), falling back to the capture dimensions until the first frame. Encoded mode now shows the true transmitted resolution instead of upscaling.

js/publish/src/video/encoder.ts
```diff
@@ -55,6 +55,10 @@ export class Encoder {
 	// The video encoder config.
 	#config = new Signal<VideoEncoderConfig | undefined>(undefined);
 
+	// The resolved encoder config (codec, bitrate, dimensions), available even with no subscriber.
+	// Exposed so a local preview can re-encode with identical settings to mirror the wire output.
+	readonly resolved: Getter<VideoEncoderConfig | undefined> = this.#config;
```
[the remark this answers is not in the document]
kixelated: Fixed in dadae592, but without widening `resolved`. Instead of overloading the `VideoEncoderConfig` getter, I thread the rendition `EncoderConfig` into `Transcode` and mirror `Encoder.serve`'s GOP logic (`keyframeInterval ?? 2s`), so a custom keyframe interval is reflected in the preview while `resolved` stays a clean WebCodecs config.

demo/web/src/publish.html
```diff
@@ -24,6 +24,13 @@
 	-->
 	<moq-publish-ui>
 		<moq-publish url="%VITE_RELAY_URL%" name="me.hang">
+			<!--
+				The preview can be a <video> (raw camera/screen source, shown here) or a <canvas>.
+				A <canvas> with `preview="encoded"` on <moq-publish> shows a decoded copy of the encoded
+				video, i.e. what a viewer actually receives (codec artifacts and all), at the cost of an
+				extra encode + decode. Swap the line below to try it:
+				<canvas style="width: 100%; height: auto; border-radius: 4px; margin: 0 auto;"></canvas>
+			-->
```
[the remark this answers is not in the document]
kixelated: Fixed in dadae592: the demo comment now says to do both — add `preview="encoded"` to `<moq-publish>` and replace the `<video>` with a `<canvas>`.

demo/web/src/publish.html
```diff
@@ -24,6 +24,14 @@
 	-->
 	<moq-publish-ui>
 		<moq-publish url="%VITE_RELAY_URL%" name="me.hang">
+			<!--
+				The preview element can be a <video> (raw camera/screen source, shown here) or a <canvas>.
+				A <canvas> draws the frames directly. To preview what a viewer actually receives over the
+				network (a decoded copy of the encoded video, codec artifacts and all), do BOTH: add
+				preview="encoded" to the <moq-publish> element above, and replace the <video> below with:
+				<canvas style="width: 100%; height: auto; border-radius: 4px; margin: 0 auto;"></canvas>
```
kixelated: I think you put this underneath and leave it commented out instead. 

<!- Alternatively, use canvas.... 
-->

js/publish/src/preview.ts
```diff
@@ -0,0 +1,206 @@
+import { Time } from "@moq/net";
+import { Effect, type Getter, Signal } from "@moq/signals";
+import type * as Video from "./video";
+
+// What the canvas preview renders.
+// - `source`: the raw captured frames, drawn directly (cheap, no extra codec work).
+// - `encoded`: a decoded copy of the encoded video, so the preview shows the same codec
+//   artifacts a viewer would receive. This costs a full extra encode + decode pass.
+export type Mode = "source" | "encoded";
```
kixelated: Should we add "none" to make it easy to toggle off?

kixelated: Good call, added in aa635298. `Mode` is now `"none" | "source" | "encoded"`; `preview="none"` stops the canvas render (and skips the encode/decode in encoded mode) and hides a `<video>` preview.

demo/web/src/publish.html
kixelated: Done in aa635298: the `<video>` stays active and the `<canvas>` now sits underneath as a commented-out, ready-to-uncomment alternative.

PR #1692, opened 2026-06-12
feat(ffi): surface auth connect errors in language wrappers

kixelated:
Follow-up to #1649, which added `Unauthorized` (HTTP 401) and `Forbidden` (HTTP 403) variants to the moq-ffi `MoqError` so callers can distinguish an auth rejection from a retryable transport failure. That PR landed the Rust side but left the language wrappers without an ergonomic way to act on the new variants (the [Cross-Package Sync](../blob/main/CLAUDE.md) row for `rs/moq-ffi`). This closes that gap.

## Summary

Parity helpers next to the existing `isShutdown`:

- **Swift** — `MoqError.isAuth` (true for `.Unauthorized` / `.Forbidden`)
- **Kotlin** — `MoqException.isAuth`
- **Go** — `moq.IsAuthError(err)`, using `errors.Is` over the generated `ErrMoqErrorUnauthorized` / `ErrMoqErrorForbidden` sentinels
- **Python** — re-export `MoqError` from the `moq` package so callers can `except moq.MoqError.Unauthorized` without reaching into `moq_ffi`

Plus an auth-error note in each wrapper's guide (`doc/lib/{swift,kt,go,py}`) and the C library doc (`doc/lib/c`), which documents the distinct return codes + `moq_error()` strings since libmoq has no named C constants for them.

The variants themselves are already emitted by every binding (flat UniFFI error), so this is purely the ergonomic + documentation surface.

## Test plan

- [ ] CI `just ci` runs `just {py,kt,go} check` on the Linux runner (these wrappers were skipped for #1649 since it didn't touch them); Go compiles `IsAuthError` against freshly-regenerated bindings that now include `ErrMoqErrorForbidden`.
- [ ] `swift check` is macOS-only and skips on the Linux runner; the `isAuth` helper mirrors the existing `isShutdown` exactly.
- [x] `gofmt` clean; Python isort/`__all__` ordering consistent with existing style.

(Written by Claude)

### DIFF
--- a/doc/lib/c/index.md
+++ b/doc/lib/c/index.md
@@ -111,6 +111,8 @@ if (rc < 0) {
 
 `moq_error()` returns the reason for the most recent failed call **on the calling thread**, including detail the numeric code can't carry (which URL failed to parse, why a decode failed, etc.). The returned pointer is valid until the next libmoq call on that thread, so copy it if you need to keep it. It is only meaningful after a call returned a negative code; check the code first. Errors delivered through status callbacks carry their code directly, so read `moq_error()` from inside the callback if you want the matching reason.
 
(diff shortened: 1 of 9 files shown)


kixelated:
@Qizot do you need this?

Qizot:
@kixelated yes, please

PR #1698, opened 2026-06-12
feat: certificate pinning for native and browser clients

kixelated:
## Summary

Trusting a self-signed or privately-issued peer previously meant disabling TLS verification entirely. This adds proper pinning to both the native and browser clients so callers can trust a specific certificate without turning verification off.

The unifying idea: a server already reports its certificate's SHA-256 fingerprint as **hex** (native `cert_fingerprints()`, the `/certificate.sha256` endpoint). Now **both** clients accept that same hex value directly — no insecure runtime fetch, no manual byte-wrangling.

### Native (`moq-native` / `moq-ffi` / `py`)

- New `tls::Client.fingerprint` field (`--tls-fingerprint`, `MOQ_CLIENT_TLS_FINGERPRINT`): pin the peer's leaf certificate to one or more hex SHA-256 hashes, bypassing the CA chain, hostname match, and expiry. The native equivalent of WebTransport `serverCertificateHashes`. `FingerprintVerifier` is ungated from the quinn/noq backends so it applies to every transport, and matches against a set to support rotation.
- The existing PEM-root knob (`--tls-root`) is now reachable from the FFI too.
- FFI: `MoqClient.set_tls_roots(paths)` and `set_tls_fingerprints(hashes)` alongside `set_tls_disable_verify`.
- Python: `Client(url, tls_roots=..., tls_fingerprints=...)`.
- `moq-cli` picks up `--tls-root` / `--tls-fingerprint` for free via the flattened config.

### Browser (`js/net`)

- `connect()`'s `webtransport.serverCertificateHashes` values now accept a **hex string** (decoded automatically), not just raw bytes.
- New `webtransport.serverCertificate` option: pin from a PEM string or DER bytes with the SHA-256 computed in JS.
- Exported `certificateHash(cert)` helper and `Hex.fromBytes` (counterpart to `toBytes`).

## Motivation / scope

This came out of looking at [pipecat#4629](https://github.com/pipecat-ai/pipecat/pull/4629). Worth being precise about what helps where:

- **pipecat's serve mode** (bot = server, browser pins via `serverCertificateHashes`) is a **browser-side** flow. The native client change doesn't touch it; the JS ergonomics here do (hex-string + `serverCertificate` instead of hand-built hash structs).
- **pipecat's client mode** (bot dials a relay as a native `moq.Client(tls_verify=...)`) only had a bool. `tls_fingerprints` / `tls_roots` is the upgrade if that relay is ever self-signed.

So neither half is strictly *required* by that PR, but both close a real "verify or disable, nothing in between" gap on their respective clients.

## Notes for reviewers

- **Public API**: additive only — `#[non_exhaustive]` config field + new FFI methods (native); widened option types (existing `{value: BufferSource}` callers still compile) + new exports (JS). Non-breaking, hence `main`.
- Hex chosen everywhere to match `cert_fingerprints()` / `/certificate.sha256`.
- swift/kt/go have no hand-written client wrappers, so uniffi auto-generates `setTlsRoots` / `setTlsFingerprints` on the next `moq-ffi-v*` mirror. libmoq exposes no client TLS-verify surface today, so nothing to mirror there.

## Test plan

- [x] `cargo test -p moq-native` — `--tls-fingerprint` TOML/CLI parsing, the TOML→CLI merge-clobber regression, and a `FingerprintVerifier` roundtrip (match / mismatch / invalid-hex)
- [x] `bun test` in `js/net` — `certificateHash` over DER and PEM against a known SHA-256 vector, hex round-trip (145/145 pass)
- [x] clippy + fmt (nix) on `moq-native` + `moq-ffi`; tsc + biome on `js/net`
- [ ] End-to-end against a live self-signed peer (native needs a maturin rebuild of `moq_ffi`)

(Written by Claude)

### DIFF
--- a/js/net/src/connection/connect.ts
+++ b/js/net/src/connection/connect.ts
@@ -288,11 +288,12 @@ type WebTransportHash = NonNullable<WebTransportOptions["serverCertificateHashes
 
 // Strip PEM armor and base64-decode to the raw DER bytes.
 function pemToDer(pem: string): Uint8Array<ArrayBuffer> {
-	const body = pem
-		.replace(/-----BEGIN CERTIFICATE-----/g, "")
-		.replace(/-----END CERTIFICATE-----/g, "")
-		.replace(/\s+/g, "");
-	const binary = atob(body);
+	const match = pem.match(/-----BEGIN CERTIFICATE-----([\s\S]+?)-----END CERTIFICATE-----/);
+	if (!match) {
+		throw new Error("invalid PEM certificate: missing -----BEGIN/END CERTIFICATE----- armor");
+	}
+
+	const binary = atob(match[1].replace(/\s+/g, ""));
 	const der = new Uint8Array(binary.length);
 	for (let i = 0; i < binary.length; i++) {
 		der[i] = binary.charCodeAt(i);
@@ -352,10 +353,9 @@ async function connectWebTransport(
 		...webtransport,
 	};
 
-	const hashes = await resolveCertificateHashes(options);
-	if (hashes) {
-		finalOptions.serverCertificateHashes = hashes;
-	}
+	// Accumulate caller-provided pins first, then append anything we fetch below,
+	// so a fetched fingerprint never clobbers hashes passed in via options.
+	const hashes = (await resolveCertificateHashes(options)) ?? [];
 
 	// Only perform certificate fetch and URL rewrite when polyfill is not needed
 	// This is needed because WebTransport is a butt to work with in local development.
@@ -373,15 +373,14 @@ async function connectWebTransport(
 		const fingerprintText = await Promise.race([fingerprint.text(), cancel]);
 		if (fingerprintText === undefined) return undefined;
 
-		finalOptions.serverCertificateHashes = (finalOptions.serverCertificateHashes || []).concat([
-			{
-				algorithm: "sha-256",
-				value: Hex.toBytes(fingerprintText),
-			},
-		]);
+		hashes.push({ algorithm: "sha-256", value: Hex.toBytes(fingerprintText) });
 
 		finalUrl = new URL(url);
 		finalUrl.protocol = "https:";
+	}
+
+	if (hashes.length > 0) {
+		finalOptions.serverCertificateHashes = hashes;
 	}
 
 	const quic = new WebTransport(finalUrl, finalOptions);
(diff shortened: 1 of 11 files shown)


kixelated:
Addressed the fingerprint-length nitpick in 34239b1f: `Client::build()` now rejects any `--tls-fingerprint` that doesn't decode to exactly 32 bytes (new `Error::FingerprintLength`), so a typo'd or truncated hash fails at config time with a clear message instead of silently never matching at handshake. Added a `build_rejects_wrong_length_fingerprint` test.

(Written by Claude)

js/net/src/connection/connect.ts
```diff
@@ -263,13 +283,60 @@ async function handshakeAlpn(url: URL, session: WebTransport, version: Ietf.Ietf
 	});
 }
 
+// One entry of the DOM `serverCertificateHashes`, derived without naming the lib type.
+type WebTransportHash = NonNullable<WebTransportOptions["serverCertificateHashes"]>[number];
+
+// Strip PEM armor and base64-decode to the raw DER bytes.
+function pemToDer(pem: string): Uint8Array<ArrayBuffer> {
+	const body = pem
+		.replace(/-----BEGIN CERTIFICATE-----/g, "")
```
kixelated: We should error if this armor isn't found.

js/net/src/connection/connect.ts
```diff
@@ -282,9 +349,14 @@ async function connectWebTransport(
 			Ietf.ALPN.DRAFT_16,
 			Ietf.ALPN.DRAFT_15,
 		],
-		...options,
+		...webtransport,
 	};
 
+	const hashes = await resolveCertificateHashes(options);
+	if (hashes) {
+		finalOptions.serverCertificateHashes = hashes;
```
kixelated: Any chance we end up overridding webtransport provided hashes?

js/net/src/connection/connect.ts
kixelated: Fixed in aecc668f: `pemToDer` now matches against the `-----BEGIN/END CERTIFICATE-----` armor and throws `invalid PEM certificate: missing ... armor` if it's absent, instead of silently `atob`-ing arbitrary input. Added a test. (Written by Claude)

js/net/src/connection/connect.ts
kixelated: No — caller-provided hashes are preserved. They flow through `resolveCertificateHashes(options)` and the http:// fetched fingerprint is appended (it was already `concat`, not a replace). aecc668f restructures it to accumulate into one array with a single assignment site so that's structurally obvious rather than relying on reading both spots. (Written by Claude)

PR #1701, opened 2026-06-12
moq-mux: ingest and export legacy non-browser broadcast audio over MPEG-TS mp2, ac-3 & e-ac-3

arielmol:
Import and export legacy broadcast audio over the MPEG-TS bridge: MP2
(stream types 0x03/0x04), AC-3 (ATSC 0x81) and E-AC-3 (ATSC 0x87).
Carriage is verbatim: complete, well-formed codec frames pass through
byte-exact, each as one hang frame in its own group, never decoded or
transcoded; malformed input is rejected rather than mis-described. The
catalog types each track honestly ("mp2" / "ac-3" / "ec-3") so TS gear
on the other side picks it up; browsers cannot decode these codecs and
skip the rendition.

## Why hang's enum and not a catalog extension

Discussed on Discord (2026-06-12). Adding the variants is additive:
`AudioCodec` is `#[non_exhaustive]` and older readers parse unknown
codec strings as `Unknown(...)`, so nothing matched these strings
before. Per that conversation, the doc comments mark all three variants
as legacy codecs: TS bridge only, not decodable in browsers, not for
new content.

## Design

- **One shared engine, three parsers.** A `pub(crate)` `codec/legacy.rs`
  owns the track lifecycle (unique track, catalog rendition, whole-frame
  publish, `description = None`); each codec contributes a header parser
  and a static `Descriptor` (track suffix, catalog codec, min header
  bytes, parse fn). A single `LegacyStream` in `ts/import.rs` serves all
  three. No traits, no facades: the only caller is the TS demuxer.
- **Samples live in the parsed header, not per codec**: E-AC-3 varies
  them (256 x numblks); MP2/AC-3 are constant (1152/1536). The PTS
  advance accumulates per frame.
- **Frames may split across PES packets.** ISO 13818-1 does not require
  audio/PES alignment, so the trailing fragment is carried over and
  reassembled with the next PES. The reassembled frame keeps the PTS of
  the PES it began in (a PES PTS only covers access units starting in
  it); a regression test pins that PTS behavior across PES reassembly.
- **Faithful egress.** MP2 re-announces its stream_type from the catalog
  sample rate (>= 32 kHz are MPEG-1 rates -> 0x03, below -> 0x04); the
  mapping stays faithful because the parser rejects the unofficial
  MPEG-2.5, which has no TS stream type, so the derivation stays
  bijective. AC-3/E-AC-3 are 1:1 to 0x81/0x87, paired with the
  "AC-3"/"EAC3" registration descriptors. The PCR falls back to an audio
  PID in audio-only programs.
- **Only AAC anchors the audio-burst jitter run**: a legacy PES opening
  it would inflate AAC's published jitter by the inter-PID PTS offset
  (regression test included).
- **Explicit scope rejections** with diagnostic errors: MPEG-2.5, E-AC-3
  dependent substreams (7.1+) and nonzero substream ids, reserved
  fscod2/strmtyp, AC-3 bsid > 8, frame lengths shorter than their own
  header.

## Known limitations (deferred)

- **A corrupt audio header still aborts the whole import**, video and
  SCTE-35 included: same fail-fast contract as the existing ADTS path.
  Packet-level TEI protection only exists on the SCTE-35 path (the
  mpeg2ts reader parses the TEI bit but never consults it). Follow-up:
  frame-level drop-and-resync; the SCTE section reassembler already
  models the "never fatal" alternative in this file.
- **A PES lost between a split frame's cut and its continuation** joins
  mismatched bytes and fails the next parse. Reassembly covers
  consecutive PES; recovery from loss is not attempted.
- **The catalog is frozen at the first frame** (same as AAC): a
  mid-stream sample-rate/channel change is not re-described; per-frame
  parameters (bitrate, padding, frame size) are read per frame and pass
  through fine.
- **The ATSC AC-3 audio descriptor (tag 0x81) is not emitted on export;
  only the registration descriptor is**, which suffices for
  identification. The Kyrion emits both, visible in the checked-in
  TSDuck dump.
- E-AC-3 multi-substream layouts and DVB-style AC-3 (0x06 + descriptor)
  are future iterations.

## Public API changes

- `hang::catalog::AudioCodec`: new `Mp2`, `Ac3`, `Ec3` variants
  (additive on a `#[non_exhaustive]` enum); `FromStr` now types
  "mp2" / "ac-3" / "ec-3".
- `moq-mux`: **no new public API.** `codec::{legacy,mp2,ac3,eac3}` are
  all `pub(crate)` (the AAC precedent: its ADTS parser is a private
  module).
- No JS changes needed: `js/hang` types the codec as a plain string.

## Testing

Byte-exact TS -> MoQ -> TS round-trips for all three codecs, against
real encoder captures and synthetic clips:

- `kyrion_mpeg2av_ac3.ts`: a real Kyrion capture in ATSC compliance
  mode: MPEG-2 1080i video (clock-only), AC-3 (bsid 6, stereo) + MP2,
  SCTE-35 cues, a dedicated PCR PID, PSIP tables, and a dirty mid-stream
  start. The TSDuck companion holds two dumps (cues, CRC32 OK; and the
  PMT, evidencing the encoder itself pairs 0x81 with the "AC-3"
  registration descriptor).
- The existing Kyrion capture covers the two-MP2-programs case.
- `ac3.ts` / `eac3.ts`: ffmpeg 5.1 clips covering what the hardware
  doesn't produce: the lfeon path (6 channels) and audio-only programs.
- Unit tests pin the header parsers (including the lfeon edge where the
  bit lands on the last byte-6 bit), every scope rejection's error text,
  PES reassembly PTS semantics, and the AAC jitter anchoring.

`cargo test -p moq-mux` 221/221, `cargo test -p hang` 14/14, fmt and
`clippy --all-targets -- -D warnings` clean.

Docs updated: `doc/bin/cli.md`, `doc/lib/rs/crate/moq-mux.md`.


arielmol:
> 🧹 Nitpick comments (2)

Okay rabbit. As usual no big deal, happy to patch alongside @kixelated reviews/comments.

PR #1713, opened 2026-06-13
moq-net: revert origin builder/Session toward main's shape

kixelated:
## Summary

The dev origin reshape made the session builders take a required `OriginProducer` per side, default both to a shared auto-created origin, and surface `Session::publisher()` / `Session::consumer()`. Exposing the origin handles through the `Session` turned out to be the wrong shape (a handle you can hold but that silently does nothing when unset). This reverts the public API back to something close to `main`:

- **`with_publish(OriginConsumer)` / `with_subscribe(OriginProducer)`** — renamed from `with_publisher`/`with_consumer`; the publish side takes an `OriginConsumer` again (you keep the producer and publish into it). No `impl Into<Option<_>>`: you either hand it a real handle or you don't call the setter.
- **Both halves default to `None`.** An unset publish side announces nothing; an unset subscribe side issues no `ANNOUNCE_PLEASE` (the lite subscriber task simply isn't spawned). No more shared auto-origin.
- **Dropped `Session::publisher()` / `Session::consumer()`.** Callers keep their own origin handles. `moq-ffi` and `moq-wasm` now hold the handles they wired (or a fresh one) rather than reading them back from the `Session`.

`lite::start` / `ietf::start` take `Option` origins again. The lite path leaves an unset half idle; the ietf (moq-transport interop) path substitutes a fresh origin for an unset half, matching its prior behavior.

Net `-48` lines: removing the accessors and the connect/accept threading.

## Migration

`with_publisher(p)` → `with_publish(p.consume())`, `with_consumer(p)` → `with_subscribe(p)`, across moq-cli, moq-relay, moq-gst, moq-boy, moq-bench, moq-hls, moq-rtc, hang examples, moq-native (+ wrappers, tests, examples), moq-ffi, libmoq, moq-wasm.

## Test plan

- [x] `cargo test -p moq-net -p moq-native -p moq-relay --all-features` (moq-net 374 + relay 129/58 + native suites pass)
- [x] `cargo check --all-targets` across moq-net/native/cli/relay/rtc/boy/bench/hls/hang/moq-ffi/libmoq
- [x] `moq-gst` (nix) and `moq-wasm` (wasm32 target) compile
- [x] fmt + clippy clean (via nix)

Targets `dev` per Branch Targeting (breaking moq-net API change). CI's `Check` job is currently red workspace-wide on an unrelated `libva`/`cros-libva` build gap from the native-codecs work; not introduced here.

(Written by Claude)


rs/moq-net/src/client.rs
```diff
@@ -75,11 +69,9 @@ impl Client {
 
 	/// Perform the MoQ handshake as a client negotiating the version.
 	pub async fn connect<S: web_transport_trait::Session>(&self, session: S) -> Result<Session, Error> {
-		let publisher = self.publish.clone();
-		let consumer = self.consume.clone();
-		let publish = publisher.consume();
-		let consume = consumer.clone();
-		let consumer_view = consumer.consume();
+		if self.publish.is_none() && self.subscribe.is_none() {
+			tracing::warn!("not publishing or consuming anything");
```
kixelated, line 73: This should be an error, like on main.

PR #1718, opened 2026-06-14
feat(moq-video): Windows D3D11 zero-copy GPU capture + hardware H.264 encoder

kixelated:
## Summary

Makes the Windows capture/encode path stay on the GPU end to end, so Windows joins macOS (VideoToolbox) as a zero-copy platform. Before this, Windows capture handed back CPU NV12 deinterleaved to I420 for openh264 only.

- **Capture** (`capture/mediafoundation.rs`): creates a shared hardware Direct3D11 device + `IMFDXGIDeviceManager`, drives the source reader with `MF_SOURCE_READER_D3D_MANAGER` + advanced video processing, and hands back NV12 `ID3D11Texture2D` surfaces.
- **Frame** (`frame.rs`): new `Frame::Texture` variant (texture + device + subresource). Cloning is an `AddRef`, so the surface flows capture → encode without a copy. `to_i420()` downloads via a staging texture for the software fallback.
- **Encoder** (`encode/backend/mediafoundation.rs`): new hardware H.264 encoder MFT (`MFTEnumEx` HARDWARE). It adopts the D3D11 device from the first texture frame via `MFT_MESSAGE_SET_D3D_MANAGER`, which preserves the existing capture/encode seam (the `Frame` carries its device, the backend binds lazily). Drives the async-MFT event pump and emits Annex-B directly (no AVCC rewrite, unlike VideoToolbox). Registered hardware-first in Auto mode, mirroring VideoToolbox.

The new path: camera → MF advanced video processor → NV12 texture on the GPU → hardware H.264 MFT → Annex-B, no system-memory round trip.

### Graceful fallback

Both halves degrade: with no D3D11 device, capture uses the source reader's software video processor (`Frame::I420`) and the MFT uploads NV12 from system memory. A GPU-less box still works via openh264. In practice the I420 → hardware-MFT case is near-unreachable: no GPU means `MFTEnumEx(HARDWARE)` finds nothing and it falls back to openh264 anyway.

## Public API

No public API or wire change. `Frame` is `pub(crate)`; `capture::Config` / `encode` surface unchanged. `encode::Kind::Named("mediafoundation")` is a new valid selector but additive.

## Why `dev`

Major feature that changes Windows Auto-mode to prefer the new MFT over openh264, and it's runtime-unvalidated. No breaking change, but per the branch-targeting rules it should settle on `dev` first.

## Test plan

- [x] Cross-compiled for `x86_64-pc-windows` from macOS (rustup-stable std + `cargo-zigbuild`), clean build, zero warnings.
- [x] Host (macOS) `cargo build` + `cargo test -p moq-video`: all 10 tests pass, the macOS path is untouched.
- [x] `cargo fmt --check` clean.
- [ ] **Runtime validation on real Windows hardware is still pending** (same status as the NVENC backend). The async-MFT pump, the "SPS/PPS inline at each IDR" assumption, and the I420 → hardware sysmem upload all need to be exercised on a Windows box with a real camera + GPU.

> [!WARNING]
> Because Auto mode now prefers the hardware MFT, a runtime bug in the encoder would break Windows capture versus today's working openh264. If you'd rather be cautious until validated, the MFT can be gated behind `Kind::Hardware` / `Kind::Named("mediafoundation")` instead of being hardware-first.

(Written by Claude)

### DIFF
--- a/rs/moq-video/Cargo.toml
+++ b/rs/moq-video/Cargo.toml
@@ -62,7 +62,21 @@ objc2-screen-capture-kit = "0.3.2"
 objc2-video-toolbox = "0.3.2"
 
 [target.'cfg(target_os="windows")'.dependencies]
-# Native Media Foundation webcam capture (replaces nokhwa). An `IMFSourceReader`
-# with its video processor enabled converts the camera's native format to NV12,
-# which we deinterleave to I420. Links only system import libs, no vendor SDK.
-windows = { version = "0.62", features = ["Win32_Foundation", "Win32_System_Com", "Win32_Media_MediaFoundation"] }
+# Native Media Foundation webcam capture + hardware H.264 encode (replaces
+# nokhwa). An `IMFSourceReader` backed by a D3D11 device hands us NV12 textures
+# that feed a hardware encoder MFT zero-copy; without a GPU it falls back to CPU
+# NV12 deinterleaved to I420 for openh264. Links only system import libs, no
+# vendor SDK. The Direct3D11 / Dxgi features cover the shared D3D11 device and
+# the `IMFDXGIBuffer` surfaces wrapping each captured texture.
+windows = { version = "0.62", features = [
+	"Win32_Foundation",
+	"Win32_System_Com",
+	"Win32_Media_MediaFoundation",
+	"Win32_Graphics_Direct3D",
+	"Win32_Graphics_Direct3D10",
+	"Win32_Graphics_Direct3D11",
+	"Win32_Graphics_Dxgi",
+	"Win32_Graphics_Dxgi_Common",
+	"Win32_System_Ole",
+	"Win32_System_Variant",
+] }
--- a/rs/moq-video/Cargo.toml
+++ b/rs/moq-video/Cargo.toml
@@ -8,6 +8,11 @@ license = "MIT OR Apache-2.0"
 version = "0.0.3"
 edition = "2024"
 rust-version.workspace = true
+# Not publishable to crates.io while the `vaapi` feature pulls cros-codecs from
+# git; flip back once that fork is published (tracked separately). This also lets
+# cargo-deny's allow-wildcard-paths exempt the git dep, which it only does for
+# unpublished crates.
+publish = false
 
 keywords = ["quic", "http3", "webtransport", "media", "video"]
 categories = ["multimedia", "multimedia::video", "multimedia::encoding"]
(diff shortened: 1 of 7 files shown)


kixelated:
## Self-review (high-effort, 7 finder angles + verification)

Ran a review over the diff. Memory-safety holds up — several scary-sounding candidates were chased down and cleared (see bottom). Fixed everything actionable in `c24f84d6`.

### Fixed
- **Event pump could hang.** `handle_event` ignored everything but NeedInput/HaveOutput and never checked event status, so an async `MEError` (or a stalled MFT) would block the capture thread forever. Now checks `IMFMediaEvent::GetStatus()` and propagates the failure.
- **Magic `601`/`602` + wrong comment.** `METransformNeedInput`/`METransformHaveOutput` *are* exposed in `windows` 0.62; switched to the named constants and removed the incorrect comment.
- **Per-output re-query.** `process_output` called `GetOutputStreamInfo` on every output (non-provides-samples path). Cached the flag + buffer size in `start()`, refreshed after a stream-format change.
- **Duplication.** `ComGuard` / `mf_err` / `pack_2x32` were copied across capture and encode → extracted to a shared `mf` module. NV12 chroma interleave/deinterleave → shared `frame::{interleave_uv, deinterleave_uv}`; the upload buffer is now built through a slice rather than raw pointer math.

### Considered and left as-is
- **Auto prefers the unvalidated MFT over openh264** (the one real judgment call). `backend::open` only falls back on `open()` error, not a runtime `encode()` error, so a broken MFT would break Windows capture. Left hardware-first deliberately — that's the point of the zero-copy path — but it's a one-line change to gate behind `Kind::Hardware`/`Named` if runtime validation argues for caution.
- **NV12 UV-plane offset (`pitch * height`)** in the staging download: standard mapped-NV12 layout, but an assumption to confirm on real hardware.

### Cleared (raised by finders, not real)
- `ptr::read`/`ManuallyDrop` in `process_output`: not a double-free — fields are `ManuallyDrop`, ownership taken exactly once.
- Mid-stream `Texture`↔`I420` switch breaking the lazily-bound device: can't happen — `producer.rs` resets the encoder whenever it releases the camera, so frame variant is stable for an encoder's lifetime.
- `enumerate_encoder` leaking un-activated `IMFActivate`s: released on slice drop.
- Captured-texture recycling: the `Frame` and the DXGI surface buffer both hold COM refs, so the reader pool can't reclaim it under the MFT.

Still compile-verified only (zigbuild cross from macOS) + host tests green; runtime validation on a real Windows box remains the open item.

(Written by Claude)

PR #1736, opened 2026-06-14
fix(moq-net): read frames in place, without minting a per-poll FrameConsumer

kixelated:
## Summary

`GroupConsumer::poll_read_frame` / `poll_read_frame_chunks` called `poll_get_frame` → `frame.consume()` on **every poll**, then dropped that `FrameConsumer` whenever the frame's data wasn't complete yet (`poll_read_all` still `Pending`). A `FrameConsumer` is a kio consumer handle, so that create+drop flips the frame's consumer count `0→1→0` each poll, and kio wakes the state's waiters on **both** the "first consumer appears" and "last consumer drops" transitions — the same waiters the read itself registered on. So **every poll re-woke itself** — a silent busy spin.

- On a multi-threaded runtime the producer fills the frame concurrently, so the spin ends in microseconds: wasted CPU, no visible hang. This is why it's been latent.
- On a **single-thread executor** (e.g. the wasm/`@moq/net`-over-WASM consume path on `dev`) the consumer's self-wake loop starves the producer, so the frame never completes and the spin runs away into a **hard browser freeze** (~22M re-polls / ~45M wakes on one frame).

## Fix

Read the frame **in place** instead of through a consumer handle:

- **kio:** add `Producer::poll_ref` — a read-only counterpart to `Producer::poll` that registers a waiter on a read condition without taking a `Mut` (no `modified` flag, no consumer-count side effects).
- **`model/frame`:** `FrameProducer::poll_read_all` reads the producer's own buffer once the frame is finished, via `poll_ref`. Stateless (always offset 0), so any number of parallel readers are fine.
- **`model/group`:** `GroupState::poll_frame_read_all` reads the cached `FrameProducer` directly; `poll_read_frame` / `poll_read_frame_chunks` use it and no longer mint a `FrameConsumer`. `GroupConsumer` stays a plain `derive(Clone)` with no extra state.

## Public API

`GroupConsumer` unchanged. One additive kio method (`Producer::poll_ref`). `FrameProducer::poll_read_all` is `pub(crate)`.

> Note: the kio `Producer::poll_ref` addition is additive; coordinating with separate in-flight kio work on the `Mut`/wake semantics. That work is complementary — this PR removes the consumer-count churn regardless.

## Test plan

- [x] `cargo test -p moq-net --lib` — 346 pass; `cargo test -p kio`
- [x] `cargo clippy -p moq-net -p kio` clean, `cargo fmt`
- [x] Verified the freeze is gone end-to-end in a real browser via the wasm consume path (heartbeat kept ticking while a probe drained hundreds of frames and completed cleanly — vs. an instant freeze on frame 1 before). The `dev` wasm PR (#1726) carries the same fix and is where the freeze was first observed.

(Written by Claude)

### DIFF
--- a/rs/kio/src/producer.rs
+++ b/rs/kio/src/producer.rs
@@ -111,6 +111,34 @@ impl<T> Producer<T> {
 		Poll::Pending
 	}
 
+	/// Poll-based **read-only** access with waker registration.
+	///
+	/// Like [`Self::poll`] but hands `f` a [`Ref`] instead of a [`Mut`], so it
+	/// never flags the state modified and never wakes consumers. Use it to wait on
+	/// a read condition (e.g. a `fin` flag) from the producer side without creating
+	/// a [`Consumer`] — creating/dropping a consumer churns the consumer count and
+	/// wakes the value waiters, which would spin a polling reader.
+	pub fn poll_ref<F, R>(&self, waiter: &Waiter, mut f: F) -> Poll<Result<R, Ref<'_, T>>>
+	where
+		F: FnMut(&Ref<'_, T>) -> Poll<R>,
+	{
+		let state = self.state.lock();
+		let state = Ref { state };
+
+		if let Poll::Ready(res) = f(&state) {
+			return Poll::Ready(Ok(res));
+		}
+
+		if state.state.closed {
+			return Poll::Ready(Err(state));
+		}
+
+		let mut state = state.state;
+		waiter.register(&mut state.waiters);
+
+		Poll::Pending
+	}
+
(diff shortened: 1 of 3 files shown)


kixelated:
Superseded by the kio fixes #1735 (read-only `Producer::poll` predicate) and #1739 (split `State.waiters` so `consume()`/`Consumer::drop` no longer wake value readers). Those fix the consumer-count-churn spin at the source, so the original `frame.consume()`-per-poll code no longer freezes — verified in-browser. No moq-net workaround needed.

(Written by Claude)

PR #1737, opened 2026-06-14
feat(moq-video): Windows hardware H.264 encode via Media Foundation

kixelated:
## Summary

Adds a **Media Foundation Transform (MFT) hardware H.264 encoder** for Windows — the peer to the existing VideoToolbox (macOS) / NVENC / VAAPI (Linux) hardware backends. Before this, Windows `moq-video` only had the openh264 software encoder.

The new backend (`rs/moq-video/src/encode/backend/mediafoundation.rs`):

- Enumerates the GPU's hardware H.264 encoder MFT via `MFTEnumEx` (Intel QSV / NVENC / AMD VCE, whichever the GPU exposes) and activates the first that accepts NV12 in, H.264 out.
- Drives the **async MFT event model** (`METransformNeedInput` / `METransformHaveOutput`) synchronously by pumping the `IMFMediaEventGenerator` queue on the existing capture/encode thread — no extra worker or callback. Hardware MFTs are always async, so this is required.
- Requests **low-latency mode** via `ICodecAPI` (1-in-1-out, no B-frames/lookahead) plus CBR rate control and per-keyframe force-IDR.
- Emits **Annex-B with in-band SPS/PPS** on keyframes, prepending the cached sequence header only when the encoder doesn't already include it — so output is self-contained avc3, identical in shape to every other backend. The CDN/relay stays media-agnostic.

It slots into the `HARDWARE` candidate list, so `Kind::Auto` now selects it on Windows and falls back to openh264 when no hardware encoder is present. The CLI's existing `--hardware` / `--software` flags already cover it; no CLI change needed.

Supporting changes:
- Extracted the shared COM + Media Foundation init (`ComGuard`) out of the capture module into `src/win.rs` so capture and encode reuse it.
- Added the `Win32_System_Ole` / `Win32_System_Variant` windows-crate features (for `ICodecAPI::SetValue` / `VARIANT`).
- Fixed the encoder-backend lists in `doc/bin/cli.md` and the crate docs (which were also missing VAAPI).

## Test plan

Verified on a real Windows machine with a hardware encoder (the same box this was written and tested on, not the Mac it was originally coded for):

- [x] `cargo build -p moq-video` + `cargo build -p moq-cli --features capture` — clean
- [x] `cargo clippy -p moq-video --all-targets` — no warnings
- [x] `cargo fmt -p moq-video --check` — clean
- [x] `cargo test -p moq-video --lib` — 7/7 pass
- [x] **Hardware encode** (`#[ignore]`d test `mediafoundation_emits_annexb_keyframe`, run with `--ignored`): 30 synthetic frames → 30 Annex-B packets, 1-in-1-out confirmed. First access unit NAL types `[9, 7, 8, 5]` (AUD, SPS, PPS, IDR) — a self-contained keyframe. Testing caught and fixed a duplicate-SPS/PPS bug in the first cut.

The hardware test is `#[ignore]`d because GitHub's Windows CI runners are GPU-less VMs with no hardware encoder MFT, where it would correctly fail to find one. Run it on demand with `cargo test -p moq-video --lib -- --ignored`.

Not done: full publish → browser playback (needs a relay + web subscriber spun up). The encoder output is byte-for-byte the Annex-B format the avc3 importer already consumes from the other backends, so integration risk is low.

Targeting `dev` per the branch-targeting rules (sizable feature on the wire-adjacent `moq-video` crate).

(Written by Claude)

### DIFF
--- a/doc/bin/cli.md
+++ b/doc/bin/cli.md
@@ -90,9 +90,9 @@ moq-cli publish --url https://relay.example.com --broadcast cam.hang capture --n
 
 Video capture uses a native per-platform backend (AVFoundation on macOS, V4L2 on
 Linux, Media Foundation on Windows) and picks a hardware H.264 encoder
-(VideoToolbox on macOS, NVENC on Linux NVIDIA, VAAPI on Linux Intel/AMD) when one
-is present, falling back to the built-in software encoder (openh264); force either
-with `--hardware` / `--software`. `--camera` takes a bare integer as a device index, otherwise a
+(VideoToolbox on macOS, NVENC on Linux NVIDIA, VAAPI on Linux Intel/AMD, Media
+Foundation on Windows) when one is present, falling back to the built-in software
+encoder (openh264); force either with `--hardware` / `--software`. `--camera` takes a bare integer as a device index, otherwise a
 device path (Linux) or name (a friendly-name substring on Windows, the
 AVFoundation `uniqueID` on macOS). Audio capture uses cpal (CoreAudio / WASAPI /
 ALSA) and encodes Opus.
--- a/rs/moq-video/Cargo.toml
+++ b/rs/moq-video/Cargo.toml
@@ -76,7 +76,16 @@ objc2-screen-capture-kit = "0.3.2"
 objc2-video-toolbox = "0.3.2"
 
 [target.'cfg(target_os="windows")'.dependencies]
-# Native Media Foundation webcam capture (replaces nokhwa). An `IMFSourceReader`
-# with its video processor enabled converts the camera's native format to NV12,
-# which we deinterleave to I420. Links only system import libs, no vendor SDK.
-windows = { version = "0.62", features = ["Win32_Foundation", "Win32_System_Com", "Win32_Media_MediaFoundation"] }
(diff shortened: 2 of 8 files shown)


kixelated:
Closing this. While investigating the merge conflict against `dev`, I found that `dev` already has a Windows hardware H.264 encoder (`encode/backend/mediafoundation.rs` + `src/mf.rs`), and it's more complete than this PR: it adds a **zero-copy Direct3D11 texture path** (capture hands the encoder a GPU NV12 texture via a DXGI device manager) with a CPU fallback, plus `MF_E_TRANSFORM_STREAM_CHANGE` handling and async event-status error surfacing. This PR was CPU-only.

I built and ran `dev`'s version on real Windows hardware (it had never been tested on Windows): it compiles clean, and both paths work — 30/30 camera frames flow through as D3D11 textures, and output is valid self-contained Annex-B (`[AUD, SPS, PPS, IDR]`). So this PR is redundant.

The one thing worth keeping is test coverage — `dev` has no encoder tests. I'll open a small follow-up PR against `dev` with the two verified tests (CPU path + the zero-copy camera→texture integration).

(Written by Claude)

PR #1758, opened 2026-06-16
fix(libmoq)!: standardise the pkg-config paths

termoose:
Standardise the pkg-config paths, noticed this issue when trying to build FFmpeg with support for `libmoq`:
```bash
PKG_CONFIG_PATH=/moq/target/lib/pkgconfig ./configure ... --enable-libmoq
```
as seen in my pull request for FFmpeg: https://code.ffmpeg.org/FFmpeg/FFmpeg/pulls/23263.

This should produce a more standard directory structure for the `.pc` file, library files and headers:
```bash
target/{target}/
├── include/moq.h
├── lib/
│   ├── libmoq.a
│   └── pkgconfig/moq.pc
└── release/
    └── libmoq.a
```

Marked as breaking, since it might break builds that depend on the old directory structure, but the previous `moq.pc` file also pointed at the wrong include directory, so not sure if this ever worked properly?

I was also not able to link on Mac OS without `-framework CoreServices`, so that was added. Target for `windows` and fallback (linux etc.) is untouched.

Added a small chore to add the `@moq/json` to the `devDependencies`, not sure why they were removed but `just web` gives me a broken frontend without it.

### DIFF
--- a/js/hang/package.json
+++ b/js/hang/package.json
@@ -27,6 +27,7 @@
 		"zod": "^4.4.3"
 	},
 	"devDependencies": {
+		"@moq/json": "workspace:^",
 		"@types/audioworklet": "^0.0.100",
 		"@types/bun": "^1.3.14",
 		"@typescript/lib-dom": "npm:@types/web@^0.0.350",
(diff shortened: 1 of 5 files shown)


kixelated:
@termoose lemme know if this works for you. I switched the path to be relative instead of absolute so it works for Nix.

termoose:
Works great! Thanks for the assistance! Updated the FFmpeg PR to accomodate these changes now 🤞

PR #1766, opened 2026-06-16
fix(moq-mux): reject frames before the first keyframe instead of dropping

kixelated:
## Summary

A MoQ group must start with a keyframe. The container `Producer` papered over this with a `with_lenient_start` opt-in that **silently dropped** non-keyframes arriving before the first group. That let a malformed stream produce an undecodable group with no signal, and the leniency was load-bearing in ways that weren't obvious at the call sites.

This removes `with_lenient_start` and makes the invariant explicit, with a single legitimate exception (MPEG-TS, which can join mid-GOP).

## What changed

- **`Producer` owns and enforces the invariant** ([producer.rs](rs/moq-mux/src/container/producer.rs)): a non-keyframe with no open group is rejected with a new typed `Error::MissingKeyframe`. `write` now returns `crate::Error` directly (the bound `crate::Error: From<C::Error>` holds for every container error, incl. `fmp4::Error` via the `Cmaf` variant), so importers just `?`-propagate the error rather than re-deriving the decision. Removed `with_lenient_start`.
- **New typed error** ([error.rs](rs/moq-mux/src/error.rs)): `Error::MissingKeyframe` plus `Error::is_missing_keyframe(&anyhow::Error)` to detect it through an anyhow chain.
- **Codec importers** ([h264](rs/moq-mux/src/codec/h264/import.rs), [h265](rs/moq-mux/src/codec/h265/import.rs)) reset per-frame state before writing, so a propagated `MissingKeyframe` leaves them ready for the next access unit. The h264 avc3 path additionally errors on a keyframe it **cannot configure** (no inline SPS, no avcC seed) — the genuinely undecodable case, kept distinct from `MissingKeyframe` so it is never swallowed.
- **TS is the sole exception** ([ts/import.rs](rs/moq-mux/src/container/ts/import.rs)): a transport stream legitimately joins mid-GOP, so `Stream::write` ignores `MissingKeyframe`. This is the single choke point, so it also covers the post-`seek` discontinuity case.
- **FLV propagates it** ([flv/import.rs](rs/moq-mux/src/container/flv/import.rs)): a valid FLV is expected to open on a keyframe, so a leading delta surfaces as a malformed-stream error to the caller.

## Why

The old behavior silently swallowed a real protocol violation. Making it a typed error that the `Producer` returns by default — with mid-stream tolerance opt-in at exactly one importer — means a group can never start without a keyframe, the one format where mid-GOP join is normal (TS) is explicitly the special case, and the keyframe decision lives in one place instead of being re-derived at every call site.

## Public API note

Removing `pub fn with_lenient_start` from `moq-mux`'s container `Producer` is technically a breaking change, but the only callers were internal to the crate. `Producer::write`/`finish`/`seek`/`finish_group` now return `crate::Error` instead of the (always-`crate::Error` in practice) associated `C::Error`. `Error::MissingKeyframe` (a `#[non_exhaustive]` enum) and `Error::is_missing_keyframe` are additive. moq-mux isn't in the explicit breaking-change list, so this targets `main` — happy to retarget `dev` if reviewers prefer.

## Test plan

- [x] `cargo test -p moq-mux` — 236 passed, incl. `survives_midstream_join` and `kyrion_dirtystart_extracts_real_cues`
- [x] New: `keyframe_without_sps_errors`, `delta_before_init_returns_missing_keyframe`; updated `first_frame_must_be_keyframe` to expect `MissingKeyframe`
- [x] `cargo clippy -p moq-mux --all-targets` clean
- [x] `cargo fmt --check` clean (via nix toolchain)
- [x] Dependents compile (`hang`, `moq-cli`, `moq-native`)

No cross-package sync needed: the JS side (`js/hang`) has no equivalent leniency concept.

(Written by Claude)

### DIFF
--- a/rs/moq-mux/src/codec/h264/import.rs
+++ b/rs/moq-mux/src/codec/h264/import.rs
@@ -267,12 +267,8 @@ impl<E: CatalogExt> Import<E> {
 			.as_mut()
 			.context("not initialized; call initialize() first")?;
 
-		// A delta with no open group can't anchor a group (which must start with a
-		// keyframe). Reject it; importers that join mid-stream ignore MissingKeyframe.
-		if !keyframe && !track.has_group() {
-			return Err(crate::Error::MissingKeyframe.into());
-		}
-
+		// A delta before the first keyframe is rejected by the producer with
+		// MissingKeyframe; importers that join mid-stream ignore that error.
 		track.write(crate::container::Frame {
 			timestamp: pts,
 			payload: data.to_vec().into(),
@@ -441,12 +437,9 @@ impl<E: CatalogExt> Import<E> {
 
 		let track = self.track.as_mut().context("avc3 track not created")?;
 
-		// A delta with no open group can't anchor a group (which must start with a
-		// keyframe). Reject it; importers that join mid-stream ignore MissingKeyframe.
-		if !keyframe && !track.has_group() {
-			return Err(crate::Error::MissingKeyframe.into());
-		}
-
+		// Per-frame state is reset above, so a delta before the first keyframe (which
+		// the producer rejects with MissingKeyframe) leaves the importer ready for the
+		// next access unit. Importers that join mid-stream ignore that error.
 		track.write(crate::container::Frame {
 			timestamp: pts,
 			payload,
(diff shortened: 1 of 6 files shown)


kixelated:
@arielmol I missed `with_lenient_start` in the original PR.

I'll fix this on the `dev` branch, but IMO we should error on invalid input instead of skipping. The TS importer can be a special case that allows joining mid-stream, but like we should tell the user they doing stuff wrong instead of silently accepting it IMO.

I think on the dev branch I'll add a `MissingKeyframe` error that the TS importer will ignore.

PR #1770, opened 2026-06-16
feat(moq-mux): cap pending AU growth and add discontinuity()

arielmol:
## Summary

Two additive, internal hardening changes to the moq-mux codec importers.
No wire-protocol or breaking API change, so this targets `main`.

- **Bound the in-progress access unit.** H.264/H.265/AV1 append every
  NAL/OBU to a pending buffer and only flush on a slice/frame; a stream
  that never sends one (repeated AUD/SEI/SPS/PPS, or non-frame OBUs) grew
  it without limit. Every append, including the SPS/PPS/VPS re-inserted
  before a keyframe, now goes through a checked helper that bails past
  `MAX_PENDING_FRAME_BYTES` (16 MiB). moq-gst's per-buffer limit does not
  bound accumulation across buffers, so this is the right layer for it.
- **Add `discontinuity()`** to signal a timeline boundary (e.g. a
  GStreamer FLUSH). It closes the open group via
  `Producer::finish_group()` so the next frame starts a fresh group, and
  for H.264/H.265/AV1 also drops the partial access unit. It does not use
  `seek(0)` (which would rewrite the group sequence) and does not touch
  lenient_start, so it stays forward compatible with the stricter keyframe
  contract in #1766.

## Public API (all additive, non-breaking)

- `Framed::discontinuity()`
- `discontinuity()` on the `h264`/`h265`/`av1`/`vp8`/`vp9`/`aac`/`opus`
  codec importers.

## Not in this PR

The moq-gst `Pad::flush()` caller that will invoke `Framed::discontinuity()`
ships with the moqsink promotion (it lives on a separate branch). This PR is
the moq-mux half only.

## Test plan

- [x] `cargo test -p moq-mux`: 243 passed (10 new: cap + discontinuity per
  codec, an end-to-end no-leak test, and a group-boundary test).
- [x] The group-boundary test was verified faithful: deleting
  `finish_group()` from `discontinuity()` makes it fail (2 frames vs 1).
- [x] `cargo clippy -p moq-mux --all-targets` clean, `cargo fmt --check` clean.

(Written by Claude)

### DIFF
--- a/rs/moq-mux/src/codec/aac/import.rs
+++ b/rs/moq-mux/src/codec/aac/import.rs
@@ -79,6 +79,13 @@ impl<E: CatalogExt> Import<E> {
 		Ok(())
 	}
 
+	/// Signal a timeline discontinuity: close the current group so the next
+	/// frame starts a fresh group.
+	pub fn discontinuity(&mut self) -> anyhow::Result<()> {
+		self.track.finish_group()?;
+		Ok(())
+	}
+
 	pub fn decode<T: Buf>(&mut self, buf: &mut T, pts: Option<crate::container::Timestamp>) -> anyhow::Result<()> {
 		let pts = self.pts(pts)?;
 
--- a/rs/moq-mux/src/codec/av1/import.rs
+++ b/rs/moq-mux/src/codec/av1/import.rs
@@ -29,11 +29,34 @@ pub struct Import {
 	jitter: MinFrameDuration,
 }
 
+/// Cap on a single in-progress temporal unit. Without a frame OBU the importer
+/// never flushes `chunks`, so a stream of non-frame OBUs (sequence headers,
+/// metadata) would grow it unboundedly; bail instead.
+const MAX_PENDING_FRAME_BYTES: usize = 16 * 1024 * 1024;
+
 #[derive(Default)]
 struct Frame {
 	chunks: BytesMut,
 	contains_keyframe: bool,
 	contains_frame: bool,
+}
+
+impl Frame {
+	fn append_obu(&mut self, obu: &[u8]) -> anyhow::Result<()> {
+		let next = self.chunks.len() + obu.len();
+		anyhow::ensure!(
+			next <= MAX_PENDING_FRAME_BYTES,
+			"pending AV1 frame exceeds {MAX_PENDING_FRAME_BYTES} bytes without a frame OBU"
+		);
+		self.chunks.extend_from_slice(obu);
+		Ok(())
+	}
+
+	fn clear_partial(&mut self) {
+		self.chunks.clear();
+		self.contains_keyframe = false;
+		self.contains_frame = false;
+	}
 }
 
 impl Import {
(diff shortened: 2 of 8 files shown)


arielmol:
Well the rabbit left no comments? 🐰

rs/moq-mux/src/codec/aac/import.rs
```diff
@@ -79,6 +79,13 @@ impl<E: CatalogExt> Import<E> {
 		Ok(())
 	}
 
+	/// Signal a timeline discontinuity: close the current group so the next
+	/// frame starts a fresh group.
+	pub fn discontinuity(&mut self) -> anyhow::Result<()> {
+		self.track.finish_group()?;
```
kixelated, line 85: ? each frame should be its own group

rs/moq-mux/src/codec/av1/import.rs
```diff
@@ -29,13 +29,36 @@ pub struct Import {
 	jitter: MinFrameDuration,
 }
 
+/// Cap on a single in-progress temporal unit. Without a frame OBU the importer
+/// never flushes `chunks`, so a stream of non-frame OBUs (sequence headers,
+/// metadata) would grow it unboundedly; bail instead.
+const MAX_PENDING_FRAME_BYTES: usize = 16 * 1024 * 1024;
```
kixelated, line 35: Maybe we should export `moq_net::MAX_FRAME_SIZE` and use that instead?

rs/moq-mux/src/codec/av1/import.rs
```diff
@@ -418,6 +441,17 @@ impl Import {
 		Ok(())
 	}
 
+	/// Signal a timeline discontinuity: drop the in-progress temporal unit and
+	/// close the current group so the next frame starts a fresh group with a
+	/// keyframe. Tolerant before the track exists.
+	pub fn discontinuity(&mut self) -> anyhow::Result<()> {
```
kixelated, line 447: IDK I really don't like this TS stuff leaking everywhere. Partially writing a frame is just gross.

I'm architecting moq-mux on `dev` now. There's a separate `av1::Split` module that finds frame boundaries, while the dumb `av1::Import` takes entire frames (not a stream). TS and stdin can use `av1::Split`, while everybody else uses `av1::Import` directly instead.

rs/moq-mux/src/codec/av1/import.rs
arielmol: > Maybe we should export `moq_net::MAX_FRAME_SIZE` and use that instead?

Agreed

rs/moq-mux/src/codec/aac/import.rs
arielmol: Right, for AAC/Opus each frame already calls finish_group() after write() so discontinuity(..) here just refinishes a closed group already (no-op), I'll drop discontinuuty from the audio importers and keep it only on video

rs/moq-mux/src/codec/av1/import.rs
arielmol: Your call, just two questions:
 1. is it ok if I land just the pending AU cap here for now?  
 2. once Split exists, where should a FLUSH hook in? SPlit or still surfaced via Framed? (moqsink will need this)

PR #1771, opened 2026-06-17
moq-gst: replace moqsink with a rewritten, unit-tested core

arielmol:
## Summary

Replaces the `moqsink` element with a rewritten core that separates the timeline,
per-pad state, and session lifecycle into independently testable pieces. The old sink
interleaved connection, pads, timing, finalization, and errors, with no dedicated unit
tests and ~16% line coverage. The new sink module has a dedicated unit suite (timeline,
segment classification, session lifecycle, flush, per-pad generations); the moq-gst
crate's suite is 59 unit + 3 element tests.

It is a compatible superset of the old element: same element name, same `sink_%u`
request pads, the same seven codec caps (H.264/H.265 byte-stream/au, AV1, VP8, VP9,
AAC raw, Opus), and the same three writable properties (`url`, `broadcast`,
`tls-disable-verify`). It adds three read-only diagnostics: `connected`, `moq-version`,
`estimated-send-bitrate`.

New behavior: shared running-time continuity (no per-pad anchoring), explicit pad
membership with per-pad generations, per-pad error isolation, B-frame support
(decode-order frames, no PTS monotonicity filter), a frame-size cap, and FLUSH handling
(timeline re-anchor: drop to NoSegment, recover on the next SEGMENT).

History note: the branch builds the core as a parallel `moqsinkspike` element, then
promotes it to `moqsink` (rename + remove the old), then applies post-review polish.
Squash-merge is fine.

## Session handoff (non-blocking)

The session handoff is a non-blocking unbounded channel: `chain` accepts each buffer
toward the session worker and returns, never blocking the streaming thread. On the
connected steady path the worker writes into the MoQ producers, whose cache owns
retention and eviction, so loss under congestion is governed by MoQ cache/retention
semantics rather than by stalling the encoder. `moqsink` is therefore not a GStreamer
backpressure boundary; pacing a non-live/file source belongs upstream (for example
`identity sync=true`).

This replaces an earlier bounded-channel design, where a send blocked on a full channel
could not be interrupted by an out-of-band `FLUSH_START` and so could wedge a
seek/teardown. With a non-blocking handoff there is no blocked send, so the
flush-cancellation apparatus is gone: the bounded channel, `send_or_flush`/`SendOutcome`,
and the per-pad flush `watch` gates that cut a blocked send (plus the `AddPad`/`DropPad`
blocking sends). `FlushSignal` and the per-pad generations stay; `FlushSignal` now carries
only the timeline re-anchor. FLUSH re-anchors a pad's timeline (drop to NoSegment, recover
on the next SEGMENT), not an anti-deadlock cut.

## Behavior differences vs the old moqsink (intentional, not regressions)

- **Stricter caps.** AAC requires `mpegversion=4` + `stream-format=raw`; H.264/H.265
  require `stream-format=byte-stream` + `alignment=au`; a non-unit-rate SEGMENT
  invalidates the timeline. The old sink validated none of this in code (only the pad
  template), so it accepted inputs the rewrite now rejects per-pad. The seven codec
  *types* are unchanged.
- **Absolute running-time origin.** Timestamps use shared running time (for A/V
  alignment) instead of the old per-pad `reference_pts` rebasing, so a late-joining live
  pipeline emits a large first timestamp rather than ~0.

## Testing

- Unit + lint: 59 unit + 3 element tests (crate-wide), `cargo clippy -D warnings`,
  `cargo fmt`.
- Live e2e: a fleet of senders (x86_64 + aarch64) publishing to a real relay;
  subscribing back confirmed decoded frames with no errors.
- Transport failure: connect-failure (hermetic element test); mid-stream relay death
  (SIGKILL the local relay -> session error on the bus + non-zero exit ~35 s after the
  kill, consistent with QUIC's ~30 s idle timeout for a silent peer; a graceful close is
  instant; this latency is shared with the old sink).
- Memory-leak, local soak (dev box, 35 min): `x264 -> moqsink` in isolation held RSS
  flat (24 consecutive identical samples) with zero leaked GstMiniObjects at teardown.
- Memory-leak, overnight bisection (target node, about 12-14 h): `msdk -> moqsink` vs
  `msdk -> fakesink` (differ only in the sink) showed identical flat RSS slopes
  (about +0.07 KB/min, R^2~=0.1), so moqsink adds no measurable leak over the encoder.

## Notes

- Public API: no change to any `pub` item in library crates; internal to `rs/moq-gst`
  (a plugin), plus three property rows in `doc/bin/gstreamer.md`.
- Targets `main` (no wire or public library API change).
- Known/accepted: `moq_net::Session::closed()` always returns `Err`, so a clean remote
  close surfaces as a bus ERROR rather than EOS.
- Future work (out of scope): building the session producers up front and writing
  directly (removing the data channel and its connect-window residual buffer), non-VCL
  NAL accumulation in `moq-mux::import`, `Framed::discontinuity()` for an AU split across
  buffers, stats/observability, a latency benchmark harness.

(Co-Written by Claude)


arielmol:
To the holy rabbit:

 - The blocking_send on AddPad/DropPad is a known limitation, already called out in the comment  and marked  as future work.

 - For the connect-failure test I'm keeping `.invalid`. The `127.0.0.1:1` suggestion assumes TCP refused semantics,  but this is QUIC over UDP, so a closed port won't refuse _fast_, it'd just sit until the handshake timeout and get flakier.  `.invalid` is a reserved TLD that never resolves, so it fails immediately.

 - The `.ok()` on the timestamp overflow is fine, it only triggers on ABSURD durations and decode_frame already takes an optional timestamp. Can add a warning log if you'd rather see it.

For CI fails, will fix it asap.

arielmol:
Update on the byte round-trip for https://github.com/moq-dev/moq/pull/1771 done. I ran a VP8 payload round-trip: videotestsrc(ball) -> vp8enc -> moqsink -> relay -> moqsrc, capturing encoded buffers before moqsink and after moqsrc. The harness writes one file per GStreamer buffer and checks received hashes against a contiguous slice of the published hashes.

150 buffers published, all 150 distinct; 149 received; every received buffer matched the published VP8 payload byte-for-byte and in order. In this run the missing buffer was the tail not draining before teardown, not a payload mismatch.

PR #1778, opened 2026-06-18
Route moq-boy status and command tracks through moq-json

kixelated:
## Summary

moq-boy hand-rolled its JSON tracks: the Rust side serialized the status struct, deduped against the last string, and appended a one-frame group per change; the JS side read raw frames and wrote raw command JSON. This PR routes both JSON tracks through the existing snapshot/delta helper (`moq-json` / `@moq/json`), the same one the hang catalog already uses.

- **Status track** (Rust publishes → JS consumes): `StatusPublisher` now wraps `moq_json::Producer<Status>`, dropping the manual serialize / `last_json` dedupe / `append_group` logic. The JS viewer reads it with `Json.Consumer`, reusing the existing `GameStatusSchema` for validation instead of a separate `safeParse`.
- **Command track** (JS publishes → Rust consumes): each viewer's command track is written via a per-track `Json.Producer` and read with `moq_json::Consumer<RawCommand>`, so snapshots and RFC 7396 merge-patch deltas are reconstructed into a full command before dispatch.

The status track now gets **merge-patch deltas** (delta ratio left at the default `8`): its per-frame stats and latency change incrementally, so deltas are a clear win, while unchanged values are still skipped. Malformed-command and invalid-status handling stays lenient (warn and continue); a real track error still ends the stream. `serde_json` is no longer a direct dependency of `moq-boy`.

Both ends of each track ship together, so the wire-format shift (snapshot-per-group → snapshot+deltas) is internal to moq-boy.

## Test plan

- [x] `cargo build -p moq-boy` and `cargo clippy -p moq-boy` clean
- [x] `bun run check` (tsc) passes for `js/moq-boy`
- [x] Biome clean on `js/moq-boy/src/index.ts`
- [ ] Manual: run moq-boy + web viewer, confirm status (buttons/latency/stats) renders and button/reset input still controls the emulator

(Written by Claude)

---
_Generated by [Claude Code](https://claude.ai/code/session_014YHhHt7bc3hqfiCMNB3gNE)_


js/moq-boy/src/index.ts
```diff
@@ -225,18 +226,21 @@ export class Game {
...
 			for (;;) {
-				const json = await Promise.race([effect.cancel, statusTrack.readJson()]);
-				if (!json) break;
-
-				const result = GameStatusSchema.safeParse(json);
-				if (!result.success) {
-					console.warn("Invalid status JSON:", result.error);
+				let status: GameStatus | undefined;
+				try {
+					status = (await Promise.race([effect.cancel, consumer.next()])) || undefined;
```
kixelated: Do we need promise.race any longer? Can we cancel the Json.Consumer instead?

kixelated: Good call, dropped it in 4b4a004. The race was redundant: `effect.cleanup(() => statusTrack.close())` already closes the track on teardown, and closing it unblocks a pending `consumer.next()` (it returns `undefined`, since `recvGroup`/`readFrame` check `state.closed`), so the loop ends on its own. This matches the watch audio/video decoder loops, which read the consumer directly and rely on the subscription closing. The loop body here is a synchronous `status.set`, so there's no long-running work that needed `effect.cancel` to interrupt. (Written by Claude)

---

PR #1781, opened 2026-06-18
fix(hang): omit empty HEVC constraint component in codec string

### LINKED ISSUE #1780: hang: HEVC codec string with all-zero constraint flags emits a trailing "." that fails to parse ("expected int")
## Summary

For HEVC streams whose `general_constraint_indicator_flags` are **all zero**, `hang`'s `H265`
`Display` produces a codec string ending in a dangling `.` (an empty constraint field), e.g.
`hvc1.1.60.L150.`. The matching `FromStr` then tries to parse that empty field as a hex integer and
fails, which surfaces to callers as `json: expected int`.

This breaks the catalog round-trip for such streams: catalog generation on publish **and** the
consumer side (`subscribe_catalog` / `subscribe_media`, and `moq-cli subscribe --format fmp4`) all
error with `json: expected int`. All-zero constraint flags are common in live HEVC from hardware
encoders, so this is hit in practice when publishing a real camera via `moq-cli … publish … fmp4`
and subscribing.

## Reproduce

Unit level (in `rs/hang/src/catalog/video/h265.rs`):

```rust
// Display does not round-trip for all-zero constraint flags:
let v = H265 {
    in_band: false, profile_space: 0, profile_idc: 1,
    profile_compatibility_flags: [0x60, 0, 0, 0],
    tier_flag: false, level_idc: 150,
    constraint_flags: [0, 0, 0, 0, 0, 0],
};
assert_eq!(v.to_string(), "hvc1.1.60.L150."); // trailing '.'
H265::from_str("hvc1.1.60.L150.").unwrap();    // panics: Err parsing "" as int
```

End to end: publish any HEVC stream whose SPS has zero constraint flags via `fmp4` and subscribe →
the catalog track fails with `json: expected int` (observed with `moq-relay` v0.12.11 +
`moq-cli` v0.7.32).

## Root cause

`rs/hang/src/catalog/video/h265.rs`:

- `Display` always appends `.{constraints}`; when all constraint bytes are zero, `constraints` is
  `""`, so the string ends with a stray `.`.
- `FromStr` parses every `.`-separated field after the level as a hex byte, so the empty trailing
  field hits `u8::from_str_radix("", 16)` → error.

## Fix

Emit the constraint component (with its leading `.`) only when there are significant bytes, and let
`FromStr` tolerate an empty trailing field so streams already published with the dangling `.` still
parse. The three existing tests still pass; added a zero-constraint round-trip test.

```diff
@@ impl fmt::Display for H265 {
-		// Skip the trailing "0" elements
-		let skip = self.constraint_flags.iter().rev().skip_while(|b| **b == 0).count();
-		let constraints = self
-			.constraint_flags
-			.iter()
-			.take(skip)
-			.map(|b| format!("{b:X}"))
-			.collect::<Vec<_>>()
-			.join(".");
+		// Constraint flags: hex encoded, trailing-zero bytes omitted. When every
+		// byte is zero the component is omitted ENTIRELY (carry its own leading
+		// '.' only when non-empty); otherwise the string ends in a dangling '.'
+		// and an empty final field that `FromStr` rejects ("expected int"). This
+		// happens for streams whose general_constraint_indicator_flags are all
+		// zero (e.g. some IP-camera HEVC encoders).
+		let significant = self.constraint_flags.iter().rev().skip_while(|b| **b == 0).count();
+		let constraints = if significant == 0 {
+			String::new()
+		} else {
+			let body = self
+				.constraint_flags
+				.iter()
+				.take(significant)
+				.map(|b| format!("{b:X}"))
+				.collect::<Vec<_>>()
+				.join(".");
+			format!(".{body}")
+		};
 
 		write!(
 			f,
-			"{}.{}{}.{}.{}{}.{}",
+			"{}.{}{}.{}.{}{}{}",
@@ impl FromStr for H265 {
 		let parts = parts.enumerate();
 		for (i, constraint) in parts {
+			// Tolerate an empty trailing field (a dangling '.') emitted by older
+			// encoders/muxers so we still parse their streams once this is fixed.
+			if constraint.is_empty() {
+				continue;
+			}
 			if i >= 6 {
 				return Err(Error::InvalidCodec);
 			}
 			constraint_flags[i] = u8::from_str_radix(constraint, 16)?;
 		}
```

New test:

```rust
#[test]
fn test_h265_zero_constraints() {
    let decoded = H265 {
        in_band: false, profile_space: 0, profile_idc: 1,
        profile_compatibility_flags: [0x60, 0, 0, 0],
        tier_flag: false, level_idc: 150,
        constraint_flags: [0, 0, 0, 0, 0, 0],
    };
    assert_eq!(decoded.to_string(), "hvc1.1.60.L150");
    assert_eq!(H265::from_str("hvc1.1.60.L150").unwrap(), decoded);
    assert_eq!(H265::from_str("hvc1.1.60.L150.").unwrap(), decoded); // tolerate dangling '.'
}
```

Happy to open a PR if useful.

quidamschwarz:
Fixes #1780.

HEVC streams whose `general_constraint_indicator_flags` are all zero produced a codec string ending
in a dangling `.` (e.g. `hvc1.1.60.L150.`), which `H265::from_str` then rejected while parsing the
empty trailing field as a hex int. This breaks the catalog round-trip for such streams — both
catalog generation on publish and `subscribe_catalog`/`subscribe_media` on consume (surfaces to
callers as `json: expected int`). All-zero constraint flags are common in live HEVC from hardware
encoders, so it's hit publishing a real camera via `moq-cli … publish … fmp4`.

### Changes — `rs/hang/src/catalog/video/h265.rs`
- **`Display`**: emit the constraint component (and its leading `.`) only when there are significant
  bytes, so all-zero flags yield `hvc1.1.60.L150` instead of `hvc1.1.60.L150.`.
- **`FromStr`**: tolerate an empty trailing field, so streams already published with the dangling
  `.` still parse (version skew while producers update).
- **Test**: `test_h265_zero_constraints` (zero-constraint round-trip + dangling-`.` parse). The
  existing `test_h265` / `test_h265_long` / `test_h265_out_of_band` cases are unchanged.

`cargo test -p hang` passes (4/4 in the `h265` module).

### DIFF
--- a/rs/hang/src/catalog/video/h265.rs
+++ b/rs/hang/src/catalog/video/h265.rs
@@ -47,19 +47,29 @@ impl fmt::Display for H265 {
 			.collect::<Vec<_>>()
 			.join("");
 
-		// Skip the trailing "0" elements
-		let skip = self.constraint_flags.iter().rev().skip_while(|b| **b == 0).count();
-		let constraints = self
-			.constraint_flags
-			.iter()
-			.take(skip)
-			.map(|b| format!("{b:X}"))
-			.collect::<Vec<_>>()
-			.join(".");
+		// Constraint flags: hex encoded, trailing-zero bytes omitted. When every
+		// byte is zero the component is omitted ENTIRELY (carry its own leading
+		// '.' only when non-empty); otherwise the string ends in a dangling '.'
+		// and an empty final field that `FromStr` rejects ("expected int"). This
+		// happens for streams whose general_constraint_indicator_flags are all
+		// zero (e.g. some IP-camera HEVC encoders).
+		let significant = self.constraint_flags.iter().rev().skip_while(|b| **b == 0).count();
+		let constraints = if significant == 0 {
+			String::new()
+		} else {
+			let body = self
+				.constraint_flags
+				.iter()
+				.take(significant)
+				.map(|b| format!("{b:X}"))
+				.collect::<Vec<_>>()
+				.join(".");
+			format!(".{body}")
+		};
 
 		write!(
 			f,
-			"{}.{}{}.{}.{}{}.{}",
+			"{}.{}{}.{}.{}{}{}",
 			match self.in_band {
 				true => "hev1",
 				false => "hvc1",
@@ -116,6 +126,11 @@ impl FromStr for H265 {
 
 		let parts = parts.enumerate();
 		for (i, constraint) in parts {
+			// Tolerate an empty trailing field (a dangling '.') emitted by older
+			// encoders/muxers so we still parse their streams once this is fixed.
+			if constraint.is_empty() {
+				continue;
+			}
 			if i >= 6 {
 				return Err(Error::InvalidCodec);
 			}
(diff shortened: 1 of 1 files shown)


kixelated:
thanks, I'll fix the unrelated CI failure

PR #1788, opened 2026-06-19
feat(moq-net): encode Hop ID as fixed-width 64-bit on lite-05

kixelated:
Implements [moq-dev/drafts#33](https://github.com/moq-dev/drafts/pull/33) ("Encode Hop ID as fixed-width 64-bit integer") for **moq-lite-05**.

Hop IDs are randomly assigned, so a variable-length integer would almost never be shorter than its fixed-width equivalent. Switching to a fixed-width big-endian 64-bit integer also recovers the 2 bits a 62-bit varint would have cost (the value can now use the full 64-bit space).

## Summary

Applies to the three moq-lite wire fields that carry Hop ID values, gated to lite-05+ so lite-04 and earlier keep the varint:

- **ANNOUNCE_OK** `Hop ID` → `(64)`
- **ANNOUNCE** `Hop ID` list → `(64)`
- **ANNOUNCE_INTEREST** `Exclude Hop` → `(64)`

### Rust (`rs/moq-net`)
- New `Version::hops_fixed_width()` predicate. Older versions are listed explicitly so future versions default forward (per the version-matching convention).
- `Origin`'s `Encode`/`Decode` are now concrete over `lite::Version` so they can branch: the fixed-width path uses big-endian `put_u64`/`get_u64` and carries the full 64-bit space (no 62-bit cap); older versions keep the varint + cap.
- `exclude_hop` routes through `Origin` to share the same branch.

### JS mirror (`js/net`)
- Added `Reader.u64()` / `Writer.u64()` fixed-width helpers (+ `setBigUint64`).
- `hopsFixedWidth(version)` predicate mirroring Rust.
- `OriginSchema` widened 62 → 64 bits. `randomOrigin` deliberately stays masked to 62 bits, because the same id must still varint-encode if a session negotiates lite-04.

### Not included
The **moqt relay-hops** half of the draft PR (`HOP_PATH` list of 8-byte entries, `EXCLUDE_HOP` type `0x40B58` → `0x40B59` length-prefixed) targets the moq-transport KVP parameters, which are not implemented anywhere in this repo yet, so there is nothing to change.

## Branch targeting
Targets **dev** per the branch-targeting rules (wire-protocol change under `rs/moq-net`).

## Test plan
- [x] `cargo test -p moq-net` (389 pass), incl. new tests: full `u64::MAX` origin round-trip, fixed-width byte count, `exclude_hop` round-trip across `[0, 7, 1<<53, u64::MAX]`.
- [x] Full Rust workspace builds (moq-relay / moq-bench included).
- [x] `bun test` in `js/net` (180 pass), incl. new `u64` fixed-width stream test; integration test exercises the AnnounceOk + hops flow end-to-end.
- [x] clippy + `cargo fmt --check` (via nix), biome, and `tsc --noEmit` all clean.

(Written by Claude)

### DIFF
--- a/rs/moq-net/src/model/origin.rs
+++ b/rs/moq-net/src/model/origin.rs
@@ -14,16 +14,17 @@ use crate::{
 	coding::{Decode, DecodeError, Encode, EncodeError},
 };
 
-/// A relay origin, identified by a 62-bit varint on the wire.
+/// A relay origin (Hop ID), a randomly-assigned identifier carried in the hop chain.
 ///
 /// `id` must be non-zero for a real origin; `id == 0` is reserved as a
 /// placeholder for Lite03-style hops where the actual value isn't carried.
-/// Encoding a value outside the 62-bit range (>= 2^62) will fail at the
-/// varint layer; [`Origin::random`] picks a valid random nonzero id.
+/// On the wire lite-05+ encodes the id as a fixed-width 64-bit integer (the full
+/// 64-bit space), while older versions used a 62-bit varint. [`Origin::random`]
+/// picks a random nonzero id.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct Origin {
-	/// Non-zero 62-bit identifier. Encoded as a QUIC varint on the wire.
+	/// Non-zero identifier. Encoded as a fixed-width 64-bit integer on the wire (lite-05+).
 	pub id: u64,
 }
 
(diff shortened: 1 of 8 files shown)


rs/moq-net/src/model/origin.rs
```diff
@@ -64,25 +65,37 @@ impl fmt::Display for Origin {
...
+	fn decode<R: bytes::Buf>(r: &mut R, version: crate::lite::Version) -> Result<Self, DecodeError> {
+		if version.hops_fixed_width() {
+			if r.remaining() < 8 {
+				return Err(DecodeError::Short);
+			}
+			// The full 64-bit space is valid here (no 62-bit cap).
+			Ok(Self { id: r.get_u64() })
+		} else {
+			let id = u64::decode(r, version)?;
+			if id >= 1u64 << 62 {
```
kixelated: This should be impossible?

kixelated: Right, that was dead code. A lite `Version` varint decodes via `decode_quic`, which masks the top 2 bits off, so the value already caps at 2^62-1 and the check could never fire. Dropped it in 562a96f4. (Written by Claude)

PR #1789, opened 2026-06-19
feat(relay): close sessions when the token/cert expires

kixelated:
## Summary

Authentication was a one-time gate. The JWT `exp` claim was checked only in `Key::decode` at connect time, the relay's `AuthToken` then **dropped** `claims.expires`, and the session was held open indefinitely via a bare `session.closed().await?`. mTLS client certificates were validated only during the TLS handshake and never re-checked. So a credential that expired mid-session kept working until the client disconnected.

This enforces the credential's expiry for the whole session: the relay now closes the connection once the JWT `exp` (or the mTLS certificate's `notAfter`) passes, forcing the client to reconnect with a fresh credential.

## What changed

- **`rs/moq-relay/src/auth.rs`** — added `expires: Option<SystemTime>` to `AuthToken` (the struct is `#[non_exhaustive]`, so this is additive). Populated from `claims.expires` in `finalize()`, so both the JWT path and the auth-API path carry it.
- **`rs/moq-relay/src/connection.rs`** — `Connection::run` races `session.closed()` against `tokio::time::sleep` until `token.expires` (sleeps forever via `pending()` when there's no expiry), closing with `Error::Unauthorized` when the deadline passes. The mTLS branch sets `token.expires` from the peer certificate.
- **`rs/moq-native`** — new `peer_certificate_expiry()` helper that downcasts the backend's `peer_identity()` to `Vec<CertificateDer>` and parses the leaf's `notAfter` with `x509-parser` (already in the lockfile). Exposed on `QuinnRequest`, `NoqRequest`, and the `Request` wrapper; other backends return `None`.
- **`doc/bin/relay/auth.md`** — notes that `exp` (and mTLS `notAfter`) is enforced for the session lifetime, not just at connect.

## Reviewer notes

- **Public API (additive, non-breaking):** `AuthToken.expires` field (on a `#[non_exhaustive]` struct); new `pub fn peer_certificate_expiry()` on `moq_native::Request`, `QuinnRequest`, `NoqRequest`. New optional dep `x509-parser`, gated behind the `quinn`/`noq` features. Targets `main` since there are no wire or breaking changes.
- **Clock caveat:** the timer computes the delay from wall-clock `SystemTime` but `tokio::time::sleep` runs on the monotonic clock, so a large wall-clock jump after connect isn't tracked. This matches how `exp` was originally validated (`SystemTime::now()`). Exactness under clock skew would need a periodic re-check instead; happy to switch if preferred.

## Test plan

- [x] New unit tests: JWT `exp` survives `finalize`; cert `notAfter` parses correctly; `None`/wrong-type identity yields `None` without panicking.
- [x] `cargo fmt --check`, `clippy` (both crates, `noq` feature on), and all lib tests pass via nix.

(Written by Claude)

### DIFF
--- a/rs/moq-native/src/noq.rs
+++ b/rs/moq-native/src/noq.rs
@@ -515,25 +515,14 @@ impl NoqRequest {
 		}
 	}
 
-	/// Whether the peer presented a client certificate that rustls validated
-	/// against the configured `tls.root` during the handshake.
-	pub fn has_peer_certificate(&self) -> bool {
+	/// The client certificate chain the peer presented, if any, validated by
+	/// rustls against the configured `tls.root` during the handshake.
+	pub fn peer_identity(&self) -> Option<crate::tls::PeerIdentity> {
 		let conn = match self {
 			NoqRequest::Raw { connection, .. } => connection,
 			NoqRequest::WebTransport { request, .. } => request.conn(),
 		};
-		conn.peer_identity().is_some()
-	}
-
-	/// The `notAfter` of the peer's end-entity certificate, if it presented one
-	/// that rustls validated during the handshake. Used to close the session
-	/// once the certificate expires.
-	pub fn peer_certificate_expiry(&self) -> Option<std::time::SystemTime> {
-		let conn = match self {
-			NoqRequest::Raw { connection, .. } => connection,
-			NoqRequest::WebTransport { request, .. } => request.conn(),
-		};
-		crate::tls::peer_certificate_expiry(conn.peer_identity())
+		crate::tls::PeerIdentity::from_any(conn.peer_identity())
 	}
 
 	/// Reject the session with a status code.
--- a/rs/moq-native/src/noq.rs
+++ b/rs/moq-native/src/noq.rs
@@ -525,6 +525,17 @@ impl NoqRequest {
 		conn.peer_identity().is_some()
 	}
 
+	/// The `notAfter` of the peer's end-entity certificate, if it presented one
+	/// that rustls validated during the handshake. Used to close the session
+	/// once the certificate expires.
+	pub fn peer_certificate_expiry(&self) -> Option<std::time::SystemTime> {
+		let conn = match self {
+			NoqRequest::Raw { connection, .. } => connection,
+			NoqRequest::WebTransport { request, .. } => request.conn(),
+		};
+		crate::tls::peer_certificate_expiry(conn.peer_identity())
+	}
+
 	/// Reject the session with a status code.
 	pub async fn close(
 		self,
--- a/rs/moq-native/src/server.rs
+++ b/rs/moq-native/src/server.rs
@@ -618,7 +618,6 @@ impl Request {
 	/// Only the Quinn and noq backends support mTLS; other backends always
 	/// return `None`. Use it to grant elevated access or to close the session
 	/// once the certificate expires (see [`crate::tls::PeerIdentity::expiry`]).
-	#[cfg(any(feature = "quinn", feature = "noq"))]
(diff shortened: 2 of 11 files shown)


rs/moq-native/src/noq.rs
```diff
@@ -525,6 +525,17 @@ impl NoqRequest {
 		conn.peer_identity().is_some()
 	}
 
+	/// The `notAfter` of the peer's end-entity certificate, if it presented one
+	/// that rustls validated during the handshake. Used to close the session
+	/// once the certificate expires.
+	pub fn peer_certificate_expiry(&self) -> Option<std::time::SystemTime> {
```
kixelated: Should this be `peer_identity() -> Option` instead so it's more generic? And can replace the `has_peer_certificate` function too?

Is there a better type we could return? Like a x509 cert type. Then moq-relay can trivially check it?

rs/moq-relay/src/connection.rs
```diff
@@ -99,8 +99,26 @@ impl Connection {
 
 		tracing::info!(version = %session.version(), transport, "negotiated");
 
-		// Wait until the session is closed.
-		session.closed().await?;
+		// The credential (JWT `exp` or client cert `notAfter`) is only checked at
+		// connect time, so hold the session open no longer than the credential is
+		// valid. Without an expiry, sleep forever and just wait for close.
+		let expiry = async {
```
kixelated: Is there a simpler way to write all of this stuff?

rs/moq-native/src/noq.rs
kixelated: Done. Collapsed both into a single, more generic `peer_identity() -> Option<PeerIdentity>` (replacing `has_peer_certificate`, which is now `peer_identity().is_some()`). `PeerIdentity` owns the validated cert chain and exposes `chain()` plus `expiry()`, so moq-relay just calls `identity.expiry()`. I kept the returned type as the owned DER chain rather than an `x509_parser::X509Certificate`, since that type borrows from the DER bytes (self-referential) and would force x509-parser into moq-relay; `expiry()` does the parse internally instead. (Written by Claude)

rs/moq-relay/src/connection.rs
kixelated: Simplified: dropped the async block + `tokio::select!` in favor of `let-else` for the no-expiry case and `tokio::time::timeout(remaining, session.closed())` for the rest. (Written by Claude)

rs/moq-native/src/server.rs
```diff
@@ -612,30 +612,25 @@ impl Request {
...
+	/// against a configured [`crate::tls::Server::root`] during the handshake.
 	///
-	/// Only the Quinn and noq backends support mTLS; other backends always return `false`.
-	pub fn has_peer_certificate(&self) -> bool {
+	/// Only the Quinn and noq backends support mTLS; other backends always
+	/// return `None`. Use it to grant elevated access or to close the session
+	/// once the certificate expires (see [`crate::tls::PeerIdentity::expiry`]).
+	#[cfg(any(feature = "quinn", feature = "noq"))]
+	pub fn peer_identity(&self) -> Option<crate::tls::PeerIdentity> {
 		match self.kind {
```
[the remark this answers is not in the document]
kixelated: Fixed in 09ca272: `peer_identity()` and `PeerIdentity` are no longer feature-gated (x509-parser is now a normal dep), and the `_ => None` no-backend fallback is restored, so `--no-default-features --features iroh,websocket` compiles again. Verified the websocket-only relay build. The tls.rs `chain()` semver nitpick is documented in 0a39bdee. (Written by Claude)

Issue #1799, opened 2026-06-19
Proposal: broadcast contribution profile for MoQ (faithful TS / satellite replacement)


t0ms:
# Proposal: a broadcast contribution profile for MoQ (faithful TS / satellite replacement)

## Summary

MoQ today is a media-aware platform: ingest demuxes a transport stream into elementary
access units, carries them as catalog + container tracks, and (for browsers) repackages to
CMAF/WebCodecs. That is exactly right for browser playback, ABR, transcode, and WebRTC.

It is *not* sufficient for the broadcast contribution / distribution use case, where the goal
is to replace a satellite link: take a broadcaster's native MPEG-TS feed, carry it across MoQ,
and hand a downstream IRD / MVPD a stream that is byte-faithful and carries the metrics and
error-correction operators expect from SRT / Zixi / RIST.

This issue proposes a **broadcast contribution profile** and lays out the work as a set of
child tasks. It deliberately presents two architectural options **neutrally** and asks for a
direction decision before we split the children into separate issues.

A working reference implementation of most of this exists privately (proof-of-concept, see
"Reference implementation" below). It can be shared on request.

## Relationship to #1798

#1798 ("moq-cli ts round-trip drops multi-PPS H.264 (and DTS)") is a **prerequisite for the
media-aware lane**, not a duplicate of this. Those are real bugs in the demux/remux path and
worth fixing regardless of the direction chosen here. This proposal is about the larger
question that #1798 surfaced: *can a demux/remux path ever be a faithful broadcast carrier, or
do we also need an opaque lane?*

## The core architectural question

Demux to access units and remux back to TS is a **re-origination**, not a passthrough. Even
after #1798 is fixed (multi-PPS retained, DTS preserved), a remuxed stream still loses or
alters everything the access-unit model does not carry:

- ancillary elementary streams / metadata: CEA-608/708 captions in SEI, AFD / bar data,
  teletext, DVB subtitles, SCTE-35 splice info, private-data PIDs;
- SI tables beyond PAT/PMT: SDT, EIT, TDT/TOT, NIT, CAT, and PMT descriptors (language, AC-3,
  etc.) - the remux rebuilds a minimal PAT/PMT;
- exact PCR values & placement, continuity counters, stuffing, CBR shaping / null padding,
  precise PTS/DTS (rescaling drift), MPTS / multi-program structure, and any CAS / encryption.

For browser delivery none of that matters. For an IRD doing satellite replacement, byte
[...]

PR #1815, opened 2026-06-19
feat(moq-mux): generic verbatim MPEG-TS carriage (mpegts catalog section)

kixelated:
## Summary

Generalizes the SCTE-35-specific MPEG-TS path into a **generic per-PID verbatim lane**. Any PMT-listed elementary stream the demuxer doesn't decode (SCTE-35, DVB AC-3, teletext/subtitles, private data, ...) is now carried byte-for-byte instead of being dropped, described in a new `mpegts` catalog section. SCTE-35 becomes one instance of this.

This is the "tier 2" lane from the discussion under #1799: keep full demux for browser/ABR, but stop discarding the ancillary streams a contribution feed carries — and faithfully reconstruct them (PIDs, descriptors, PES stream_ids, program signaling) on export.

## Catalog: the `mpegts` section

One unified per-track map keyed by MoQ track name, plus the program-level descriptors:

```rust
pub struct Mpegts {
    pub tracks: BTreeMap<String, Track>,        // media AND verbatim tracks
    pub program_descriptors: Vec<Descriptor>,   // PMT program_info, verbatim
}
pub struct Track {
    pub pid: u16,                               // original MPEG-TS PID
    pub descriptors: Vec<Descriptor>,           // PMT ES descriptors (language, registration, ...)
    pub verbatim: Option<Verbatim>,             // Some ⇒ carried undecoded
}
pub struct Verbatim {
    pub stream_type: u8,
    pub framing: Framing,                       // Pes | Section
    pub stream_id: Option<u8>,                  // original PES stream_id (PES framing)
}
pub struct Descriptor { pub tag: u8, pub data: Bytes /* base64 */ }
```

Decoded media keeps its codec config in the base `video`/`audio` sections; the `mpegts` entry records only its MPEG-TS identity. Empty section serializes to `{}`, so a non-TS broadcast is byte-identical. `Import`/`Export` are generic over a small `Catalog` trait so the media-only `()` catalog and the `Ext` (mpegts) catalog share one code path.

## Behaviour

- **Import** routes every undecoded PMT PID to a verbatim track (PES-framed via the normal PES reassembly; section-framed — SCTE-35 and other private sections — intercepted before the reader). It records each track's PID, PMT ES descriptors, the program-level descriptors, and the verbatim PES `stream_id`.
- **Export** rebuilds the PMT from the media renditions + `mpegts.tracks`: preserves PIDs, re-emits captured ES + program descriptors verbatim (deriving the SCTE-35 'CUEI' registration only when none were captured), re-emits each verbatim PES under its original `stream_id` (falling back to `private_stream_1`), and repacketizes per `framing`. It refreshes track metadata across catalog snapshots until the PMT is built, so PID/descriptors/stream_id filled by the importer across several publishes aren't dropped.

## Public API (moq-mux library)

- **Added:** `container::ts::catalog::{Mpegts, Track, Verbatim, Framing, Descriptor, Ext, Catalog (trait)}`.
- **Removed (breaking):** `container::ts::scte35` module; `Export::with_scte35` → `Export::with_ts`.

Per [Branch Targeting], a breaking library-API change would normally target `dev`; on `main` at the maintainer's request (the SCTE-35 surface is recent and unreleased in this form).

## Validated on a real CNN 1080i contribution feed

`CNNiEMEA2.ts`, 80 MB prefix, import → export → re-import is identical by PID:

- **Media** (H.264 1080i + MP2) decoded; PIDs + PMT descriptors captured, including the **audio ISO-639 language (0x0a)** and video (0x28) descriptors.
- **3 program-level descriptors** round-trip.
- **Five verbatim streams the old demux-only path dropped**, all byte-exact with PIDs and stream_ids preserved:

| PID | stream_type | framing | stream_id | descriptors | frames |
|---|---|---|---|---|---|
| 0x7b | 0x06 | PES | 0xBD | lang, **AC-3 (0x6a)**, max-bitrate | 233 |
| 0x8d / 0x8e / 0x8f | 0x86 | Section | — | — | 66/66/67 (3× SCTE-35) |
| 0x83 | 0x06 | PES | 0xBD | lang, **teletext (0x56)** | 1683 |

(Video decode for this feed's two-PPS H.264 requires #1812; the two PRs are independent and compose cleanly.)

This is faithful at the elementary-stream + PMT level (content + signaling), not byte-for-byte at the transport level — PCR cadence, continuity counters, stuffing, and SI tables outside the PMT are regenerated. Literal byte fidelity (and CAS, exact PCR for TR 101 290) is the opaque whole-mux lane (Option B in #1799), future work.

## Out of scope / follow-ups

- **CLI wiring** to carry verbatim streams through `moq pub/sub`.
- **PCR-PID preservation** and **MPTS/multi-program** (SPTS-scoped today).
- A mid-session PMT *version change* that alters ES/program descriptors is recorded once, not refreshed (rare).

## Test plan

- [x] `cargo test -p moq-mux` (237), incl. `private_pes_carried_verbatim`, `carries_all_sections_verbatim`, `mpegts` catalog (de)serialization, SCTE-35 fixture round-trips
- [x] `cargo clippy -p moq-mux --all-targets -- -D warnings`, `cargo fmt --check`, and `cargo doc -D warnings` clean
- [x] `moq-cli` / `moq-relay` / `hang` / `libmoq` build
- [x] Real CNN 1080i feed round-trips identical by PID (with #1812 for video)

## Cross-package sync

No `js/hang` / `js/net` mirror: the `mpegts` section is a moq-mux-side extension; browser consumers ignore the unknown section. Base `hang` catalog and the `moq-net` wire are unchanged.

(Written by Claude)


kixelated:
Addressed the review feedback in `54d2cf5`/`b055063`:

- **Gate derived CUEI on section framing** — done; only a section-framed `0x86` derives the program-level CUEI registration now.
- **Select SCTE-35 by `stream_type`, not map order** (import_test) — already handled in the unified-model reshape: tests now find the cue track via `verbatim.stream_type == 0x86`.
- **`#[non_exhaustive]` on `Ext`** — done.
- **`StreamType::from_u8` can `Err`** — export now skips (with a warning) a verbatim stream whose recorded `stream_type` isn't a known TS type instead of aborting. Round-tripped streams always parse (the type came from the import's PMT); this only guards a hand-built catalog.

**Deferred (wants a maintainer call):** *preserve the original PES `stream_id` for verbatim PES streams.* Today tier-2 carriage regenerates PES headers and preserves the elementary-stream **payload** byte-for-byte (validated: teletext/AC-3/SCTE all round-trip exactly) plus the PMT signaling; the PES `stream_id` is not preserved, so a verbatim PES is rewritten with the audio `stream_id`. Faithful preservation means adding a `stream_id` field to the `mpegts` `Verbatim` record and threading it through import→catalog→export — a catalog-schema change I'd rather confirm before making. PES-header fidelity is otherwise an explicit non-goal of this lane (that's the opaque whole-mux lane in #1799).

(Written by Claude)

kixelated:
Follow-up in `147cdef` / `740249e`:

- **Preserve the PES `stream_id`** (the deferred item). The original `stream_id` is now captured on import into `mpegts` `Verbatim.stream_id` and re-emitted on export, instead of relabeling every non-video stream as audio (`0xC0`). DVB teletext / subtitles / DVB AC-3 ride `private_stream_1` (`0xBD`); rewriting them to audio is rejected by strict broadcast demuxers and TR 101 290 analyzers. Absent (a non-TS source) falls back to `private_stream_1`. Validated on a real feed: AC-3 and teletext both preserve `0xBD` across the round-trip. (No special handling for the no-optional-PES-header stream_ids like `private_stream_2`/padding — those belong to the opaque whole-mux lane; the realistic verbatim PES all use header-bearing ids.)
- **Clear stale section route on PMT remap** (the outside-diff note). `ensure_stream` now finishes/drops a prior `sections` entry when a later PMT remaps a section PID to a PES codec/verbatim stream, so it isn't intercepted as a section forever. Skipped the suggested `Stream::Ignored`-replacement half: codec support is fixed at construction, so an Ignored PID can't later become decodable, and replacing it would just re-warn on every repeated PMT.

(Written by Claude)

rs/moq-mux/src/codec/legacy.rs
```diff
@@ -79,6 +79,11 @@ impl<E: CatalogExt> Import<E> {
 		})
 	}
 
+	/// The underlying MoQ track producer (for its name, etc.).
+	pub fn track(&self) -> &moq_net::TrackProducer {
```
kixelated: Can we not expose the Producer? Just the `name()` would be fine.

rs/moq-mux/src/container/ts/catalog.rs
```diff
@@ -0,0 +1,243 @@
...
+
+use std::collections::BTreeMap;
+
+use bytes::Bytes;
+use serde::{Deserialize, Serialize};
+
+use crate::catalog::hang::CatalogExt;
+
+/// Serialize [`Bytes`] as a base64 string in the catalog JSON.
+mod base64_bytes {
```
kixelated: serde_from or serde_as instead?

rs/moq-mux/src/container/ts/catalog.rs
```diff
@@ -0,0 +1,243 @@
...
+pub struct Mpegts {
+	/// Per-track MPEG-TS info, keyed by MoQ track name. Media tracks record their
+	/// PID and PMT descriptors; undecoded tracks add a [`Verbatim`] carriage record.
+	#[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
+	pub tracks: BTreeMap<String, Track>,
+
+	/// PMT program-level descriptors (`program_info`), carried verbatim. Export
+	/// re-emits these; the SCTE-35 'CUEI' registration is derived when absent.
+	#[serde(default, skip_serializing_if = "Vec::is_empty")]
+	pub program_descriptors: Vec<Descriptor>,
```
kixelated, line 36: `descriptors` ?

rs/moq-mux/src/container/ts/catalog.rs
```diff
@@ -0,0 +1,243 @@
...
+}
+
+/// Carriage record for an undecoded elementary stream carried byte-for-byte.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Default)]
+#[serde(rename_all = "camelCase")]
+#[non_exhaustive]
+pub struct Verbatim {
+	/// PMT `stream_type` to re-announce (0x86 SCTE-35, 0x06 private PES, 0x05
+	/// private sections, ...).
+	pub stream_type: u8,
```
kixelated, line 83: Is it worth having an enum (with unknown(u8) or no?

rs/moq-mux/src/container/ts/catalog.rs
```diff
@@ -0,0 +1,243 @@
...
+	/// subtitles/DVB AC-3, 0xC0-0xDF audio). Preserved so export re-emits the PES
+	/// under its real id rather than relabeling it, which strict broadcast demuxers
+	/// and TR 101 290 analyzers reject. `None` for section framing or a non-TS
+	/// source; export then falls back to `private_stream_1`.
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub stream_id: Option<u8>,
+
+	/// How the verbatim bytes are timestamp-framed as MoQ frames.
+	#[serde(default)]
+	pub container: hang::catalog::Container,
```
kixelated: Just hard-code hang::legacy::Container? Seems dumb to support CMAF.

rs/moq-mux/src/container/ts/catalog.rs
```diff
@@ -0,0 +1,243 @@
...
+/// Implement this for an application extension to compose MPEG-TS carriage with
+/// additional sections.
+pub trait Catalog: CatalogExt {
+	/// The section to record MPEG-TS details into, or `None` for an extension that
+	/// doesn't carry them.
+	///
+	/// Keep this stable per catalog: an importer samples support once at
+	/// construction, so a result that flips between `Some` and `None` mid-stream
+	/// would disable verbatim carriage or fail.
+	fn mpegts_mut(&mut self) -> Option<&mut Mpegts>;
```
kixelated, line 158: Why `mut`? Is this needed?

rs/moq-mux/src/container/ts/catalog.rs
```diff
@@ -0,0 +1,243 @@
...
+	pub mpegts: Mpegts,
+}
+
+impl CatalogExt for Ext {}
+
+/// An extension that can carry an `mpegts` catalog section.
+///
+/// Implement this for an application extension to compose MPEG-TS carriage with
+/// additional sections.
+pub trait Catalog: CatalogExt {
```
kixelated, line 151: And why do we need another trait?

rs/moq-mux/src/container/ts/export.rs
```diff
@@ -313,42 +331,87 @@ impl<E: scte35::Catalog> Export<E> {
...
 			if self.tracks.contains_key(name) {
 				continue;
 			}
-			let kind = scte35_kind(config, name)?;
-			let source = ExportSource::for_scte35(&self.broadcast, name, config, self.latency)?;
+			// Skip (don't abort) a verbatim stream whose recorded stream_type isn't a
+			// known TS stream type, so it can't be announced in the PMT. Round-tripped
+			// streams always parse (the type came from the import's PMT); this guards a
+			// hand-built catalog. Mirrors the `active` filter above.
+			if StreamType::from_u8(verbatim.stream_type).is_err() {
```
kixelated: Do we really need to gate based on StreamType?

rs/moq-mux/src/container/ts/export.rs
```diff
@@ -313,42 +331,87 @@ impl<E: scte35::Catalog> Export<E> {
...
 		self.tracks.retain(|name, _| active.contains_key(name));
 		Ok(())
 	}
 
+	/// Pick a PID for a new track: the original PID recorded in the `mpegts` section
+	/// when it's free, else the next free PID from [`FIRST_ES_PID`]. PID `0x0000`
+	/// (PAT), [`PMT_PID`], `0x1FFF` (null), and any already-assigned PID are avoided
+	/// so cross-references survive when possible.
+	fn alloc_pid(&self, name: &str, tracks: &std::collections::BTreeMap<String, catalog::Track>) -> u16 {
+		let mut used: std::collections::HashSet<u16> = self.tracks.values().map(|t| t.pid).collect();
```
kixelated: IDK If we need to construct a HashSet every time, just iterate over the tracks.

kixelated:
Thanks — addressed your comments + a self-review in `09a1d505`. CI doc-link failure (`unresolved link to \`ts\``) is also fixed.

**Changed as suggested:**
- **base64** (`catalog.rs:19`): now `#[serde_as] ... #[serde_as(as = "Base64")]` via `serde_with`, matching `rs/hang`; deleted the hand-rolled module.
- **`Verbatim.container`** (`:116`): dropped — verbatim frames are always Legacy-framed, hard-coded now.
- **`legacy::Import`** (`legacy.rs:83`): exposes `name() -> &str` instead of the `Producer`.
- **`alloc_pid` HashSet** (`export.rs:397`): gone — PID assignment is a two-pass over the desired tracks (prefer original, then fill free), no per-call set.
- **`from_u8` gate** (`export.rs:360`): de-duplicated into a single `verbatim_ok` predicate (it was in two places). It's a graceful skip of an unknown `stream_type` rather than aborting the whole export — only reachable from a hand-built catalog, since round-tripped types always parse. Happy to drop it entirely if you'd rather a bad catalog error loudly.

**Correctness fix from the review (you + CodeRabbit both flagged the class):** export was *latching* per-track metadata from the first catalog snapshot. The importer fills a track's PID, ES descriptors, and PES `stream_id` across **several** publishes (the codec importer creates the rendition; `record_media_track` adds PID/descriptors on the first frame; `VerbatimStream` backfills `stream_id`). Export now **refreshes** existing tracks' PID/kind/descriptors on every snapshot until the PMT is built, so none of that is dropped.

**Kept (with reasoning — let me know if you disagree):**
- **`stream_type` as `u8`** (`:100`) rather than an enum-with-`Unknown(u8)`: it's a raw PMT value we carry verbatim and never interpret (only `0x86`→CUEI and `from_u8` to re-announce). `mpeg2ts` already owns the `StreamType` enum; mirroring it in the catalog is maintenance for no gain. An `Unknown(u8)` enum would also serialize less cleanly.
- **The `Catalog` trait** (`:173`): it's what lets `Import`/`Export` be generic over `()` (media-only, zero `mpegts` overhead — what `moq-cli` uses today) **and** `Ext` with one code path; the `()` impl returns `None`. Without it, every TS import/export carries the `mpegts` machinery or we duplicate code paths. If you'd prefer to drop the trait and always use `Ext` (section omitted when empty), I can — it simplifies the generics but makes the section always-present in the type.
- **`program_descriptors`** (`:53`): the PMT `program_info` descriptors (e.g. the SCTE-35 CUEI registration, program-level registration), carried verbatim so export re-emits them rather than re-deriving.
- **`mut` (`:180`)**: `clippy -D warnings` runs in CI, so there's no unused `mut` — the one you spotted is mutated (the test builds the section then inserts into it).

**Known minor edges I left** (rare, happy to fix if you want): a *mid-session PMT version change* that alters a track's ES descriptors or adds CUEI to an already-non-empty `program_info` is recorded once and not refreshed. Unusual in practice; the old SCTE path re-detected CUEI per-PMT.

(Written by Claude)

kixelated:
Follow-up (`caf90517`):

- **`from_u8` gate: removed.** A verbatim stream imported from TS always has a known `stream_type` (mpeg2ts parsed it from the PMT), so the gate was unreachable for real data — it only guarded a hand-built catalog. `build_psi`'s existing `from_u8` conversion is the backstop. (This moots the "warn on skip" suggestion — there's no skip path anymore.)

- **Dropping the `Catalog` trait: investigated, recommend keeping it.** Tried making `Import`/`Export` concrete over `Ext`, but the trait turns out to be load-bearing: the crate's generic unified import layer (`import::Framed::new` / `import::Import::new`) takes one base `Producer<()>` and feeds it to *every* format importer including TS; the trait (`mpegts_mut() -> Option<_>`, `None` for `()`) is what lets TS accept that base catalog with zero overhead for media-only callers. Going concrete-`Ext` forces `Producer<Ext>` up through both unified importers and every caller (moq-cli, libmoq/ffi) — a crate-wide cascade. Reverted; happy to do the full cascade if preferred.

(Written by Claude)

arielmol:
Re-read at 09a1d505 round-trip is narrower than carriage. Import carries undecoded stream_type verbatim, but export skips anything StreamType::from_u8() doesn't recognize, so an exotic stream type imports but drops on export, for example KLV crosses the bridge but canot be remuxed.

arielmol:
Oh reviews in cascade, too fast for me :)

arielmol:
bradhards' KLV will work transparently with this

PR #1818, opened 2026-06-19
feat(moq-relay): add a `Relay` embed entry point + flatten-friendly config loader

kixelated:
## What

Extracts the wiring that `main.rs` inlined into a reusable **`Relay`** type and a generic **config loader**, so the relay can be embedded in a larger binary with almost no boilerplate — e.g. a node that runs the relay *plus* in-process workers against its shared origin.

`main.rs` itself is rewritten to use the new API (~80 lines → ~6), dogfooding it with no behavior change.

## Why

The relay is already a library, but embedding it meant copy-pasting ~80 lines of `main.rs` (build server/client/auth/cluster/web, the `serve` accept loop, the `tokio::select!`) plus re-listing every sub-config to add one flag. A downstream binary ([moq-pro's `moq-edge`](https://github.com/moq-dev/moq-pro)) wants exactly this: run the relay and host in-process workers (a VOD recorder) that consume `cluster.origin` directly, with no localhost QUIC hop. This gives it a clean seam instead of a fork of `main`.

## API

```rust
let relay = moq_relay::Relay::new(config).await?;   // build the stack
let origin = relay.origin().clone();                // the one OriginProducer
tokio::spawn(my_worker(origin));                    // attach in-process work
relay.run().await                                   // cluster + web + accept loop + jemalloc
```

- **`Relay::new(config).await?`** builds the full stack (QUIC server, cluster + origin + stats, auth, web).
- **`Relay::origin()`** exposes the `OriginProducer` every session and cluster peer publishes into, so embedders can attach producers/consumers before running. `Relay::cluster()` gives the full handle.
- **`Relay::run().await`** drives the cluster mesh, web server, QUIC accept loop, and (under `jemalloc`) the heap profiler, returning the first error. It logs `listening` and notifies **systemd readiness**, so embedders inherit the `Type=notify` contract for free. The private `serve` loop moved into the new `relay` module.
- **`load_config::<T>(|c| c.relay.file.clone())`** lets a downstream binary **flatten** `Config` into its own clap parser (adding flags alongside every relay flag) and reuse the exact `CLI → TOML → CLI` merge. `Config::parse_and_merge` now delegates to the same generic `merge_from_args`, so there's a single merge implementation.

## Tests

- New `embedder_can_flatten_config` exercises all four corners of the embed contract at once: clap `flatten` + the positional `file`, **serde `flatten` of `Config`** (its `deny_unknown_fields` must not break embedding — confirmed it doesn't), and the CLI re-apply landing an embedder-specific flag.
- Doc-tests on `Relay` and `load_config` show the intended usage and are compiled/run.
- All **125 lib tests + 3 doc-tests pass**, `clippy` clean, `cargo fmt` clean (default and `jemalloc` features build).

## Notes

- No behavior change to the standalone binary: `Relay::run` reproduces `main`'s exact `select!` arms, ordering, "listening" log, and systemd notify.
- `Relay::new` is `async` only because binding the optional iroh endpoint is async.

### DIFF
--- a/rs/moq-relay/src/config.rs
+++ b/rs/moq-relay/src/config.rs
@@ -1,3 +1,4 @@
+use anyhow::Context;
 use clap::Parser;
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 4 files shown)


rs/moq-relay/src/config.rs
```diff
@@ -87,14 +88,68 @@ impl Config {
...
+/// # Ok::<(), anyhow::Error>(())
+/// ```
+///
+/// `file` extracts the optional TOML path from the parsed value (e.g.
+/// `|c| c.relay.file.clone()`). Unlike [`Config::load`] this does NOT initialize
+/// the logger; call `config.relay.log.init()` yourself once you have the value.
+///
+/// The clap+TOML clobber pitfall documented on [`Config::load`] applies to the
+/// flattening config too: type any extra TOML-overridable flag as `Option<T>`.
+pub fn load_config<T, F>(file: F) -> anyhow::Result<T>
```
kixelated, line 127: kinda gross. Do we need to export this? Or can we make the API better?

rs/moq-relay/src/config.rs
```diff
@@ -87,14 +88,68 @@ impl Config {
...
+
+/// Shared implementation behind [`Config::parse_and_merge`] and [`load_config`].
+/// Merge order: CLI args -> TOML file (if named) -> CLI args re-applied so
+/// explicit flags / env vars win over the file.
+fn merge_from_args<I, A, T, F>(args: I, file: F) -> anyhow::Result<T>
+where
+	I: IntoIterator<Item = A>,
+	A: Into<std::ffi::OsString> + Clone,
+	T: Parser + serde::de::DeserializeOwned,
+	F: FnOnce(&T) -> Option<String>,
```
kixelated, line 143: Why is F a function? instead of a file name or some shit.

kixelated:
Closing: the `Relay::new().run()` abstraction is too rigid for our embedder (moq-edge). It owns `Web` internally, and `Web::run` builds its axum router privately with a websocket catch-all (`/{*path}`), so there's no seam to add custom HTTP endpoints — which we need. Going with explicit wiring (copy-paste) in moq-edge instead, for full control over the web server and serve loop.

PR #1825, opened 2026-06-20
fix(moq-video): NVENC falls back instead of aborting when no NVIDIA driver

kixelated:
## Summary

On a Linux host with no NVIDIA driver, `moq-video`'s NVENC backend **aborts the process** instead of falling back to software. cudarc (`CudaContext::new`) and the NVENC SDK both dlopen their driver libraries lazily and **`panic!`** when the library is missing (`cudarc`'s `panic_no_lib_found`, the SDK's "failed to dlopen the NVIDIA encode library"). Under release's `panic = "abort"` that panic aborts; in tests (Cargo forces unwind) it shows up as a hard test failure.

So `Kind::Auto` / `Kind::Hardware` — the default for `moq-cli capture` — crashes on any GPU-less Linux box, contrary to the documented "falls back to software (see backend::open)" behavior.

This was surfaced by a `Kind::Auto` round-trip test in #1802 panicking on the GPU-less CI runner.

## Fix

Probe the driver libraries with `libloading` *before* calling into cudarc / the SDK, and return `Error::Codec` if either is absent so `backend::open` moves on to the next candidate (openh264):

- `libcuda.so.1` / `libcuda.so` (cudarc's CUDA driver), and
- `libnvidia-encode.so.1` / `libnvidia-encode.so` (matches the SDK's own candidate list).

A pre-check (not `catch_unwind`) is required because release builds use `panic = "abort"`, where `catch_unwind` can't help. `libloading` is added as an `nvenc`-feature-gated optional dep, mirroring the existing `cudarc` / `nvidia-video-codec-sdk` gating.

## Test plan

- [x] `cargo fmt --check` clean; non-NVENC `cargo build -p moq-video` unaffected (macOS).
- [x] New `#[cfg(all(target_os = "linux", feature = "nvenc"))]` test `missing_driver_errors_instead_of_panicking`: on a host without the driver, `Nvenc::open` returns `Err` rather than panicking. This runs on the GPU-less Linux CI runner (`cargo test -p moq-video --all-features`), which is exactly where the abort happened before.
- [ ] Could not compile-check the `cfg(linux)` path locally (macOS box; nix/rustup cross-toolchain split blocked a zigbuild). Relying on the Linux CI runner to compile + run the new test.

## Note

VAAPI (`moq-vaapi`, dlopen'd libva) may have the same panic-on-missing-library shape; if so, `Kind::Auto` would still abort at the VAAPI candidate after NVENC falls back. Worth a follow-up check, but out of scope here.

(Written by Claude)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4337,6 +4337,7 @@ dependencies = [
  "bytes",
  "cudarc",
  "dispatch2",
+ "libloading 0.8.9",
  "moq-mux",
  "moq-net",
  "moq-vaapi",
--- a/rs/moq-video/Cargo.toml
+++ b/rs/moq-video/Cargo.toml
@@ -26,7 +26,7 @@ doctest = false
 # links on a GPU-less builder and still starts on machines without the NVIDIA
 # driver, falling back to software (see backend::open). `cuda-12020` pins the CUDA
 # API version so the build needs no CUDA toolkit.
-nvenc = ["dep:nvidia-video-codec-sdk", "nvidia-video-codec-sdk/dynamic-loading", "cudarc/fallback-dynamic-loading", "cudarc/cuda-12020"]
+nvenc = ["dep:nvidia-video-codec-sdk", "nvidia-video-codec-sdk/dynamic-loading", "cudarc/fallback-dynamic-loading", "cudarc/cuda-12020", "dep:libloading"]
 # Intel/AMD VAAPI hardware encoder (Linux). Off by default; enable for a build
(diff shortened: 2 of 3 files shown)


kixelated:
Folded into #1819 (commit b0e768283): that PR makes hardware encoders always-on, which is what turns this NVENC panic into a crash on every GPU-less Linux box, so the fallback fix belongs there rather than as a separate PR racing it. Same probe (libcuda + libnvidia-encode via libloading) + the GPU-less test, adapted to the always-on dep layout (libloading as a non-optional target.linux dep). Verified in podman: `cargo test -p moq-video --all-features --locked` passes with the driver absent.

(Written by Claude)

PR #1828, opened 2026-06-20
feat(moq-srt): serve broadcasts back out over SRT (m=request egress)

kixelated:
## Summary

`moq-srt` was ingest-only. This adds the reverse direction, turning it into a bidirectional SRT gateway.

A caller dialing the SRT listener with `streamid=#!::r=<broadcast>,m=request` now gets the broadcast re-muxed from the origin back to MPEG-TS and streamed to it, so a plain SRT player can watch any broadcast the origin carries:

```bash
ffplay 'srt://host:9000?streamid=#!::r=cam0,m=request'
vlc    'srt://host:9000?streamid=#!::r=cam0,m=request'
```

Routing now parses the SRT stream-id `m=` mode: `m=request` is egress, everything else (including absent / raw OBS-style ids) stays ingest, so existing publish behavior is unchanged. Requests don't claim the `ActivePaths` guard, so any number of players can pull one broadcast, and they wait for the broadcast to be announced (a player may connect before the publisher does).

The TS muxer (`moq_mux::container::ts::Export`) already existed and re-injects SPS/PPS/VPS on keyframes; this PR is just the SRT-side wiring plus its mirror of the existing ingest `Publisher`.

## Changes

- `src/listen.rs`: `resolve_path` → `resolve_route` (returns broadcast path + `ConnectionMode`); `serve` → `serve_publish`; new `serve_request` packs the muxer output into 1316-byte (7×188) SRT messages.
- `src/ts.rs`: new `Subscriber`, the egress mirror of `Publisher`, wrapping `Export`.
- `src/error.rs`: new `Mux(#[from] moq_mux::Error)` variant.
- README / module docs updated for both directions.

## Public API

New `pub` surface on `moq-srt` (already a `dev`-targeted crate): `ts::Subscriber` is `pub` within the crate but `ts` is a private module, so the only externally visible change is the new `Error::Mux` variant (additive; enum is `#[non_exhaustive]`). New behavior, no breaking changes. Targets **dev** per branch rules.

## On SRT latency

The default 200 ms SRT receive latency can drop the front-loaded keyframe burst on a low-latency link when a player tunes in (ffmpeg logs `non-existing PPS 0` until the next keyframe). This is standard SRT live behavior, not a mux issue, and it is already tunable: SRT negotiates the effective latency as `max(caller, listener)`, and for egress the caller is the receiver, so a player can raise its own buffer with `?latency=N` in the URL (e.g. `?latency=2000`). The gateway-side floor remains `--srt-latency` / `Config::latency`.

## Test plan

- [x] `cargo build/test/clippy/fmt -p moq-srt` (via nix) green; 7 unit tests incl. new mode-parsing cases.
- [x] E2E loopback with ffmpeg: published H.264+AAC TS via `m=publish`, pulled it back via `m=request`. ffprobe identified `h264 320x240 + aac 48kHz`; a capture decoded 119 video + 242 audio frames.
- [ ] Tested against VLC / a hardware SRT decoder (only ffmpeg/ffprobe exercised so far).

(Written by Claude)

### DIFF
--- a/rs/moq-srt/README.md
+++ b/rs/moq-srt/README.md
@@ -1,12 +1,19 @@
 # moq-srt
 
-SRT contribution ingest gateway for Media over QUIC.
+SRT gateway for Media over QUIC, both directions.
 
-SRT carries MPEG-TS. This crate runs an SRT listener, demuxes each connection's
-transport stream with [`moq-mux`](../moq-mux), and publishes the result into a
-MoQ origin as ordinary broadcasts. It's the contribution-ingest analogue of
-`moq-hls`'s import and `moq-rtc`'s WHIP. Pure Rust: SRT is provided by
-`srt-tokio`, with no libsrt or ffmpeg dependency.
+SRT carries MPEG-TS. This crate runs an SRT listener and routes each connection
+by its stream id `m=` mode:
+
+- `m=publish` (the default): ingest. Demux the connection's transport stream
+  with [`moq-mux`](../moq-mux) and publish it into a MoQ origin as an ordinary
+  broadcast. The contribution-ingest analogue of `moq-hls`'s import and
+  `moq-rtc`'s WHIP.
+- `m=request`: egress. Re-mux a broadcast from the origin back to MPEG-TS and
+  stream it to the caller, so `vlc srt://...` and `ffmpeg -i srt://...` can play
+  any broadcast the origin carries (H.264/H.265 video, AAC/AC-3/MP2 audio).
+
+Pure Rust: SRT is provided by `srt-tokio`, with no libsrt or ffmpeg dependency.
 
 ## Library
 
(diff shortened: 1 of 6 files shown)


kixelated:
## Code review (high-effort, recall-biased) + fixes

Ran an 8-angle review. Deduped to three real findings:

**1. (fixed) Egress task leak on `m=request`.** `serve_request` parked on `announced_broadcast` — which never returns `None` for an in-scope path on the local origin — without polling the socket. A caller requesting a stream that's never published, or one that disconnects while waiting, leaked a tokio task holding an open SRT socket (unbounded on an unauthenticated listener). The `m=publish` path was unaffected since it reads the socket immediately.

Fixed in `ed569f6` by racing the announce wait against a socket-closed future (`wait_closed`), so a caller hang-up reaps the task. Verified e2e: a request for a non-existent broadcast followed by client disconnect now logs `closed before its broadcast was available` and the task ends, instead of parking forever.

**2. (accepted) `Instant::now()` send timestamps vs media PTS.** SRT packets are stamped with wall-clock-now rather than the frame PTS, so on tune-in a keyframe burst is handed to SRT with near-identical timestamps and TSBPD releases it as a burst. This is the same tune-in pacing tradeoff discussed on the PR (raise the receiver buffer with `?latency=N`); a proper fix needs a PTS-exposing API change in `moq_mux::Export`, out of scope here. Left as-is intentionally.

**3. (accepted) `m=bidirectional` / no-`m=` receiver routes to ingest.** A caller that isn't `m=request` is treated as a publisher and waits for TS it'll never send. This is documented ("non-request means ingest"), and it self-heals on disconnect (the `ActivePaths` guard releases), same as any idle publisher. No change.

Non-findings confirmed clean: the 1316-byte (7×188) slicing never splits a TS packet mid-way (SRT also reassembles reliably + receivers resync on 0x47); `Error::Mux` is additive on a `#[non_exhaustive]` enum (non-breaking); no em-dashes; `ts::Subscriber` is crate-visible only (private `ts` module).

`cargo build/test/clippy/fmt -p moq-srt` green (via nix); 7 unit tests pass.

(Written by Claude)

t0ms:
Filed #1839 to generalise this SRT egress into a pluggable TS output sink (UDP/RTP/FEC/2022-7) and add the PCR-aware pacing this PR deferred. (Written by Claude)

PR #1824, opened 2026-06-20
feat(moq-rtmp): RTMP / enhanced-RTMP ingest gateway + E-RTMP codecs

kixelated:
Adds **`moq-rtmp`**, an RTMP contribution-ingest gateway (sibling of `moq-srt`), plus the **enhanced-RTMP codec support** in the `moq-mux` FLV demuxer it builds on.

## moq-mux FLV (import + export)

The FLV demuxer previously handled only legacy H.264 + AAC and dropped/rejected the enhanced-RTMP (E-RTMP) FourCC payloads. Both directions now handle:

- **Video**: HEVC (`hvc1`), AV1 (`av01`), VP9 (`vp09`), and the AVC FourCC (`avc1`)
- **Audio**: Opus (`Opus`), AC-3 (`ac-3`), E-AC-3 (`ec-3`), and AAC (`mp4a`)

FLAC/MP3 enhanced audio are dropped (no MoQ catalog codec). Each codec reuses the existing config parsers via new `pub(crate)` helpers (`h265::config_from_hvcc`, `av1::config_from_av1c`, `vp9::config_from_keyframe`); VP9 / AC-3 / E-AC-3 configure in band from the first frame. **No public API change** — purely additive behavior.

## moq-rtmp (new crate)

Runs an RTMP server (OBS, ffmpeg, hardware encoders) and publishes each stream into a MoQ origin. Pure Rust via `rml_rtmp` (handshake / chunk / AMF, no librtmp); RTMP audio/video messages are re-wrapped as FLV tags and fed to `moq_mux::container::flv::Import`, so both legacy and enhanced codecs flow through.

**Auth is a `moq-native`-style `Server` / `Request`** (no callback): `Server::accept()` yields a pending publish; the caller inspects `app()` / `stream_key()`, then calls `request.accept(origin, path)` or `request.reject(reason)`. An embedder (e.g. a relay verifying the stream key as a JWT and scoping the origin per token) owns the policy. The unauthenticated `run(origin, Config)` convenience powers the binary's `serve` / `publish` modes and spawns one task per connection.

Docs: `doc/bin/rtmp.md` + index entry.

## Test plan

- [x] `cargo test -p moq-mux` (268 pass) — incl. new enhanced VP9/Opus import + VP9/Opus export round-trip tests
- [x] `cargo test -p moq-rtmp` (9 pass) — incl. an **end-to-end test** driving a real `rml_rtmp` *client* through handshake → connect → publish against `Server::accept`
[...]

PR #1829, opened 2026-06-20
feat(moq-rtmp): RTMP play mode (egress)

kixelated:
Builds on #1824 (RTMP ingest) to make `moq-rtmp` **bidirectional**. A player now connects to `rtmp://host/<app>/<key>` and the gateway streams the broadcast back down. VLC, ffplay, and mpv can play it (browsers can't, Flash is dead). The same URL round-trips: push to `rtmp://h/live/cam0`, pull it back from `rtmp://h/live/cam0`.

It's the egress counterpart to the publish path, and the sibling of `moq-hls`'s export.

## What changed

**Public API (`moq-rtmp`, breaking, hence `dev`):**

- `Server::accept()` / `accept_stream()` now return an enum `Request::{ Publish(Publish<S>), Play(Play<S>) }` (`#[non_exhaustive]`). The old `Request` struct is renamed `Publish`; `Play` is new. Match on it, inspect `app()` / `stream_key()`, then `accept`/`reject` the right direction. The crate is v0.0.1 with no external consumers, so impact is nil.

**Egress plumbing:**

- `flv::TagReader` — the inverse of the existing `flv::tag()` / `file_header()`. Splits the FLV byte stream `moq-mux`'s `Export` produces back into individual tags, so each body goes out as an RTMP audio/video message.
- `Play::accept` subscribes to the broadcast from an `OriginConsumer`, builds a `moq_mux::container::flv::Export`, and pumps. It waits for the broadcast to be announced (so a player can connect just before the publisher), cancelling cleanly if the viewer disconnects first.
- `play_pump` uses `tokio::io::split` to send media and service client input (acks/pings/teardown) concurrently in one `select!`; broadcast EOF → `finish_playing`.
- `listen::run` serves plays from `origin.consume()` — publishers ingest into the origin, players are served out of it. Plays don't claim an `ActivePaths` slot, so any number of viewers can pull the same broadcast at once.

All codecs flow both ways (legacy H.264/AAC + E-RTMP HEVC/AV1/VP9/Opus/AC-3), since muxing lives in `moq-mux`. Legacy-only players reject the E-RTMP codecs.

## Cross-package sync

- `doc/bin/rtmp.md` and `rs/moq-rtmp/README.md` updated for the egress direction (play examples, routing round-trip, auth/embedding notes). No other rows in the sync table apply (no wire/catalog/FFI changes).

## Test plan

- [x] `cargo test -p moq-rtmp --all-features` (15 pass), including a new **end-to-end play test**: publish a real broadcast via the FLV importer → `Server` play → a real `rml_rtmp` *client* receives the muxed avcC sequence header + keyframe back. Plus 3 `TagReader` unit tests.
- [x] `cargo check -p moq-rtmp --no-default-features` (the library-embedding path)
- [x] `cargo clippy -p moq-rtmp --all-features --all-targets` clean
- [x] `cargo fmt` (via nix devshell)
- [ ] Not yet exercised against a real player (ffplay / VLC / mpv); the e2e test drives a real RTMP client, but a live player round-trip would be the final confidence check.

(Written by Claude)

### DIFF
--- a/doc/bin/rtmp.md
+++ b/doc/bin/rtmp.md
@@ -1,24 +1,31 @@
 ---
 title: moq-rtmp
-description: RTMP / enhanced-RTMP -> MoQ contribution ingest gateway
+description: RTMP / enhanced-RTMP <-> MoQ gateway (ingest and egress)
 ---
 
 # moq-rtmp
 
-`moq-rtmp` ingests [RTMP](https://en.wikipedia.org/wiki/Real-Time_Messaging_Protocol)
-(the protocol OBS, ffmpeg, and most hardware encoders speak) and publishes each
-stream into Media over QUIC as an ordinary broadcast.
+`moq-rtmp` bridges [RTMP](https://en.wikipedia.org/wiki/Real-Time_Messaging_Protocol)
+(the protocol OBS, ffmpeg, and most hardware encoders and players speak) and
+Media over QUIC, in **both directions**:
+
+- **Publish (ingest):** an encoder pushes a stream in, and `moq-rtmp` publishes it
+  into MoQ as an ordinary broadcast.
+- **Play (egress):** a player pulls `rtmp://host/<app>/<key>`, and `moq-rtmp`
+  subscribes to that broadcast from MoQ and streams it back down. VLC, ffplay, and
+  mpv can play it (browsers can't -- Flash is dead).
 
 RTMP carries media as FLV-format audio/video messages. `moq-rtmp` runs the RTMP
 handshake and chunk/AMF session (via the pure-Rust
-[`rml_rtmp`](https://crates.io/crates/rml_rtmp), no librtmp), re-wraps each
-message as an FLV tag, and feeds it to `moq-mux`'s FLV demuxer, which routes the
-media onto MoQ tracks. It's the contribution-ingest sibling of `moq-srt`
-(SRT/MPEG-TS) and `moq-rtc` (WHIP).
+[`rml_rtmp`](https://crates.io/crates/rml_rtmp), no librtmp). On ingest it re-wraps
+each message as an FLV tag and feeds it to `moq-mux`'s FLV demuxer; on egress it
+muxes the broadcast back to FLV with `moq-mux` and sends the tags out as RTMP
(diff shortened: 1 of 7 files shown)


kixelated:
## Self-review (high effort, recall-biased)

Ran an 8-angle review (3 correctness + reuse/simplification/efficiency + altitude + conventions). One real correctness bug found and fixed in 0a03d4b; the rest were design-acceptable behaviors or minor cleanups.

### Confirmed bug (fixed)

**`Play::accept` desynced the RTMP deserializer while waiting for the broadcast.** The pre-broadcast wait used `drain_input`, which read client bytes off the socket and **discarded** them. RTMP is one continuous chunk stream whose parser state lives in the `ServerSession`; skipping any bytes (a partial chunk from a single `read`, or a client `Set Chunk Size`) desynchronizes the deserializer that `play_pump` then resumes using. A player that sends a `Set Buffer Length` / window-ack right after `play` (ffplay does) could desync, breaking the clean `PlayStreamFinished` teardown or erroring out playback.

Fix: `feed_input` now pumps those bytes through `session.handle_input` (keeping the deserializer in sync) and leaves any response packets queued in `work` for `play_pump` to flush, instead of writing them mid-wait. The only await is the read, so dropping the future when the broadcast arrives stays cancellation-safe; it also returns early on a `play` teardown during the wait.

### Considered and not changed

- **No timeout on the broadcast wait.** A player requesting a never-announced path holds the connection open until it disconnects. This is intended ("a player can connect slightly before the publisher") and matches how RTMP servers behave; gating is the firewall/auth story already documented.
- **`FlvExport` runs at `Duration::ZERO` latency.** Correct for live; a mid-GOP joiner waits for the next keyframe, standard RTMP behavior.
- **Raw timestamp passthrough** (no rebasing to 0). Correct for a live stream; ffmpeg-based players base off the first received timestamp.
- **Minor cleanups** (near-identical `Publish::reject`/`Play::reject`; test handshake duplicated between `run_client` and `play_client_collect`; `TagReader::push` copies each chunk) — low value, left for a future pass to keep this PR focused.

Re-verified after the fix: 15 tests pass (incl. the e2e play loop), `clippy --all-targets` clean, builds with and without the `server` feature, `fmt` applied.

(Written by Claude)

PR #1833, opened 2026-06-20
feat(mux,hls): segment-target export mode (one moof per HLS segment)

kixelated:
## Problem

The HLS exporter rolls a video segment on every "independent" fragment (once per GOP), and the moq-mux importer derives a fragment's keyframe flag from the container sample flags. A publisher that over-fragments — notably ffmpeg `-movflags …+frag_every_frame` (what the demo `pub` uses) — writes **one moof per frame** with per-fragment sample flags that mark ~every frame as a sync sample. So the exporter emitted a fragment per frame and the packager rolled a **segment per frame** → zero-duration (`#EXTINF:0.00000`), unplayable HLS.

Concretely, ffmpeg writes for the P-frame fragments `tfhd.default_sample_flags = 0x02000000` (sync) with no per-sample/first-sample override, so the importer faithfully reads every P-frame as a keyframe. (ffprobe still shows correct I/P because it parses the slice headers, not the container flags.)

## Fix

Add an opt-in **segment-target** mode that combines whole GOPs (and any sub-GOP input fragmentation) into exactly **one `moof`+`mdat` per segment**, rolled on a keyframe once the segment reaches the target duration (audio, having no keyframes, rolls purely on the target). This is the shape a VOD packager wants (no LL-HLS parts) and is robust to noisy per-fragment keyframe flags: the per-keyframe and `fragment_duration` part triggers are bypassed, so segments are target-sized rather than per-frame.

- **moq-mux**: `Export::with_segment_target(Option<Duration>)`; `should_flush` gains a segment-target branch + a shared `buffer_span` helper. Default `None` keeps the existing per-GOP / LL-HLS-part behavior — **live/RTMP unaffected**.
- **moq-hls**: `export::Config::segment_target` plumbed to the exporter and to a `fragment_per_segment` mode on `SegmentStore` (each fragment becomes its own segment). Default `None` preserves the live gateway's behavior.
- **test**: `segment_target_combines_gops_into_one_fragment` — three GOPs → two target-sized, keyframe-aligned, single-moof segments.

## Validation

`cargo test -p moq-mux fmp4` → 17 pass (16 existing + new). `cargo clippy -p moq-mux -p moq-hls` clean. Built against the dev rev moq-pro currently pins (`5f047b3d`). The consumer side (moq-pro's VOD recorder) sets `segment_target: Some(~2s)` to opt in.

### DIFF
--- a/rs/moq-hls/src/export/mod.rs
+++ b/rs/moq-hls/src/export/mod.rs
@@ -35,6 +35,14 @@ pub struct Config {
 	pub latency: Duration,
 	/// Target segment duration for audio renditions (video rolls on GOPs).
 	pub audio_segment_target: Duration,
+	/// VOD segment mode. When set, each rendition emits exactly ONE moof+mdat per
+	/// segment, rolled on a keyframe once the segment reaches this duration (audio:
+	/// purely on duration). Combines GOPs / sub-GOP input fragments into one fragment
+	/// per segment -- the right shape for recording (no LL-HLS parts), and robust to
+	/// publishers that over-fragment (e.g. ffmpeg `frag_every_frame`, whose per-frame
+	/// moofs otherwise explode into per-frame segments). `None` (default) keeps the
+	/// per-GOP + LL-HLS-part behavior used by the live gateway.
+	pub segment_target: Option<Duration>,
 }
 
 impl Default for Config {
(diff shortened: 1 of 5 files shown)


kixelated:
Wrong approach: a time-based segment target merges GOPs and can roll at non-keyframe boundaries, which would require transcoding to re-time. Segments must be exactly one GOP, rolled at keyframe boundaries. Re-diagnosing: the container's video keyframe flags are actually correct, so the per-frame segmentation comes from elsewhere in the publish/import path; will redo with a GOP-aligned fix that combines each GOP's per-frame CMAF fragments into one moof.

PR #1841, opened 2026-06-20
feat(moq-net): per-track durable cache on TrackState (RAM + disk/remote tiers)

kixelated:
## Summary

A per-track durable cache in `moq-net`: keep recent groups past the live window and serve them
back on a FETCH, spilling to local disk and optionally remote object storage (`object_store`). It
lives in moq-net so any consumer of a track (relay, edge, archiver) gets durable caching for free.

This started as a design doc (`rs/moq-archive/DESIGN.md`, since deleted) and is now implemented.
The final shape, after review, puts the cache **on the shared `TrackState`** rather than as a
separate handle wired onto each endpoint:

- **RAM is the live window.** No second in-memory copy of recent groups: the cache reuses
  `TrackState.groups`, the buffer the track already keeps for live subscribers. A group is
  serialized to `cache::Group` and handed to the disk tier only when it ages out of that window.
  Inserting a group *is* caching it; the old spawned populate-subscriber is gone.
- **Two retention gates** in `evict_expired`, both sized by `TrackInfo::cache`. A group is evicted
  when it trips **either**:
  - *wall-clock* — received (an `Instant` stamped on arrival, never on the wire) more than the
    window ago. The hard memory backstop against timestamp abuse.
  - *media-time* — its last frame timestamp is more than the window behind the live media edge.
    Bounds a startup stampede where a burst of buffered media arrives "all at once".
- **Serving.** `fetch_group` resolves a live miss from the disk/remote tiers via an async lookup
  spawned on the first poll. A store hit resolves the fetch; a store **miss chains upstream** —
  the lookup task queues the request for a `TrackDynamic` (a wire FETCH for a relay) when one
  exists, so the fetch resolves once upstream serves the group into the live window. Queuing only
  after the store misses keeps the store the fast path and avoids a redundant upstream fetch. With
  no handler, a miss is `NotFound`. `get_group` stays synchronous and live-only. Because the cache
  is on the shared state, every `TrackConsumer` of the track serves from it automatically.
- **Tiers.** Disk and remote are `object_store`, target-gated to `cfg(not(target_arch = "wasm32"))`
  so native builds get them with no flag and wasm drops the server-side stack automatically. The
  disk byte format (`segment.rs`: self-describing segments + rollup), the multi-tier index
  (`index.rs`), and the object_store glue (`store.rs`) are unchanged; the flush task writes one
  disk segment per eviction pass and rolls the oldest disk segments up into large remote objects.

See `rs/moq-net/CACHE.md` for the full write-up.

## Public API (moq-net, library crate — targets `dev`)

New:
- module `moq_net::cache`: `Disk` (+ `new`/`with_remote`), `Bounds`, `Limit`, `Frame`, `Group`
  (+ `read`/`produce`/`size`/`ts_first`/`ts_last`), `Batch`, and the `segment` / `store` submodules.
  `Disk` / `store` are native-only (target-gated).
- `TrackProducer::with_cache(cache::Disk) -> Self` (native-only).

Relative to `dev` there are no removals (the cache is new on this branch). `TrackFetch` gains a
private field. `TrackInfo::cache` is **not** removed yet — it still sources the retention window;
making retention purely local (and dropping the wire field) is deferred.

## Cross-package sync

The cache is local policy, never on the wire, so `js/net` and `doc/concept` need no matching change.
moq-cli/moq-relay flags for `with_cache` are noted as follow-up in CACHE.md.

## Test plan

- [x] `cargo test -p moq-net --lib` (427 pass), including the new `get_group_does_not_read_disk`,
      `fetch_group_serves_evicted_group_from_disk`, `fetch_chains_upstream_on_cache_miss`,
      `fetch_cache_miss_without_dynamic_is_not_found`, and `tiers_evict_then_fetch_back`, plus the
      retained segment/index/store unit tests.
- [x] `cargo clippy -p moq-net --all-targets --all-features`, `cargo fmt`, rustdoc (`-D warnings`),
      `cargo check --no-default-features`, and the native dependents (`moq-relay`, `moq-cli`).
- [ ] wasm target build (gating verified by inspection; the local env's `getrandom` wasm backend
      isn't configured, so CI is the real check).

(Written by Claude)


rs/moq-archive/DESIGN.md
```diff
@@ -0,0 +1,325 @@
...
+  *arrival* order (preserves out-of-order delivery; `next_group()` would skip late arrivals,
+  which we do not want for an archive).
+- `GroupConsumer { sequence }`, `GroupConsumer::read_frame() -> Result<Option<Bytes>>` drains
+  frames in order; `Frame { size, timestamp: Option<Timestamp> }` carries the media timestamp
+  when present (moq-lite-05+).
+- `GroupConsumer::finished() -> Result<u64>` resolves once the group is complete, returning
+  the final frame count. This is our "safe to flush" signal.
+
+**Serving side (produce a track + answer fetches):**
+- `TrackProducer::dynamic() -> TrackDynamic`.
```
kixelated: I think we just take a TrackDynamic.

rs/moq-archive/DESIGN.md
```diff
@@ -0,0 +1,325 @@
...
+- `TrackDynamic::requested_group() -> Result<GroupRequest>` blocks until a consumer FETCHes a
+  group that is not in the producer's cache. While any `TrackDynamic` handle is alive, the
+  miss waits to be served instead of failing fast with `NotFound`.
+- `GroupRequest::sequence() -> u64`, `GroupRequest::accept(info) -> GroupProducer`. We fill the
+  returned `GroupProducer` with `create_frame` / `write` / `finish` from storage, then
+  `GroupProducer::finish()`.
+- To expose the track over a session, wrap the producer in a `BroadcastProducer` and publish
```
kixelated: Note the caller will do this. For example, moq-relay might first try to use moq-archive to handle any dynamic requests.

I think moq-archive also needs it's own dynamic() to make that work? Because if something isn't on disk or on S3, we'll have to ask the original publisher for it.

That might be an argument for stuffing this logic directly into `moq_net::TrackProducer` and `moq_net::TrackConsumer`? It feels a bit gross but IDK the API is far more friendly. We would also definitely know when data is evicted from RAM cache too.

rs/moq-archive/DESIGN.md
```diff
@@ -0,0 +1,325 @@
...
+served from memory without touching disk.
+
+## Storage layout
+
+Everything is an `object_store` key, so the same code paths work for a local dir
+(`LocalFileSystem`) and a bucket (`AmazonS3`). Proposed key scheme, rooted at a configurable
+prefix and namespaced by broadcast/track:
+
+```
+<root>/<broadcast>/<track>/segments/<segment-id>      # concatenated groups
```
kixelated: The broadcast and track names can have slashes. I guess we need to escape them.

rs/moq-archive/DESIGN.md
```diff
@@ -0,0 +1,325 @@
...
+therefore keeps a map of open buffers and only flushes a group on its own `finished()`; it
+never assumes contiguity. The index is keyed by sequence but appended in completion order, and
+the reader's `BTreeMap` makes lookup order-independent. FETCH is inherently random-access
+(consumer asks for an arbitrary old seq), so the read path has no ordering assumptions either.
+Sequence gaps (a group that was lost upstream and never recorded) are legal: a FETCH for a gap
+returns `NotFound`.
+
+## Open questions
+
+1. **`frame_start` granularity.** moq-lite-05 FETCH can request "group N starting at frame K".
```
kixelated: We might remove this from moq-lite-05. The current API doesn't even support it.

rs/moq-archive/DESIGN.md
```diff
@@ -0,0 +1,325 @@
...
+   The cheap path: ranged-GET the whole group, parse, skip K frames in memory (groups are
+   bounded at 32 MB, so this is fine). The optimization: store per-frame offsets in the index
+   for a partial ranged GET. Recommend the cheap path for v1, add per-frame offsets only if
+   profiling demands it.
+2. **Restart/recovery.** Rebuild the `BTreeMap` by listing + reading index objects on startup.
+   Need a crash-consistency story: write the segment object first, then its index entries, so a
+   half-written segment is simply never indexed (and is GC'd by a startup sweep of unindexed
+   segments).
+3. **Aggregation/compaction shape.** When promoting disk -> S3, do we copy segments 1:1 or
```
kixelated: I think we should merge multiple groups at each rollup step. So RAM can be fragmented, but each rollup becomes less fragmented.

ex. store for up to 30s in RAM, flush in 10s segments to disk. Store on disk for up to 5m, flush to S3 in 1m batches? 

IDK if we want to get an LRU involved or we just eyeball RAM usage though. I feel like an LRU is correct in general for both RAM and disk though. We could use the `used()` and `unused()` state too I guess. No sense keeping `unused` stuff in RAM if we could combine it with the next flush to disk.

rs/moq-archive/DESIGN.md
```diff
@@ -0,0 +1,325 @@
...
+   concatenate many small disk segments into one big S3 object (rewriting offsets in the
+   index)? Concatenation is better for S3 request economics but adds a rewrite step. Lean
+   toward 1:1 in v1, compaction later.
+4. **Serving the *latest* group / live edge.** v1 answers FETCH for past groups. Should the
```
kixelated: We always keep the latest group in RAM.

rs/moq-archive/DESIGN.md
```diff
@@ -0,0 +1,325 @@
...
+   archive also serve a live `subscribe` (replay newest groups as they land) so it can stand in
+   for a departed origin? That is closer to DVR and probably a follow-up.
+5. **Index for a hot, long archive.** A multi-day archive has a large index. JSONL + in-RAM
```
kixelated: We should try to nail down the index file format for sure.

rs/moq-archive/DESIGN.md
kixelated: Captured this as the lead open question and kept v1 wiring the chain in `moq-archive` (caller owns the `TrackProducer`, archive takes the `TrackDynamic` plus an upstream `TrackConsumer` to forward misses + record). Agreed the moq-net-native version is the friendlier API and the only one that reliably knows *when* a group is evicted from the RAM cache, which is the natural flush trigger.

The way I'd frame the decision: do we want `moq_net::TrackProducer` to accept a pluggable storage backend (a `trait Cache` it consults on a miss and notifies on eviction), so the archive becomes "attach storage to a track" rather than "wire a fallback chain"? That keeps storage code out of moq-net but gives moq-net the one hook (eviction) we can't get from the outside. If you're good with that shape I'll redesign the public API around it before any implementation; if you'd rather keep moq-net pure for now, the chain version stands. Which way do you want to go?

(Written by Claude)

---

kixelated:
Folded all the review comments into the latest commit:

- **Take a `TrackDynamic`** (not a `TrackProducer`): the archive is now a composable link in a cache chain (relay → archive → origin); caller owns the producer.
- **Miss fallback / chaining**: on a storage miss the archive forwards to an upstream `TrackConsumer` instead of returning `NotFound`; whether this should instead live inside `moq_net` core types is now the lead open question (replied on that thread for your call).
- **Rollup**: each tier concatenates multiple units from the tier above (RAM 30s → 10s disk segments → 1m S3 objects), so fragmentation drops downward. Resolves the old 1:1-vs-concatenate question toward concatenate-at-every-rollup.
- **Eviction**: LRU + byte budget per tier, not just age; **always keep the latest group in RAM**; use `used()`/`unused()` to flush unused groups early.
- **Index format nailed down**: per-segment `postcard` footer (group offset table) + per-track manifest, Parquet/Iceberg-shaped, replacing the JSONL sketch.
- **Slashes**: percent-encode broadcast/track names into single opaque path segments.
- **Sub-group `frame_start`**: dropped (likely removed from moq-lite-05, unsupported by the current API); the archive serves whole groups.

Also added a prior-art survey (BookKeeper ledgers, Kafka KIP-405 segments, Haystack/Bitcask, Parquet/Iceberg) and why no single embeddable crate covers batch + index + S3 tiering, in answer to the earlier "is there an existing library?" question.

(Written by Claude)

---

kixelated:
Added a **Usage mockups** section with concrete call sites, and you're right that the relay doesn't fit the per-track API. Three scenarios:

**A. Standalone / VOD node** — works with `serve(TrackDynamic)` today, because the archive *is* the publisher: it `publish_broadcast`s, and for each `requested_track` it `accept`s the producer, takes `producer.dynamic()`, and serves from storage. No path collision since the live broadcast is gone.

**B. Why moq-relay can't use it as-is** — the relay only holds a single `OriginProducer` and forwards whole broadcasts; it never constructs a `TrackProducer` or calls `.dynamic()`. That happens *inside* moq-net's `lite`/`ietf` subscriber fan-out, so there's no point in relay code to hand the archive a `TrackDynamic`. Forcing it would mean republishing every broadcast through an archive-owned `BroadcastProducer` and re-accepting every track, i.e. reimplementing fan-out around the archive.

**C. The moq-net seam (recommended)** — a pluggable `trait Cache` that moq-net consults on a miss and notifies on eviction, attached at the origin where it already owns the per-track RAM cache. Then the whole relay integration is:

```rust
let origin = Origin::random().produce().with_cache(archive);
```

moq-net catches evictions (the natural flush trigger, which the chain version can't see) and serves misses before they cost an upstream FETCH. The public `Archive` becomes a `Cache` impl plus the standalone `record`/`serve` helpers.

This is the same fork as the open question I flagged, now made concrete: **do you want me to redesign the API around a `moq_net::Cache` trait (relay-friendly, needs a small moq-net addition), or keep the per-track chain version (standalone/VOD only for now)?** I'd recommend the `Cache` trait. Pushed in `93046420`.

(Written by Claude)

---

kixelated:
## Extensive review (self-review with three finder passes + verification)

Ran a high-recall review over `origin/dev...HEAD` (correctness, removed-behavior, cross-file, reuse/simplification/efficiency, altitude, conventions). Fixes pushed in `71deaf25`. Below: the public-API surface, what was fixed, and two open design calls for you.

### Public API surface (all additive → not breaking; `dev` is correct)

New under `moq_net::cache` (`pub mod cache`):
- `Config` (`#[non_exhaustive]`, `Default`), `Bounds`, `Limit`, `Frame`, `Group`, `Batch`
- `Producer` (intentionally **not** `Clone`), `Consumer` (`Clone`); `Group::read` / `Group::produce` bridges
- `cache::segment`: `encode`, `rollup`, `group_from_blob`, `Segment`, `GroupEntry` (`#[non_exhaustive]`), `Error`
- `cache::store` (behind `cache-tiered`): `Store`, `Error`
- `TrackProducer::with_cache`, `TrackConsumer::with_cache`

No existing signature changed. The one planned **breaking** change, removing the wire field `TrackInfo.cache`, is still pending and is `dev`-appropriate.

### Correctness findings — fixed

1. **Disk tier could wipe itself.** `index::promotion` drained the *entire* disk tier when the low watermark was unset (floor of zero), so one over-max trip evicted even the newest flushed groups on the no-remote path. Now keeps the newest segment, mirroring the RAM tier's "keep latest" rule.
2. **Crash-consistency.** `store::flush`/`compact` mutated the index *before* the `put`/upload. A failed write stranded sequences pointing at a nonexistent object. Reordered to write-then-index (added `Index::next_id()` to reserve the key); a failure now leaves the index intact.
3. **Overflow guard.** `store::get` computed `offset + length` unchecked; a corrupt footer could overflow `u64`. Now `checked_add`, matching `segment::blob`.
4. **fetch vs get inconsistency.** `fetch_group`'s cache path surfaced a `produce()` rebuild error (e.g. timescale mismatch on a not-yet-accepted wire consumer) as a hard fetch failure, while `get_group` treated it as a miss. Now both rebuild synchronously and miss-through on error.

### Cleanup — applied

- `store_of` now `expect`s a configured remote tier instead of silently reading disk for a remote location.
- `index` is internal orchestration (only the gated store uses it) → gated behind `cache-tiered` and dropped from the public surface; `segment::GroupEntry` got `#[non_exhaustive]`.
- Varint coding correctly reuses `VarInt::{encode,decode}_quic` (not duplicated); both error enums use `thiserror` + `#[from]` + `#[non_exhaustive]`; no em dashes.

### Two open decisions for you (not unilaterally changed)

**A. `TrackProducer::with_cache` populate disables demand teardown.** The populate path subscribes internally, so `producer.unused()` never resolves while a cache is attached, and a relay that drops idle tracks by demand (`lite`/`ietf` subscriber, `session.rs`) won't drop a cached one. That's *intended* for "keep recording when idle" (archive/DVR), but wrong for a transient relay cache. Options: keep as-is (documented now); populate on group **eviction** instead (no subscription, but needs a sync frame-snapshot primitive on groups); or populate from the relay's existing fan-out consumer. This is the same lifecycle question I flagged earlier — worth settling before wiring the relay.

**B. Same populate loop head-of-line-blocks** on each group's finish (`Group::read` resolves only when a group finishes), so a stalled group delays caching of later finished ones. Fixable with concurrent reads (`FuturesUnordered`) if we keep the subscribe approach — I held off pending the decision in A, since the populate design may change.

Both caveats are now documented on `with_cache`. Everything else is fixed; default suite 427 pass, `cache-tiered` 39 pass, clippy/rustdoc/fmt clean.

(Written by Claude)

---

rs/moq-net/src/model/cache/mod.rs
```diff
@@ -0,0 +1,605 @@
...
+}
+
+impl Group {
+	/// Total size of the group's frame payloads in bytes.
+	pub fn size(&self) -> u64 {
+		self.frames.iter().map(|f| f.payload.len() as u64).sum()
+	}
+
+	/// The first frame's media timestamp, if any. Used as the group's lower time bound.
+	pub fn ts_first(&self) -> Option<Timestamp> {
```
kixelated, line 187: Note that timestamp is optional, and only supported in moq-lite-05. If we actually need this timing information, we also need to store an Instant when the frame was received (wall clock timestamp). 

I don't know if it should be a separate field, or if timestamp should be an enum (producer supplied, or received time).

rs/moq-net/src/model/cache/mod.rs
```diff
@@ -0,0 +1,605 @@
...
+
+	/// The last frame's media timestamp, if any. Used as the group's upper time bound.
+	pub fn ts_last(&self) -> Option<Timestamp> {
+		self.frames.last().and_then(|f| f.timestamp)
+	}
+
+	/// Drain a live [`GroupConsumer`](crate::GroupConsumer) into a cached group, reading every
+	/// frame's payload and timestamp. Resolves once the group is finished, so this is how the
+	/// producer side snapshots a finished group for caching.
+	pub async fn read(mut group: crate::GroupConsumer) -> Result<Self, crate::Error> {
```
kixelated, line 199: I think we can make this more efficient with a `poll` variant? So we don't create the Vec until we know the exact number of frames.

rs/moq-net/src/model/cache/mod.rs
```diff
@@ -0,0 +1,605 @@
...
+			let timestamp = frame.timestamp;
+			let payload = frame.read_all().await?;
+			frames.push(Frame { timestamp, payload });
+		}
+		Ok(Self { sequence, frames })
+	}
+
+	/// Rebuild a live [`GroupConsumer`](crate::GroupConsumer) from this cached group, for serving a
+	/// fetch. `timescale` must match the track's: each frame timestamp is validated against it.
+	pub fn produce(&self, timescale: impl Into<Option<Timescale>>) -> Result<crate::GroupConsumer, crate::Error> {
```
kixelated, line 212: Maybe timescale should be part of `Group`, and rename it to `GroupInfo`? Kinda weird that it's a separate argument. 

And why do we even need the timescale passed in? Shouldn't we know it from the track cache object that created this group?

rs/moq-net/src/model/cache/mod.rs
```diff
@@ -0,0 +1,605 @@
...
+	config: Config,
+	/// Groups keyed by sequence, so the first entry is the oldest and the last is the latest.
+	ram: BTreeMap<u64, Group>,
+	ram_bytes: u64,
+}
+
+impl State {
+	/// The time span between the oldest group's first frame and the newest group's last frame.
+	/// Zero unless both ends carry a timestamp, so a track without media timestamps applies no
+	/// duration pressure (byte bounds still apply).
```
kixelated: We should use wall clock timestamp when no timescale is used.

In fact, that might even be a separate bounds. The producer can lie about media timestamps, but they can't lie about wall clock arrival timestamps.

rs/moq-net/src/model/cache/store.rs
```diff
@@ -0,0 +1,211 @@
...
+	/// The backing object store failed.
+	#[error(transparent)]
+	Store(#[from] object_store::Error),
+}
+
+/// A tiered durable store: a disk object store, an optional remote one, and the index mapping
+/// group sequences to their location. Bands flushed from the RAM tier land here; old disk segments
+/// roll up into the remote tier, or are evicted when there is none.
+pub struct Store {
+	disk: Arc<dyn ObjectStore>,
```
kixelated, line 42: disk should also be optional.

Maybe the stores should be split and take a <T: ObjectStore>? IDK

rs/moq-net/src/model/track.rs
```diff
@@ -691,6 +692,31 @@ impl TrackProducer {
...
+	///
+	/// - The subscriber counts as a consumer, so [`unused`](Self::unused) never resolves while a
+	///   cache is attached. A relay that drops idle tracks via demand will not drop a cached one;
+	///   it stays alive until it ends or the producer is dropped. Intended for "keep recording
+	///   when idle", but it disables demand-driven teardown for this track.
+	/// - Groups are drained in arrival order and [`cache::Group::read`] resolves only once a group
+	///   finishes, so a stalled group head-of-line-blocks the caching of later finished ones.
+	pub fn with_cache(self, mut cache: cache::Producer) -> Self {
+		let mut subscriber = self.subscribe(None);
+		web_async::spawn(async move {
```
kixelated: Why?

Why not just insert into the cache when inserting a group???

Also we should use the cache for dynamic requests too. It should be an Option on State or something.

rs/moq-net/src/model/track.rs
```diff
@@ -972,6 +1002,19 @@ impl TrackConsumer {
 		&self.name
 	}
 
+	/// Attach a read-through cache: `get_group` / `fetch_group` resolve locally on a cache hit.
+	/// Share the [`cache::Producer::consume`] handle of the cache a [`TrackProducer`] fills to
+	/// serve a track's recent groups without a wire fetch.
+	pub fn with_cache(mut self, cache: cache::Consumer) -> Self {
```
kixelated: Do we even need this? Could it be automatic if the producer is configured with one?

rs/moq-net/src/model/track.rs
```diff
@@ -1167,12 +1240,20 @@ impl GroupRequest {
 pub struct TrackFetch {
 	state: kio::Consumer<TrackState>,
 	sequence: u64,
+	/// A group already rebuilt from the read-through cache. When set, the fetch resolves from it
+	/// instead of polling the live state.
+	cached: Option<GroupConsumer>,
```
kixelated: I think we put the Cache in `TrackState` instead. And we check it on the first poll instead.

Also, isn't the remote cache async? How do we know if this is cached already?

rs/moq-net/Cargo.toml
```diff
@@ -13,6 +13,9 @@ keywords = ["quic", "http3", "webtransport", "media", "live"]
 categories = ["multimedia", "network-programming", "web-programming"]
 
 [features]
+# Disk and remote cache tiers (object_store). Off by default so RAM-only and
+# wasm builds stay dependency-free.
+cache-tiered = ["dep:object_store"]
```
kixelated: IDK maybe call it `fs` instead. It would be nice to have default for non-WASM.

kixelated:
These cohere into a real redesign of how the cache attaches to a track, and I think it's the right one. Synthesis + answers, grouped (covering all nine comments):

## The big one: cache on `TrackState`, insert-on-produce, serve-on-poll (track.rs 709, 1008, 1245)

Agreed, and it's strictly better than what's there — it deletes the spawned-subscriber populate, which is exactly what caused the two issues I flagged in review (the populate subscription pinned `unused()` so a cached track never tore down, and it head-of-line-blocked on each group's finish). Both vanish. Shape:
- `TrackState` gets `cache: Option<cache::Producer>` — the cache the track fills.
- `create_group` / `append_group` insert the finished group into the cache directly, no spawn, no second subscription. [709 — "why not just insert into the cache when inserting a group"]
- `poll_fetch` consults the cache on a miss, so `fetch_group` *and* `TrackDynamic` requests serve from it. [1245, and "use the cache for dynamic requests too"]
- The consumer-side `TrackConsumer::with_cache` / cached `get_group` bolt-on goes away; it's all driven from `TrackState`. [1008 — "could it be automatic" → yes]

**The async wrinkle you flagged [1245]:** we can't synchronously know whether a group is in the disk/remote tier — that read is async. So it can't be a sync check in `poll_fetch`. Model: on the first poll of a fetch (RAM miss), kick off an async cache lookup; the fetch stays `Pending` until it resolves; on hit the group is produced into the track (same path as `GroupRequest::accept`, so the cache is effectively a built-in dynamic handler); on miss it falls through to the upstream wire FETCH / `NotFound`. Resolution chain: RAM (sync) → disk/remote (async) → dynamic/upstream. "Check on first poll" = "start the async lookup on first poll", which is clean once the cache lives on `TrackState`.

## Timestamps: wall-clock vs media (mod.rs 195, 256, 220)

You're right that media `timestamp` is optional (lite-05 only) and a producer can lie about it, so eviction shouldn't depend on it. Proposed model:
- Record a **wall-clock arrival time** when a group is cached and key **retention/eviction on that** — trustworthy, always present. [256] This also simplifies today's `span()` (drop the media-ts duration logic).
- Keep per-frame media `Option<Timestamp>` for lossless replay only (the consumer needs it), not for eviction.
- **timescale on the group, not a `produce()` arg** [220]: yes — once the cache is on `TrackState`, the track knows its timescale (`TrackInfo.timescale`), so `produce()` reads it from there and the arg disappears. Renaming the cached `Group` → `GroupInfo` (sequence + timescale + frames) is reasonable.
- Two genuine calls for you: **(a)** wall-clock as a separate `received: Instant` field vs folding `timestamp` into an enum `{ Media(Timestamp), Received(Instant) }` [195]; **(b)** whether wall-clock and media are separate bounds (trust wall-clock for eviction, optionally bound media for seeking) [256]. I lean: a separate `received` field + a single wall-clock retention bound, media kept only for replay.

## Smaller ones

- **store.rs 33 — disk optional + generic:** I'll make `disk` optional (flush straight to remote). On `<T: ObjectStore>` vs `Arc<dyn>`: disk and remote are usually *different* backends (local FS + S3), which a single `T` can't express — so `dyn` is the right call for mixed tiers. Keeping `dyn`.
- **Cargo.toml 18 — `fs` feature:** the target-gate already gives "on by default for non-wasm, off for wasm, no flag" — i.e. the "default for non-WASM" you want, without a feature that could be wrongly enabled on wasm. A named `fs` feature only works if it's *not* in `default` (else wasm breaks) or if `moq-wasm` sets `default-features = false`. Prefer the target-gate (no footgun), or want an explicit `fs` feature for opt-out on native?
- **mod.rs 207 — poll variant for `Group::read`:** noted, but with insert-on-produce we capture frames as they're written rather than re-reading a finished group, so this `read` loop likely disappears. Folding in.

**Two decisions I need before the rewrite:** the timestamp shape (separate `received` field vs enum; separate bounds or not), and the feature question (target-gate vs `fs`). With those, I'll land the `TrackState`-integrated version as one change (deleting the spawn + the lifecycle/HOL issues). Proceed on my leanings, or hold for your calls?

(Written by Claude)

---

rs/moq-net/src/model/track.rs
```diff
@@ -1020,21 +1095,49 @@ impl TrackConsumer {
...
+			other => {
+				#[cfg(not(target_arch = "wasm32"))]
+				{
+					// A group past the final sequence can never exist in any tier, and an aborted
+					// track is terminal, so skip the cache and report those synchronously below.
+					let exhausted = state.abort.is_some() || state.final_sequence.is_some_and(|fin| sequence >= fin);
+					if !exhausted && let Some(tiers) = &state.cache {
+						let timescale = state.info.as_ref().and_then(|info| info.timescale);
+						let lookup =
+							spawn_cache_lookup(tiers, self.state.clone(), sequence, options.priority, timescale);
```
kixelated: Not a fan of async tasks, and we're so close to getting rid of all of them in model. Going you instead make something we poll to completion?

rs/moq-net/src/model/track.rs
```diff
@@ -1020,21 +1095,49 @@ impl TrackConsumer {
...
-			Poll::Pending => state.fetches.push_back(GroupRequested {
-				sequence,
-				priority: options.priority,
-			}),
+			// Live miss. If a durable cache is attached and the group could still exist, spawn an
+			// async lookup across its disk/remote tiers. The returned `TrackFetch` resolves from it
+			// on a hit; on a miss the lookup task chains upstream (queues for a `TrackDynamic`) and
+			// the fetch falls through to that live decision.
+			other => {
+				#[cfg(not(target_arch = "wasm32"))]
```
kixelated: Way too much wasm gating. We should do the gating right before the fetch and return a 404 instead if WASM.

rs/moq-net/src/model/cache/mod.rs
```diff
@@ -0,0 +1,468 @@
...
+/// with [`Disk::new`], optionally [`with_remote`](Disk::with_remote), then attach via
+/// [`crate::TrackProducer::with_cache`].
+#[cfg(not(target_arch = "wasm32"))]
+#[derive(Clone, Debug)]
+#[non_exhaustive]
+pub struct Disk {
+	/// The object store for the disk tier (e.g. a `LocalFileSystem`).
+	pub store: Arc<dyn ObjectStore>,
+	/// Key prefix under which segments are written.
+	pub prefix: Path,
```
kixelated, line 102: Don't we need another prefix for remote?

rs/moq-net/src/model/cache/mod.rs
```diff
@@ -0,0 +1,468 @@
...
+	/// Retention bounds on the disk tier; exceeding them rolls up to `remote` (or evicts).
+	pub bounds: Bounds,
+	/// Optional remote store the disk tier rolls up into when over its bounds.
+	pub remote: Option<Arc<dyn ObjectStore>>,
```
kixelated: I would put this in a separate pub struct Remote. It's just weird having remote inside of Disk.

And maybe <T: ObjectStore> , IDK.

rs/moq-net/src/model/cache/store.rs
```diff
@@ -0,0 +1,267 @@
...
+
+	/// Phase 1 of compaction, run **under the store lock**: if the disk tier is over bounds, snapshot
+	/// a rollup of the oldest segments (read their bytes and build the rolled object), reserving the
+	/// remote segment id. With no remote tier, drop the oldest disk segments inline instead. Returns
+	/// the snapshot to [upload](Rollup::upload), or `None` when within bounds or evicted inline.
+	///
+	/// Disk reads happen here (local, fast); only the remote upload is slow, and it runs after this
+	/// returns so the lock can be released across it. The disk segments stay in place and indexed
+	/// until [`Self::apply_compaction`], so a concurrent fetch still reads them.
+	pub async fn plan_compaction(&mut self) -> Result<Option<Rollup>, Error> {
```
kixelated: I really really really don't like holding a lock while doing something async. Could we put reading/deleting to the async `upload` task instead? This plan is just to identify which IDs to promote.

kixelated: And we should keep an index of stuff on disk, only consulting the disk when it comes time to fetch or upload.

Issue #1837, opened 2026-06-20
moq-video: remaining work (codecs, platforms, decode, HW validation)


kixelated:
Tracking issue for the gaps in `moq-video` after the ffmpeg removal + native-codec rework on `dev`. Audit done against `rs/moq-video` on the `dev` branch.

Current state:
- **Encode**: H.264 via openh264 (software, all platforms), VideoToolbox (macOS), Media Foundation (Windows), NVENC + VAAPI (Linux, feature-gated; NVENC HW-validated on an RTX 3070 Ti, VAAPI still unverified). H.265 via VideoToolbox (macOS), Media Foundation (Windows), and NVENC (Linux, feature-gated, HW-validated).
- **Decode**: H.264 via VideoToolbox (macOS), Media Foundation/DXVA (Windows), openh264 (software). H.265 via Media Foundation/DXVA (Windows) and VideoToolbox (macOS, #1859); no software fallback. H.264/H.265/AV1 via NVDEC (Linux, #2145 / #2178), with zero-copy NVDEC to NVENC transcode and GPU resize fanout (#2158, `moq transcode`). In `[Unreleased]`.
- **Capture**: camera on all three platforms (AVFoundation / V4L2 / Media Foundation); screen capture on macOS (ScreenCaptureKit) and Windows (DXGI Desktop Duplication).

## Building on top of moq-video (renderer, mobile, universal zero-copy, public frame API)

Motivation, from an audit of downstream consumers: n0's `iroh-live` (Rust, `wgpu` renderer, iroh P2P) and Software Mansion's `moq-kit` (native Swift/Kotlin mobile SDK on `moq-ffi`) both reimplement the raw-frame media layer `moq-video` already owns. They have to, because `moq-video` today exposes zero-copy *ingest* (capture -> encode on macOS/Windows) but no zero-copy *egress*: a decoded `Frame` can only leave as CPU I420 (`into_i420()`) or be fed back into our own encoder (the transcode path). There is no way to get the GPU handle out, and no renderer. So `decode -> render` forces a CPU download + re-upload (a double copy), which is the exact inverse of the ingest path.

Closing the items below lets a consumer build a renderer or an alternate encoder/filter *on top of* `moq-video` instead of forking the whole native layer. It also gives us end-to-end zero-copy in one crate (we currently have only the front half; `iroh-live` has only the back half).

### Frame surface API (get the GPU handle out)
[...]

PR #1844, opened 2026-06-21
fix(moq-video): probe NVIDIA driver libs before NVENC init (avoid abort on GPU-less box)

kixelated:
## What

Reviewing the just-merged #1840 surfaced a pre-existing robustness bug that my NVENC H.265 change widened: **`Kind::Auto` aborts the process on a box without the NVIDIA driver instead of falling back to software.**

### Root cause (verified empirically on a GPU-less Linux box)
- cudarc (`fallback-dynamic-loading`) and `nvidia-video-codec-sdk` (`dynamic-loading`) resolve their entry points via `dlopen` and **`panic!` when the library is missing**, rather than returning an error. cudarc's `culib()` calls `panic_no_lib_found(...)`.
- The workspace builds with **`panic = "abort"`** (both `[profile.dev]` and `[profile.release]`).
- So `Nvenc::open` → `CudaContext::new(0)` panics → **process aborts**, before `backend::open` can fall through to openh264. The `.map_err(...)?` never runs.

This defeats the crate's central packaging goal ("one portable binary reaches the GPU at runtime, falls back to software where the driver is absent"). It's reachable via the default `publish_capture` path (`Options::kind` defaults to `Auto`). #1840 made it newly reachable for H.265, since NVENC now advertises that codec.

Proof: a temporary `Auto`+H264 test aborted with a cudarc panic backtrace at `CudaContext::new`; this box has no `libcuda` in `ldconfig`.

### Fix
`Nvenc::open` now `dlopen`-probes `libcuda` / `libnvidia-encode` up front (via `libloading`, already in the tree through cudarc) and returns a clean `Err` if either is absent, so the fallback chain proceeds to openh264. A driver-present-but-GPU-absent box still fails through the normal `CUresult` path, which was already handled.

Probing presence (not `catch_unwind`) is the right tool here precisely because `panic = "abort"` makes unwinding-based recovery impossible.

### Test
`auto_h264_falls_back_without_driver` (Linux + `nvenc` feature) asserts `backend::open` with `Auto` returns a backend rather than aborting. It holds on a GPU box (NVENC opens) and a driverless box (openh264 fallback) alike, guarding the panic regression. Full `cargo test -p moq-video --features nvenc`: 14 pass.

## Related (not fixed here)
The **VAAPI** backend almost certainly has the identical latent bug: cros-libva `dlopen`s `libva` and likely panics on a miss, so `Auto`+`--features vaapi` on a libva-less box would also abort. I couldn't build/verify it in this environment (no libva headers for cros-libva's bindgen), so I left it out rather than fix blind. Tracked as a follow-up in #1837. (Written by Claude)

---
_Generated by [Claude Code](https://claude.ai/code/session_01KVFm4YtH5u71sZzW6uaZC5)_

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4401,6 +4401,7 @@ dependencies = [
  "cudarc",
  "dispatch2",
  "hang",
+ "libloading 0.9.0",
  "moq-mux",
  "moq-net",
  "moq-vaapi",
--- a/rs/moq-video/Cargo.toml
+++ b/rs/moq-video/Cargo.toml
@@ -26,7 +26,7 @@ doctest = false
 # links on a GPU-less builder and still starts on machines without the NVIDIA
 # driver, falling back to software (see backend::open). `cuda-12020` pins the CUDA
 # API version so the build needs no CUDA toolkit.
-nvenc = ["dep:nvidia-video-codec-sdk", "nvidia-video-codec-sdk/dynamic-loading", "cudarc/fallback-dynamic-loading", "cudarc/cuda-12020"]
+nvenc = ["dep:nvidia-video-codec-sdk", "nvidia-video-codec-sdk/dynamic-loading", "cudarc/fallback-dynamic-loading", "cudarc/cuda-12020", "dep:libloading"]
 # Intel/AMD VAAPI hardware encoder (Linux). Off by default; enable for a build
 # that should use VAAPI where present. The `moq-vaapi` crate (moq-dev/vaapi) builds
 # against its own vendored libva headers and dlopen's libva at runtime, so a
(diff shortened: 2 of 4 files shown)


kixelated:
Closing as superseded by #1819 ("make hardware encoders always-on"), which landed on `dev` after this PR opened and independently includes the same fix: a `driver_libs_present()` probe of `libcuda` / `libnvidia-encode` before NVENC init (same panic-under-`panic = "abort"` rationale) plus a `missing_driver_errors_instead_of_panicking` regression test. The driver-probe is already on `dev`, so there's nothing left here to merge.

Note: #1819 made the Linux hardware encoders always-on (dropped the `nvenc`/`vaapi` feature gates), so the VAAPI equivalent of this bug is now reachable on every libva-less box too. Tracked in #1837. (Written by Claude)

---

PR #1856, opened 2026-06-21
refactor(moq-relay): composable web router (routes() + serve())

kixelated:
## What

Split `Web::run()` into two reusable pieces so an embedder can compose the relay's web router instead of being stuck with the fixed, privately-built one:

- **`Web::routes(state, &config) -> Router`** — the default routes (`/health`, `/announced`, `/fetch`, the WebSocket polyfill catch-all) with state + CORS applied, returned state-erased so a caller can `.merge()`/`.nest()` extra routes onto it.
- **`Web::serve(config, router)`** — owns the listener machinery (HTTP/HTTPS bind, optional mTLS client-cert extraction, hot cert reload) and the landing-page fallback, serving whatever router it's handed.
- **`Web::run()`** stays as a thin convenience wrapper: `serve(config, routes(state, &config))`. The standalone binary is unchanged.
- New **`WebConfig.certificate_endpoint`** (default `true`) optionally drops `/certificate.sha256` (only meaningful for self-signed certs, which a client pins as a WebTransport `serverCertificateHash`).

## Why

I'm building WHIP ingest into an embedded relay (`moq-edge`). WHIP signalling is a plain HTTPS `POST`, so it ideally rides the relay's existing `:443` web listener rather than a second TLS port. Today the router is built privately inside `run()`, and the WebSocket catch-all + TLS/mTLS/reload machinery aren't reachable from an embedder. This change exposes both halves so the embedder can do `routes(...).merge(my_whip_routes)` then `serve(...)`.

## Notes for reviewers

- Pure refactor; no behavior change for the standalone binary (`run()` produces the same router + listeners as before).
- CORS moved from a single global layer to being applied inside `routes()` (scoped to the default GET routes), so an embedder's merged routes bring their own CORS (e.g. a WHIP `POST` endpoint) without the relay's GET-only policy interfering. The landing-page fallback is applied in `serve()` so it covers the merged router too.
- The `/{*path}` WebSocket catch-all coexists with more-specific embedder routes (e.g. `/whip/{*path}`) via matchit's static-over-catchall precedence, same as the existing `/fetch/{*path}`.

### DIFF
--- a/rs/moq-relay/src/web.rs
+++ b/rs/moq-relay/src/web.rs
@@ -99,31 +99,55 @@ pub struct HttpsConfig {
 	pub root: Vec<PathBuf>,
 }
 
-/// Shared state passed to all web handler routes.
-pub struct WebState {
+/// Shared state passed to all web handler routes. An internal detail: callers
+/// build a [`Web`] from its parts via [`Web::new`] rather than constructing this.
+pub(crate) struct WebState {
 	/// The authenticator for verifying incoming requests.
-	pub auth: Auth,
+	pub(crate) auth: Auth,
 	/// The cluster state for resolving origins.
-	pub cluster: Cluster,
+	pub(crate) cluster: Cluster,
 	/// TLS certificate information served at `/certificate.sha256`.
-	pub tls_info: Arc<std::sync::RwLock<moq_native::tls::Info>>,
+	pub(crate) tls_info: Arc<std::sync::RwLock<moq_native::tls::Info>>,
 	/// Monotonically increasing connection counter for WebSocket sessions.
-	pub conn_id: AtomicU64,
+	#[cfg_attr(not(feature = "websocket"), allow(dead_code))]
+	pub(crate) conn_id: AtomicU64,
 }
 
 /// Run a HTTP server using Axum
 pub struct Web {
-	state: WebState,
+	state: Arc<WebState>,
 	config: WebConfig,
 }
 
 impl Web {
-	pub fn new(state: WebState, config: WebConfig) -> Self {
+	/// Build a web server from its parts. `tls_info` is the relay's TLS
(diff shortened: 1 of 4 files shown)


rs/moq-relay/src/web.rs
```diff
@@ -49,6 +49,21 @@ pub struct WebConfig {
...
+
+	/// If true (default), expose `/certificate.sha256` (the TLS cert
+	/// fingerprint clients pin as a WebTransport `serverCertificateHash`).
+	///
+	/// Only meaningful for self-signed certs. An embedder serving a
+	/// publicly-trusted cert can drop the endpoint with `certificate_endpoint =
+	/// false` (`--web-certificate-endpoint=false`) since browsers validate the
+	/// chain normally and never read the fingerprint.
+	#[arg(
+		long = "web-certificate-endpoint",
```
kixelated: rename to `web-fingerprint`

rs/moq-relay/src/web.rs
```diff
@@ -122,29 +137,51 @@ impl Web {
...
+	///
+	/// Includes the WebSocket polyfill catch-all (`/{*path}`, when
+	/// `config.ws`) and CORS scoped to its own GET routes, but NOT the
+	/// landing-page fallback (that is global, so [`serve`] sets it once across
+	/// the merged router). Extra routes a caller merges in keep their own
+	/// layers and bring their own CORS as needed (e.g. a WHIP POST endpoint).
+	///
+	/// `/certificate.sha256` is included unless `config.certificate_endpoint`
+	/// is false.
+	pub fn routes(state: Arc<WebState>, config: &WebConfig) -> Router {
```
kixelated: `Arc<WebState>` is gross. I hope it's not actually public?

PR #1862, opened 2026-06-21
feat(moq-hls): pause/resume export with discontinuity markers

kixelated:
## What

Adds a pause/resume capability to the HLS `export::Broadcaster`, plus a segment-level discontinuity marker.

- **`Broadcaster::set_paused(bool)` / `is_paused()`** — a shared `watch<bool>` that gates every rendition pump. While paused, a pump stops calling `next_fragment()` entirely, so the live media produced during the pause is dropped by the relay (the exporter's latency budget skips stale groups on resume), **not buffered** — a long pause costs no memory. Renditions discovered while paused inherit the current state.
- **Continuity across a pause** — segment sequence numbering and the init segment live on the `SegmentStore`, so resuming continues the *same* playlist rather than restarting. A paused pump still notices the broadcast closing and finalizes the track (`store.finish()`) promptly.
- **Discontinuity** — `SegmentStore::mark_discontinuity()` forces a fresh segment on resume and tags it; the new `SegmentMeta.discontinuity` flag is rendered as `#EXT-X-DISCONTINUITY` by `render_media`.

## Why

A downstream recorder (moq-pro's in-process VOD module) wants to pause a recording when its rule is disabled and resume the **same** session when re-enabled — with a discontinuity at the seam — instead of tearing down and starting a new recording. The pause/resume + discontinuity primitives have to live here, in the export pump, since that's what pulls groups.

## Notes for reviewers

- The pump deliberately does **not** cancel `next_fragment()` mid-flight (it isn't cancel-safe); entering a pause costs at most one extra fragment (~`part_target`), which records right up to the pause point.
- Based on `09ec6350c` (the moq-rtmp gateway head that moq-pro pins, on `dev`); `dev` hasn't touched the export module since, so this is a clean 4-file diff against `dev`.
- Tests: `cargo test -p moq-hls` (added `discontinuity_precedes_resumed_segment`).


rs/moq-hls/src/export/rendition.rs
```diff
@@ -105,12 +126,51 @@ async fn run_pump(
...
+			resumed = true;
+			tokio::select! {
+				// Resume request, or the Broadcaster (and its sender) being dropped.
+				res = paused.changed() => {
+					if res.is_err() {
+						return Ok(()); // Broadcaster gone: stop pumping.
+					}
+				}
+				// The broadcast ending while paused still finalizes the track.
+				_ = kio::wait(|w| closed.poll_closed(w)) => return Ok(()),
```
kixelated, line 156: isn't there a closed().await helper?

rs/moq-hls/src/export/mod.rs
```diff
@@ -54,20 +54,43 @@ pub struct Broadcaster {
 	/// Current rendition count, bumped on every catalog sync so handlers can wait
 	/// for the catalog to populate before rendering a playlist.
 	ready: watch::Sender<usize>,
+	/// Pause flag shared with every rendition pump. While true, the pumps stop
+	/// draining the broadcast; renditions discovered later inherit the current
+	/// value (they `subscribe()` to this sender).
+	paused: watch::Sender<bool>,
```
kixelated, line 60: This is a code smell. Can you spike using kio::Producer/kio::Consumer instead of tokio primatives?

PR #1870, opened 2026-06-22
moq-gst: replace moqsink with a GstAggregator-based core

kixelated:
## Summary

Rebuilds the `moqsink` element on **`GstAggregator`** instead of a bare `gst::Element` with a hand-rolled queue. This is an alternative core to the bounded-channel design in #1771: it keeps that PR's media foundation and replaces the transport seam.

It builds directly on @arielmol's commits from #1771 (the rewritten timeline, codec/caps handling, B-frame support, frame-size cap, EOS aggregation, and tests), so this PR is diffed against `main` rather than stacked on that branch. The final commit swaps the session core for Aggregator; the 14 commits beneath it are ArielM's and carry his authorship.

## Why

The #1771 core put a **bounded** mpsc channel between the GStreamer streaming threads and the async session, for backpressure. A blocking send on that channel cannot be interrupted by `FLUSH_START` (which arrives out of band on another thread), so it would wedge a seek/flush/state-change. The fix there was a flush-cancellation apparatus: per-pad `watch<bool>` gates, an out-of-band `FlushSignal`, biased cancellable sends (`send_or_flush`), and per-pad generations to discard stale in-flight messages.

`GstAggregator` already owns per-pad input queues, FLUSH/EOS/SEGMENT handling, and a single aggregate thread, so none of that is needed. Funneling serialized events and buffers onto one thread also removes the generation bookkeeping and the cross-thread failure map.

A second simplification falls out: `GroupProducer::write_frame` is **synchronous** (it appends and evicts in memory, never blocking on the network). So the aggregate thread writes frames straight into the moq producers under a `RUNTIME.enter()` guard. There is **no data channel at all** — the async task only connects, holds the session, and reports status. Because the producers are created up front, frames buffered before the connection completes are sent once it does (a "future work" item in #1771, free here).

## What changed

- **`sink/imp.rs`** — now a `GstAggregator` subclass. `aggregate()` drains each pad with `pop_buffer()` and writes; `sink_event` rejects unsupported caps; `flush()` re-anchors timelines; `start`/`stop` own the session and producers.
- **`sink/session.rs`** — slimmed from ~2035 to ~240 lines: just `Status` and a connect/lifecycle task.
- **`sink/pad.rs`** (new) — per-pad media state (caps -> producer, SEGMENT policy, frame import) extracted from the old `Pad`/`PadSet`, plus 19 hermetic unit tests.
- **`sink/timeline.rs`** — unchanged.
- **`sink/mod.rs`** — `@extends gst_base::Aggregator`.
- **`Cargo.toml`** — adds `gstreamer-base` (no version-feature bump; stays at the existing 1.14 baseline).

**Deleted** relative to the #1771 core: the bounded channel + `DATA_CHANNEL_BOUND`, `send_or_flush`/`SendOutcome`, `FlushSignal` + flush watches + `toggle_pad_flush`/`reset_pad_flushes`, per-pad generations, the `Status` failure map, `PadSet`, the manual chain/event pad functions, and the request-pad rollback dance.

## Public surface / branch targeting

No public library API or wire change. Same element name (`moqsink`), `sink_%u` request pads, the seven codec caps, the three writable properties, and the three read-only diagnostics (`connected`, `moq-version`, `estimated-send-bitrate`). Internal to the `rs/moq-gst` plugin, so this targets **`main`** (same as #1771). One cosmetic change: the element now carries an unused `src` pad (it is SINK-flagged and never pushes).

## Testing

- `cargo test -p moq-gst`: 31 unit + 3 element tests pass, including `frame_through_h264_emits_a_frame` (a real IDR access unit flows caps -> producer -> published track, hermetically).
- `cargo clippy -p moq-gst --all-targets` clean; `cargo fmt --check` clean.
- `gst-inspect-1.0 moqsink`: registers `Sink/Network/MoQ`, hierarchy `+----GstAggregator`, both pad templates `GstAggregatorPad`, all properties.
- **Live relay**: `videotestsrc ! vtenc_h264 ! h264parse ! moqsink` to a local `moq-relay` logged `session connected` -> `finalized on EOS: ["sink_0", "catalog"]` -> `posting EOS` -> clean `session terminated`.

## Not yet validated (same boundary #1771 drew)

- A subscribe-back byte round-trip against a relay.
- A genuinely stalled sparse pad. Aggregator fires `aggregate()` when every pad has a buffer; for continuous A/V that is fine, but a stalled non-EOS pad could hold up others. If sparse data tracks matter, `set_force_live(true)` is the knob (raises the floor to GStreamer 1.22).
- FLUSH/seek on a live pipeline.

(Written by Claude)


rs/moq-gst/src/sink/imp.rs
```diff
@@ -1,37 +1,29 @@
...
-static CAT: LazyLock<gst::DebugCategory> = LazyLock::new(|| {
-	gst::DebugCategory::new(
-		"moq-sink",
-		gst::DebugColorFlags::empty(),
-		Some("MoQ Sink (async element)"),
-	)
-});
+/// Reject a frame past the MoQ frame limit (moq-net's MAX_FRAME_SIZE, 16 MiB): it could not be
+/// consumed anyway, and copying it would let hostile input drive an unbounded allocation.
+const MAX_FRAME_BYTES: usize = 16 * 1024 * 1024;
```
kixelated: We should have moq-net export this constant instead.

rs/moq-gst/src/sink/imp.rs
```diff
@@ -63,72 +48,27 @@ impl TryFrom<Settings> for ResolvedSettings {
...
-
 #[derive(Default)]
 pub struct MoqSink {
 	settings: Mutex<Settings>,
-	session: Mutex<Option<SessionHandle>>,
+	/// The live session (connect task + status). Replaced on every start, so the property getters and
+	/// the aggregate thread always read the current one.
+	session: Mutex<Option<Session>>,
+	/// Producers written from the aggregate thread. Created in `start`, dropped in `stop`.
+	broadcast: Mutex<Option<moq_net::BroadcastProducer>>,
```
kixelated: Make one Mutex for this state?

Also... why a `Mutex` without `Arc`? I don't follow what the mutex is doing?

rs/moq-gst/src/sink/imp.rs
```diff
@@ -183,388 +153,248 @@ impl GstObjectImpl for MoqSink {}
 
 impl ElementImpl for MoqSink {
 	fn metadata() -> Option<&'static gst::subclass::ElementMetadata> {
-		static ELEMENT_METADATA: LazyLock<gst::subclass::ElementMetadata> = LazyLock::new(|| {
```
kixelated, line 186: lul add us back

rs/moq-gst/src/sink/imp.rs
```diff
@@ -183,388 +153,248 @@ impl GstObjectImpl for MoqSink {}
...
+	/// current caps. Every failure path is per-pad (drop the buffer); none tears down the session.
+	fn write_buffer(
+		&self,
+		agg_pad: &gst_base::AggregatorPad,
+		broadcast: Option<&moq_net::BroadcastProducer>,
+		catalog: Option<&moq_mux::catalog::Producer>,
+		buffer: gst::Buffer,
+	) {
+		// Bound the per-frame allocation before copying.
+		if buffer.size() > MAX_FRAME_BYTES {
```
kixelated: honestly, should we just let moq-net fatally error if we ever do this instead? Why silently warn on something obviously wrong?

rs/moq-gst/src/sink/imp.rs
```diff
@@ -183,388 +153,248 @@ impl GstObjectImpl for MoqSink {}
...
+	/// finalized pad returns `Ok(false)` and the catalog is taken on the first call.
+	fn finalize_all(&self) -> Result<Vec<String>> {
+		let mut order = Vec::new();
+		{
+			let mut pads = self.pads.lock().unwrap();
+			for (name, pad) in pads.iter_mut() {
+				match pad.finalize() {
+					Ok(true) => order.push(name.clone()),
+					Ok(false) => {}
+					Err(err) => gst::warning!(CAT, "finalize {name}: {err:?}"),
```
kixelated: why warn instead of returning err?

rs/moq-gst/src/sink/imp.rs
```diff
@@ -183,388 +153,248 @@ impl GstObjectImpl for MoqSink {}
...
-				}
-			}
-			ControlMessage::DropPad { pad_name } => {
-				runtime.pads.remove(&pad_name);
-				eos_pads.remove(&pad_name);
-			}
-			ControlMessage::Eos { pad_name } => {
-				eos_pads.insert(pad_name);
+		let Ok(map) = buffer.map_readable() else {
+			gst::warning!(CAT, "failed to map buffer on pad {}", agg_pad.name());
```
kixelated: why warn? Can we return an error?

rs/moq-gst/src/sink/pad.rs
```diff
@@ -0,0 +1,568 @@
...
+	/// the track stays finalized. Isolated to the pad, so the session and other pads keep going.
+	failed: bool,
+	state: PadState,
+	segment_info: Option<SegmentInfo>,
+	/// Kept only to map a buffer PTS to a running time.
+	segment: Option<gst::FormattedSegment<gst::ClockTime>>,
+}
+
+impl Pad {
+	pub(super) fn new() -> Self {
```
kixelated: wedon't need pub(super) on every method if `Pad` itself is `pub(super)`

rs/moq-gst/src/sink/pad.rs
```diff
@@ -0,0 +1,568 @@
...
+pub(super) fn caps_supported(caps: &gst::CapsRef) -> bool {
+	let Some(s) = caps.structure(0) else { return false };
+	matches!(
+		s.name().as_str(),
+		"video/x-h264" | "video/x-h265" | "video/x-av1" | "video/x-vp8" | "video/x-vp9" | "audio/mpeg" | "audio/x-opus"
+	)
+}
+
+/// FramedFormat::Avc3/Hev1 consume Annex-B access units, so H.264/H.265 caps must be byte-stream/au.
+fn require_byte_stream_au(s: &gst::StructureRef) -> Result<()> {
```
kixelated: Can't caps enforce this?

rs/moq-gst/src/sink/timeline.rs
```diff
@@ -0,0 +1,138 @@
...
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum SegmentDecision {
+	Accept,
+	Reject(&'static str),
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum FrameDecision {
```
kixelated: Do we really need this enum? Can we use an Option?

rs/moq-gst/src/sink/timeline.rs
```diff
@@ -0,0 +1,138 @@
...
+pub struct SegmentInfo {
+	/// Only TIME segments map to a media timeline (not BYTES/DEFAULT).
+	pub time_format: bool,
+	pub rate: f64,
+	/// Running-time anchor of the segment; continuity is judged on this, not on `start`.
+	pub base_nanos: u64,
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum SegmentDecision {
```
kixelated: Do we really need this enum? Can we use a Result... or boolean?

arielmol:
> A genuinely stalled sparse pad. Aggregator fires aggregate() when every pad has a buffer... set_force_live(true) is the knob (raises the floor to GStreamer 1.22)

I'd gently unbundle from "same boundary." it isn't an unvalidated edge in #1771  it's structurally absent. the Fanout in Element has no aggregate() barrier, each pad's chain runs independently to its own track, so **a stalled non-EOS pad never holds up the others**. there's nothing to force_live around (no need to upgrade gst to 1.22+).

> subscribe-back byte round-trip against a relay  & FLUSH/seek on a live pipeline

Indeed on both, what I do have is live end2end from about 15+ of the fleet (x86/aarch64) so transport is exercised end to end. I just haven't asserted byte equality, but doable, ... will do! 😄

arielmol:
Update on the byte round-trip for #1771 done. I ran a VP8 payload round-trip: videotestsrc(ball) -> vp8enc -> moqsink -> relay -> moqsrc, capturing encoded buffers before moqsink and after moqsrc. The harness writes one file per GStreamer buffer and checks received hashes against a contiguous slice of the published hashes.

150 buffers published, all 150 distinct; 149 received; every received buffer matched the published VP8 payload byte-for-byte and in order. In this run the missing buffer was the tail not draining before teardown, not a payload mismatch.

rs/moq-gst/src/sink/imp.rs
kixelated: Done. moq-net now enforces the limit in `GroupProducer::create_frame` (before `produce()` allocates), so the sink dropped its own constant and check and relies on moq-net returning `FrameTooLarge`. Making the constructor itself fallible (to also cover direct callers) is queued as a separate `dev` follow-up.

rs/moq-gst/src/sink/imp.rs
kixelated: Done. Consolidated `session`/`broadcast`/`catalog`/`pads`/`eos` into a single `Mutex<Option<State>>`. It is a bare `Mutex` (not `Arc<Mutex>`) because glib already owns and shares the subclass instance across GStreamer threads, so only interior mutability is needed. Added a comment saying so.

rs/moq-gst/src/sink/imp.rs
kixelated: Done. Restored Luke and Steve to the element-metadata authors (Ariel kept).

rs/moq-gst/src/sink/imp.rs
kixelated: Done. moq-net returns `FrameTooLarge` and the sink propagates it (failing that pad) instead of silently warning; the sink-side size check and constant are gone.

rs/moq-gst/src/sink/imp.rs
kixelated: Done. `finalize_all` now propagates the first pad/catalog finalize error (still logging each) instead of only warning; on EOS it becomes an element error.

rs/moq-gst/src/sink/imp.rs
kixelated: Done. `write_buffer` now returns `FlowError::Error` on a buffer-map failure instead of warning and continuing.

rs/moq-gst/src/sink/pad.rs
kixelated: Done. Changed the whole `Pad` surface to plain `pub` (the private `mod pad` is the real visibility gate).

rs/moq-gst/src/sink/pad.rs
kixelated: Right. Removed the runtime byte-stream/au and AAC mpegversion/stream-format checks; the pad template pins those fields so negotiation enforces them. Kept the AAC codec_data check (not a template field).

rs/moq-gst/src/sink/timeline.rs
kixelated: Done. Removed `FrameDecision`; `frame_micros` now returns `Result<u64, &'static str>` (the reason is kept for logs).

rs/moq-gst/src/sink/timeline.rs
kixelated: Done. Removed `SegmentDecision`; `classify_segment` now returns `Result<(), &'static str>`.

rs/moq-gst/src/sink/timeline.rs
```diff
@@ -0,0 +1,138 @@
...
+	Accept,
+	Reject(&'static str),
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum FrameDecision {
+	/// Emit at this MoQ timestamp (micros).
+	Emit(u64),
+	Drop(&'static str),
+}
```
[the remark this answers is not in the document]
kixelated: The `SegmentDecision`/`FrameDecision` enums were replaced with `Result`, so `#[non_exhaustive]` no longer applies. `SegmentInfo` now has a doc comment and lives in a private `mod timeline`, so it is crate-internal, not external API.

rs/moq-gst/Cargo.toml
```diff
@@ -20,6 +20,7 @@ doctest = false
 anyhow = { version = "1", features = ["backtrace"] }
 bytes = "1"
 gst = { package = "gstreamer", version = "0.23" }
+gst-base = { package = "gstreamer-base", version = "0.23" }
```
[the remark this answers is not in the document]
kixelated: Leaving this as a direct version on purpose: `gst`, `anyhow`, `bytes`, `url`, and `tracing` are all specified per-crate in this Cargo.toml, so `gst-base` matching its sibling `gst` is the consistent choice rather than workspace-managing only `gstreamer-base`.

PR #1873, opened 2026-06-22
feat(moq-net,moq-srt): meter payload usage in the model; bill non-MoQ gateway traffic

kixelated:
## What

Records ingress/egress usage (`bytes` / `frames` / `groups`) for the non-MoQ gateways (moq-srt, moq-rtc, moq-rtmp), which were blind spots: usage was only counted at the moq-lite / IETF `publisher.rs` / `subscriber.rs` wire boundary, so anything a gateway drove through the model directly went uncounted.

It does this by making the **broadcast** the single place a meter is attached, and having the model propagate it down to every track:

1. **Move per-track payload counting into the transport-agnostic model.**
2. **Attach the meter at the broadcast level** and propagate it through *every* track-creation path (`create_track`, the dynamic handler, fetch, `reserve_track`). One `set_meter` per broadcast bills the whole thing.
3. **Use that one mechanism everywhere** — the lite/IETF relay handlers and the moq-srt gateway both attach at the broadcast, so there's no per-track `set_meter` scattered through the data path.

## 1. Payload counting in the model

- New `model/meter.rs`: a shared `Usage` struct (three atomics) plus a cheap, cloneable `Meter(Option<Arc<Usage>>)` handle. **No trait, no dynamic dispatch** — a bump is a direct `fetch_add`.
- The stats `Counters` holds its payload counters as an `Arc<Usage>` and reads them back when it snapshots.
- The model bumps the meter at the real data-flow points: `create_group`/`append_group` (groups), `append_frame` (frames + bytes), and the egress duals on the consumer side.

## 2. Broadcast-level propagation

The per-track `set_meter` can't reach a gateway's tracks: a gateway hands a broadcast to a `moq-mux` importer/exporter that creates tracks and opens subscriptions *internally* (dozens of `broadcast.track(name).subscribe()` / `broadcast.unique_track(...)` sites). The same is true of the relay, which serves tracks **on demand** via the dynamic handler. So the meter rides the broadcast and flows down:

- `BroadcastProducer::set_meter` -> `create_track` / `unique_track`, **and** the dynamic path `BroadcastDynamic -> TrackRequest -> accept`, **and** the fetch path `TrackDynamic -> GroupRequest -> accept`, **and** `reserve_track` (**ingress**).
- `BroadcastConsumer::set_meter` -> `track()` -> `TrackConsumer` -> `subscribe()` -> `TrackSubscriber` (**egress**).
- `PublisherStats::meter()` / `SubscriberStats::meter()` / `BroadcastStats::subscriber_meter()` expose the sink to attach; the broadcast-lifetime guard keeps the counters alive for the snapshot task.

`TrackProducer::set_meter` is now `pub(crate)` (internal propagation only); `TrackSubscriber::set_meter` and an obsolete `GroupProducer::set_meter` are removed. The single public way to meter is at the broadcast.

## 3. Both the relay and moq-srt attach at the broadcast

- **lite/IETF relay**: the six former per-track `set_meter` calls collapse to one broadcast-level attach each: ingress in `start_announce` / `restart_announce`, egress on the per-subscribe `BroadcastConsumer`. IETF `start_publish` now creates its track through the broadcast. The per-track `SubscriberTrack`/`PublisherTrack` guards stay for the `subscriptions` lifecycle counter only.
- **moq-srt** (the proof for the gateways): `Server::with_stats(StatsHandle)` (mirrors `moq_native`) rides each `Request`; `Publish`/`Subscribe::accept` open the ingress/egress guard and thread its meter to the broadcast seam in `ts.rs`. So an SRT `m=publish` / `m=request` records usage even though `ts::Import` / `ts::Export` own the tracks. The unauthenticated `run()` stays no-op; embedders (e.g. moq-pro) use `with_stats`. moq-rtc/moq-rtmp follow the same pattern.

## Semantics

- **Egress is per-consumer**: N viewers of a track count N times (each `BroadcastConsumer` clone carries its own handle over the same `Usage`). **Ingress is single-writer**: counts once. For the relay, ingress (from upstream) and egress (to downstream) are independent roles, so a middle node counts both directions without double-counting either.
- The lite **fetch** path is now metered through the model (the fetched group's frames/bytes), where before it counted separately. A fetch re-serves an existing group, so it bumps frames/bytes but not `groups`.
- The model counts **raw (decompressed)** bytes where the handlers counted post-compression **wire** bytes. In practice only the hang catalog track sets `compress`, so the difference is noise; a follow-up will cache compressed bytes in the model so the count is the wire size again.

## Testing

- `moq-net`: **405 lib tests** pass, including new ones for ingress-once, egress-per-viewer, broadcast→track propagation, the no-op default, the stats-layer gateway path, and **dynamic-track / fetch / reserve_track meter inheritance**.
- `moq-native`: **62 e2e tests** pass (moq-lite 01-03 + IETF transport 14-18, fetch, compression, negotiation), plus 62 more unit/integration tests across the crate.
- `moq-srt`: builds + 7 lib tests pass.
- clippy + `fmt --check` clean.

> Note: validated with the repo's `rustc 1.96` (the earlier "rustc 1.91" note was a stale toolchain); moq-srt fmt/clippy ran via system cargo after a disk-pressure event broke nix's pty mid-session. CI's nix run is the source of truth.

## Cross-package sync

- `js/net`: not mirrored — the stats aggregator / meter is relay-side Rust only; the browser client has no equivalent. (No `doc/concept` wire change: this is API, not wire.)

## Follow-ups

- Wire moq-rtc (WHIP/WHEP) and moq-rtmp the same way (add `Server::with_stats`, attach the broadcast meter at their create/consume seam). moq-edge lives in moq-pro.
- Optionally thread `sessions` (presence) / `broadcasts` (per-viewer) lifecycle counters for the gateways; this PR wires `announced` + payload.
- Cache compressed bytes in the model so the byte count is the wire size for compressed tracks.

(Written by Claude)


rs/moq-net/src/ietf/publisher.rs
```diff
@@ -156,6 +156,12 @@ impl<S: web_transport_trait::Session> Publisher<S> {
 			}
 		};
 
+		// Attach the publisher-side payload meter so the model counts
+		// groups/frames/bytes as they're served to this subscriber (egress is
+		// per-viewer). `track_stats` (the PublisherTrack guard, subscriptions
+		// lifecycle) stays owned by this function for the subscription's duration.
+		track.set_meter(track_stats.meter());
```
kixelated: It needs to be possible to do this before subscribing.

rs/moq-net/src/ietf/publisher.rs
```diff
@@ -309,7 +299,8 @@ impl<S: web_transport_trait::Session> Publisher<S> {
 		let mut stream = Writer::new(stream, version);
 
 		stream.encode(&msg).await?;
-		track_stats.group();
+		// Groups/frames/bytes are counted by the model via the subscriber's meter
```
kixelated: useless comment

rs/moq-net/src/model/meter.rs
```diff
@@ -0,0 +1,103 @@
...
+/// Which side is counted is set by *where* the handle is attached:
+/// - **Ingress** lives on a [`crate::TrackProducer`] and propagates into each
+///   [`crate::GroupProducer`] it creates. A frame is produced once, so it counts
+///   once.
+/// - **Egress** lives on a [`crate::TrackSubscriber`] and propagates into each
+///   [`crate::GroupConsumer`] it hands out. Each subscriber carries its own
+///   handle to the *same* shared `Usage`, so N subscribers reading the same
+///   cached frame each bump it (matching per-viewer egress).
+#[derive(Clone, Default)]
+pub struct Meter(Option<Arc<Usage>>);
```
kixelated: Why is this Option? Can we just make it always set; it's only a few atomic operations even if nobody is reading it. IDK just merge Meter/Usage

rs/moq-net/src/ietf/publisher.rs
```diff
@@ -142,6 +142,13 @@ impl<S: web_transport_trait::Session> Publisher<S> {
 			}
 		};
 
+		// Attach the publisher-side payload meter to the broadcast consumer so the
+		// model counts groups/frames/bytes served to this subscriber. This handle is
+		// local to this subscription (per-viewer: N viewers count N times) and the
+		// meter rides into the track subscription below. `track_stats` (the
+		// PublisherTrack guard, subscriptions lifecycle) stays owned by this function.
+		broadcast.set_meter(track_stats.meter());
```
kixelated: Is there any way we could call `with_meter()` before the await?

rs/moq-net/src/ietf/subscriber.rs
```diff
@@ -462,12 +464,19 @@ impl<S: web_transport_trait::Session> Subscriber<S> {
 					.expect("an empty hop chain has room for one entry");
 				// moq-transport carries no broadcast epoch on the wire; stamp the current
 				// time so the instance is still ordered against any future re-announce.
-				let broadcast = BroadcastInfo {
+				let mut broadcast = BroadcastInfo {
 					hops,
 					..Default::default()
 				}
 				.produce();
```
kixelated: I think `.with_meter(self)` is better, make it work like a builder.

Should we put the meter in BroadcastInfo maybe??? IDK if that would work, but it's worth thinking about.

rs/moq-net/src/model/track.rs
```diff
@@ -470,6 +473,12 @@ pub struct TrackProducer {
 	name: Arc<str>,
 	state: kio::Producer<TrackState>,
 	prev_subscription: Option<Subscription>,
+
+	// Ingress usage meter, propagated into every group this producer creates.
+	// Bumped once per group here and once per frame/byte in the group; a no-op
+	// until a stats sink is attached via [`Self::set_meter`]. Clones share it
+	// (parallel producers of one track feed the same sink).
+	meter: Meter,
```
kixelated: I think this needs to be a Vec. So if the track is inserted into multiple broadcasts, it updates all of them?

rs/moq-net/src/stats.rs
```diff
@@ -918,24 +974,34 @@ pub struct SubscriberTrack {
 }
 
 impl SubscriberTrack {
+	/// A [`Meter`] over this track's shared ingress payload counters, to attach to
+	/// the model `TrackProducer` (via `set_meter`) so groups/frames/bytes are
+	/// counted as media is produced. No-op for a disabled aggregator.
+	pub fn meter(&self) -> Meter {
+		match &self.entry {
+			Some(entry) => entry.subscriber[self.tier.idx()].meter(),
+			None => Meter::default(),
```
kixelated: looks like an `unwrap_or_default` to mee

rs/moq-net/src/lite/subscriber.rs
```diff
@@ -445,11 +446,17 @@ impl<S: web_transport_trait::Session> Subscriber<S> {
...
+		// lifecycle guard that keeps these counters alive lives in the announce loop's
+		// `stats_guards`, so this is the non-bumping accessor (no double `announced`).
+		let abs = self.origin.absolute(&path).to_owned();
 		let broadcast = BroadcastInfo {
 			hops,
 			epoch: BroadcastInfo::epoch_from_wire(epoch),
 		}
-		.produce();
+		.produce()
+		.with_meter(self.stats.broadcast(&abs).subscriber_meter());
```
kixelated, line 459: is it a problem that we're not using the subscriber handle? This is a little concerning because it might be masking a bug that we're not counting subscribers?

PR #1884, opened 2026-06-23
feat(moq-mux): clear consumer buffer when group timestamps rewind

kixelated:
## Summary

Adds **always-on** rewind detection to the container `Consumer` (`rs/moq-mux`, mirrored in the `js/hang` `Consumer`). When a newer group's timestamps jump strictly backwards past the live edge, the publisher is reneging the buffered tail. The consumer drops the reneged groups, resumes at the rewound timeline, and bumps a **discontinuity counter**. In the browser watch path, that counter auto-flushes the audio and video buffers and re-anchors the playback clock, so the rewind takes effect end-to-end with no app involvement.

Prototype for [#1614](https://github.com/moq-dev/moq/issues/1614): a Python voice agent has no way to drop already-written audio when the user interrupts the bot, so the listener keeps hearing the old utterance for hundreds of ms. Rather than a new producer `flush()`/`cancel()`, this handles it on the watch side: a backwards group timestamp is the reneg signal. The existing `AudioProducer::reset_epoch()` already emits it for a relative timeline (the agent writes `timestamp_us=0`, so the next group re-anchors below the buffered tail while the group sequence keeps climbing).

## How it works

Group **sequence** is monotonic, so the unambiguous tell is *sequence climbs while timestamp drops*. The subtlety is that groups arrive out of order, so a single sequence floor is wrong: a late new-epoch group can have a **lower** sequence than the group that triggered detection.

So a reset records a boundary of three values and classifies any group by `(sequence, timestamp)`:

| sequence | verdict |
|---|---|
| `≤ prev_max` (old live-edge group) | old, drop |
| `≥ reset_group` (detection group) | new, keep |
| in the ambiguous span between them | keep iff `timestamp < reset_timestamp` |

The ambiguous span exists because `prev_max` (the old epoch's max among *delivered* groups) can underestimate the true tail when an old straggler is still in flight. The timestamp clause splits that span: new gap-fillers sit below `reset_timestamp`, old stragglers above it.

The live edge, the active boundary, and the discontinuity counter are bundled into one `Rewind` struct (Rust) / class (JS).

## Downstream flush (js/watch)

Clearing the container buffer is not the whole story: the audio ring buffer and the decoded-video pipeline also hold media that must be dropped. Both decoders now read `discontinuity` off each `consumer.next()` and, when it increments, flush before decoding the first frame of the new utterance:

- **Audio**: flush the ring buffer + `sync.reset()`.
- **Video**: `sync.reset()`, clear the playback `timestamp` (else the output guard late-rejects the rewound lower-timestamp frames), clear the decode buffer, and bump a generation so in-flight decodes from the reneged timeline are dropped on output rather than rendering mid-utterance. The held frame stays until the new keyframe renders (no black flash).

This is the same flush a manual `element.reset()` did, now automatic. Audio-only, A/V, and video-only buffered rewinds are all handled. The Rust native render path (FFI / `AudioConsumer`) is not wired yet.

## Public API

- `rs/moq-mux`: `Consumer::discontinuity() -> u64`. The `Reset`/`Rewind` types are private.
- `js/hang`: `Consumer.discontinuity` getter, plus a `discontinuity` field on the `next()` return (additive).

Always-on, no configuration. It only fires on a strictly-backwards group timestamp, which does not occur in normal (forward) streams, so existing streams are unaffected — confirmed by the pre-existing consumer tests passing unchanged.

## Test plan

- `cargo test -p moq-mux --lib container::consumer` — 35/35 pass, including `reset_classifies_out_of_order_groups` (the `(55,58,90)` example), `reset_keeps_out_of_order_new_group` (out-of-order gap-fillers kept, asserts `discontinuity() == 1`), `backwards_timestamp_resets_buffer`, `backwards_timestamp_always_resets`.
- `clippy -p moq-mux` clean; `rustfmt` clean (system toolchain; nix's tty was erroring locally, so worth a `just fix` through nix before merge).
- `js/hang` + `js/watch`: `tsc` and `biome` clean; the 78 `js/watch` unit tests pass. The watch wiring is compile- and unit-checked, not yet exercised end-to-end in a browser.

## Notes for reviewers

- **Base branch**: `main`. The branch is forked from `main`; the change only fires on a backwards group timestamp (no effect on forward streams). It does establish a consumer-side reneg-via-timestamp convention, so say so if you would rather it settle on `dev`.
- The watch flush lives in each decode loop rather than `MultiBackend`, so it is ordered before the new frame is decoded (a signal round-trip would land a microtask later). Easy to move to `MultiBackend` orchestration if you prefer.
- Keeping a late gap-filler does not guarantee it gets *played*: if it arrives after the playhead passed the detection group, the normal latency/jitter logic drops it like any reorder. The boundary just stops us discarding it as cross-epoch garbage.

(Written by Claude)

### DIFF
--- a/js/hang/src/container/consumer.ts
+++ b/js/hang/src/container/consumer.ts
@@ -11,6 +11,15 @@ export interface ConsumerProps {
 	format: Format;
 	/** Target latency in milliseconds, controlling how aggressively slow groups are skipped (default: 0). */
 	latency?: Signal<Time.Milli> | Time.Milli;
+	/**
+	 * Treat a backwards jump in group timestamps as a signal to drop the buffered tail.
+	 *
+	 * When a newer group's frames land at least this many milliseconds before the live
+	 * edge, the publisher is reneging everything buffered after that point (e.g. a voice
+	 * agent interrupted mid-utterance). The consumer drops the stale groups and resumes
+	 * from the rewound group. Disabled (undefined) by default.
+	 */
+	resetThreshold?: Signal<Time.Milli> | Time.Milli;
 }
 
 interface Group {
@@ -20,13 +29,51 @@ interface Group {
 	done?: boolean; // Set when #runGroup finishes reading all frames
 }
 
+/**
+ * A recorded rewind boundary.
+ *
+ * After a backwards timestamp jump, groups can still arrive out of order, so a single
+ * sequence floor is not enough: a late new-epoch group can have a lower sequence than the
+ * group that triggered detection. This keeps just enough state to classify any group by
+ * its (sequence, timestamp).
+ */
+class Reset {
+	/** Highest-sequence old-epoch group seen at detection. At or below this is old: drop. */
+	readonly prevMax: number;
+	/** The group whose backwards timestamp triggered detection. At or above this is new: keep. */
+	readonly group: number;
+	/** That group's timestamp; in the ambiguous span, old stragglers sit at or above it. */
+	readonly timestamp: Time.Micro;
+
+	constructor(prevMax: number, group: number, timestamp: Time.Micro) {
+		this.prevMax = prevMax;
+		this.group = group;
+		this.timestamp = timestamp;
+	}
+
+	/** Classify by sequence alone: true=old, false=new, undefined=ambiguous (resolve by timestamp). */
+	bySequence(sequence: number): boolean | undefined {
+		if (sequence <= this.prevMax) return true;
+		if (sequence >= this.group) return false;
+		return undefined;
+	}
+
+	/** Whether a group belongs to the reneged old epoch and should be dropped. */
+	isStale(sequence: number, timestamp: Time.Micro): boolean {
+		return this.bySequence(sequence) ?? timestamp >= this.timestamp;
+	}
+}
+
 /** Reads frames from a MoQ track in order, buffering groups and skipping slow ones to meet the latency target. */
 export class Consumer {
 	#track: Moq.Track;
 	#format: Format;
 	#latency: Signal<Time.Milli>;
+	#resetThreshold?: Signal<Time.Milli>;
 	#groups: Group[] = [];
 	#active?: number; // the active group sequence number
+	#high?: { group: number; timestamp: Time.Micro }; // live edge: max delivered ts + its group
+	#reset?: Reset; // the active rewind boundary, if any
 
 	// Wake up the consumer when a new frame is available.
 	#notify?: () => void;
@@ -42,6 +89,7 @@ export class Consumer {
 		this.#track = track;
 		this.#format = props.format;
 		this.#latency = Signal.from(props.latency ?? Moq.Time.Milli.zero);
+		this.#resetThreshold = props.resetThreshold !== undefined ? Signal.from(props.resetThreshold) : undefined;
 
 		this.#signals.spawn(this.#run.bind(this));
 		this.#signals.cleanup(() => {
@@ -65,9 +113,21 @@ export class Consumer {
 				this.#active = consumer.sequence;
 			}
 
-			if (consumer.sequence < this.#active) {
-				console.warn(`skipping old group: ${consumer.sequence} < ${this.#active}`);
-				// Skip old groups.
+			// Normally we drop anything behind the cursor. With an active reset the cursor isn't
+			// a valid floor (a late new-epoch group can sit below it); defer to the boundary and
+			// admit ambiguous groups so #runGroup can rule on them once their timestamps arrive.
+			let drop: boolean;
+			if (this.#reset) {
+				const verdict = this.#reset.bySequence(consumer.sequence);
+				if (verdict === undefined) drop = false;
+				else if (verdict) drop = true;
+				else drop = consumer.sequence < this.#active;
+			} else {
+				drop = consumer.sequence < this.#active;
+			}
+
+			if (drop) {
+				console.warn(`skipping old group: ${consumer.sequence}`);
 				consumer.close();
 				continue;
 			}
@@ -121,7 +181,10 @@ export class Consumer {
 						this.#notify?.();
 						this.#notify = undefined;
 					} else {
-						// Check for latency violations if this is a newer group.
+						// Newer group: resolve it against an active reset (dropping a reneged
+						// straggler), else detect a new rewind, then check latency.
+						if (this.#classifyStale(group)) return;
+						this.#checkReset(group);
 						this.#checkLatency();
 					}
 				}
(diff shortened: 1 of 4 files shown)


js/hang/src/container/consumer.ts
```diff
@@ -11,6 +11,15 @@ export interface ConsumerProps {
...
 	latency?: Signal<Time.Milli> | Time.Milli;
+	/**
+	 * Treat a backwards jump in group timestamps as a signal to drop the buffered tail.
+	 *
+	 * When a newer group's frames land at least this many milliseconds before the live
+	 * edge, the publisher is reneging everything buffered after that point (e.g. a voice
+	 * agent interrupted mid-utterance). The consumer drops the stale groups and resumes
+	 * from the rewound group. Disabled (undefined) by default.
+	 */
+	resetThreshold?: Signal<Time.Milli> | Time.Milli;
```
kixelated: Do we need this? Can we just hard-code it to zero?

js/hang/src/container/consumer.ts
```diff
@@ -20,13 +29,51 @@ interface Group {
...
+
 /** Reads frames from a MoQ track in order, buffering groups and skipping slow ones to meet the latency target. */
 export class Consumer {
 	#track: Moq.Track;
 	#format: Format;
 	#latency: Signal<Time.Milli>;
+	#resetThreshold?: Signal<Time.Milli>;
 	#groups: Group[] = [];
 	#active?: number; // the active group sequence number
+	#high?: { group: number; timestamp: Time.Micro }; // live edge: max delivered ts + its group
```
kixelated: Should this be bundled with Reset? To keep the related varibles with each other?

js/hang/src/container/consumer.ts
kixelated: And we can just always enable this functionality. I don't see why we should ever make it optional.

rs/moq-mux/src/container/consumer.rs
```diff
@@ -40,6 +40,63 @@ pub struct Consumer<F: Container> {
...
+	high_water: Option<(u64, Timestamp)>,
+
+	// How far a newer group's timestamps must fall *below* the live edge to count as an
+	// explicit reneg of the buffered tail rather than normal reordering. `None` disables
+	// rewind detection (the default), preserving the plain sequence-ordered behavior.
+	reset_threshold: Option<std::time::Duration>,
+
+	// The active rewind boundary, if any. Out-of-order groups are classified against it
+	// so a late new-epoch group is kept while a reneged old-epoch straggler is dropped.
+	reset: Option<Reset>,
```
kixelated: Same comments as the JS side.

rs/moq-mux/src/container/consumer.rs
```diff
@@ -64,6 +124,23 @@ impl<F: Container> Consumer<F> {
...
+	/// When a newer group (higher sequence) carries frames whose timestamps land at
+	/// least `threshold` before the latest delivered frame, the publisher is reneging
+	/// everything buffered after that point (e.g. a voice agent interrupted
+	/// mid-utterance, or any producer that re-anchors its timeline). The consumer drops
+	/// the stale groups, advances its floor past them so out-of-order stragglers are
+	/// ignored, and resumes from the rewound group.
+	///
+	/// Disabled by default. Set `threshold` above the worst-case cross-group reordering
+	/// (B-frames, jitter) so an ordinary stream never trips it.
+	pub fn with_reset_threshold(mut self, threshold: std::time::Duration) -> Self {
```
kixelated: Always on please, don't allow a configurable threshold. If the timestamp goes backwards but groups go forward, reset.

js/hang/src/container/consumer.ts
```diff
@@ -198,6 +261,70 @@ export class Consumer {
...
+			const first = g.frames.at(0);
+			const stale = verdict ?? (first !== undefined && reset.isStale(g.consumer.sequence, first.timestamp));
+			if (stale) {
+				g.consumer.close();
+				g.frames.length = 0;
+			}
+			return !stale;
+		});
+
+		console.warn(`buffer reset: group timestamps rewound (prevMax ${reset.prevMax}, group ${reset.group})`);
```
kixelated, line 307: We actually have to notify downstream consumers to also reset. Like we need to throw out any media data queued in the audio ring buffer or waiting to be rendered.

Maybe we need to include a discontinuity counter or something?

js/hang/src/container/consumer.ts
kixelated: Done in 748f0020d — removed the threshold and the prop entirely. Always-on now: a reset fires whenever a newer group's first frame timestamp is strictly below the live edge.

(Written by Claude)

js/hang/src/container/consumer.ts
kixelated: Done — bundled the live edge, the active boundary, and the new discontinuity counter into a single `Rewind` struct (Rust) / class (JS), so the related state lives together. `Reset` stays as the per-boundary value type.

(Written by Claude)

rs/moq-mux/src/container/consumer.rs
kixelated: Applied to the Rust side too: dropped `reset_threshold` + `with_reset_threshold`/`set_reset_threshold`, bundled into a `Rewind` struct, always-on.

(Written by Claude)

rs/moq-mux/src/container/consumer.rs
kixelated: Done — always-on, no configurable threshold. The trigger is exactly that: a newer group (sequence forward) whose first frame timestamp is strictly `< live_edge` (timestamp backwards). All the existing tests (B-frames, out-of-order groups, latency skips) still pass, so it does not false-trigger on normal streams.

(Written by Claude)

js/hang/src/container/consumer.ts
kixelated: Added the discontinuity counter. It bumps on every reset and is exposed via `Consumer::discontinuity()` (Rust) and the `next()` return + a `discontinuity` getter (JS). So the read that returns the first frame of the new timeline also reports the bumped count, which is the signal downstream needs.

I stopped short of wiring it through `js/watch` in this PR, because that flush is more than a local change: `backend.reset()` already does the right thing (it flushes the audio ring buffer via `audioDecoder.reset()` and re-anchors the clock via `sync.reset()`), but it lives at the **backend** level, whereas the container `Consumer` is created per-track *inside* each decoder. So routing a per-track discontinuity up to the shared audio+video+sync reset is a coordination decision I did not want to guess at.

Two questions before I wire it:
1. In the watch decoder loop, when `next()` reports a discontinuity bump, should it call `backend.reset()` (sync re-anchor + ring flush), or just flush its own decoder/ring locally and leave sync alone? The voice-agent case wants the clock re-anchored, so I lean toward `backend.reset()`.
2. Same treatment for the Rust native render path (the `AudioConsumer` / FFI side), or browser-only for now?

Happy to do the wiring in this PR or as a follow-up once you pick the coordination point.

(Written by Claude)

kixelated: Wired it up for audio in d089bf8fb. The audio decoder reads `discontinuity` off each `consumer.next()`, and when it increments it calls `this.#ring?.reset()` + `this.source.sync.reset()` before decoding the first frame of the new utterance. That is exactly what `MultiBackend.reset()` does, now triggered automatically off the wire signal instead of the app having to call `element.reset()`.

I did it synchronously inside the decode loop rather than via a signal up to `MultiBackend` so the flush is ordered before the new frame is decoded (a signal round-trip would land a microtask later). The decoder already holds `this.source.sync`, so it can re-anchor the shared clock directly.

Coverage:
- Audio-only (the voice-agent case): fully handled.
- A/V: audio drives the shared `sync.reset()`, so both tracks re-anchor; video also drops the reneged groups inside its own container consumer.
- Remaining gap: a video-only buffered stream that rewinds has no audio track to trigger the re-anchor. Easy to add the same hook to the video decode loop if you want it now, but I held off since it is niche and the video downstream queue is just a single held frame.

Let me know if you would rather this live in `MultiBackend` (surface a `discontinuity` signal off the decoder and have the backend orchestrate `reset()`) instead of in the decode loop.

(Written by Claude)

kixelated: Video done too in 586a92c99, mirroring audio. The video decode loop reads `discontinuity` off `next()` and on a change re-anchors `sync`, clears the decode buffer, and clears the playback `timestamp`.

Two video-specific wrinkles I had to handle that audio did not have:
- Clearing `timestamp` is load-bearing: the output guard at `decoder.ts:246` drops frames whose timestamp is below the current one, so without clearing it the rewound (lower-timestamp) frames would all be late-rejected.
- A generation counter (the discontinuity count) is captured when a frame is decoded and re-checked after the `sync.wait` in the output callback, so an in-flight frame from the reneged timeline that was parked waiting for its render time is dropped instead of rendering in the middle of the new utterance.

The held frame is left in place so the last picture shows until the new keyframe renders (no black flash).

So now: audio-only, A/V, and video-only buffered rewinds are all handled. Remaining: the Rust native render path (the FFI/`AudioConsumer` side), which I left out since #1614 is browser. 78 `js/watch` unit tests pass; tsc + biome clean.

(Written by Claude)

rs/moq-mux/src/container/consumer.rs
```diff
@@ -220,6 +336,104 @@ impl<F: Container> Consumer<F> {
...
+			if min >= live_edge {
+				return Ok(false);
+			}
+
+			Reset {
+				prev_max,
+				group: newest.group.sequence,
+				timestamp: min,
+			}
+		};
```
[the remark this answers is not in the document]
kixelated: Fixed in a36a1783f. `poll_reset` now scans newer pending groups from the back and takes the highest-sequence one that actually rewound (`min < live_edge`), so a forward or not-yet-buffered newest group no longer masks a lower-sequence rewind. Added `reset_detected_behind_forward_newest_group` to cover it.

(Written by Claude)

js/hang/src/container/consumer.ts
```diff
@@ -198,11 +266,81 @@ export class Consumer {
...
+
+		// Resume from the earliest survivor; if none, from the rewound group.
+		this.#active = this.#groups[0]?.consumer.sequence ?? reset.group;
+		this.#rewind.liveEdge = { group: reset.group, timestamp: group.latest ?? start };
+		this.#updateBuffered();
+
+		// Wake up any consumer waiting for a new frame.
+		this.#notify?.();
+		this.#notify = undefined;
+	}
```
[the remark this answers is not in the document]
kixelated: Both fixed in a36a1783f. Added `#checkBufferedReset` (run from `next()`) to re-check buffered groups once delivery advances the live edge past them, mirroring the Rust scan. And the post-reset live edge is now anchored at `start` (the rewind point) instead of `group.latest`, matching Rust.

(Written by Claude)

PR #1889, opened 2026-06-23
feat(moq-net): rework payload compression, kept compressed in RAM

kixelated:
Implements [moq-dev/drafts#39](https://github.com/moq-dev/drafts/pull/39) (moq-lite payload compression) on lite-05, with one wire refinement folded in (algorithm explicit on the wire rather than flagless — the draft is being updated to match).

## What changed

**Codec** (`model/compression.rs`)
- `Compression { Deflate, Zstd }` — deflate is the mandatory baseline, zstd optional (new `zstd` dep, `experimental` feature for magicless frames).
- Group-scoped streaming `Encoder`/`Decoder`: one stream per group, sliced per frame. deflate uses raw DEFLATE with the RFC 7692 sync-flush trim; zstd uses magicless, checksum-less frames. The decoder enforces a **cumulative per-group** decompressed bound (decompression-bomb guard).

**Negotiation + wire**
- New SETUP `Compression` parameter (`0x3`) advertises the algorithms an endpoint can **decode**, default `[Zstd, Deflate]`. Negotiated per hop.
- `TRACK_INFO` carries `compression: Option<Compression>` — the algorithm actually used, which **must** be one the receiver advertised (else a protocol violation).

**Model — payloads stay compressed in RAM**
- **Cooked API** compresses on write / decompresses on read, so origins *and* relays cache compressed bytes:
  - `GroupProducer::write_frame(payload)` / `write_frame_at(ts, payload)` compress whole-frame into the per-group stream.
  - `GroupConsumer::read_frame` / `next_frame` decompress (per-consumer decoder).
- **Raw API** is the size-upfront streaming path a relay uses to forward compressed frames without inflating:
  - `FrameProducer` / `FrameConsumer` + `GroupProducer::create_frame(size)` / `append_frame` store/stream bytes verbatim.
  - `GroupConsumer::read_frame_raw` / `next_frame_raw` pass the stored slice through untouched.
- The egress (`lite/publisher.rs`) passes through to a hop that speaks the cached algorithm (zero CPU) and transcodes/decompresses otherwise.

## Public API (all on `dev`, breaking OK)

- `Compression`: now `{ Deflate, Zstd }` (no `None` variant — absence is `Option<Compression>`); adds `encoder()`/`decoder()`/`select()` + `Encoder`/`Decoder`; removes the old whole-payload `compress`/`decompress`. **Breaking.**
- `lite::Setup` gains `compression: Vec<Compression>`; `lite::TrackInfo.compression` is now `Option<Compression>` (was an always-present algorithm) with validation. **Breaking** (wire + type).
- model `TrackInfo`: `compress: bool` → `compression: Option<Compression>` (`with_compress(bool)` kept; adds `with_compression`). **Breaking** (field), ergonomic builder preserved.
- `GroupProducer::write_frame_at`; `GroupConsumer::compression()` / `read_frame_raw` / `next_frame_raw`. **Additive** (default behavior on uncompressed tracks unchanged).

## Cross-package sync

Per the table, `rs/moq-net` wire/API changes pair with `js/net` and `doc/concept`. Those mirrors are a **deliberate follow-up** — lite-05 is still WIP, so the JS side can lag briefly. No other rows apply.

## Test plan

- [x] `cargo test -p moq-net` — 406 tests (codec round-trips for both algorithms incl. cross-frame redundancy + decompression-bomb, SETUP decoder-list + TRACK_INFO `Option` wire round-trips, model cooked/raw split)
- [x] `cargo test -p moq-native --test broadcast` — 64 tests, incl. new **deflate + zstd** subscribe e2e and the existing **deflate timed fetch** e2e
- [x] `cargo test --workspace` — all green (38 binaries)
- [x] `cargo fmt --all --check`, `cargo clippy --workspace --all-targets` — clean
- [ ] `just check` via the pinned nix toolchain — **not run here** (the dev environment had no pty for `nix develop`); please confirm in a real terminal before merge

(Written by Claude)


rs/moq-net/src/model/track.rs
```diff
@@ -36,11 +36,13 @@ pub const DEFAULT_CACHE: Duration = Duration::from_secs(5);
...
-	/// The publisher honors it by negotiating a codec in TRACK_INFO; codec-less
-	/// peers (older drafts) ignore it and send frames verbatim.
-	#[cfg_attr(feature = "serde", serde(default, skip_serializing_if = "std::ops::Not::not"))]
-	pub compress: bool,
+	/// The algorithm this track's cached frame payloads are stored in, or `None`
+	/// for plaintext. This is the *in-RAM* encoding at this hop, not an end-to-end
+	/// property: a publisher sets it to request compression (the model compresses on
+	/// write to honor it), and a subscriber sets it from the algorithm named in
+	/// TRACK_INFO. It is local and per-hop, so it is kept out of the catalog.
+	#[cfg_attr(feature = "serde", serde(skip))]
```
kixelated, line 44: I guess support it in serde, make it a string

rs/moq-net/src/model/track.rs
```diff
@@ -113,9 +115,17 @@ impl Default for TrackInfo {
 }
 
 impl TrackInfo {
-	/// Mark this track's frames as worth compressing, returning `self` for chaining.
-	pub fn with_compress(mut self, compress: bool) -> Self {
-		self.compress = compress;
+	/// Request compression for this track's frame payloads, returning `self` for
+	/// chaining. `true` picks the default algorithm ([`Compression::default`], the
+	/// universal baseline); use [`Self::with_compression`] to name one explicitly.
+	pub fn with_compress(self, compress: bool) -> Self {
```
kixelated, line 121: delete

rs/moq-net/src/model/group.rs
```diff
@@ -185,19 +202,58 @@ impl GroupProducer {
...
-	/// A helper method to write a frame from a single byte buffer.
-	///
-	/// If you want to write multiple chunks, use [Self::create_frame] to get a frame producer.
-	/// But an upfront size is required.
-	pub fn write_frame<B: Into<Bytes>>(&mut self, frame: B) -> Result<()> {
-		let data = frame.into();
-		let mut frame = self.create_frame(data.len())?;
-		frame.write(data)?;
+	/// The parent track's in-RAM compression algorithm, or `None` for plaintext.
+	pub fn compression(&self) -> Option<Compression> {
```
kixelated, line 200: Why is this public?

rs/moq-net/src/model/group.rs
```diff
@@ -185,19 +202,58 @@ impl GroupProducer {
...
+		})?;
+		frame.write(stored)?;
 		frame.finish()?;
 		Ok(())
 	}
 
-	/// Create a frame with an upfront size.
+	/// Compress one whole frame through the per-group encoder, under the state lock
+	/// so the single stream stays in frame order.
+	fn encode(&mut self, payload: &[u8]) -> Result<Bytes> {
```
kixelated, line 239: I think we grab the lock twice, once in encode and once in create_frame? Can we avoid this?

rs/moq-native/tests/broadcast.rs
```diff
@@ -247,20 +247,16 @@ async fn lite05_fetch_roundtrip(scheme: &str) {
...
-		let frame = moq_native::moq_net::Frame {
-			size: payload.len() as u64,
-			timestamp: Some(Timestamp::new(us, Timescale::MICRO).unwrap()),
-		};
-		let mut writer = group.create_frame(frame).expect("failed to create frame");
-		writer
-			.write(bytes::Bytes::from(payload))
+		let timestamp = Timestamp::new(us, Timescale::MICRO).unwrap();
+		group
+			.write_frame_at(timestamp, bytes::Bytes::from(payload))
```
kixelated, line 258: Can you switch the order of the arguments? Feels more natural.

kixelated: Also it's a little weird that the timestamp is optional. It feels very easy to misuse, as we'll get a runtime error if we forget to write one.

Maybe make the timestamp/timescale required for moq-lite in general? For older transports that don't have a timestamp/timescale, we can use milliseconds and Instant::now() or something IDK.

rs/moq-net/src/lite/publisher.rs
```diff
@@ -500,14 +549,22 @@ impl<S: web_transport_trait::Session> Publisher<S> {
 		// a stale/invalid SUBSCRIBE isn't counted as a viewer.
 		let track_stats = self.stats.broadcast(&absolute).publisher_track(&track);
 
+		// Resolve the peer's decoders once (it sent its SETUP at session start); a
+		// pre-lite-05 peer sends none, so it gets verbatim frames.
+		let egress = Egress {
```
kixelated, line 554: Could we rename this to `Setup` (or `Extensions`) instead? And make it session wide? So we can handle more extensions in the future? 

`compression()` would be an async function on it (our encode compression). I don't see why you need `our_compression`.

NOTE: The peer advertises the decompression algorithms. 

Also we shouldn't call .await here. You want to call peer_compression().await riiiight before we need the actual value so we can do as much work as possible. But yeah I don't think it makes a difference in this case.

rs/moq-net/src/lite/publisher.rs
```diff
@@ -693,14 +756,16 @@ impl<S: web_transport_trait::Session> Publisher<S> {
...
-		let compression = if track.info().await?.compress && version.has_timestamps() {
-			Compression::Deflate
-		} else {
-			Compression::None
-		};
+		// Same per-peer choice as a live subscription: the algorithm we serve with is
+		// chosen from the track's (intent) algorithm; passthrough only when it equals
+		// the fetched group's *actual* stored encoding (a relay's compressed cache),
+		// not the origin's plaintext.
+		let algo = pick_compression(
```
kixelated, line 763: Make this a method on the `Setup` class.

rs/moq-net/src/lite/setup.rs
```diff
@@ -59,6 +62,19 @@ pub struct Setup {
...
+	/// names it in TRACK_INFO. Empty means no compression support.
+	pub compression: Vec<Compression>,
+}
+
+impl Setup {
+	/// The compression algorithms an endpoint advertises by default: both `zstd`
+	/// (preferred for its better ratio on small repetitive payloads) and the
+	/// mandatory `deflate` baseline. The same list is used as our encoder set.
+	pub(crate) fn default_compression() -> Vec<Compression> {
+		vec![Compression::Zstd, Compression::Deflate]
```
kixelated, line 76: make this a const. 

&[Compression]

rs/moq-net/src/lite/setup.rs
```diff
@@ -141,6 +187,23 @@ impl PeerSetup {
 			}
 		}
 	}
+
+	/// Await the peer's advertised decoder list, blocking until its SETUP arrives.
+	///
+	/// The publisher uses this to pick the algorithm to compress a track with for
+	/// this peer. Empty means the peer can't decompress, so everything is verbatim.
+	pub async fn compression(&self) -> Vec<Compression> {
+		let mut rx = self.0.subscribe();
```
kixelated, line 196: can you use kio please.

rs/moq-net/src/lite/subscriber.rs
```diff
@@ -71,15 +74,18 @@ pub(super) struct Subscriber<S: web_transport_trait::Session> {
 	version: Version,
 	/// The peer's advertised SETUP (lite-05+), set when its Setup stream is read.
 	peer_setup: super::PeerSetup,
+	/// The algorithms we advertised as decoders, used to validate a received
+	/// TRACK_INFO compression algorithm.
+	compression: Vec<Compression>,
```
kixelated, line 79: This should be a constant. We always advertise the same thing.

rs/moq-net/src/model/compression.rs
```diff
@@ -1,122 +1,415 @@
...
+/// most-preferred advertised `decoders` that this sender can also `encode`, or
+/// `None` if they share nothing. The result is named explicitly on the wire, so
+/// the two ends don't need to compute the same value.
+pub(crate) fn select(decoders: &[Compression], encoders: &[Compression]) -> Option<Compression> {
+	decoders.iter().copied().find(|d| encoders.contains(d))
+}
+
+/// Encodes a group's frame payloads into one shared compressed stream, one slice
+/// per frame. Hold one per group (the stream is reset at each group boundary).
+pub struct Encoder(EncoderInner);
```
kixelated, line 105: This feels like a trait. Or at least enum directly, instead of EncodeInner.

I think a Zstd struct and Deflate struct is the write call regardless of an enum or trait. I don't really like the floating functions.

rs/moq-net/src/model/compression.rs
```diff
@@ -1,122 +1,415 @@
...
+	if out.len() == out.capacity() {
+		out.reserve(out.capacity().max(64));
+	}
+}
+
+/// DEFLATE one frame: feed the payload, sync-flush, and strip the redundant
+/// trailing `00 00 FF FF` the sync flush emits (RFC 7692); the decoder re-inserts
+/// it. The window carries over to the next frame.
+fn deflate_frame(c: &mut Compress, payload: &[u8]) -> Vec<u8> {
+	let mut out = Vec::with_capacity(payload.len() + 16);
```
kixelated, line 173: Should we additionally send the decompressed size over the wire? So we know the buffer size to allocate?

rs/moq-net/src/model/compression.rs
```diff
@@ -1,122 +1,415 @@
...
+			break;
+		}
+	}
+
+	out
+}
+
+/// Inverse of [`zstd_frame`].
+fn zstd_unframe(d: &mut zstd::stream::raw::Decoder<'static>, slice: &[u8], produced: &mut u64) -> Result<Vec<u8>> {
+	let mut out = Vec::with_capacity(slice.len() * 2 + 16);
```
kixelated, line 286: Would it help if we knew the output size in advance?

rs/moq-net/src/model/compression.rs
```diff
@@ -1,122 +1,415 @@
...
+		if remaining == 0 {
+			break;
+		}
+	}
+
+	out
+}
+
+/// Inverse of [`zstd_frame`].
+fn zstd_unframe(d: &mut zstd::stream::raw::Decoder<'static>, slice: &[u8], produced: &mut u64) -> Result<Vec<u8>> {
```
kixelated, line 285: what is `produced`?

rs/moq-net/src/model/compression.rs
```diff
@@ -1,122 +1,415 @@
...
+	// RFC 7692: drop the trailing empty stored block the sync flush emits; the
+	// decoder re-inserts it.
+	debug_assert!(out.ends_with(&[0, 0, 0xff, 0xff]), "deflate sync flush trailer");
+	out.truncate(out.len().saturating_sub(4));
+	out
+}
+
+/// Inverse of [`deflate_frame`]: re-append the `00 00 FF FF` trailer and inflate.
+fn deflate_unframe(d: &mut Decompress, slice: &[u8], produced: &mut u64) -> Result<Vec<u8>> {
+	let mut input = Vec::with_capacity(slice.len() + 4);
```
kixelated, line 202: Please don't allocate just to add 4 bytes to the end, it's disgusting.

Can't we call decompress_vec on a const MAGIC_BYTES at the end insead? 

Make a constnat for 0, 0, 0xff, 0xff

PR #1891, opened 2026-06-23
feat(moq-native): default QUIC backend to noq instead of quinn

kixelated:
## Summary

Switches the default-compiled and default-selected QUIC backend from **quinn** to **noq** across `moq-native` and all binaries. `quinn` stays available as an opt-in `--features quinn` escape hatch, so nothing is removed.

- **Backend selection** (`client.rs`, `server.rs`): flipped the auto-detect priority so `noq` wins when multiple backends are compiled (was `quinn > noq > quiche`, now `noq > quinn > quiche`).
- **Feature defaults** flipped to `noq`, adding a `noq` feature where it was missing: `moq-relay`, `moq-cli`, `moq-boy`, `moq-bench`, `moq-hls`, `moq-rtc`, `moq-srt`, `moq-rtmp`. `moq-ffi`, `libmoq`, and `moq-gst` inherit the new default via `default-features = true`.
- **Crypto-provider wiring fix (latent bug):** quinn routed its provider through `quinn?/rustls-aws-lc-rs` / `quinn?/rustls-ring`, but noq had no equivalent. `web-transport-noq` forced `aws-lc-rs` via its own default features, so a `--features noq,ring` build silently pulled in aws-lc-rs anyway. Set `web-transport-noq` `default-features = false` (workspace root) and route the provider through moq-native's `aws-lc-rs`/`ring` features, exactly mirroring quinn.
- **Logging/docs:** added a `noq=info` tracing directive alongside `quinn=info`; updated `doc/lib/rs/...` to name noq as the default.

## Behavioral notes

- **Congestion control changes to BBR3.** noq explicitly configures BBR3; quinn ran default CUBIC (it carried a `TODO: validate BBR before enabling`). So the effective default congestion controller for the project is now BBR3.
- **Maturity is the real trade-off**, not capability: `web-transport-noq` is 0.2.0 and `noq` is 1.0.0, both younger than the battle-tested `web-transport-quinn 0.11` / `quinn`. Functionality is at parity (mTLS, preferred_address, QUIC-LB CID generation, bloom/retry, platform-verifier all present and tested).

## Public API / breaking-change notes

No public Rust API signatures change. This is a default-feature + wire-stack swap, hence targeting **dev**: release binaries (moq-relay, moq-cli) now ship noq by default, and the default congestion controller changes.

## Test plan

- [x] `cargo check --workspace` (default = noq) clean
- [x] `moq-native` quinn opt-in still builds (`--no-default-features --features quinn,aws-lc-rs,websocket`)
- [x] `--features noq,ring,websocket` now builds correctly (previously mis-wired)
- [x] noq backend integration tests pass: `noq_raw_quic`, `noq_webtransport`, `noq_mtls`
- [x] `taplo format --check` + `cargo fmt --check` (via nix) clean

(Written by Claude)


yury:
@kixelated also enable fast-apple feature, so it will use batched udp (private apple api) you tried to patch on quinn ( as I remember)

kixelated:
> @kixelated also enable fast-apple feature, so it will use batched udp (private apple api) you tried to patch on quinn ( as I remember)

I think I tried turning it on, then turned it off off when I was testing quinn-udp a while back but I don't remember. Might as well turn it on now.

yury:
it is not bad boost on my internal tests. Consider revisit that feature flag. In my experiments I have to fork to enable it right

kixelated:
> it is not bad boost on my internal tests. Consider revisit that feature flag. In my experiments I have to fork to enable it right

Screenshot 2026-06-24 at 12 28 49 PM

IDK if it's worth fixing the feature flag in these upstream crates?

yury:
Screenshot 2026-06-25 at 7 01 31 AM

kixelated:
Yeah, my point is that Quinn and Noq don't have a path to enable it without a patch, which is kind of concerning?

yury:
Do not know really, but my bet is it is a private Apple API (no docs, no examples), so they do that just for experiments. I tested it on iOS/macOS  automatic private api usage workflow, and the way they implement it - it passes automatic private api usage - doesn’t catch that. I believe a patch from you they will accept  (noq).

PR #1893, opened 2026-06-23
moq-gst: moqsink on a bare Element with direct (no-channel) writes

kixelated:
## Summary

A third moqsink rewrite for comparison, the **bare `gst::Element` with direct, no-channel writes** (option B from the discussion). The direct-write core was coded independently, but it builds on the same media foundation @arielmol established in #1771 (timeline/running-time policy, codec/caps factory, B-frames, FLUSH/EOS, per-pad error isolation). Ariel's #1771 commits are merged into this branch so his authorship is credited on merge, and he is listed in the element's author metadata.

Each request pad has its own chain function that writes buffers **straight into the moq producers from the streaming thread**. There is no intermediate channel and no worker task: `moq_net`'s producer writes are synchronous (an in-memory append bounded by group eviction), so the streaming thread never blocks on the network. A thin async task owns only connect and the session lifetime.

## Why bare Element instead of GstAggregator (#1870)

GstAggregator *synchronizes* pads: it will not call `aggregate()` until every non-EOS pad has a buffer, so a sparse pad batches its siblings into bursts (the ~2s stalls measured on #1870). moqsink wants the opposite, **independent per-pad passthrough** (each pad is its own moq track, delivered independently). A bare Element gives that for free: each pad's chain runs on its own thread and never waits on another pad. It also stays on the GStreamer **1.14 baseline** (no `force_live` / 1.22).

## Why no channel (vs #1771)

#1771 keeps the bare Element but hands buffers to a worker over an unbounded channel. Since `GroupProducer::write_frame` is synchronous and non-blocking, that handoff is not load-bearing, so this PR drops it: the chain writes the producer directly. The result is fewer moving parts (no `DataMsg`, no worker loop, no per-pad generations) while keeping pad independence.

## Structure

- `sink/imp.rs`: bare Element. Request pads with chain + event functions; one `Mutex<Option<State>>` holds the producers + per-pad media, held only briefly per buffer. EOS aggregated across pads. Oversized frames are left to moq-net, whose `create_frame` rejects them (`FrameTooLarge`) before allocating the group slot, invalidating just that pad.
- `sink/pad.rs`: per-pad media (caps -> producer, SEGMENT/running-time policy, frame import) + unit tests.
- `sink/session.rs`: connect + transport lifetime + observable `Status`, no media.
- `sink/timeline.rs`: pure SEGMENT/running-time policy.
- Three read-only diagnostics: `connected`, `moq-version`, `estimated-send-bitrate`.

## Public surface / branch targeting

No public library API or wire change. Same element name (`moqsink`), `sink_%u` pads, codec caps, and writable properties. No `gstreamer-base` dependency (it is a bare Element). Internal to the `rs/moq-gst` plugin, so this targets `main`.

## Testing

- `cargo test -p moq-gst`: 34 unit + 3 element tests; `cargo clippy --all-targets` and `cargo fmt --check` clean.
- Live relay: `videotestsrc ! vtenc_h264 ! moqsink.sink_0` plus `audiotestsrc ! opusenc ! moqsink.sink_1` to a local `moq-relay` logged `session connected` -> two independent pads -> `finalized on EOS: ["sink_0", "sink_1", "catalog"]` -> EOS.

## Relationship to the other PRs

- #1771 (Ariel): bare Element + unbounded channel + worker. Proven, deployed, 1.16-safe.
- #1870 (mine): GstAggregator. Smaller, but synchronizes pads (sparse-pad stall) and `force_live` is 1.22.
- This PR (B): bare Element + direct writes. Independent pads like #1771, no channel/worker, 1.14 baseline. Newer than #1771, so less battle-tested.

(Written by Claude)


rs/moq-gst/src/sink/imp.rs
```diff
@@ -13,25 +14,15 @@ use bytes::Bytes;
...
-		gst::DebugColorFlags::empty(),
-		Some("MoQ Sink (async element)"),
-	)
-});
+use super::pad::{Pad, caps_supported};
+use super::session::{CAT, RUNTIME, ResolvedSettings, Session};
+
+/// Reject a frame past the MoQ frame limit (moq-net's wire MAX_FRAME_SIZE, 32 MiB): it could not be
+/// sent anyway, and copying it would let hostile input drive an unbounded allocation. moq-net keeps the
+/// constant `pub(crate)`; a follow-up to make its frame constructor fallible would let this check go.
```
kixelated: This should be merged now, so you can remove MAX_FRAME_BYTES

rs/moq-gst/src/sink/imp.rs
```diff
@@ -63,65 +47,54 @@ impl TryFrom<Settings> for ResolvedSettings {
...
-		let _ = self.sender.send(ControlMessage::Shutdown);
-		RUNTIME.spawn(async move {
-			if let Err(err) = self.join.await {
-				gst::warning!(CAT, "session task ended with error: {err:?}");
+impl State {
+	/// Finalize every live producer once, catalog last; runs on EOS and on stop. Idempotent. A pad
+	/// finalize error is logged and surfaced as the returned `Err`; a catalog error takes precedence.
+	fn finalize_all(&mut self) -> Result<Vec<String>> {
+		let mut order = Vec::new();
+		let mut failure = None;
```
kixelated: IDK if it would be cleaner to have this be `Ok(Vec::new)` to start with and we only push when ok_mut() (is that even a function...?)

rs/moq-gst/src/sink/imp.rs
```diff
@@ -247,324 +253,207 @@ impl ElementImpl for MoqSink {
...
+		if state.session.errored() {
+			return Err(gst::FlowError::Error);
+		}
+
+		let media = state.pads.entry(pad.name().to_string()).or_insert_with(Pad::new);
+		if media.is_failed() {
+			return Ok(gst::FlowSuccess::Ok); // drop quietly; the pad already reported its failure
+		}
 
+		let map = buffer.map_readable().map_err(|_| {
```
kixelated: Can we do this before grabbing the lock?

rs/moq-gst/src/sink/imp.rs
```diff
@@ -247,324 +253,207 @@ impl ElementImpl for MoqSink {
...
+		let _rt = RUNTIME.enter();
+		let mut guard = self.state.lock().unwrap();
+		let Some(state) = guard.as_mut() else {
+			return Err(gst::FlowError::Flushing); // not started
+		};
+		if state.session.errored() {
+			return Err(gst::FlowError::Error);
+		}
+
+		let media = state.pads.entry(pad.name().to_string()).or_insert_with(Pad::new);
```
kixelated: This will be occupied 99% of the time right? Maybe don't use `entry()` to avoid allocating a string every iteration.

rs/moq-gst/src/sink/imp.rs
kixelated: Done in 8c0e920. Removed the constant and the pre-check. `moq-net`'s `create_frame` now rejects oversized frames with `FrameTooLarge` *before* it preallocates the group slot (`group.rs:184`), so an oversized buffer just invalidates the pad like any other bad bitstream, which is what `pad.rs` already documented ("an oversized frame, rejected by moq-net"). (Written by Claude)

rs/moq-gst/src/sink/imp.rs
kixelated: Done. `finalize_all` now starts as `Ok(Vec::new())` and pushes names while it's still `Ok` (via `result.as_mut()`), switching to the first error and leaving the partial order behind. Dropped the separate `failure` accumulator. One behavioral nuance: it's now first-error-wins rather than catalog-takes-precedence; both are teardown finalize errors so the distinction is cosmetic, and I updated the doc comment to match. (Written by Claude)

rs/moq-gst/src/sink/imp.rs
kixelated: Done. The `map_readable` + `copy_from_slice` now happen before the lock is taken (neither needs shared state), so the per-pad lock is held only for the producer write. Trade-off: a buffer arriving on an already-failed pad now incurs the copy before we discover it's failed, but that's an error path and checking `is_failed` needs the lock anyway. (Written by Claude)

rs/moq-gst/src/sink/imp.rs
kixelated: Done. The hot path now does `contains_key` + `get_mut` (no allocation) and only builds the owned `String` key on the rare first-buffer insert. (Written by Claude)

rs/moq-gst/src/sink/imp.rs
```diff
@@ -247,324 +253,207 @@ impl ElementImpl for MoqSink {
...
+			let name = pad.name();
+			if let Some(mut media) = state.pads.remove(name.as_str())
+				&& let Err(err) = media.finalize()
+			{
+				gst::warning!(CAT, "finalize on release {name}: {err:?}");
+			}
+			state.ended.remove(name.as_str());
 		}
 		let _ = self.obj().remove_pad(pad);
 	}
```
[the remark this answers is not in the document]
kixelated: Addressed in 8c0e920. Extracted the EOS-aggregation logic into a shared `maybe_post_eos` helper and call it from `release_pad` after `remove_pad`, so releasing the last still-active pad now posts the element EOS for the pads that already ended (the `!sink_pads.is_empty()` guard still prevents a spurious EOS when the last pad is released without any having ended). (Written by Claude)

rs/moq-gst/src/sink/pad.rs
```diff
@@ -0,0 +1,598 @@
...
+		Self {
+			framed: None,
+			caps: None,
+			failed: false,
+			state: PadState::NoSegment,
+			segment_info: None,
+			segment: None,
+			no_segment_reported: false,
+		}
+	}
```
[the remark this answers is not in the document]
kixelated: Added in 8c0e920. (Written by Claude)

rs/moq-gst/src/sink/timeline.rs
```diff
@@ -0,0 +1,120 @@
...
+
+/// The facts about a SEGMENT that decide timeline continuity.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct SegmentInfo {
+	/// Only TIME segments map to a media timeline (not BYTES/DEFAULT).
+	pub time_format: bool,
+	pub rate: f64,
+	/// Running-time anchor of the segment; continuity is judged on this, not on `start`.
+	pub base_nanos: u64,
+}
```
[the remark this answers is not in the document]
kixelated: Added in 8c0e920. (Written by Claude)

PR #1894, opened 2026-06-23
design: usage stats in the model (BroadcastInfo-carried, origin-attributed)

kixelated:
A design doc only — the proposed architecture for per-broadcast usage stats, to settle the shape before implementing. It supersedes the `with_meter` / `set_meter` approach currently on [#1873](https://github.com/moq-dev/moq/pull/1873).

## The idea

Put the usage sinks **in `BroadcastInfo`**, set them **at construction**, and let the immutable `Arc<BroadcastInfo>` carry them to every track/group/frame. `Usage` is atomics, so the model bumps through a shared `&Arc<Usage>` — no mutation, so **no `with_meter` setter and no `Arc::make_mut`**.

- **Ingress** is baked in when the publisher builds the broadcast (it has its session's sink).
- **Egress** is attributed by the per-session `OriginConsumer`: when it hands out a `BroadcastConsumer` it builds that consumer's own `Arc<BroadcastInfo>` with this session's egress sink — so **per-tier egress survives**, with zero mutation, and `with_meter` disappears from every handler/gateway.
- **Live viewer/publisher counts** move into the model: a `BroadcastConsumer` is one live viewer while it has ≥1 outstanding `TrackConsumer`. This replaces the stats-layer `SessionBroadcasts` sentinel and fixes the fragile split where the payload sink came from a non-bumping `subscriber_meter()` while the lifecycle guard lived elsewhere.
- `Arc<TrackInfo>` is threaded to the group too, so `timescale` stops being passed separately.

Full writeup, type sketches, flows, and a 3-phase plan are in [`rs/moq-net/DESIGN-stats.md`](rs/moq-net/DESIGN-stats.md).

## Deliberately out of scope

- The binary `Tier` (internal/external) split, and the **published stats schema** — both unchanged here. The intent behind tiers is real (billable customer traffic vs cluster forwarding), but encoding it as a fixed enum is rigid; moving attribution to per-auth-root keying is a cleaner model that reshapes the output format, so it's a separate change.

Looking for a sign-off on the shape (especially construction-time attribution and the origin-injection hook) before I sink the ~1500-line implementation.

(Written by Claude)

### DIFF
--- a/rs/moq-net/DESIGN-stats.md
+++ b/rs/moq-net/DESIGN-stats.md
@@ -1,0 +1,161 @@
+# Usage stats in the model
+
+Status: proposed. Supersedes the `with_meter` / `set_meter` approach currently on
+the `stats-in-model` branch (PR #1873).
+
+## Why
+
+We want every transport that drives the model types (moq-lite, IETF, and the
+non-MoQ gateways: moq-srt, moq-rtc, moq-rtmp) to record per-broadcast usage
+uniformly, without each one re-instrumenting its data path. The current branch
+gets there, but the *attachment* mechanism is awkward:
+
+- `Broadcast{Producer,Consumer}::with_meter(sink)` is a setter the relay/gateway
+  calls after constructing or obtaining a broadcast. It exists only because the
+  sink isn't known at construction.
+- On the relay subscriber side the payload sink is pulled from a *separate*,
+  non-bumping `subscriber_meter()` while the lifecycle guard (which counts the
+  subscriber) lives elsewhere in a `stats_guards` map. Two handles for one
+  broadcast is fragile and, as noted in review, can mask a "we forgot to count
+  this" bug.
+- There's no model-level notion of "how many viewers/publishers are live"; the
+  stats layer reconstructs it with a per-session `SessionBroadcasts` sentinel.
+
+## The idea
+
+Put the usage sinks **in `BroadcastInfo`**, set them **at construction**, and let
+the immutable `Arc<BroadcastInfo>` carry them down to every track, group, and
+frame. `Usage` is atomics, so the model bumps through a shared `&Arc<Usage>` —
+there is no mutation, so there is no setter (`with_meter`) and no `Arc::make_mut`.
+
+```rust
+// usage.rs
+pub struct Usage {
+    // payload, bumped as media flows
+    groups:  AtomicU64,
+    frames:  AtomicU64,
+    bytes:   AtomicU64,
+    // lifecycle, bumped as broadcast handles open/close (live = opened - closed)
+    opened:  AtomicU64,
+    closed:  AtomicU64,
+}
+
+// One per direction. Lives in BroadcastInfo, pub so a stats layer can read it.
+pub struct BroadcastStats {
+    pub producer: Arc<Usage>, // ingress
+    pub consumer: Arc<Usage>, // egress
+}
+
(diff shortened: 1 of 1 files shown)


rs/moq-net/DESIGN-stats.md
```diff
@@ -0,0 +1,161 @@
...
+
+```rust
+// usage.rs
+pub struct Usage {
+    // payload, bumped as media flows
+    groups:  AtomicU64,
+    frames:  AtomicU64,
+    bytes:   AtomicU64,
+    // lifecycle, bumped as broadcast handles open/close (live = opened - closed)
+    opened:  AtomicU64,
```
kixelated, line 39: rename to `active`?

We only care about BroadcastConsumers that have at least one TrackConsumer instance open. Otherwise we're just counting the number of sessions where a broadcast is announced, which I guess is cool, but IDK.

How about announced: Gauge, and active: Gauge. Gauge is two atomics, open and closed.

rs/moq-net/DESIGN-stats.md
```diff
@@ -0,0 +1,161 @@
...
+pub struct BroadcastStats {
+    pub producer: Arc<Usage>, // ingress
+    pub consumer: Arc<Usage>, // egress
+}
+
+// broadcast.rs — now immutable after construction.
+pub struct BroadcastInfo {
+    pub hops:  OriginList,
+    pub epoch: SystemTime,
+    pub stats: BroadcastStats,
```
kixelated, line 53: Maybe make this Arc instead of producer/consumer? So there's one item on the heap instead of two.

rs/moq-net/DESIGN-stats.md
```diff
@@ -0,0 +1,161 @@
...
+(or a small trait object) supplied by the relay when it scopes the origin per
+session. Default `None` => no-op sinks. The stats layer stays the only thing that
+knows about tiers.
+
+## Live viewer / publisher counts in the model
+
+The model owns "is this broadcast being watched / published," replacing
+`SessionBroadcasts`:
+
+- A `BroadcastConsumer` counts as **one live viewer while it has ≥ 1 outstanding
```
kixelated, line 106: yeah some Arc<BroadcastActive> that holds an Arc<BroadcastInfo>. when all TrackConsumers are dropped, the BroadcastActive is dropped and we update the Gauge. Maybe a bool on if it's a producer or consumer, or just make a separate type to avoid the branch IDK. Or have Arc<Usage> like the plan suggests and ignore my previous comment.

rs/moq-net/DESIGN-stats.md
```diff
@@ -0,0 +1,161 @@
...
+  `Arc<Usage>` pair) instead of vending `with_meter`-style guards; the publish
+  loop reads them; `SessionBroadcasts` goes away (model owns the counts). The
+  current `BroadcastStats` type here is renamed `BroadcastHandle` to free the
+  name.
+- **handlers / gateways**: stop calling `with_meter`. Ingress builds the broadcast
+  with its sink; egress just uses a stats-aware origin.
+
+## Out of scope (deliberately)
+
+- **The `Tier` (internal/external) split.** It's kept as-is here. The intent
```
kixelated, line 143: I think Origin should flag if a producer/consumer is internal/external? IDK there's something wrong with the Origin abstraction that needs to be untangled but I don't know how.

PR #1895, opened 2026-06-23
feat(moq-net): usage stats in the model (BroadcastInfo-carried, origin-attributed)

kixelated:
Implements the design in [`rs/moq-net/DESIGN-stats.md`](https://github.com/moq-dev/moq/pull/1894/files) (PR #1894), which supersedes the `with_meter` / `set_meter` approach on #1873. Per-broadcast usage sinks live in `BroadcastInfo`, are set at construction, and ride the immutable `Arc<BroadcastInfo>` down to every track, group, and frame. `Usage` is atomics, so the model bumps through a shared `&Arc<Usage>` with no mutation, no setter, and no `Arc::make_mut`.

This takes over #1873 (which should be closed in favor of this).

## What changed

The design's three phases, each landed as its own commit:

1. **The model carries usage stats.** New `model/usage.rs`: `Usage` (groups/frames/bytes payload + opened/closed lifecycle atomics) and `BroadcastStats` (one `Usage` per direction). `BroadcastInfo` gains a `stats` field; `Arc<BroadcastInfo>` + `Arc<TrackInfo>` are threaded through Broadcast/Track/Group; payload bumps move into the model (ingress at create_group/append_frame, egress as the consumer reads). The stats-layer `BroadcastStats` is renamed `BroadcastHandle`.

2. **Origin attributes egress, ingress baked at construction.** The lite/IETF subscriber loops build the broadcast with `stats.producer` already set. The per-session `OriginConsumer::with_egress` provider stamps each `BroadcastConsumer` it yields (announce stream, `request_broadcast`, dynamic accept) with that session's egress sink, so per-tier egress survives with zero mutation. The handler/gateway per-frame/byte/group bumps are deleted; the stats layer reads the shared `Arc<Usage>` instead of vending payload guards.

3. **Model-tracked live viewer/publisher counts.** A `Live` refcount over each sink: a `BroadcastConsumer` is one live viewer while it has ≥1 outstanding `TrackConsumer` (the token rides into the `TrackSubscriber`, so a subscription counts for its whole life); a broadcast with ≥1 live `TrackProducer` is one publisher (the publisher token rides the shared broadcast state, so the relay's dynamic-accept ingress path is covered). `SessionBroadcasts` is gone; the publish loop maps `opened - closed` onto the existing `broadcasts` / `broadcasts_closed` fields, so the **published schema is unchanged**.

## Semantics

- **Egress is per-consumer** (N viewers count N times); **ingress is single-writer** (counts once). A relay counts both directions without double-counting either.
- The lite **fetch** path is metered through the model: a fetch re-serves an existing group, so it bumps frames/bytes but not `groups`.
- The model counts **raw (decompressed)** bytes vs. the old post-compression wire bytes. Only the hang catalog track sets `compress`, so the difference is noise; a follow-up will cache compressed bytes so the count is the wire size again. (Confirmed acceptable.)

## Testing

`cargo test -p moq-net` passes (400 lib + 4 integration), including new model tests for ingress-once/egress-per-viewer, origin egress-stamping, and live viewer/publisher counts. `clippy`/`doc`/`fmt` clean; dependent crates (`moq-mux`, `hang`, `moq-relay`, `moq-native`, `moq-srt`, `moq-json`, `moq-ffi`) compile.

> Note: this environment's egress policy blocks the `kixelated/nvidia-video-codec-sdk` git fork and the sandbox has no IPv6, so the full workspace `just check` and the `moq-native` end-to-end relay suite can't run here. CI's nix run is the source of truth for those.

## Cross-package sync

- `js/net`: not mirrored. The stats aggregator / usage sinks are relay-side Rust only; the browser client has no equivalent. No `doc/concept` wire change (this is API, not wire).

## Breaking changes (target `dev`)

Public API in `rs/moq-net`: `BroadcastInfo` gains `stats`; `Broadcast{Producer,Consumer,Dynamic}` carry `Arc<BroadcastInfo>`; `GroupProducer::new` takes `Arc<TrackInfo>` + `Arc<BroadcastInfo>` (was `Option<Timescale>`); the stats-layer `BroadcastStats` is renamed `BroadcastHandle`; `SessionBroadcasts` / `BroadcastSubscription` and the `publisher_broadcasts` / `subscriber_broadcasts` / per-track `frame`/`bytes`/`group` methods are removed; `Counters` swaps its payload + broadcasts atoms for a shared `Arc<Usage>`. `TrackProducer::new(name, info)` keeps its signature (standalone tracks get a shared no-op broadcast).

## Follow-ups

- Wire moq-rtc (WHIP/WHEP) and moq-rtmp through the same origin egress / ingress-at-construction seam (the mechanism exists; their origins just need `with_egress` and an ingress-baked broadcast).
- Cache compressed bytes in the model so the byte count is the wire size for compressed tracks.

(Written by Claude)

### DIFF
--- a/rs/moq-net/src/ietf/publisher.rs
+++ b/rs/moq-net/src/ietf/publisher.rs
@@ -300,7 +300,9 @@ impl<S: web_transport_trait::Session> Publisher<S> {
 		msg: ietf::GroupHeader,
 		priority: u8,
 		mut group: GroupConsumer,
-		track_stats: std::sync::Arc<crate::PublisherTrack>,
+		// Held for the group's lifetime so it counts as one subscription; payload
+		// is metered by the model when the frames are read.
+		_track_stats: std::sync::Arc<crate::PublisherTrack>,
 		version: Version,
 	) -> Result<(), Error> {
 		let mut stream = session.open_uni().await.map_err(Error::from_transport)?;
@@ -309,7 +311,6 @@ impl<S: web_transport_trait::Session> Publisher<S> {
 		let mut stream = Writer::new(stream, version);
 
 		stream.encode(&msg).await?;
-		track_stats.group();
 
 		loop {
 			let frame = tokio::select! {
@@ -333,7 +334,6 @@ impl<S: web_transport_trait::Session> Publisher<S> {
 
 			// Write the size of the frame.
 			stream.encode(&frame.size).await?;
-			track_stats.frame();
 
 			if frame.size == 0 {
 				// Have to write the object status too.
@@ -349,9 +349,7 @@ impl<S: web_transport_trait::Session> Publisher<S> {
 
 					match chunk? {
 						Some(mut chunk) => {
-							let n = chunk.len() as u64;
 							stream.write_all(&mut chunk).await?;
-							track_stats.bytes(n);
 						}
 						None => break,
 					}
--- a/rs/moq-net/src/ietf/publisher.rs
+++ b/rs/moq-net/src/ietf/publisher.rs
@@ -19,22 +19,16 @@ pub(super) struct Publisher<S: web_transport_trait::Session> {
 	origin: OriginConsumer,
 	control: Control,
 	stats: StatsHandle,
-	/// Per-session egress broadcast-subscription tracker. Each downstream
-	/// subscription holds a guard so `broadcasts - broadcasts_closed` counts
-	/// the distinct sessions (viewers) watching each broadcast.
-	broadcasts: crate::SessionBroadcasts,
 	version: Version,
 }
 
 impl<S: web_transport_trait::Session> Publisher<S> {
 	pub fn new(session: S, origin: OriginConsumer, control: Control, stats: StatsHandle, version: Version) -> Self {
-		let broadcasts = stats.publisher_broadcasts();
 		Self {
 			session,
 			origin,
 			control,
 			stats,
-			broadcasts,
 			version,
 		}
 	}
@@ -125,9 +119,8 @@ impl<S: web_transport_trait::Session> Publisher<S> {
 
 		// Per-track subscription guard (bumps `subscriptions`). Taken before
 		// validation so a stale/invalid SUBSCRIBE still counts as an attempt,
-		// matching the lite path. The per-(session, broadcast) `broadcasts`
-		// sentinel that counts viewers is taken only once the subscription is
-		// validated and active, below.
+		// matching the lite path. The viewer count (`broadcasts`) is owned by the
+		// model: `broadcast.track()` hands the subscription a live-viewer token.
 		let track_stats = std::sync::Arc::new(self.stats.broadcast(&absolute).publisher_track(&track_name));
 
 		// We just received a subscribe for this exact namespace, so the peer must have already
@@ -155,10 +148,6 @@ impl<S: web_transport_trait::Session> Publisher<S> {
 				return Ok(());
 			}
 		};
-
-		// Subscription is now active: count this session as a viewer of the
-		// broadcast. Dropping this guard (subscription end) releases it.
-		let _broadcast_sub = self.broadcasts.subscribe(&absolute);
 
 		// Send SubscribeOk on the stream
 		stream.writer.encode(&ietf::SubscribeOk::ID).await?;
--- a/rs/moq-net/src/ietf/session.rs
+++ b/rs/moq-net/src/ietf/session.rs
@@ -27,6 +27,15 @@ pub fn start<S: web_transport_trait::Session>(
 		// nothing, and an empty subscribe origin issues no SUBSCRIBE_NAMESPACE.
 		let publish = publish.unwrap_or_else(|| OriginProducer::empty(Origin::random()).consume());
 		let subscribe = subscribe.unwrap_or_else(|| OriginProducer::empty(Origin::random()));
+
+		// Attach this session's egress sink provider so the model meters every
+		// broadcast the publisher serves downstream, keyed by absolute path and tier.
+		let publish = {
+			let stats = stats.clone();
+			publish.with_egress(std::sync::Arc::new(move |abs: &str| {
+				stats.broadcast(abs).egress_usage()
+			}))
+		};
 		let res = match version {
 			Version::Draft14 | Version::Draft15 | Version::Draft16 => {
 				let Some(setup) = setup else {
--- a/rs/moq-net/src/model/broadcast.rs
+++ b/rs/moq-net/src/model/broadcast.rs
@@ -1,13 +1,23 @@
 use std::{
 	collections::{HashMap, VecDeque, hash_map},
-	sync::Arc,
+	sync::{Arc, LazyLock},
 	task::{Poll, ready},
 	time::{Duration, SystemTime, UNIX_EPOCH},
 };
 
 use crate::{Error, TrackConsumer, TrackProducer, TrackRequest, TrackWeak};
 
-use super::{OriginList, TrackInfo};
+use super::{BroadcastStats, OriginList, TrackInfo};
+
+/// Shared no-op broadcast metadata for tracks created outside a broadcast.
+///
+/// A standalone [`TrackProducer::new`] track has no broadcast to inherit from, so
+/// it points at this default instance; its usage sinks have no reader, so bumping
+/// them is harmless. Shared so a fresh standalone track is a cheap `Arc` clone.
+pub(crate) fn noop_broadcast() -> Arc<BroadcastInfo> {
+	static NOOP: LazyLock<Arc<BroadcastInfo>> = LazyLock::new(|| Arc::new(BroadcastInfo::default()));
+	NOOP.clone()
+}
 
 /// Wall-clock base for the broadcast epoch: 2020-01-01T00:00:00 UTC, expressed as
 /// seconds since the Unix epoch. The wire carries milliseconds since this base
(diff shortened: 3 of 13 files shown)


rs/moq-net/src/ietf/publisher.rs
```diff
@@ -300,7 +289,9 @@ impl<S: web_transport_trait::Session> Publisher<S> {
 		msg: ietf::GroupHeader,
 		priority: u8,
 		mut group: GroupConsumer,
-		track_stats: std::sync::Arc<crate::PublisherTrack>,
+		// Held for the group's lifetime so it counts as one subscription; payload
+		// is metered by the model when the frames are read.
+		_track_stats: std::sync::Arc<crate::PublisherTrack>,
```
kixelated, line 294: This should be in model now right?

rs/moq-net/src/ietf/session.rs
```diff
@@ -27,6 +27,15 @@ pub fn start<S: web_transport_trait::Session>(
 		// nothing, and an empty subscribe origin issues no SUBSCRIBE_NAMESPACE.
 		let publish = publish.unwrap_or_else(|| OriginProducer::empty(Origin::random()).consume());
 		let subscribe = subscribe.unwrap_or_else(|| OriginProducer::empty(Origin::random()));
+
+		// Attach this session's egress sink provider so the model meters every
+		// broadcast the publisher serves downstream, keyed by absolute path and tier.
+		let publish = {
+			let stats = stats.clone();
+			publish.with_egress(std::sync::Arc::new(move |abs: &str| {
+				stats.broadcast(abs).egress_usage()
```
kixelated, line 36: Why isn't this done in origin?

rs/moq-net/src/model/broadcast.rs
```diff
@@ -1,13 +1,23 @@
...
 };
 
 use crate::{Error, TrackConsumer, TrackProducer, TrackRequest, TrackWeak};
 
-use super::{OriginList, TrackInfo};
+use super::{BroadcastStats, OriginList, TrackInfo, Usage, usage::Live};
+
+/// Shared no-op broadcast metadata for tracks created outside a broadcast.
+///
+/// A standalone [`TrackProducer::new`] track has no broadcast to inherit from, so
```
kixelated, line 14: Delete TrackProducer::new and delete this. All tracks must be created/owned by a broadcast.

rs/moq-net/src/model/broadcast.rs
```diff
@@ -133,17 +158,20 @@ impl BroadcastState {
...
 
 impl BroadcastProducer {
 	/// Create a producer for the given broadcast metadata. Prefer [`BroadcastInfo::produce`].
 	pub fn new(info: BroadcastInfo) -> Self {
-		Self {
-			info,
-			state: Default::default(),
+		let info = Arc::new(info);
+		let state = kio::Producer::<BroadcastState>::default();
+		if let Ok(mut s) = state.write() {
```
kixelated: Don't make a default state then grab the lock. Set the init state instead

rs/moq-net/src/model/broadcast.rs
```diff
@@ -179,8 +207,9 @@ impl BroadcastProducer {
 		info: impl Into<Option<TrackInfo>>,
 	) -> Result<TrackProducer, Error> {
 		let info = info.into().unwrap_or_default();
-		let track = TrackProducer::new(name, info);
 		let mut state = BroadcastState::modify(&self.state)?;
+		let publisher = state.publishers.enter();
+		let track = TrackProducer::for_broadcast(name, info, self.info.clone(), Some(publisher));
```
kixelated, line 212: rename to new and make pub(super)

rs/moq-net/src/model/broadcast.rs
```diff
@@ -323,7 +355,10 @@ impl BroadcastDynamic {
 		}))?;
 
 		let name = state.request_order.pop_front().expect("predicate guaranteed a request");
-		let pending = state.requests.remove(&name).expect("request_order out of sync");
+		let mut pending = state.requests.remove(&name).expect("request_order out of sync");
+		// Serving this request makes the broadcast a live publisher (the relay's
+		// ingress tracks are created on this path).
+		pending.set_publisher(state.publishers.enter());
```
kixelated, line 361: This set publisher stuff is sus

rs/moq-net/src/model/broadcast.rs
```diff
@@ -388,15 +424,35 @@ impl BroadcastDynamic {
...
 	}
 
+	/// Replace this consumer's egress usage sink, attributing everything it
+	/// delivers to a particular session/tier.
+	///
+	/// The shared channel is untouched (so [`Self::is_clone`] and origin routing
+	/// are unaffected); only this handle's `Arc<BroadcastInfo>` is swapped for one
+	/// with a different `stats.consumer`. The origin calls this once per session
+	/// when it hands the consumer out.
+	pub(crate) fn set_consumer_sink(&mut self, usage: Arc<Usage>) {
```
kixelated, line 447: IDK about this

rs/moq-net/src/model/group.rs
```diff
@@ -169,20 +172,22 @@ impl GroupProducer {
...
 			info,
 			state: kio::Producer::default(),
-			timescale,
+			track,
+			broadcast,
 		}
 	}
 
 	/// The parent track's negotiated timescale, or `None` for untimed tracks.
 	pub fn timescale(&self) -> Option<Timescale> {
```
kixelated: Should delete this

rs/moq-net/src/model/group.rs
```diff
@@ -342,7 +359,18 @@ impl std::ops::Deref for GroupConsumer {
 impl GroupConsumer {
 	/// The parent track's negotiated timescale, or `None` for untimed tracks.
 	pub fn timescale(&self) -> Option<Timescale> {
```
kixelated: Delete

rs/moq-net/src/model/origin.rs
```diff
@@ -1179,6 +1179,10 @@ impl BroadcastRequest {
 /// handler drops before serving it.
 pub struct BroadcastRequested {
 	inner: Requested,
+	// When set, stamp the resolved (dynamically served) consumer with this
+	// session's egress sink for the absolute path. The already-announced path
+	// (`Ready`) is stamped by `get_broadcast` instead.
+	stamp: Option<(EgressSink, PathOwned)>,
```
kixelated, line 1185: Maybe we have Arc<OriginInfo> instead? IDK

rs/moq-net/src/model/usage.rs
```diff
@@ -0,0 +1,159 @@
...
+	groups: AtomicU64,
+	frames: AtomicU64,
+	bytes: AtomicU64,
+	opened: AtomicU64,
+	closed: AtomicU64,
+}
+
+impl Usage {
+	/// Record one group.
+	pub(crate) fn add_group(&self) {
```
kixelated, line 32: Maybe all should be pub for custom stats.

rs/moq-net/src/model/usage.rs
```diff
@@ -0,0 +1,159 @@
...
+/// While at least one [`LiveToken`] is outstanding the handle counts as one live
+/// viewer/publisher: the `0 -> 1` transition bumps [`Usage::open`] and the last
+/// `1 -> 0` drop bumps [`Usage::close`], so `opened - closed` is the live count.
+/// Clones of a handle share one `Live`, so they are a single logical
+/// viewer/publisher.
+#[derive(Debug)]
+pub(crate) struct Live {
+	usage: Arc<Usage>,
+	count: AtomicUsize,
+}
```
kixelated, line 97: Maybe use Weak<LiveToken>? If it fails to upgrade, we make a new LiveToken?

kixelated:
Thanks for the thorough pass. Pushed `27730d3` for the clear-cut ones:

- **`BroadcastProducer::new`** (default-then-lock) — now builds `BroadcastState` with the publisher count set up front.
- **`GroupProducer::timescale` / `GroupConsumer::timescale`** — both deleted. The one caller (lite fetch) now reads the timescale from `track.info()` (the `TrackInfo` it already fetches for compression), so the group keeps `Arc<TrackInfo>` only for the model's own use.

On the bigger ones I'd like your call before I sink the effort, since a couple ripple beyond moq-net:

1. **Delete `TrackProducer::new` / `noop_broadcast` (all tracks broadcast-owned).** I'm on board with the principle, but it's ~88 call sites across `moq-mux` (container producer/consumer), `moq-json`, and `moq-net`, and most of those create *standalone* tracks for muxing — making them broadcast-owned is a `moq-mux` design change, not a mechanical rename. It also reverses the design doc's "keep `new` so the ~80 deferred sites are untouched". **Do it in this PR, or land this and do the broadcast-owned migration as a follow-up?** (And I read broadcast.rs:212 as "rename `for_broadcast` → `new`, `pub(super)`" once standalone `new` is gone — confirm?)

2. **Subscriptions in the model** (ietf/publisher.rs:294). Agreed — `subscriptions`/`subscriptions_closed` can move to the model the same way `broadcasts` did (a token on a `subscriptions` `Usage`), which lets me drop the `publisher_track`/`subscriber_track` guards entirely. This PR or a follow-up?

3. **Egress wiring "why not in origin?"** (ietf/session.rs:36). The closure is `stats.broadcast(abs).egress_usage()` — the model deliberately can't depend on the stats layer, so the stats→usage mapping has to be supplied from outside. I put it in `session::start` because that's where the `StatsHandle` already lives, but it's duplicated lite/ietf. I can instead attach it in **moq-relay** when it scopes the per-session origin (pulling the stats coupling out of moq-net's session entirely), or factor a shared helper. Preference?

4. **"set publisher stuff is sus"** (broadcast.rs:361). The producer token rides the shared `BroadcastState` so the relay's dynamic-accept ingress path (`poll_requested_track`) is covered as well as `create_track`/`reserve_track`. What's bugging you — the multiple attach points, the `Default` no-op `Live`, or the whole approach? Happy to reshape if you have a shape in mind.

Exploratory ones I'll treat as follow-ups unless you want them now: make `Usage`'s bump methods `pub` for custom stats (usage.rs:32), `Weak<LiveToken>` (usage.rs:97), `Arc<OriginInfo>` (origin.rs:1185), and the `set_consumer_sink` clone-to-swap (broadcast.rs:447).

(Written by Claude)

---

kixelated:
Correction on the timescale deletions: CI caught that `hang::container::Frame::encode` uses `GroupProducer::timescale` (to decide whether to attach a per-frame timestamp), so I kept that one and deleted only `GroupConsumer::timescale` (`3ab234f`). Fully removing `GroupProducer::timescale` means threading the timescale into `hang`'s `Frame::encode` (and its moq-mux callers) instead — happy to do that if you want it gone, just flagging it's a cross-crate change rather than a local delete.

(Written by Claude)

---

PR #1896, opened 2026-06-23
feat(moq-net): mandatory timestamps for moq-lite-05

kixelated:
## Summary

Makes per-track **timescale** and per-frame **timestamps** mandatory on moq-lite-05, gives both Rust and JS a proper timestamp type, and wires real timestamps through the media/transport paths.

Behavioral contract:

- **Every track is timed.** `TrackInfo.timescale` defaults to **milliseconds** (Rust and JS).
- **moq-lite-05 always carries timestamps**: `TRACK_INFO` encodes the timescale as a mandatory non-zero varint, and every frame is prefixed with a zigzag-delta timestamp at the track's scale.
- **moq-transport carries timestamps** as per-object extension headers — Timestamp (`0x06`, absolute) in Timescale (`0x08`) units, the MOQ Object Properties registry ids shared with draft-ietf-moq-loc — decoded on receive and emitted on send.
- **Timeless sources** (a peer/track with no real presentation time) fall back to wall-clock: Rust `write_frame_now`/`create_frame_now`, JS `Timestamp.now()`.
- **Timestamps auto-convert** into the track's timescale.

### Rust

- `Timescale` (default `MILLI`) and `Timestamp` (value + scale) carry their own scale; `Timestamp::now()` / `From<Instant>` are millisecond-scale and anchored at **2020-01-01** (a timestamp is non-negative + roughly monotonic, not a real clock — the smaller magnitude trims the first frame's varint).
- `Frame.timestamp` is a required `Timestamp`; `GroupProducer::create_frame` / `write_frame` (and `TrackProducer::write_frame`) require it, with `create_frame_now` / `write_frame_now` for the wall-clock case.
- `TrackInfo.timescale` and `lite::TrackInfo.timescale` are non-optional.
- Surfaced a real bug: the moq-mux **LOC** and **fMP4** export paths wrote the net frame with size only (wall-clock), dropping the media PTS — both now pass it through.
- IETF publisher/subscriber encode/decode the Timestamp/Timescale object properties (`ietf::{encode,decode}_object_time`).

### JS (@moq/net + consumers)

- New `Timestamp` (value + scale) and `Timescale` types mirroring Rust; `Timestamp.now()` is `performance.now()` ms.
- `Frame.timestamp` is required; `writeFrame(frame)` / `readFrame()` both use the `Frame` type (symmetric). No `writeFrameNow` — timeless callers pass `Timestamp.now()`.
- Model `TrackInfo` gains a `timescale` (default ms): the lite publisher advertises it and emits each frame converted to it; the subscriber reconstructs a `Timestamp` at the wire scale and carries it into the accepted track (so a relay re-publishes at the same scale). hang/loc/publish pass `Timestamp.fromMicros`.

## Public API changes (`rs/moq-net`, breaking — targets `dev`)

- `Frame.timestamp`: `Option<Timestamp>` → `Timestamp`; size-only `From<uN>` shortcuts removed.
- `GroupProducer`: `create_frame(Frame)` requires a timestamp; new `create_frame_now(size)`; `write_frame(timestamp, data)` + `write_frame_now(data)`; `timescale()` → `Timescale`.
- `GroupConsumer::timescale()` → `Timescale`; `TrackProducer::write_frame(timestamp, data)` + `write_frame_now(data)`.
- `TrackInfo.timescale` / `lite::TrackInfo.timescale`: `Option<Timescale>` → `Timescale`.
- `Timescale`: `Default` = `MILLI`; `Timestamp::now()` / `From<Instant>` now millisecond-scale, 2020-anchored.
- `ietf::{encode_object_time, decode_object_time}` added.

## Test plan

- [x] `cargo test -p moq-net -p moq-mux --lib` (400 + 288), incl. `ietf::group` object-property round-trips; `cargo clippy` + `cargo doc -D warnings` clean.
- [x] `@moq/net` 189 / `@moq/json` 32 / hang container 26 / `@moq/loc` 6; `tsc` + biome clean across net/hang/loc/json/publish/watch.
- [ ] moq-native integration tests can't *run* in the sandbox (no IPv6 → `[::]:0` bind), but pass in CI.

## Follow-ups

- `doc/concept` note on the lite-05 mandatory timestamp + moq-transport object properties.
- Optional: drop the per-process wall-clock jitter on the Rust anchor if epoch-obfuscation is no longer wanted.

(Written by Claude)

### DIFF
--- a/js/net/src/group.ts
+++ b/js/net/src/group.ts
@@ -8,11 +8,10 @@ export const MAX_GROUP_CACHE_BYTES = 32 * 1024 * 1024;
 export const MAX_GROUP_FRAMES = 1024;
 
 /**
- * A frame buffered in a {@link Group}: its presentation timestamp and payload bytes.
+ * A frame buffered in a {@link Group}: its presentation {@link Timestamp} and payload bytes.
  *
- * The timestamp is in milliseconds; wall-clock accuracy doesn't warrant finer units for
- * transport-level timing, and the wire layer converts it into the track's timescale.
- * A future `Timestamp` type (mirroring Rust) will let a track pick its own scale.
+ * The timestamp carries its own scale, so a track can pick its units; the wire layer
+ * converts it into the track's negotiated timescale.
  */
 export interface Frame {
 	/** The frame payload. */
--- a/js/net/src/group.ts
+++ b/js/net/src/group.ts
@@ -1,5 +1,5 @@
 import { Signal } from "@moq/signals";
-import { Micro } from "./time.ts";
+import { Milli } from "./time.ts";
 
 /** Maximum bytes of frames cached in a group before old frames are evicted from the front. */
 export const MAX_GROUP_CACHE_BYTES = 32 * 1024 * 1024;
@@ -10,13 +10,13 @@ export const MAX_GROUP_FRAMES = 1024;
 /**
  * A frame buffered in a {@link Group}: its presentation timestamp and payload bytes.
  *
- * The timestamp is in microseconds (the unit the hang container and WebCodecs use); the
- * wire layer converts it into the track's timescale. Frames written without an explicit
- * timestamp (catalogs, JSON state) are stamped with wall-clock now.
+ * The timestamp is in milliseconds; wall-clock accuracy doesn't warrant finer units for
+ * transport-level timing, and the wire layer converts it into the track's timescale.
+ * A future `Timestamp` type (mirroring Rust) will let a track pick its own scale.
  */
 export interface Frame {
-	/** Presentation timestamp in microseconds. */
-	timestamp: Micro;
+	/** Presentation timestamp in milliseconds. */
+	timestamp: Milli;
 	/** The frame payload. */
 	data: Uint8Array;
 }
@@ -77,11 +77,12 @@ export class Group {
 	}
 
 	/**
-	 * Writes a frame to the group at the given presentation timestamp (microseconds).
-	 * @param timestamp - Presentation time in microseconds
+	 * Writes a frame to the group.
 	 * @param data - The frame payload
+	 * @param timestamp - Presentation time in milliseconds; defaults to wall-clock now
+	 *   (`performance.now()`) for data with no presentation time of its own.
 	 */
-	writeFrame(timestamp: Micro, data: Uint8Array) {
+	writeFrame(data: Uint8Array, timestamp: Milli = Milli.now()) {
 		if (this.state.closed.peek()) throw new Error("group is closed");
 
 		this.#cacheBytes += data.byteLength;
@@ -104,17 +105,9 @@ export class Group {
 		if (this.#mirrors) {
 			for (const mirror of this.#mirrors) {
 				if (mirror.state.closed.peek()) this.#mirrors.delete(mirror);
-				else mirror.writeFrame(timestamp, data);
+				else mirror.writeFrame(data, timestamp);
 			}
 		}
-	}
-
-	/**
-	 * Writes a frame stamped with wall-clock now, for data with no presentation time of
-	 * its own (a JSON catalog, control state) or a source whose protocol can't carry one.
-	 */
-	writeFrameNow(data: Uint8Array) {
-		this.writeFrame(Micro.now(), data);
 	}
 
 	/**
@@ -126,7 +119,7 @@ export class Group {
 	 */
 	mirror(): Group {
 		const dst = new Group(this.sequence);
-		for (const frame of this.state.frames.peek()) dst.writeFrame(frame.timestamp, frame.data);
+		for (const frame of this.state.frames.peek()) dst.writeFrame(frame.data, frame.timestamp);
 		// Inherit the evicted prefix: frames dropped before this copy was made are a gap
 		// for its reader too, so reading them throws CacheFull.
 		dst.state.offset = this.state.offset;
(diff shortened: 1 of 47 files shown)


js/net/src/lite/publisher.ts
```diff
@@ -400,11 +411,23 @@ export class Publisher {
...
+			const timestamps = supportsTrackStream(this.version);
+			let prevTs = 0n;
+
 			try {
 				for (;;) {
 					const frame = await Promise.race([group.readFrame(), stream.closed]);
 					if (!frame) break;
 
+					if (timestamps) {
+						const ts = BigInt(Math.round(Milli.now()));
```
kixelated: I'm missing context, why are we not using the timestamp provided by the app?

kixelated: Good catch — it was a stopgap. The JS frame model stored bare `Uint8Array`s with no per-frame timestamp (hang embedded its timestamp *inside* the payload), so at the publisher there was no app-provided time to read, and wall-clock was just there to keep the mandatory lite-05 timestamp on the wire.

Fixed in df988c7: the model now carries a per-frame microsecond timestamp (`Group`/`TrackProducer` take `writeFrame(timestamp, data)`, with `writeFrameNow(data)` for genuinely timeless data like catalogs/JSON). The lite publisher advertises a microsecond timescale and delta-encodes each frame's real timestamp, and the subscriber decodes it (it was previously read and discarded). hang/loc/publish pass their µs timestamps through.

(Written by Claude)

---

js/net/src/lite/publisher.ts
```diff
@@ -374,8 +383,10 @@ export class Publisher {
 			return new TrackInfoMessage({
 				priority: info.priority,
 				ordered: info.ordered,
-				// This implementation doesn't produce per-frame timestamps yet.
-				timescale: 0,
+				// Lite05 mandates per-frame timestamps. Model frames carry a microsecond
+				// timestamp (the app's, or wall-clock for timeless data), so we advertise
+				// microseconds and emit each frame's own timestamp in `#runGroup`.
+				timescale: MICRO_TIMESCALE,
```
kixelated: This should be set by each track independently. It's a huge waste of bytes to use microseconds if we don't need it.

ex. anything using writeFrameNow should probably use milliseconds, because clock accuracy is shit anyway.

js/net/src/group.ts
```diff
@@ -118,26 +142,31 @@ export class Group {
...
+		this.writeFrameNow(new Uint8Array([bool ? 1 : 0]));
 	}
 
 	/**
-	 * Reads the next frame from the group.
-	 * @returns A promise that resolves to the next frame or undefined
+	 * Reads the next frame's payload from the group.
+	 * @returns A promise that resolves to the next payload or undefined
 	 */
 	async readFrame(): Promise<Uint8Array | undefined> {
```
kixelated: MAke this return `Frame` instead of a separate Timed method.

js/net/src/group.ts
```diff
@@ -62,22 +77,23 @@ export class Group {
...
 
 	/**
-	 * Writes a frame to the group.
-	 * @param frame - The frame to write
+	 * Writes a frame to the group at the given presentation timestamp (microseconds).
+	 * @param timestamp - Presentation time in microseconds
+	 * @param data - The frame payload
 	 */
-	writeFrame(frame: Uint8Array) {
+	writeFrame(timestamp: Micro, data: Uint8Array) {
```
kixelated: Can you put timestamp second and make it `?`. If undefined, we use `performance.now()`.

Or honestly just require it. have the caller do performance.now() if they don't care about timestamps. Milli not Micro.

But really we should have a proper Timestamp type like Rust. I think it would be a big improvement anyway, as type aliases are convient but very easy to get wrong.

kixelated: I mean I'm okay with writeFrameNow only if it's marked deprecated and we use it as a signal that we need to actually port everything over to mandatory timestamps (like the PR title btw).

rs/moq-net/src/model/track.rs
```diff
@@ -43,13 +43,12 @@ pub struct TrackInfo {
...
-	/// reported in TRACK_INFO and the publisher zigzag-delta encodes
-	/// per-frame timestamps at that scale on the wire; rejecting a frame at
-	/// the wrong scale prevents silent corruption.
-	#[cfg_attr(feature = "serde", serde(default, skip_serializing_if = "Option::is_none"))]
-	pub timescale: Option<Timescale>,
+	/// Every track is timed; this defaults to [`Timescale::MILLI`]. On Lite05+ it is
+	/// reported in TRACK_INFO and the publisher zigzag-delta encodes per-frame
+	/// timestamps at this scale on the wire. Protocols whose wire can't carry it
+	/// (pre-Lite05 moq-lite, IETF moq-transport) fall back to wall-clock milliseconds.
+	#[cfg_attr(feature = "serde", serde(default, skip_serializing_if = "is_default_timescale"))]
```
kixelated: remove `skip_serializing_if`

js/net/src/lite/publisher.ts
kixelated: Switched the publisher to advertise **milliseconds** (1000) instead of microseconds, and `writeFrame`'s default is now `performance.now()` in ms — so the byte-waste is gone and timeless data uses ms as you said.

True *per-track* timescale (different scales per track) I left for the follow-up: with the current branded-number approach the per-frame value can't carry its own scale, so per-track only works cleanly once we add the proper `Timestamp` type (value + scale). For now every JS track is ms. Flagged it in the commit + PR.

(Written by Claude)

---

js/net/src/group.ts
kixelated: Done in f024ebb: `writeFrame(data, timestamp?)` — timestamp second and optional, defaulting to `performance.now()`, in **Milli** (not Micro). I removed `writeFrameNow` entirely (the optional default replaces it), so the deprecation note above is moot.

Left the proper `Timestamp` type (value + scale, like Rust) as a follow-up per your call — agreed it's the real fix for the alias footgun and what unlocks per-track timescales.

(Written by Claude)

---

PR #1897, opened 2026-06-24
feat(json): group-scoped DEFLATE compression with browser support

kixelated:
Moves payload compression up to the **application layer** (`moq-json`) instead of the moq-lite wire (the alternative explored in #1874 / #1889). The relay stays media-agnostic, and `@moq/json` gets a matching, wire-interoperable implementation so a Rust producer and a browser consumer can talk.

## Approach: shared-window DEFLATE per group

Each group is compressed as **one raw DEFLATE ([RFC 1951](https://www.rfc-editor.org/rfc/rfc1951.html)) stream**, sync-flushed at each frame boundary. Every frame is a self-delimited slice, but later frames reuse the snapshot (and earlier deltas) as window context, so a delta restating known keys is almost all back-references. Measured on a realistic catalog (snapshot + 5 deltas): deltas drop from ~44 B (independent) to **~12–16 B** (shared window), ~38% smaller overall.

Each frame's wire slice is just the DEFLATE bytes with the fixed 4-byte sync-flush marker `00 00 ff ff` **stripped** (re-appended to decode), the [RFC 7692](https://www.rfc-editor.org/rfc/rfc7692.html#section-7.2.1) (permessage-deflate) trick, saving 4 bytes/frame. moq-net already frames each slice, so there's no length prefix; the decoder bounds its output as it inflates (in `onData` on the JS side, in the inflate loop on Rust) rather than from any declared size.

Compression is **off by default**; existing tracks (including the hang `catalog.json`) stay byte-identical on the wire.

### Why pako on the JS side (and the dead ends)

All verified empirically:

- The platform `CompressionStream("deflate-raw")` genuinely **can't flush mid-stream** (emits nothing until `close()`; its writer has no `flush()`), so it can't *encode* a per-frame shared-window stream.
- `fflate` *can* sync-flush, but its **encoder flush is buggy** on some inputs (a snapshot + 3 deltas case: even fflate's own decoder can't round-trip it). A regression test pins this; the bug is fixed upstream, so once released, swapping pako for the smaller fflate is a drop-in behind the same synchronous interface.
- **pako** encodes and decodes correctly and interoperates with `flate2` both directions.

pako is **synchronous**, so the whole codec is synchronous and pako is a normal `dependency` (~14 KB gzipped, pulled in only by `@moq/json` consumers).

## What changed

**Rust (`moq-json`)** — `compression.rs` is a per-group streaming `Encoder`/`Decoder` over `flate2`'s `Compress`/`Decompress` with `Z_SYNC_FLUSH`, raw DEFLATE at the default level, RFC 7692 tail strip/re-append, and a 64 MiB per-frame decompressed-size cap enforced as output is produced. The `Consumer` keeps a per-group decoder and replays a group's slices to rebuild a cloned consumer's window. The delta-vs-snapshot budget is measured on compressed slice sizes.

**JS (`@moq/json`)** — `compression.ts` is pako-based per-group `Encoder`/`Decoder` (plain synchronous constructors) with the same tail stripping. The encoder makes one copy into a tight owned buffer (pako backs each chunk with a ~16 KB buffer and `writeFrame` retains the reference, so returning a view would pin memory); the decoder feeds the slice + re-appended marker as two pako pushes, allocating no combined buffer. `Producer`/`Consumer` keep `compression?: boolean`; compressed writes happen inline.

## Public API

**`moq-json`**
- `ProducerConfig { delta_ratio: u32, compression: bool }` (`#[non_exhaustive]`, `Default`).
- `ConsumerConfig { compression: bool }` (`#[non_exhaustive]`, `Default`).
- `Producer::new(track, ProducerConfig)`; `Consumer::new(track, ConsumerConfig)`.
- `Error` gains `Decompress` and `TooLarge(u64)` (`Error` is `#[non_exhaustive]`).
- In-repo callers updated: `moq-mux` (catalog producer/consumer), `moq-boy`.

**`@moq/json`**
- `Config.compression?: boolean` (additive on/off toggle). Left as a single generic `Config<T>` for producer and consumer; mirroring the Rust `ProducerConfig`/`ConsumerConfig` split is deferred to a follow-up if wanted.
- `pako` added as a normal `dependency`.

## Branch targeting

Targets **`main`**: compression is off by default and unreleased, the catalog wire bytes are unchanged, and neither `moq-json` nor `js/json` is among the crates/packages the breaking-change rule routes to `dev`. The compressed-frame wire format and the config API changed across revisions of this same (unreleased) PR.

## Cross-package sync

`rs/moq-json` ↔ `js/json` are in sync on the wire format. `moq-json` is not on the wire-sync table and the catalog format is unchanged, so no `doc/` updates are required.

## Test plan

- [x] `cargo test -p moq-json` — 34 tests incl. group round-trip, cross-frame-context shrink, compressed snapshot/delta round-trips, late-joiner, mid-group cloned-consumer window rebuild, windowed-delta-beats-raw size check, a chunk-spanning large-frame round-trip, and codec garbage rejection.
- [x] `cargo clippy -p moq-json --tests` + `cargo fmt -p moq-json` — clean. `moq-mux` builds.
- [x] `bun test js/json` — 52 tests incl. group round-trip, cross-frame-context shrink, compressed snapshot/delta/late-joiner reconstruction, in-order live consumer, fresh-decoder-decodes-a-snapshot interop, inflate-past-cap rejection, a chunk-spanning large-frame round-trip, a size-shrink check, and a **pako-vs-fflate** test pinning why we use pako (with a positive control).
- [x] `tsc --noEmit` + `biome check` on `@moq/json` — clean. `@moq/hang` still type-checks.
- [x] CI (`just ci` via the pinned nix toolchain) — green.

## Wire-format interop verified

`flate2`(≈zlib) ↔ pako, both encode→decode directions, with RFC 7692 tail stripping, on small + large + the catalog cases.

(Written by Claude)

### DIFF
--- a/js/json/src/compression.ts
+++ b/js/json/src/compression.ts
@@ -15,13 +15,14 @@
  * Each slice is prefixed with its decompressed length as a QUIC varint (via `@moq/net`'s `Varint`),
  * so the decoder bounds the frame before inflating and matches the Rust producer on the wire.
  *
- * `pako` is an optional peer dependency loaded on demand, so consumers that never compress a track
- * never bundle it.
+ * pako is synchronous, so the whole codec is synchronous: only an enabled track pulls it in, but it
+ * is a normal dependency rather than a lazily loaded one.
  *
  * @module
  */
 
 import { Varint } from "@moq/net";
+import * as pako from "pako";
 
 // Maximum decompressed size of a single frame. A malicious publisher could otherwise send a tiny
 // slice that inflates hugely, so {@link Decoder.frame} stops rather than allocating without limit.
@@ -30,23 +31,6 @@ const MAX_DECOMPRESSED_FRAME = 64 * 1024 * 1024;
 
 // The trailing bytes of a DEFLATE sync flush, stripped on the wire and re-appended to decode.
 const SYNC_FLUSH_TAIL = new Uint8Array([0x00, 0x00, 0xff, 0xff]);
-
-type Pako = typeof import("pako");
-let pako: Promise<Pako> | undefined;
-
-// Load pako once, on demand. The optional peer dependency keeps it out of bundles that never
-// enable compression; a clear error points at the missing install if it's reached without it.
-function loadPako(): Promise<Pako> {
-	pako ??= import("pako")
-		.then((m) => (m as { default?: Pako }).default ?? (m as Pako))
-		.catch((err) => {
-			pako = undefined;
-			throw new Error("@moq/json compression requires the optional peer dependency `pako` to be installed", {
-				cause: err,
-			});
-		});
-	return pako;
-}
 
 // Concatenate chunks into one buffer (a single chunk passes through untouched).
 function concat(chunks: Uint8Array[], total: number): Uint8Array {
@@ -62,29 +46,22 @@ function concat(chunks: Uint8Array[], total: number): Uint8Array {
 
 /**
  * Encodes a group's frame payloads into one shared DEFLATE stream, one self-delimited slice per
- * frame. Hold one per group; create a new one at each group boundary. Build with {@link create}.
+ * frame. Hold one per group; create a new one at each group boundary.
  *
  * @public
  */
 export class Encoder {
-	#deflate: import("pako").Deflate;
-	#flush: number;
+	#deflate = new pako.Deflate({ raw: true });
 	#chunks: Uint8Array[] = [];
 	#total = 0;
 
-	private constructor(lib: Pako) {
-		this.#deflate = new lib.Deflate({ raw: true });
-		this.#flush = lib.constants.Z_SYNC_FLUSH;
+	/** Start a fresh per-group encoder with a cold window. */
+	constructor() {
 		this.#deflate.onData = (chunk) => {
 			const bytes = chunk as Uint8Array;
 			this.#chunks.push(bytes);
 			this.#total += bytes.length;
 		};
-	}
-
-	/** Start a fresh per-group encoder with a cold window. */
-	static async create(): Promise<Encoder> {
-		return new Encoder(await loadPako());
 	}
 
 	/**
@@ -96,7 +73,7 @@ export class Encoder {
 		if (payload.length === 0) return payload;
 		this.#chunks = [];
 		this.#total = 0;
-		this.#deflate.push(payload, this.#flush);
+		this.#deflate.push(payload, pako.constants.Z_SYNC_FLUSH);
 		const full = concat(this.#chunks, this.#total);
 		// Drop the trailing sync-flush marker (the decoder re-appends it) and prefix the length.
 		const deflate = full.subarray(0, full.length - SYNC_FLUSH_TAIL.length);
@@ -110,20 +87,18 @@ export class Encoder {
 
 /**
  * Decodes a group's frame slices back into the original payloads. Hold one per group; feed slices
- * in frame order (each frame builds on the earlier ones). Build with {@link create}.
+ * in frame order (each frame builds on the earlier ones).
  *
  * @public
  */
 export class Decoder {
-	#inflate: import("pako").Inflate;
-	#flush: number;
+	#inflate = new pako.Inflate({ raw: true });
 	#chunks: Uint8Array[] = [];
 	#total = 0;
 	#tooLarge = false;
 
-	private constructor(lib: Pako) {
-		this.#inflate = new lib.Inflate({ raw: true });
-		this.#flush = lib.constants.Z_SYNC_FLUSH;
+	/** Start a fresh per-group decoder with a cold window. */
+	constructor() {
 		this.#inflate.onData = (chunk) => {
 			const bytes = chunk as Uint8Array;
 			this.#total += bytes.length;
@@ -134,11 +109,6 @@ export class Decoder {
 			}
 			this.#chunks.push(bytes);
 		};
-	}
-
-	/** Start a fresh per-group decoder with a cold window. */
-	static async create(): Promise<Decoder> {
-		return new Decoder(await loadPako());
 	}
 
 	/**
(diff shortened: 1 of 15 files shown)


js/json/src/compression.ts
```diff
@@ -0,0 +1,178 @@
...
+			pako = undefined;
+			throw new Error("@moq/json compression requires the optional peer dependency `pako` to be installed", {
+				cause: err,
+			});
+		});
+	return pako;
+}
+
+// Concatenate chunks into one buffer (a single chunk passes through untouched).
+function concat(chunks: Uint8Array[], total: number): Uint8Array {
```
kixelated, line 34: Is there no way to write to a buffer directly to avoid these copies?

js/json/src/compression.ts
```diff
@@ -0,0 +1,178 @@
...
+export class Encoder {
+	#deflate: import("pako").Deflate;
+	#flush: number;
+	#chunks: Uint8Array[] = [];
+	#total = 0;
+
+	private constructor(lib: Pako) {
+		this.#deflate = new lib.Deflate({ raw: true });
+		this.#flush = lib.constants.Z_SYNC_FLUSH;
+		this.#deflate.onData = (chunk) => {
```
kixelated, line 58: Could we copy to the destination directly instead of pushing to chunks?

Any way to have deflate reuse a buffer or write directly?

js/json/src/compression.ts
```diff
@@ -0,0 +1,178 @@
...
+	 */
+	frame(payload: Uint8Array): Uint8Array {
+		if (payload.length === 0) return payload;
+		this.#chunks = [];
+		this.#total = 0;
+		this.#deflate.push(payload, this.#flush);
+		const full = concat(this.#chunks, this.#total);
+		// Drop the trailing sync-flush marker (the decoder re-appends it) and prefix the length.
+		const deflate = full.subarray(0, full.length - SYNC_FLUSH_TAIL.length);
+		const prefix = Varint.encode(payload.length);
```
kixelated: Do we still need the varint at the front? It's kind of annoying because it requires reallocating, and adds  few bytes of course.

At the very least it should be appended to the end (since we're using a framed transport), but that would require some new varint encoding.

js/json/src/compression.ts
```diff
@@ -0,0 +1,178 @@
...
+	 * Decompress the next frame's `slice` back into its payload. Empty in yields empty out. Throws
+	 * if the input is malformed, declares more than the per-frame size limit, or inflates to a length
+	 * that doesn't match its prefix.
+	 */
+	frame(slice: Uint8Array): Uint8Array {
+		if (slice.length === 0) return slice;
+
+		// The decompressed-length prefix bounds the frame before any inflation.
+		const [declared, deflate] = Varint.decode(slice);
+		if (declared > MAX_DECOMPRESSED_FRAME) {
```
kixelated: We need to enforce this within `onData`

rs/moq-json/src/compression.rs
```diff
@@ -0,0 +1,356 @@
...
+/// [`Consumer`](crate::Consumer) is told via [`Consumer::with_compression`](crate::Consumer::with_compression)
+/// only *that* a track is compressed, not how.
+#[derive(Debug, Clone, Default)]
+#[non_exhaustive]
+pub struct Compression {
+	/// DEFLATE [`Level`]. Higher is smaller but slower. Defaults to `6`.
+	///
+	/// This is a sender-only choice and need not match the consumer (the wire format is the same
+	/// at any level). Browser producers can't set it; the platform deflate picks its own level.
+	pub level: Level,
```
kixelated: IDK just use the default level for now. Make compression a bool.

rs/moq-json/src/lib.rs
```diff
@@ -53,24 +66,42 @@ impl From<serde_json::Error> for Error {
 pub type Result<T> = std::result::Result<T, Error>;
 
 /// Configuration for a [`Producer`].
+///
+/// Build from [`Default`] and override fields (the struct is `#[non_exhaustive]`, so new
+/// options stay additive): `let mut config = Config::default(); config.delta_ratio = 0;`.
 #[derive(Debug, Clone)]
+#[non_exhaustive]
 pub struct Config {
```
kixelated: This should be ProducerConfig.

rs/moq-json/src/lib.rs
```diff
@@ -321,11 +403,27 @@ impl<T> Clone for Consumer<T> {
...
 	pub fn new(track: moq_net::TrackConsumer) -> Self {
+		Self::build(track, false)
+	}
+
+	/// Create a consumer that decompresses DEFLATE frames written by a producer with
+	/// [`Config::compression`] set.
+	///
+	/// Decompression is self-describing, so no settings are needed beyond knowing the track is
+	/// compressed (the producer's level does not have to be matched).
+	pub fn with_compression(track: moq_net::TrackConsumer) -> Self {
```
kixelated: We should have DecoderConfig instead. Compression is a boolean.

kixelated:
Pushed `cc1c67ff` addressing the review. Two of your comments were questions worth answering directly:

**On copying / "have deflate reuse a buffer or write directly?"** pako's streaming API only hands output to `onData` as chunks; there's no "write into my buffer" option. I tested it: pako backs *each* frame's output with its own ~16 KB buffer (`chunkSize`), and `group.writeFrame` retains the reference without copying. So returning pako's view would pin ~16 KB per small retained frame, which is worse than copying. The encoder therefore makes **one** copy into a tight owned buffer. What I *did* remove is the second copy (the varint prefix-prepend, by dropping the prefix) and the decoder's combined-input allocation (it now feeds the slice + re-appended marker as two pako pushes).

**Dropped the front varint** entirely (both sides). moq-net already frames each slice, so it wasn't needed for delimiting, and the size cap is now enforced as output is produced (in `onData` on the JS side, in the inflate loop on the Rust side) rather than from a declared length.

**Compression is a `bool`** at the default level now; the `Compression`/`Level` types are gone.

**Config split:** per your pick I split Rust into `ProducerConfig { delta_ratio, compression }` and `ConsumerConfig { compression }` (`Consumer::new` takes `ConsumerConfig`, replacing `with_compression`). One open question: `@moq/json` still uses a single generic `Config<T>` for both the producer and consumer (it also carries the zod `schema` and `initial` value). Want me to mirror the split there as `ProducerConfig<T>` / `ConsumerConfig<T>`, or leave the JS side unified?

(Written by Claude)

---

PR #1901, opened 2026-06-24
feat(moq-relay): reuse client TLS for outbound auth HTTP; make --client-tls-* flags consistent

kixelated:
## Summary

Two related changes to the relay's client TLS story.

### 1. Reuse `--client-tls-*` for outbound auth HTTP, deprecate `--auth-tls-*`

The relay's outbound auth HTTP client (JWK fetches and the `--auth-api` / `--auth-public-api` / `--auth-key-dir` URL endpoints) now reuses the cluster client's `--client-tls-*` configuration. A single client identity covers both relay-to-relay dials and relay-to-auth-API requests, so an auth API can require mTLS and recognize the relay by the same certificate it uses for clustering.

- **Default**: the auth client builds from `--client-tls-*` (`[client.tls]`). With neither group configured, this is identical to the old default (system roots, no client identity), so standalone JWT relays are unaffected.
- **Backwards compatibility**: `--auth-tls-*` still works and overrides `--client-tls-*` when set, logging a deprecation warning. Hidden from `--help` (`hide = true` + `#[doc(hidden)]`).
- The cluster client TLS is injected into `AuthConfig` via `init(&client_tls)`, so `Auth::new(config)` (and its ~40 tests) stays unchanged.

### 2. Make the client TLS flag names consistent

`--client-tls-cert` / `--client-tls-key` carried the `client-` prefix (to avoid colliding with the server's `--tls-cert` / `--tls-key`), while `--tls-root`, `--tls-system-roots`, `--tls-fingerprint`, and `--tls-disable-verify` were bare. A user who learned `--client-tls-cert` would reasonably guess `--client-tls-root` and hit a parse error.

The four bare flags are renamed to `--client-tls-*`. The old spellings live in a private, hidden, flattened `Deprecated` struct (kept off the public `Client` surface and out of TOML): they still parse, are hidden from `--help`, and **log a deprecation warning on use** pointing at the canonical name. `build()` (every backend) and `quiche` read `effective_*()` accessors that fold the deprecated value into the canonical one. Canonical wins for scalar options; vecs concatenate. Env vars (`MOQ_CLIENT_TLS_*`) and TOML field names are unchanged.

| Field | Canonical flag | Deprecated (hidden, warns) |
|---|---|---|
| root | `--client-tls-root` | `--tls-root` |
| system_roots | `--client-tls-system-roots` | `--tls-system-roots` |
| fingerprint | `--client-tls-fingerprint` | `--tls-fingerprint` |
| disable_verify | `--client-tls-disable-verify` | `--tls-disable-verify` |
| cert / key | `--client-tls-cert` / `--client-tls-key` | (already canonical) |

Out of scope: the server keeps its own asymmetry (`--tls-cert` / `--server-tls-root`).

## Deprecation policy

Per review, deprecated surface isn't documented: removed the `--auth-tls-*` and deprecated-endpoint mentions from `doc/bin/relay/auth.md`, dropped the alias notes from the field docs, and added a `## Deprecation` convention to `CLAUDE.md` (document only the canonical surface; keep deprecated paths working but invisible; a runtime warning on *use* is fine, advertising the dead name is not).

## Public API changes

- `moq-relay`: `AuthConfig::init(self)` → `init(self, client_tls: &moq_native::tls::Client)`. Breaking for direct lib consumers, but `moq-relay` is a binary crate and not in the `dev`-targeted breaking list, so this targets `main`. `AuthTls` marked `#[doc(hidden)]`.
- `moq-native`: client TLS flag long-names changed (CLI surface, not Rust API; the `pub` struct fields are unchanged), old spellings preserved as hidden warning flags.

## Test plan

- [x] `cargo test -p moq-native --lib` (43), incl. fold + canonical-precedence tests for the deprecated flags
- [x] `cargo test -p moq-relay --lib` (125), incl. mTLS-via-`--client-tls-*` path
- [x] `cargo clippy` clean on both (incl. `--features quiche`)
- [x] `moq-relay --help` shows only `--client-tls-*`; bare `--tls-*` and `--auth-tls-*` hidden but still parse and warn on use
- [ ] `cargo test -p moq-relay --test smoke` — can't bind `[::]:0` in the sandbox (no IPv6), unrelated

(Written by Claude)


rs/moq-native/src/tls.rs
```diff
@@ -108,23 +108,33 @@ pub struct Client {
...
+	///
+	/// The bare `--tls-root` form is a deprecated alias and will be removed.
 	#[serde(skip_serializing_if = "Vec::is_empty")]
-	#[arg(id = "tls-root", long = "tls-root", env = "MOQ_CLIENT_TLS_ROOT")]
+	#[arg(
+		id = "client-tls-root",
+		long = "client-tls-root",
+		alias = "tls-root",
+		env = "MOQ_CLIENT_TLS_ROOT"
+	)]
```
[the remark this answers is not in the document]
kixelated: This is a false positive: the env var names did not change. They were already `MOQ_CLIENT_TLS_*` before this PR, and only the CLI `long` names changed (with `--tls-*` kept as deprecated aliases).

The diff lines that look like a rename are just the `env = "MOQ_CLIENT_TLS_ROOT"` / `env = "MOQ_CLIENT_TLS_FINGERPRINT"` moving onto their own line as the `#[arg(...)]` became multiline. Both sides of the diff read `MOQ_CLIENT_TLS_*`:

```
-	#[arg(id = "tls-root", long = "tls-root", env = "MOQ_CLIENT_TLS_ROOT")]
+		env = "MOQ_CLIENT_TLS_ROOT"
```

`MOQ_TLS_ROOT` / `MOQ_TLS_SYSTEM_ROOTS` / `MOQ_TLS_FINGERPRINT` / `MOQ_TLS_DISABLE_VERIFY` (without `CLIENT`) have never existed in the repo, so there's no env-var compatibility to preserve. The analysis even shows `rg 'MOQ_TLS_'` returning nothing. No change needed.

(Written by Claude)

---

doc/bin/relay/auth.md
```diff
@@ -171,6 +171,19 @@ auth_api = "https://api.moq.dev/cluster/auth"
...
 
+### Authenticating the relay to the auth API
+
+The outbound HTTP the relay makes to `--auth-api` (and the deprecated `--auth-key-dir` URL and `--auth-public-api` endpoints) reuses the cluster client's TLS configuration. The same `--client-tls-cert` / `--client-tls-key` the relay presents when dialing cluster peers also identifies it to the auth API, and `--client-tls-root` trusts a private CA on the endpoint (env `MOQ_CLIENT_TLS_*`, or `[client.tls]` in TOML). So an auth API can require mTLS and recognize the relay by the same certificate it uses for clustering.
```
kixelated: Don't document the deprecated stuff.

In fact add a note to claude.md about this.

rs/moq-native/src/tls.rs
```diff
@@ -108,23 +108,33 @@ pub struct Client {
...
+	/// and system roots disabled) is rejected, since verification could never pass.
+	///
+	/// The bare `--tls-system-roots` form is a deprecated alias and will be removed.
 	#[serde(skip_serializing_if = "Option::is_none")]
 	#[arg(
-		id = "tls-system-roots",
-		long = "tls-system-roots",
+		id = "client-tls-system-roots",
+		long = "client-tls-system-roots",
+		alias = "tls-system-roots",
```
kixelated: Can we make a separate arg so we can print a deprecation notice on the old one? In general is this a good idea?

doc/bin/relay/auth.md
kixelated: Done in 9d4cf20. Removed the `--auth-tls-*` paragraph (and the deprecated-endpoint enumeration) from `auth.md`, dropped the "deprecated alias" notes from the `moq-native` client TLS field docs, and hid the `--auth-tls-*` flags from `--help` (`hide = true` + `#[doc(hidden)]`). They still parse and still warn on use; they're just not advertised anywhere.

Added a `## Deprecation` section to `CLAUDE.md`: document only the current/canonical surface; keep deprecated paths working but invisible (hidden aliases / `#[doc(hidden)]`), with the rationale living in the commit/PR rather than docs that rot. A runtime warning on *use* is still fine since it fires on use rather than advertising the dead name.

(Written by Claude)

---

rs/moq-native/src/tls.rs
kixelated: The pattern is good in general, but I'd recommend against it here.

clap can't tell you which spelling matched a single arg's aliases, so a per-flag deprecation notice requires a *separate* `#[arg(hide = true)]` field plus merge logic. That's exactly what `--auth-tls-*` already does (separate fields, warns on use). It works there because `AuthTls` is funneled through one `to_client_tls()`.

For these `--tls-*` flags it's a lot more invasive: `root` / `system_roots` / `fingerprint` / `disable_verify` are read directly in ~8 places (`quinn.rs`, `quiche.rs`, `noq.rs`, `server.rs`, `build()`, `moq-ffi`, `moq-gst`, relay auth), so a deprecated twin field would have to be merged before every one of those reads, and it adds 4 mirror fields to the public `Client` struct that consumers build programmatically. High cost for a cosmetic rename of low-traffic flags. It also cuts against the "don't document deprecated stuff" rule we just added.

So my call: keep the bare forms as silent hidden aliases (zero clutter, old scripts keep working, new docs only show `--client-tls-*`). If we actually want to *force* migration, the cleaner lever is a hard rename, drop the alias so clap errors with a "did you mean --client-tls-root" rather than carrying twin fields. Happy to do that instead if you'd prefer the push over the soft landing.

(Written by Claude)

---

kixelated: Yes, that works well, done in cf575ee.

A private, non-pub `Deprecated` struct holds the bare `--tls-*` spellings and is `#[command(flatten)]`'d into `tls::Client` with the field itself private + `#[serde(skip)]`. So the deprecated flags keep parsing but stay entirely off the public `Client` surface (consumers still build via the canonical `pub` fields) and out of TOML.

I also have to walk back my "~8 readers" claim from before — I'd miscounted: the `quinn.rs`/`noq.rs`/`server.rs` `tls.root` reads are the *server* struct (`--server-tls-root`), not these client flags. The client flags are read in exactly two spots: `Client::build()` and `quiche.rs`. So the merge is small: `build()` (which runs for every backend) calls `warn_deprecated()` once and reads `effective_*()` accessors that fold the alias into the canonical value; `quiche` (which bypasses `build()`) reads the same accessors. Canonical wins for scalar options, vecs concatenate.

Net behavior: `--tls-root` etc. still parse, are hidden from `--help`, and now log `--tls-root is deprecated; use --client-tls-root` on use. Added tests for the fold and the canonical-wins precedence.

(Written by Claude)

---

PR #1906, opened 2026-06-24
feat(moq-net): generalize stats Tier to an arbitrary label

kixelated:
## Summary

Today the relay's stats are split into exactly two hardcoded tiers (`Tier::External` / `Tier::Internal`), with `internal/` baked into track names and an `internal: bool` driving the choice. This replaces that with an **arbitrary tier label** chosen by business logic, so a relay can bucket traffic into `internal`, `local`, `region/sjc`, or anything else.

- `moq_net::Tier` becomes a `Tier(PathOwned)` newtype wrapping a label instead of a 2-variant enum.
  - The **default tier** (`Tier::default()`) is unprefixed: `publisher.json`, `subscriber.json`, `sessions.json` (unchanged names, so existing consumers keep working).
  - A **named tier** prefixes every track with its label: `Tier::new("internal")` → `internal/publisher.json`, `Tier::new("region/sjc")` → `region/sjc/publisher.json`.
- **Fully dynamic**: per-broadcast counters and the published JSON tracks are created lazily the first tick traffic routes to a label. The default-tier tracks are pre-created so they always exist and emit `{}` while idle. The orphan-safe GC and the `closed`-before-`open` snapshot atomicity are preserved.
- **Relay wiring**: the unified auth API's response gains a `tier: String` (an arbitrary label) field. The legacy `internal: bool` field is still accepted for backward compatibility (`true` → `internal`, `false` → default tier) and is ignored when `tier` is present. `AuthToken.internal` becomes `AuthToken.tier: moq_net::Tier`. The API can return any label (or `""` for the default tier).
- **No hardcoded `internal`** (review feedback): each trusted-connection source gets its own tier flag, defaulting to `internal`:
  - `--cluster-tier` — relay-to-relay (cluster) dials
  - `--internal-tier` — the `[internal]` unauthenticated listener
  - `--auth-mtls-tier` — mTLS peers, when the auth API doesn't return a `tier`
  - Set any to `""` to record on the default unprefixed tier. `AuthToken::unrestricted()` no longer bakes the tier; the caller sets it.

So an auth service can reply `{"tier": "region"}` and that session's stats land on `region/publisher.json`.

## Public API changes

`rs/moq-net` (library):
- **BREAKING** `Tier` is no longer an enum. Removed `Tier::External`/`Tier::Internal` and the private `idx()`. Added `Tier::new(impl Into)`, `Tier::label() -> &PathOwned`, `Tier::is_default() -> bool`, `Default`, `Display` (default tier prints `external` in logs).
- `Stats::tier`, `StatsHandle`, `BroadcastStats`, and the RAII guards keep their method signatures; `StatsHandle::tier()` now returns `&Tier` instead of a `Copy` value.

`rs/moq-relay`:
- The `--auth-api` JSON response gains a `tier` (string label) field. The legacy `internal` (bool) field is **still accepted** when `tier` is absent, so this is **not** a breaking HTTP-contract change. New `AuthApiResponse::tier()` resolves the effective tier, preferring `tier`.
- `AuthToken.internal: bool` → `AuthToken.tier: moq_net::Tier` (`#[non_exhaustive]`).
- New config flags `--cluster-tier`, `--internal-tier`, `--auth-mtls-tier` (all `Option`, default `internal`). New `pub fn Auth::with_mtls_tier`.
- `Auth::resolve_mtls` now returns `(String, Tier)` (`pub(crate)`).

Targeting `dev` per the branch-targeting rules (breaking `moq-net` API). Rebased onto current `dev` (which carries the broadcast/track API refactor).

## Cross-package sync

- `rs/moq-net` stats are a relay feature, not part of `js/net`, so the only JS touched is the demo dashboard (comment-only; track names unchanged).
- `doc/bin/relay/auth.md` + `config.md`: the auth-API `tier` field (and the retained legacy `internal` bool), the new per-site tier flags, and the stats track table (`/...` placeholder).

## Test plan

- [x] `cargo test -p moq-net` — 21 stats tests (dynamic tiers, lazy track creation, idle `{}`, per-tier isolation, GC)
- [x] `cargo test -p moq-relay --lib` — auth tier-resolution, legacy `internal`-bool backward-compat, `tier`-precedence, and the `cli_does_not_clobber_toml_tiers` merge guard
- [x] `cargo clippy -p moq-net -p moq-relay --all-targets` + `cargo fmt --check` clean
- [ ] `--test smoke` integration tests can't run in the sandbox (QUIC socket binds blocked); unaffected by this change

## Deferred

- Renaming / moving the `--internal-*` listener flags under `[server]` is left for a separate PR (`[server]` is `moq_native::ServerConfig`, a shared crate), to keep that config churn out of the tier change. Direction for that follow-up: make transport security legible in the flag names. `--internal-*` does not convey that the listener is unencrypted/unverified (no TLS), so a plaintext-TCP listener wants a clearly-scary name (`--server-plaintext` / `--server-danger`) while the UDS listener (filesystem/peer-cred protected) can be `--server-uds` / `--server-local`; `--cluster-connect` should likewise be explicit about tcp/uds vs TLS.

(Written by Claude)


doc/bin/relay/config.md
```diff
@@ -212,17 +212,25 @@ interval = 1
...
-| `internal/publisher.json`   | internal (e.g. mTLS cluster peer) egress    |
-| `internal/subscriber.json`  | internal ingress                            |
-| `sessions.json`             | external connected sessions, keyed by root  |
-| `internal/sessions.json`    | internal connected sessions, keyed by root  |
+| `publisher.json`            | default-tier egress                         |
+| `subscriber.json`           | default-tier ingress                        |
+| `<label>/publisher.json`    | named-tier egress (e.g. `internal/publisher.json`) |
+| `<label>/subscriber.json`   | named-tier ingress                          |
+| `sessions.json`             | default-tier connected sessions, keyed by root |
+| `<label>/sessions.json`     | named-tier connected sessions, keyed by root |
```
kixelated: Why `<label>` and not `<tier>`?

rs/moq-relay/src/internal.rs
```diff
@@ -213,7 +213,7 @@ where
 	let token = AuthToken::unrestricted(moq_net::Path::new("").to_owned());
 	let publish = cluster.publisher(&token);
 	let subscribe = cluster.subscriber(&token);
-	let stats = cluster.stats.tier(moq_net::Tier::Internal);
+	let stats = cluster.stats.tier(moq_net::Tier::new("internal"));
```
kixelated: Could you make this a config arg?

rs/moq-relay/src/web.rs
```diff
@@ -517,9 +517,9 @@ async fn serve_announced(
 	};
 	let token = if mtls.is_some() {
 		// mTLS peers: the API returns the canonical root and the billing tier.
-		let (root, internal) = state.auth.resolve_mtls(&params.path).await?;
+		let (root, tier) = state.auth.resolve_mtls(&params.path).await?;
 		let mut token = AuthToken::unrestricted(moq_net::Path::new(&root).to_owned());
-		token.internal = internal;
+		token.tier = tier;
```
kixelated, line 522: Can this be a config flag?

kixelated:
Addressed in f3480f2.

Rather than hard-coding `internal`, each trusted-connection source now has its own tier flag, each defaulting to `internal` (set to `""` for the default unprefixed tier):

- `--internal-tier` — the `[internal]` listener (your `internal.rs` comment)
- `--cluster-tier` — relay-to-relay (cluster) dials
- `--auth-mtls-tier` — mTLS peers, used when the auth API doesn't return a `tier` (your `web.rs` comment); the API's `tier` field still overrides per-connection

`AuthToken::unrestricted()` no longer bakes the tier — the caller sets it from config. Also renamed the doc placeholder `<label>` → `<tier>` (your `config.md` comment).

On the broader note: I left the `--internal-*` listener flags as-is and did **not** move them under `[server]` (note `[server]` is `moq_native::ServerConfig`, a shared crate, so a literal move means adding relay-only fields there). Happy to do that rename/regroup as its own PR so the config churn doesn't tangle with the tier change — let me know.

(Written by Claude)

---

PR #1910, opened 2026-06-25
feat(moq-net): two-phase accept exposing the SETUP request path (lite-05 + moq-transport 14-18)

kixelated:
## Summary

A server can learn the client's request path on transports that have no request URI (raw QUIC, qmux over TCP/TLS/UDS), by reading it from the MoQ SETUP before deciding what to serve.

- **moq-net** — `Server::accept_request` returns a `Request` handle. The handshake pauses after the client's SETUP is read; you inspect `Request::path()`, set the origins to serve, then `ok()` (or `close()`). `accept()` is a thin `accept_request().ok()` wrapper, so existing callers are untouched.
- **moq-native** — `tcp://` advertises `url.path()`; `unix://` advertises the `?path=` query (the socket stays in the URL path), threaded through `moq_net::Client::with_path`.
- **moq-relay** — the internal tcp/uds listener reads `request.path()` and scopes the trusted grant's root to it. No path still means the empty root (everything), so behavior is unchanged when no path is sent.

The path is read across **moq-lite-05 and every moq-transport draft (14-18)**:
- lite-05 and moq-transport 17/18 carry it on a unidirectional Setup Stream — `accept_setup` blocks on that stream and `STOP_SENDING`s any uni stream that races ahead (group data needs a prior subscribe, so nothing legitimate precedes the SETUP), then hands the stream back to `start` so it keeps serving (PROBE for lite, GOAWAY for IETF).
- moq-transport 14-16 carry it as an in-band parameter on the bidi SETUP, read synchronously.

We bite the bullet on blocking the lite-05 / IETF-17-18 accept on the SETUP; in the common flow it's in the client's first post-handshake flight, so it isn't a full extra round-trip.

## Why no `Session::path()`

An established `Session` deliberately has no `path()`. It could only ever return the in-band SETUP path, so it'd be `None` for WebTransport/WebSocket (whose path is in the connection URL) — a confusing half-truth. The path is surfaced where you act on it: the `accept_request` -> `Request::path()` flow.

## Public API changes

All additive; no breaking signature changes. `mod lite` / `mod ietf` are private, so their `accept_setup` / `start` / `Setup` items are crate-internal, not public surface.

**`moq-net` (`pub use server::*`)**
- **New** `Server::accept_request(&self, session) -> Result<Request<S>, Error>`.
- **New** `pub struct Request<S: web_transport_trait::Session>` with `path()`, `with_publisher()`, `with_subscriber()`, `with_stats()`, `async ok()`, `close()`.
- **Behavior change (no signature change)** `Server::accept`: now `accept_request().ok()`, and on lite-05 / moq-transport-17-18 it blocks on the client SETUP before returning.

**`moq-native`** / **`moq-relay`** — none public.

Targets **dev** (SETUP/behavior change).

## Test plan
- [x] `cargo test -p moq-net -p moq-native -p moq-relay`
  - moq-net: `accept_request_reads_lite05_path`, `..._without_path_is_none`, `accept_request_skips_uni_stream_before_setup`
  - moq-relay smoke: `internal_tcp_path_reaches_server`, `internal_unix_path_reaches_server` — a client pins each of **moq-lite-05 and moq-transport 14/15/16/17/18**, publishes under `/room`, and a no-path subscriber sees `room/test`, proving the path reaches the server on every version over both qmux transports.
- [x] `cargo build --workspace`
- [x] `cargo clippy` clean
- [x] `cargo fmt` via nix
- [ ] e2e against a live relay over `tcp://` / `unix://`

## Follow-ups
- `doc/concept` note on the SETUP path.

(Written by Claude)

### DIFF
--- a/rs/moq-net/src/server.rs
+++ b/rs/moq-net/src/server.rs
@@ -65,50 +65,49 @@ impl Server {
 		self
 	}
 
-	/// Perform the MoQ handshake as a server for the given session.
+	/// Perform the MoQ handshake as a server, returning the established [`Session`].
+	///
+	/// Convenience wrapper over [`accept_request`](Self::accept_request) that
+	/// completes the handshake immediately. Use `accept_request` when you need to
+	/// inspect the client's advertised path before deciding what to serve.
 	pub async fn accept<S: web_transport_trait::Session>(&self, session: S) -> Result<Session, Error> {
+		self.accept_request(session).await?.ok().await
+	}
+
+	/// Two-phase accept: negotiate the version and, for lite-05, read the client's
+	/// SETUP so [`Request::path`] is available before the caller commits to
+	/// [`Request::ok`] or [`Request::close`]. Mirrors the WebTransport-level
+	/// `Request` in moq-native.
+	///
+	/// Every regime defers its session start to `ok()`, so origins set on the
+	/// Request (after inspecting the path) always take effect. The path is surfaced
+	/// for lite-05, whose Setup Stream carries it; other versions report `None`.
+	pub async fn accept_request<S: web_transport_trait::Session>(&self, session: S) -> Result<Request<S>, Error> {
+		// Regimes without a path to read defer to `ok()` without surfacing one.
+		let deferred = |handshake| Request {
+			server: self.clone(),
+			path: None,
+			handshake,
+		};
+
 		let (encoding, supported) = match session.protocol() {
 			Some(ALPN_18) => {
-				let v = self
-					.versions
+				self.versions
 					.select(Version::Ietf(ietf::Version::Draft18))
 					.ok_or(Error::Version)?;
-
-				// Draft-17+: SETUP is exchanged in the background by the session.
-				ietf::start(
-					session.clone(),
-					None,
-					None,
-					false,
-					self.publish.clone(),
-					self.subscribe.clone(),
-					self.stats.clone(),
-					ietf::Version::Draft18,
-				)?;
-
-				tracing::debug!(version = ?v, "connected");
-				return Ok(Session::new(session, v, None));
+				return Ok(deferred(Handshake::IetfModern {
+					session,
+					version: ietf::Version::Draft18,
+				}));
 			}
 			Some(ALPN_17) => {
-				let v = self
-					.versions
+				self.versions
 					.select(Version::Ietf(ietf::Version::Draft17))
 					.ok_or(Error::Version)?;
-
-				// Draft-17+: SETUP is exchanged in the background by the session.
-				ietf::start(
-					session.clone(),
-					None,
-					None,
-					false,
-					self.publish.clone(),
-					self.subscribe.clone(),
-					self.stats.clone(),
-					ietf::Version::Draft17,
-				)?;
-
-				tracing::debug!(version = ?v, "connected");
-				return Ok(Session::new(session, v, None));
+				return Ok(deferred(Handshake::IetfModern {
+					session,
+					version: ietf::Version::Draft17,
+				}));
 			}
 			Some(ALPN_16) => {
 				let v = self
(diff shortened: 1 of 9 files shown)


rs/moq-net/src/server.rs
```diff
@@ -65,50 +65,49 @@ impl Server {
...
+	/// Perform the MoQ handshake as a server, returning the established [`Session`].
+	///
+	/// Convenience wrapper over [`accept_request`](Self::accept_request) that
+	/// completes the handshake immediately. Use `accept_request` when you need to
+	/// inspect the client's advertised path before deciding what to serve.
 	pub async fn accept<S: web_transport_trait::Session>(&self, session: S) -> Result<Session, Error> {
+		self.accept_request(session).await?.ok().await
+	}
+
+	/// Two-phase accept: negotiate the version and, for lite-05, read the client's
```
kixelated: not just lite-05+. Also moq-transport. Just remove the mention of the version stuff.

rs/moq-net/src/server.rs
```diff
@@ -65,50 +65,49 @@ impl Server {
...
+	/// SETUP so [`Request::path`] is available before the caller commits to
+	/// [`Request::ok`] or [`Request::close`]. Mirrors the WebTransport-level
+	/// `Request` in moq-native.
```
kixelated: this documentation kinda sucks.

rs/moq-net/src/server.rs
```diff
@@ -197,18 +170,161 @@ impl Server {
...
+pub struct Request<S: web_transport_trait::Session> {
+	server: Server,
+	path: Option<String>,
+	handshake: Handshake<S>,
+}
+
+/// The handshake state captured at the pause point. Every variant defers its
+/// session start to [`Request::ok`] so origins set on the Request still apply.
+enum Handshake<S: web_transport_trait::Session> {
+	/// Modern IETF (17/18): no readable client SETUP (exchanged in the background).
```
kixelated: We have to fix this. The path is in SETUP for moq-transport too.

rs/moq-net/src/server.rs
```diff
@@ -197,18 +170,161 @@ impl Server {
...
+		self
+	}
+
+	/// Accept the session, completing the handshake.
+	pub async fn ok(self) -> Result<Session, Error> {
+		let server = self.server;
+
+		let (session, mut stream, version, request_id_max) = match self.handshake {
+			Handshake::IetfModern { session, version } => {
+				// Draft-17+: SETUP is exchanged in the background by the session.
```
kixelated: change it

rs/moq-net/src/session.rs
```diff
@@ -81,17 +82,36 @@ impl Session {
...
+	}
+
 	/// Returns the negotiated protocol version.
 	pub fn version(&self) -> Version {
 		self.version
 	}
 
+	/// The request path the peer addressed, or `None` if it provided none.
+	///
+	/// On URL-less transports (native QUIC, qmux over TCP/TLS/UDS) this is the path the
```
kixelated: Just to confirm, we expose the transport path otherwise?

rs/moq-relay/tests/smoke.rs
```diff
@@ -513,6 +516,84 @@ async fn internal_unix_round_trip() {
 	handle.abort();
 }
 
+/// Publisher and subscriber that announce/observe `broadcast` over the internal
+/// listener at `pub_url` / `sub_url`. Returns the path the subscriber sees the
+/// publisher's broadcast announced at, proving whether the request path reached
+/// the server (it scopes the publisher's grant to that root).
+async fn path_round_trip(pub_url: url::Url, sub_url: url::Url, broadcast: &str) -> String {
```
kixelated: We should also test different versions. Currently it's only testing moq-lite-05, but moq-transport-14+ should also support PATH.

rs/moq-net/src/server.rs
kixelated: Done — rewrote the summary, dropped the lite-05-specific framing. The path now comes from the SETUP across both moq-lite-05 and moq-transport (14-16). (f4a8738)

rs/moq-net/src/server.rs
kixelated: Reworded — it now leads with what it does (pause once the path is known so you can authorize/scope before serving), not the version mechanics. (f4a8738)

rs/moq-net/src/server.rs
kixelated: Fixed for moq-transport **14-16**: the client now advertises the `Path` SETUP parameter and the server reads it in the legacy bidi-SETUP branch → `Request::path()` / `Session::path()`. Smoke tests cover it.

**17/18 deferred** (TODO in code): their SETUP is decoded-and-discarded in a background uni loop (`ietf/session.rs` `run_unis`), and reading it before `start()` collides with the GOAWAY stream that loop owns (uni `Reader` vs the bidi `Stream` the start path takes). That's a focused IETF-internals change I didn't want to rush into this PR. Happy to do it next — want it here or as a follow-up?

rs/moq-net/src/server.rs
kixelated: Updated the comment (and added a TODO) to say both SETUPs are backgrounded and we should read the client's first so the path reaches `path()`. Same 17/18 follow-up as above. (f4a8738)

rs/moq-net/src/session.rs
kixelated: Correct — for WebTransport/WebSocket the path is in the connection URL, which the relay already reads at the moq-native/axum layer (`Request::url()` → auth). It just isn't on `moq_net::Session` because `web_transport_trait::Session` has no `path()`. Adding that to web-transport (so `Session::path()` covers URL transports too) is the follow-up I flagged in the PR description.

rs/moq-relay/tests/smoke.rs
kixelated: Done — `path_round_trip` is now parametrized over version and both the tcp:// and unix:// tests loop over moq-lite-05 **and** moq-transport-14 (client pinned via `--client-version`). 17/18 left out until the server surfaces their path (above). (f4a8738)

rs/moq-net/src/server.rs
kixelated: Done — 17/18 fixed properly (1bb3b31). `ietf::accept_setup` reads the client's uni SETUP before serving (STOP_SENDING-ing anything racing ahead, same as lite-05), pulls out the `Path` param, and hands the stream to `ietf::start` so it keeps serving as the GOAWAY channel — no lost GOAWAY. The client advertises its path via `run_setup`. `path()` now works across moq-lite-05 and moq-transport 14-18, and the smoke tests loop over all of them.

rs/moq-net/src/session.rs
kixelated: Agreed — dropped `Session::path()` entirely (35d101c). It could only return the SETUP path, so it'd be `None` for WebTransport/WebSocket and just mislead. The path lives on the `accept_request` -> `Request::path()` flow, where you actually act on it. No more web-transport follow-up needed.

PR #1912, opened 2026-06-26
perf(moq-json): generate merge patches with a diffing serializer (+ benchmark)

kixelated:
## Summary

Started as "benchmark JSON Merge Patch vs streaming DEFLATE", and grew into replacing the merge-patch generator with a faster serde-`Serializer`-based one.

The producer used to call `serde_json::to_value(new)` to build a full `Value` tree for every update, then diff two trees. Now a serde `Serializer` walks the new value `T` and diffs each field against the previous value as it goes: unchanged scalars and subtrees cost only a comparison (no allocation), only changed nodes are built into the patch, and the producer folds each delta into the baseline via `json_patch::merge` instead of re-materializing the whole value. No public API or wire-format change; patches are byte-identical.

## Correctness (this is the part that matters)

The serializer lives in `src/diff.rs` with a thorough test suite, since a wrong merge patch silently corrupts consumer state:

- A Value-vs-Value **reference oracle** the serializer is cross-checked against on every case, including a 40-step evolving-document roundtrip.
- **Typed structs** (the serializer's reason to exist — diff `T` without building its `Value`): `Option` + `skip_serializing_if` Some↔None transitions (a skipped field becomes a null deletion), nested structs, `Vec` wholesale replacement, unit enum variants.
- Edge cases: value set to null forces a snapshot, nested null, non-object/array roots, unchanged → empty patch, key add/remove.
- The shared **golden vectors** (`tests/vectors.json`, cross-checked with the JS impl) now run through the serializer.
- All existing **Producer** behavior tests (dedup, budget gates, frame cap, late joiner, compression) still pass.

49 tests total, all green.

## Performance

Merge-patch **generation** CPU (no DEFLATE), serializer vs old `to_value + diff`:

| workload | change |
|---|---|
| telemetry (small, many fields move) | **−15%** |
| big-static (large + big static array) | **−11%** |
| big-nested (large, few scalars move) | **−31%** |

For large docs the generation step is most of the producer's non-DEFLATE CPU, so this is a real producer win. The full-stack `Producer::update` loop dropped from ~19.6k → ~15.4k ns/tick over this PR. **Limit:** merge patch replaces arrays wholesale and serde streaming can't prune inside an array, so a big static array is rebuilt each tick regardless.

No existing crate offers a serializer-driven merge-patch diff (`json-patch` only *applies* RFC 7396; its `diff` is RFC 6902; `treediff` needs a built `Value`; `jsondiffpatch` is its own format).

## Part 1 — the original question: merge-patch + DEFLATE vs DEFLATE-only

`examples/bench.rs` (`cargo run --release -p moq-json --example bench`):

| workload | wire (merge vs deflate-only) | producer CPU | consumer CPU |
|---|---|---|---|
| small, fast-moving (~800 B) | 0.90× | +48% | −21% |
| large, mostly-static (~10 KB) | **0.36×** | +64% | **−94%** |

The shared DEFLATE window already de-duplicates most redundancy, so on small docs deflate-only is within ~11% on bytes. Merge patch trades producer CPU for consumer CPU and bytes, and the trade improves sharply as the doc grows and changes get sparser (big static doc: 3× smaller wire, 94% cheaper consumer). **DEFLATE level is the biggest single CPU lever** (level 3 ≈ −19% CPU for +11% bytes; level 9 is pure loss here).

## Also in this PR

- Lazy snapshot serialization: `Producer::update` no longer serializes a full snapshot it discards on the delta path (folded into the serializer rewrite).
- Benchmark group-boundary capture fix and corrected synthetic temp range (review fixes).

## Suggested follow-up (not in this PR)

- Expose the DEFLATE level on `ProducerConfig` (currently hardcoded to 6). Public-API addition — would want sign-off on the shape.

## Test plan

- [x] `cargo test -p moq-json` — 49 passed (27 diff + 22 producer)
- [x] `cargo clippy -p moq-json --all-targets` clean
- [x] `cargo fmt -p moq-json --check` clean
- [x] `cargo run --release -p moq-json --example bench`

(Written by Claude)


rs/moq-json/src/diff.rs
```diff
@@ -1,67 +1,516 @@
...
-	if let (Value::Object(old), Value::Object(new)) = (old, new) {
-		let mut patch = Map::new();
-		let mut forced = false;
-		diff_objects(old, new, &mut patch, &mut forced);
-		Diff {
-			patch: Value::Object(patch),
-			forced_snapshot: forced,
+/// Only object roots produce a recursive patch; any other root forces a snapshot. A merge patch that
+/// would delete a key it shouldn't (a value genuinely set to null) also forces a snapshot.
+pub fn diff<T: Serialize>(old: &Value, new: &T) -> Diff {
```
kixelated, line 32: Could old also be `&T`? Could we traverse both structs at the same time?

I think that would prevent needing to fully convert old into a Value? We could just serialize the diff inline instead?

rs/moq-json/src/diff.rs
```diff
@@ -1,67 +1,516 @@
...
-				patch.insert(key.clone(), Value::Object(sub));
+	fn finish(self) -> Result<Node, Error> {
+		let mut patch = self.patch;
+		if let Value::Object(base) = self.differ.baseline {
+			// A deletion is only possible if some key was added or the counts differ. Otherwise the new
+			// keys are exactly the baseline keys, so there's nothing to delete and we skip the scan,
+			// keeping the common path O(1) rather than O(n^2). A removed key is a clean delete (explicit
+			// null), and unlike a value set to null it does not force a snapshot.
+			if self.added_key || self.seen.len() != base.len() {
+				let seen: HashSet<&str> = self.seen.iter().map(String::as_str).collect();
```
kixelated, line 402: Gross.

I suppose this is unavoidable unless serde has a deterministic iteration order.

rs/moq-json/src/diff.rs
kixelated: Good question, and I went down this path. The blocker is that diffing needs *random access* into `old`: as the serializer streams the new value's fields, it has to look up old's value for each key/index it sees. serde is push-based and one-directional. you can drive one `Serialize` from start to finish, but you can't pause it to pull the matching field out of a second `Serialize` in lockstep. To traverse both at once you'd have to buffer one side into an indexable structure first, and that structure is basically `Value`.

The good news is the producer doesn't pay that conversion per tick. `old` (the `last` baseline) persists across updates and is updated in place via `json_patch::merge(last, &patch)`. We only call `to_value` on a snapshot (once per group), amortized across all of that group's deltas. The expensive part we set out to remove was building a `Value` for the *new* value every tick, and that's gone. building/keeping one for `old` is cheap by comparison.

Two more reasons to keep `old` as a `Value`:
- Storing it as `T` instead would force a `T: Clone` (or `DeserializeOwned`) bound so we can retain the previous value between calls. today `update` only needs `Serialize`.
- Key-based lookup handles the `skip_serializing_if` `Some`<->`None` case for free (a field present last tick and absent now is just a deletion). a positional zip of two `T` event streams would desync the moment one side omits a field.

I'm happy to prototype a both-`&T` variant in the example to get real numbers if you're curious, but I expect it to lose for the streaming-producer case for the reasons above. (Written by Claude)

---

rs/moq-json/src/diff.rs
kixelated: Exactly right. The baseline `Value::Object` iterates in sorted order (it's a `BTreeMap`), but the new value's `Serialize` impl makes no such promise. structs emit fields in declaration order, maps in whatever order the map has. So we can't assume the two key streams line up, which rules out a lockstep merge-join for detecting deletions. We have to remember which keys appeared and diff that against the baseline's key set.

The one mitigation is the guard right above it: `added_key || seen.len() != base.len()`. when no key was added and the counts match, the new keys are exactly the baseline keys, so there's nothing to delete and the whole `HashSet` scan is skipped. That's the common case (stable schema), so the gross part only runs on the rare tick where a key actually appears or disappears. (Written by Claude)

---

Issue #1923, opened 2026-06-26
moq-mux: pre-existing codec/container correctness findings (from #1918 review)


kixelated:
CodeRabbit's review of #1918 (the moq-mux backport to main) surfaced several Major-severity findings that are **pre-existing in the `dev` moq-mux lineage** — they are not regressions introduced by the backport. The two low-risk ones were fixed in #1918; the rest are tracked here because they're either risky to change blindly (could affect the byte-exact roundtrip tests), architectural, or ambiguous. They should be fixed on `dev` (and flow back to `main`) or in a focused follow-up with dedicated tests.

### Deferred findings

- **`codec/h264/split.rs` / `codec/h265/split.rs` — consecutive bare IDRs can merge into one access unit.** The IDR/keyframe arm reconciles SPS/PPS and sets `contains_idr`/`contains_slice` but never closes an already-open slice-bearing frame on the first slice of a new picture (unlike the non-keyframe slice path which checks `first_slice…`). A stream that doesn't repeat parameter sets before each keyframe, with a keyframe arriving right after a delta picture in the same `decode` chunk, can fold two AUs into one frame (and mis-flag it `keyframe`). Streaming-only; the `decode`+`flush`-per-AU tests don't surface it.

- **`codec/aac/mod.rs` — explicit (15) sample-rate index misparses.** When `freq_index == 15` the 24-bit explicit rate begins in bits already consumed from the preceding bytes; the current path reads the next three whole bytes, so both `sample_rate` and `channel_count` decode wrong. Breaks `Config::encode()` → `Config::parse()` round-trips for non-standard rates. Needs a bit-reader.

- **`container/consumer.rs` — group read errors all treated as evictions.** The `poll_read` error branch skips forward on any `F::Error`, but a malformed container payload (decode error) is not the same as a relay group eviction. Only known group-abort / old-cache errors should skip; decode errors should propagate.

- **`container/fmp4/mod.rs` — duration-less multi-sample fragments lose timestamps.** Encoding omits the sample duration whenever `Frame.duration` is `None`, but `decode()` advances DTS by 0 for those, so a fragment with frames at e.g. 0ms and 33ms but no explicit durations decodes both at 0ms. Should infer adjacent durations for non-last samples or reject such multi-sample fragments. Related to the signed/duration work in #1917.

[...]

PR #1950, opened 2026-06-30
Merge main into dev

kixelated:
## Summary

Merges `main` into `dev`. The hard part: several dev-only features had been backported to `main` against main's **older** moq-net API, so the same code exists on both sides in two API shapes. Throughout this merge, **dev's newer API is the source of truth**; main's genuine post-backport fixes are ported onto it, and main's old-API shapes are dropped.

## Genuine main fixes ported (translated to dev's API)

- **moq-mux #1923** (codec/container correctness): AAC `BitReader` parser rewrite; H.264/H.265 IDR first-slice access-unit split; fMP4 `MissingSampleDuration` guard; container `poll_aborted` (distinguishes a relay eviction from a real decode error); producer backfills **every** batched sample's duration; MSF AAC SBR/PS trailing-byte acceptance; catalog `Filter`/`Target` end-with-upstream EOF fix.
- **moq-mux #1933**: fMP4 zero-duration handling across decode/import/export + `infer_missing_durations`, with the fallback expressed at the frames' **own** timescale (dev's `Timestamp` is multi-scale, unlike main's fixed-microsecond one).
- **moq-mux #1941**: `opus::Config::encode()` returns `Result` instead of panicking on multichannel (+ `mkv::Error::Opus` ripple).
- **moq-srt #1922**: negative-pacing offsets (a reordered B-frame paces before the anchor); the `serve` binary derives the HTTP sidecar bind from the real listener.
- **moq-rtmp**: reject a publish cleanly when the broadcast path is unavailable (before acking the client); RTMPS flags gated behind `noq`/`quinn`.
- **moq-ffi**: `publish_media` gains whole-buffer container support, re-expressed in dev's `reserve_track`/`MediaDecoder` shape.
- **moq-relay**: main's #1902 client-TLS verification (`tls.rs`) + HTTPS cert arrays (#1932) + the web-embedding `routes()`/`serve()` test.
- **moq-gst**: reject multichannel Opus.
- **moq-json / JS**: adopt `moq-flate` (Rust) / `@moq/flate` (JS) extraction + main's compressed delta budgeting + drain-to-latest backlog collapse; net `Group` gains `tryReadFrame`/`tryReadFrameSequence`/`readable`/`done`; watch/publish gain the `hangz` (`catalog.json.z`) compressed catalog.
- Dependency bumps from main's cargo/bun groups (e.g. `tower-http` 0.7).

## Clean-but-broken auto-merges fixed (a naive merge would miss these)

- `kio`: dev (#1779) and main (#1913) both added an identical `Consumer::write` — kept one.
- `hang`: `compressed_track()` used main's `moq_net::Track`; reshaped to dev's `TrackInfo`.
- `moq-relay`: restored the `AuthToken` import dropped by the auto-merge; reverted the #1901 `auth.init(tls)` call to dev's 0-arg shape.
- opus `Result` change rippled into the flv tests (now `.encode().unwrap()`).

## Decisions deferred as owed follow-ups (documented in the merge commit)

- **moq-mux `select::Broadcast` reshape** (68c9ebd/#1944): not adopted — it drops dev's `Target` (resolution/bitrate ABR) capability that moq-cli depends on, so dev's `Filter`+`Target` is kept.
- **moq-rtc #1931 session runner**: not ported — dev's moq-rtc has diverged well past main; re-apply the API reshape against dev's shape later.
- **moq-hls #1945** (discontinuity sequence through fMP4): deferred — dev's hls import predates main's select-coupled shape.
- **Rust compressed catalog (#1904 / `HangZ`)**: deferred on the Rust side (the JS `@moq/flate` + `hangz` path is kept).
- **moq-relay #1901** (reuse client TLS for outbound auth HTTP): deferred — `tls.rs` carries #1902 but `auth.init` keeps dev's shape.
- **moq-json telemetry example**: removed (main-only, written against main's net API).

## Test plan

- [x] `cargo build` + `cargo test` for every crate touched by the merge: `moq-mux` (308), `moq-json` (50), `moq-flate` (7), `hang` (16), plus `moq-net`/`moq-srt`/`moq-relay`/`moq-rtmp`/`moq-rtc`/`moq-native` compiling. All green.
- [x] JS: `tsc --noEmit` + Biome clean; `bun test` for net (194), json (43), flate (9), msf (7), watch (78), publish (5), hang (34). All green.
- [x] `cargo fmt` clean; no conflict markers anywhere.
- [ ] Full-workspace CI build (the sandbox can't fetch the `nvidia-video-codec-sdk` git dep behind its egress policy, nor build the GStreamer system dep, so `moq-video`/`moq-boy`/`moq-cli`/`libmoq`/`moq-gst` were validated by compile-of-touched-code only).

https://claude.ai/code/session_01R3bVtLrzn4Q82tChYu9EJZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01R3bVtLrzn4Q82tChYu9EJZ)_


kixelated:
## CI status

**Check is green** (full cargo build + clippy + tests across the workspace, including `libmoq`/`moq-relay`). Two follow-up commits fixed `moq-cli` (`moq-ffi`/`moq-cli` aren't in the default cargo workspace, so the merge commit's `hls import`/`export` API mismatches only surfaced in the Smoke/nix build):
- `747f242` — adapt main's `hls import` to dev's moq-net API (`BroadcastInfo`, `Result`-returning `publish_broadcast`).
- `e287eab` — enable moq-hls's `server` feature for `hls export`.

**Smoke is red, but it's pre-existing dev debt, not a merge regression.** The cross-language interop matrix is green on the diagonal (`rust→rust`, `rust→gst`); every cross-client *subscriber* fails because the smoke clients under `test/smoke/clients/` call APIs that dev's own bindings already dropped:

- **C** (`subscribe.c`): `moq_origin_consume` (sync lookup) was replaced on dev by the async `moq_origin_consume_announced` callback.
- **Python** (`smoke.py`): `subscribe_catalog()` is now an awaitable returning a `CatalogConsumer`, not an async-iterable.
- **js-native / browser** (`subscribe.ts`): `Catalog.fetch` was removed from `@moq/hang`.

These clients track **main's** API (which is why nightly Smoke on `main` is green), and `origin/dev` carries the same stale clients — `git diff origin/dev HEAD -- test/smoke/` is empty apart from one `package.json` line. Pure `dev` would fail identically; it has just never been exercised because the Smoke job only runs on PRs that touch `test/smoke/**` (this PR triggered it via a `package.json` dependency bump). The matrix is all-or-nothing, so it can't go green until all three client surfaces are updated to dev's API — which is a separate Cross-Package-Sync task, not part of resolving the merge.

Recommend tracking the `test/smoke/clients/{c,python,js-native}` update to dev's API as a follow-up. Happy to do it here instead if preferred.

---

PR #1960, opened 2026-06-30
feat(moq-native): add platform TLS verifier selection

Qizot:
This PR adds an ability to set what kind of TLS verifier will be used when connecting with a server.

The previous approach didn't work on iOS and android platforms. The current approach allows the user to choose and also provides a sane defaults for each platform (it should be backwards compatible for platforms other than ios/android/macos).


kixelated:
replaced with #1968

PR #1966, opened 2026-06-30
moq-mux/moq-rtc: report + drop incompatible egress renditions

kixelated:
Stacked on #1964.

## Problem

Each egress format carries only a subset of the codecs a catalog can hold. Today an unsupported codec either:
- **hard-fails the whole session** — FLV/MPEG-TS export `bail!`s on the first codec it can't carry, so e.g. an Opus-audio broadcast can't be played over SRT *at all*, even though its H.264 video could; or
- **vanishes silently** — WHEP drops the track with only a `tracing::warn!` (the AAC-over-WebRTC case).

Either way the user gets no signal about *what* was incompatible.

## Change

Supported tracks now still egress, and the dropped ones are surfaced so a caller can flag the incompatibility.

- **`moq-mux::compat`** (new): the codec-support matrix in one place — `carries_audio` / `carries_video` per `Protocol` (Webrtc / Flv / Ts), `classify` (catalog → `Vec<DroppedTrack>`), `DroppedTrack` with friendly labels ("AAC", "VP8", …), and `dropped_for` (resolve a broadcast, read its catalog, report drops). `dropped_for` is for serve paths that own the exporter and can't surface it mid-session (RTMP/SRT sealed `accept`).
- **FLV + MPEG-TS export**: filter renditions the format can't carry out of the catalog snapshot at ingest (`bail!` → partial), so a supported track still exports; expose the dropped set via `Export::dropped`. FLV carries H.264/H.265/AV1/VP9 + AAC/Opus/AC-3/E-AC-3; TS carries H.264/H.265 + AAC/MP2/AC-3/E-AC-3. Verbatim (`mpegts`) passthrough streams are untouched.
- **moq-rtc WHEP**: `Response::dropped` carries the renditions WebRTC can't egress.

## Tests

New unit tests for the support matrix + labels. All 314 existing `moq-mux` tests (incl. FLV/TS export round-trips) pass unchanged.

### DIFF
--- a/rs/moq-mux/src/compat.rs
+++ b/rs/moq-mux/src/compat.rs
@@ -1,0 +1,216 @@
+//! Codec compatibility helpers shared by the egress muxers and the RTP egress.
+//!
+//! Each egress format (MPEG-TS, FLV/RTMP, WebRTC) can only carry a subset of the
+//! codecs a hang catalog might contain. When a broadcast carries a codec a given
+//! format can't, that rendition is *dropped*: the rest still egresses, and the
+//! gateway reports the drop so the dashboard can flag the incompatibility (e.g.
+//! "AAC audio can't go over WebRTC, use Opus").
+
+use std::time::Duration;
+
+use anyhow::Context;
+use hang::catalog::{AudioCodec, VideoCodec};
+
+use crate::catalog::hang::{Catalog, CatalogExt};
+
+/// An egress protocol, used to pick the codec set it can carry.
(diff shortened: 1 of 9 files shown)


kixelated:
Superseding this: moq main now natively carries the codecs this was about (Opus-in-TS, MP3-in-FLV via #1969/#1970), so the FLV/TS partial-export + drop-reporting isn't needed. The downstream dashboard handles codec advertising proactively. Closing.

PR #1971, opened 2026-06-30
feat(net): pluggable GroupSource for deep out-of-RAM replay window

iameli:
@kixelated So this is what Claude came up with for implementing atmoq's 72-hour replay window. Works, but let me know if this is completely wrongheaded and I'm happy to scrap it and do a FETCH-based backfill implementation instead. That said, this behavior is kinda neat, allows library consumers to bring their own backing store for use cases where you want to backfill older content right up to live.

### DIFF
--- a/rs/moq-net/src/model/track.rs
+++ b/rs/moq-net/src/model/track.rs
@@ -60,6 +60,10 @@ struct State {
 	max_sequence: Option<u64>,
 	final_sequence: Option<u64>,
 	abort: Option<Error>,
+	/// Per-track override for how long evicted-eligible groups are retained.
+	/// `None` falls back to [`MAX_GROUP_AGE`]. Lets a publisher trade memory for
+	/// a deeper late-join / replay window.
+	max_group_age: Option<std::time::Duration>,
 }
 
 impl State {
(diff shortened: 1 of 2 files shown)


iameli:
@sourcery-ai well now we gotta box :mechanical_arm:

rs/moq-net/src/model/track.rs
```diff
@@ -26,6 +26,26 @@ use std::{
...
+/// subscriber requests a `start_group` older than anything still in the in-RAM
+/// track cache (see the lite publisher's `run_track`). This lets a publisher
+/// serve a replay window far deeper than memory allows — e.g. disk-backed —
+/// without holding every group in the cache.
+///
+/// Wrapped in an `Arc` and shared with the publisher's async task, so
+/// implementations must be `Send + Sync`. [`Self::group`] may block briefly on
+/// I/O; the publisher only calls it on the backfill path, never in the hot live
+/// loop.
+pub trait GroupSource: Send + Sync {
```
kixelated, line 39: Yeah I'm not a fan of "callback" traits like this. This one in particular requires all history to be synchronous (oof if you're using S3).

On dev there's a `dynamic()` method that allows you to receive requests for old groups.

PR #1974, opened 2026-06-30
Fold the internal listener into --server-bind (one authenticated accept path)

kixelated:
## What

The relay no longer has a separate `--internal-*` listener. Its trusted-local-worker transports (plaintext qmux over TCP / Unix sockets) become **their own `--server-*` flags** alongside the QUIC `--server-bind`, and every transport shares one authenticated accept path:

```
--server-bind [::]:443                 # QUIC/WebTransport (UDP). Omit to run stream-only.
--server-tcp-bind 127.0.0.1:4444       # plaintext qmux over TCP (requires `tcp` feature)
--server-unix-bind /run/moq/moq.sock   # plaintext qmux over a Unix socket (requires `uds`, unix-only)
--server-unix-allow-uid 1001           # peer-credential allowlist for the unix listener
```

## Why

The internal listener originally existed because it was *unauthenticated full access* — a different trust model that deserved its own door. The first commit on this branch made it authenticate through the relay's `Auth`. Once it's authenticated, it differs from the normal server only in **transport**, not in auth — so a separate subsystem, config namespace, and accept loop no longer earn their keep.

## Authentication

Every transport goes through the same `Auth`; only the source of the path + JWT differs:

- **QUIC / WebSocket**: path + `?jwt=` from the request URL; a valid mTLS client cert (QUIC only) stands in for a JWT.
- **`tcp` / `unix`**: path + `?jwt=` ride the moq-lite-05 SETUP in-band. A **no-JWT** stream connection resolves through the relay's existing **public auth** (`--auth-public` / `[auth] public`), exactly like a tokenless QUIC client — no listener-specific scope. A local helper that must publish without a token is granted publicly (e.g. `--auth-public-publish .stats`).

Unix sockets can additionally be gated by peer credentials via `--server-unix-allow-uid` / `-gid` / `-pid`, enforced in the listener before the SETUP is read.

## Changes

- **`moq-native`**: `Server` gains TCP (`TcpConfig`) and Unix (`UnixConfig`) stream listeners alongside QUIC/WebSocket, each a feature/os-gated config struct flattened into `ServerConfig` (`--server-bind` stays QUIC-only, `Option<String>`). The listeners bind lazily on first `accept()` and feed one accept loop; their accept-loop tasks are owned by `StreamListeners` and aborted on `Drop`. `Request` gains `path()` (in-band SETUP path). The `unix://` listener enforces the peer-cred allowlist. A stream-only server (no `--server-bind`) doesn't open QUIC.
- **`moq-relay`**: delete `internal.rs` / `InternalConfig`. `Connection::authenticate` is one path: params from the URL (QUIC/WS) or the SETUP (stream), then `Auth::verify`. `main.rs` drops the `run_internal` arm and only treats `Error::NoBackend` from `local_addr()` as "no QUIC".
- **Docs**: `doc/bin/relay/{config,auth}.md` rewritten (the `[internal]` section becomes "Stream Listeners").

## Breaking changes

- `moq-native` (0.x): `Server` gains `tcp`/`unix` config + `Request::path()`. Kept on `main` per maintainer call — all consumers are in-tree.
- `moq-relay`: `--internal-*` flags / `[internal]` TOML removed, replaced by `--server-tcp-bind` / `--server-unix-bind` / `--server-unix-allow-*`. The old unauthenticated behaviour is reproducible with `--server-tcp-bind 127.0.0.1:4444 --auth-public ""`.

## Tests

- `cargo nextest --all-features` green (incl. the `internal_tcp_round_trip` / `internal_unix_round_trip` smoke tests, now stream-only servers driven through `Server` + `Connection` with public auth) plus config/allowlist unit tests.
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo doc --all-features`, `cargo check --no-default-features`, and shear/sort/taplo all pass.

(Written by Claude Opus 4.8)


rs/moq-native/src/server.rs
```diff
@@ -15,20 +15,63 @@ use futures::stream::FuturesUnordered;
...
+	/// [`bind`](Self::bind), by the peer's kernel-reported user ID. Empty means no
+	/// uid constraint. A connection that fails the allowlist is dropped before its
+	/// SETUP is read. Ignored by TCP/QUIC, which carry no peer credentials.
+	#[arg(
+		long = "server-unix-allow-uid",
+		env = "MOQ_SERVER_UNIX_ALLOW_UID",
+		value_delimiter = ','
+	)]
+	#[serde(default, skip_serializing_if = "Vec::is_empty")]
+	pub unix_allow_uid: Vec<u32>,
```
kixelated: Make a Unix struct instead, and feature/os gate this?

rs/moq-native/src/server.rs
```diff
@@ -116,18 +159,118 @@ impl ServerConfig {
...
+	Quic(String),
+	/// Plaintext qmux over TCP. Carries the resolved address.
+	#[cfg(feature = "tcp")]
+	Tcp(net::SocketAddr),
+	/// Plaintext qmux over a Unix socket. Carries the socket path.
+	#[cfg(all(feature = "uds", unix))]
+	Unix(PathBuf),
+	/// A `tcp://`/`unix://` entry whose transport feature is not compiled in.
+	#[cfg(not(all(feature = "tcp", feature = "uds", unix)))]
+	Unsupported(String),
```
kixelated: Why not return an error instead?

rs/moq-native/src/server.rs
```diff
@@ -116,18 +159,118 @@ impl ServerConfig {
...
+	/// back-compat. `udp://` is QUIC with the scheme stripped. `tcp://` and
+	/// `unix://` select the stream transports.
+	fn parse(entry: &str) -> crate::Result<Self> {
+		let Some((scheme, rest)) = entry.split_once("://") else {
+			// No scheme: a bare address is QUIC, as before.
+			return Ok(BindScheme::Quic(entry.to_string()));
+		};
+
+		match scheme {
+			"udp" | "quic" => Ok(BindScheme::Quic(rest.to_string())),
```
kixelated: I guess add `https` too

rs/moq-native/src/server.rs
```diff
@@ -116,18 +159,118 @@ impl ServerConfig {
...
+		Ok(quic)
+	}
+}
+
+/// A `--server-bind` entry classified by transport.
+enum BindScheme {
+	/// QUIC over UDP. Carries the address with any `udp://` scheme stripped.
+	Quic(String),
+	/// Plaintext qmux over TCP. Carries the resolved address.
+	#[cfg(feature = "tcp")]
```
kixelated: IDK I would just use separate flags for these protocols. It feels like we're overloading --bind especially since we only support one QUIC listener.

```
--server-tcp-bind
--server-unix-bind
--server-unix-allow-uid
```

kixelated: Done — switched to separate flags in 874fdae0f. `--server-bind` is QUIC-only again; the stream transports now have `--server-tcp-bind` and `--server-unix-bind` (+ `--server-unix-allow-{uid,gid,pid}`). The whole `BindScheme` URL-scheme parser and `quic_bind()` are gone.

(Written by Claude Opus 4.8)

rs/moq-native/src/server.rs
kixelated: Done — the unix knobs are now a feature/os-gated `UnixConfig` struct (`#[cfg(all(feature = "uds", unix))]`), and there's a matching `#[cfg(feature = "tcp")] TcpConfig`. Both flatten into `ServerConfig`.

(Written by Claude Opus 4.8)

rs/moq-native/src/server.rs
kixelated: Moot now — with separate flags there's no scheme classification, so `BindScheme`/`quic_bind()` (and the `Unsupported`-then-error dance) are deleted. An unknown flag is just a clap parse error.

(Written by Claude Opus 4.8)

rs/moq-native/src/server.rs
kixelated: Also moot — no URL scheme parsing anymore, so there's nothing to add `https` to. (The web/wss listener stays its own `--web-https-*` config.)

(Written by Claude Opus 4.8)

rs/moq-relay/src/main.rs
```diff
@@ -25,7 +25,8 @@ async fn main() -> anyhow::Result<()> {
 	let mut server = config.server.init()?;
 	let client = config.client.clone().init()?;
 
-	let addr = server.local_addr()?;
+	// `None` when `--server-bind` lists only stream (tcp/unix) transports.
+	let addr = server.local_addr().ok();
```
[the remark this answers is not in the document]
kixelated: Fixed in 874fdae0f — `local_addr()` now only maps `Error::NoBackend` to `None` (the stream-only case); every other error propagates with context.

(Written by Claude Opus 4.8)

rs/moq-native/src/server.rs
```diff
@@ -468,6 +667,224 @@ impl Server {
...
+				#[cfg(all(feature = "uds", unix))]
+				StreamBind::Unix(path) => {
+					let listener = crate::unix::Listener::bind(&path)
+						.await?
+						.with_protocols(versions.alpns());
+					// Loose file perms: the uid/gid/pid allow list is the real gate,
+					// and the worker usually runs as a different user than the server.
+					listener.set_mode(0o666)?;
+					tracing::info!(path = %path.display(), allow_uid = ?self.unix_allow.uid, "listening (unix)");
+					spawn_unix_loop(listener, versions, self.unix_allow.clone(), tx.clone());
```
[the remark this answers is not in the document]
kixelated: Fixed in 874fdae0f — `StreamListeners` now stores the accept-loop `JoinHandle`s and aborts them in its `Drop`. Since `Server` owns `StreamListeners`, dropping the server stops the loops and frees the bound sockets.

(Written by Claude Opus 4.8)

rs/moq-native/src/server.rs
```diff
@@ -103,6 +117,68 @@ pub struct ServerConfig {
...
+	pub bind: Option<PathBuf>,
+
+	/// Allowed peer user IDs. Empty means any uid.
+	#[arg(
+		long = "server-unix-allow-uid",
+		env = "MOQ_SERVER_UNIX_ALLOW_UID",
+		value_delimiter = ','
+	)]
+	#[serde(default, skip_serializing_if = "Vec::is_empty")]
+	pub allow_uid: Vec<u32>,
```
kixelated: Make struct Option<UnixAllow> ? If None then we don't enforce anything.

rs/moq-native/src/server.rs
```diff
@@ -468,6 +632,223 @@ impl Server {
...
+	Tcp(net::SocketAddr),
+	#[cfg(all(feature = "uds", unix))]
+	Unix(PathBuf),
+}
+
+/// Peer-credential allowlist for the `unix://` listener, built from
+/// [`UnixConfig::allow_uid`] and friends.
+#[cfg(all(feature = "uds", unix))]
+#[derive(Clone, Default)]
+struct UnixAllow {
```
kixelated: Lul just make this use clap

kixelated: Done in 9f3004ccc — `UnixAllow` is now a `clap::Args` struct with the `--server-unix-allow-*` flags, flattened into `UnixConfig` as `allow` and reused for enforcement. No more hand-rolled duplicate.

(Written by Claude Opus 4.8)

rs/moq-native/src/server.rs
kixelated: Done — the listener now takes `Option<UnixAllow>`; `None` (the default, built as `(!allow.is_empty()).then(...)`) enforces nothing, so there's no per-connection is_empty() check anymore.

(Written by Claude Opus 4.8)

kixelated: Also made the config field itself `Option<UnixAllow>` now (not just the enforcement path), in 9f3004ccc's follow-up: `None` = no allowlist. clap's flatten sets it to `Some` only when a `--server-unix-allow-*` flag is present, and the `Option` survives the TOML→CLI re-parse (the clobber test covers it).

(Written by Claude Opus 4.8)

doc/bin/relay/config.md
```diff
@@ -37,14 +37,36 @@ level = "info"
 
 ### \[server]
 
-QUIC/WebTransport server settings.
+QUIC/WebTransport server settings. Optionally add plaintext qmux stream
+listeners for trusted local workers; the relay authenticates every accepted
+connection the same way (JWT, mTLS, or public access) regardless of transport.
```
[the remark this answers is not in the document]
kixelated: Good catch — fixed. Reworded so mTLS is scoped to QUIC; stream listeners are described as JWT / public-access plus optional Unix peer-credential gating.

(Written by Claude Opus 4.8)

PR #1975, opened 2026-06-30
moq-rtmp/srt/rtc: convert to library-only crates

kixelated:
Converts the three gateway crates — `moq-rtmp`, `moq-srt`, `moq-rtc` — from lib+bin into **library-only** crates, removing their standalone binaries.

This is the first step of moving the gateway CLIs into `moq-cli`. The CLI surface itself is intentionally **not** in this PR: after design discussion it's being reshaped into a unified endpoint model (protocol-prefixed `--rtmp-connect`/`--srt-listen`/… flags + `import`/`export`), which lands in a follow-up PR together with new RTMP/SRT dial-out (client) library support.

## What changed

- **Delete the standalone binaries** (`moq-rtmp`, `moq-srt`, `moq-rtc`) and their `bin/` helpers.
- **Slim each crate to a library** — drop the binary-only deps (`moq-native`, `clap`, `axum-server`, `sd-notify`, `tower-http`, and the `noq`/`quinn`/`quiche`/`websocket`/`iroh` transport-backend feature forwards).
  - `moq-rtmp`: the `server` feature gated real library code (RTMPS — the `Conn::Tls` variant, `Server::with_tls`, `Config::tls`, `pub use rustls`), so it's renamed to **`tls`** (`default = ["tls"]`). `moq-native` becomes a dev-dependency for the RTMPS test.
  - `moq-srt` / `moq-rtc`: their `server` feature gated nothing in the library, so it's removed entirely (`default = []`).
- **Docs**: remove the standalone-binary pages (`doc/bin/rtmp.md`, `doc/bin/rtc.md`) and their nav entries; the crate READMEs and `rs/CLAUDE.md` now describe the libraries.

## Public API

- `moq-rtmp`: feature `server` → `tls` (rename). No `pub` item changes.
- `moq-srt` / `moq-rtc`: `server` feature + transport forwards removed; no `pub` item changes.
- These are `0.0.1` crates with no external consumers; the gateway library APIs (`run`, `Server`/`Request`, `Client`) are untouched.

## Test plan

- [x] `cargo check --all-targets`, `cargo clippy -- -D warnings`, `cargo fmt --check`, rustdoc (`-D warnings`).
- [x] `cargo test -p moq-rtmp` (incl. the RTMPS test, now built via the `moq-native` dev-dependency), `-p moq-srt`, `-p moq-rtc`.

(Written by Claude Opus 4.8)


kixelated:
CodeRabbit and Sourcery were both rate-limited and didn't produce a review, so I ran a local multi-angle review of the diff. Two real issues found and fixed in `6bf02fe57`:

- **SRT cert-sidecar bind regression.** The old `moq-srt` bin derived the `/certificate.sha256` web bind from `server.local_addr()`; the port computed it from the `--server-bind` string, which breaks `--server-bind [::]:0` / hostname binds (the sidecar would land on a different port than the QUIC server). Restored `local_addr()`-derived bind, and applied it to `rtmp` too for consistency (moq-native's default bind is `[::]:443`, so the unset case is unchanged).
- **Dead doc sidebar links.** `doc/.vitepress/config.ts` still pointed at the deleted `/bin/rtc` and `/bin/rtmp` pages (now folded into `/bin/cli#gateways`); removed them.

Verified clean: `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, and the gateway crate tests still pass.

Findings I deliberately left as-is: the `MOQ_RTC_URL` -> `MOQ_RTC_REMOTE` env rename (intentional CLI reshape), the `rtc` `import: bool` / positional-arg threading and the `rtmp`/`srt` double-`match` (ported verbatim, internal binary plumbing), and a pre-existing latent constraint that `moq-cli` needs a quinn/quiche/noq backend to compile (already true before this PR via the HLS path; default features include `quinn`).

(Written by Claude Opus 4.8)

doc/bin/cli.md
```diff
@@ -224,6 +224,120 @@ AudioSpecificConfig. Both pass straight through to the catalog `description`. Th
...
 
+## Gateways
+
+`moq-cli` bridges three other live protocols to MoQ: RTMP, SRT, and WebRTC
+(WHIP/WHEP). They share one grammar:
+
+```
+moq-cli <rtmp|srt> <client|server> <import|export>
+moq-cli rtc <import|export>
+```
```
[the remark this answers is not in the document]
kixelated: Added a `text` language to the fenced grammar block (`b84b4b241`).

doc/bin/cli.md
```diff
@@ -224,6 +224,120 @@ AudioSpecificConfig. Both pass straight through to the catalog `description`. Th
...
+ffplay rtmp://127.0.0.1:1935/live/cam0
+```
+
+The broadcast path is `<app>/<key>` from the RTMP URL (`rtmp://host/<app>/<key>`),
+with `--rtmp-prefix` prepended. Both legacy RTMP (H.264 + AAC, plus MP3) and
+enhanced RTMP (HEVC, AV1, VP9, Opus, AC-3, E-AC-3) work, since the codec handling
+lives in `moq-mux`'s FLV demuxer/muxer. Add `--rtmps-listen` with a cert
+(`--rtmps-tls-cert` / `--rtmps-tls-key`, or `--rtmps-tls-generate` for a
+throwaway self-signed cert) to also accept RTMPS (`rtmps://`).
+
```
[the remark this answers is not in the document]
kixelated: Fixed. The `server` examples were doubly wrong: the flag is `--server-bind` (not `--bind`) and it belongs to the `server` command, so it must precede `import`. Corrected every rtmp/srt `server` example in cli.md and both crate READMEs, and verified each parses (`b84b4b241`).

rs/moq-cli/src/rtc.rs
```diff
@@ -0,0 +1,248 @@
...
+			.await
+			.map_err(|_| anyhow::anyhow!("broadcast {} not announced", broadcast_name))?;
+		client.publish(url, broadcast).await?;
+	}
+
+	// `client.subscribe` / `publish` spawn the session in the background; block on
+	// ctrl-c instead of returning so the gateway stays up.
+	tokio::signal::ctrl_c().await?;
+	Ok(())
+}
```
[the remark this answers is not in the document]
kixelated: Leaving this as-is for now: it's ported verbatim from the old `moq-rtc` bin (which also just blocked on ctrl-c after spawning the session). Surfacing later ICE/DTLS/media-path death would need `moq_rtc::Client::{subscribe,publish}` to hand back a closed/session handle to `select!` on, i.e. a moq-rtc API change, which is out of scope for this binary-move PR. Better as a focused follow-up than folded in here.

rs/moq-srt/Cargo.toml
```diff
@@ -15,48 +15,17 @@ categories = ["multimedia", "network-programming", "web-programming"]
...
 moq-mux = { workspace = true }
-moq-native = { workspace = true, default-features = false, features = ["aws-lc-rs"], optional = true }
 moq-net = { workspace = true }
-rustls = { version = "0.23", default-features = false, features = ["aws-lc-rs"], optional = true }
 # Pure-Rust SRT (no libsrt / ffmpeg). SRT carries MPEG-TS, demuxed by moq-mux.
 srt-tokio = "0.4"
 thiserror = "2"
 tokio = { workspace = true, features = ["full"] }
-tower-http = { version = "0.7", features = ["cors", "fs"], optional = true }
 tracing = "0.1"
```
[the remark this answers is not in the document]
kixelated: Leaving this out of scope: these deps predate the PR (moq-srt already declared them locally; I only removed the binary-only ones). `anyhow`/`bytes`/`futures`/`thiserror`/`tracing` are declared per-crate across the whole workspace rather than in `[workspace.dependencies]`, so pinning only moq-srt would be inconsistent with the repo's existing convention. A workspace-wide dep-pinning pass is a separate change.

rs/moq-cli/src/rtc.rs
```diff
@@ -0,0 +1,248 @@
...
+		#[command(flatten)]
+		endpoint: Endpoint,
+	},
+}
+
+/// The WebRTC endpoint: either run a server or dial a remote, exactly one.
+#[derive(Args, Clone)]
+pub struct Endpoint {
+	/// Run a WHIP/WHEP server on this HTTP address.
+	#[arg(long = "rtc-listen", env = "MOQ_RTC_LISTEN", conflicts_with = "remote")]
```
kixelated: Make a `client/server` command for RTC, sorry.

rs/moq-cli/src/rtc.rs
```diff
@@ -0,0 +1,248 @@
...
+	#[arg(long = "rtc-remote", env = "MOQ_RTC_REMOTE", conflicts_with = "listen")]
+	remote: Option<Url>,
+
+	/// UDP socket the shared WebRTC media mux binds to (server mode). All sessions
+	/// share this one port (demuxed by ICE ufrag); `0.0.0.0:0` lets the OS pick.
+	#[arg(long = "rtc-udp-bind", env = "MOQ_RTC_UDP_BIND", default_value = "0.0.0.0:0")]
+	udp_bind: SocketAddr,
+
+	/// Optional TLS cert (PEM) for the WHIP/WHEP server. Requires `--rtc-tls-key`.
+	#[arg(long = "rtc-tls-cert", env = "MOQ_RTC_TLS_CERT", requires = "tls_key")]
```
kixelated: don't include `--rtc` in the long, could you have id = rtc-tls-cert instead?

Or ideally use moq_native::tls

rs/moq-cli/src/rtc.rs
```diff
@@ -0,0 +1,248 @@
...
+			)
+			.await
+		}
+		(None, Some(remote)) => run_client(import, remote, broadcast, public_addr, publisher, subscriber).await,
+		_ => anyhow::bail!("exactly one of --rtc-listen or --rtc-remote is required"),
+	}
+}
+
+/// Run a WHIP (import) / WHEP (export) server on `listen`.
+#[allow(clippy::too_many_arguments)]
```
kixelated: pls just make a Server struct

rs/moq-cli/src/rtc.rs
```diff
@@ -0,0 +1,248 @@
...
+		Router::new().merge(server.subscribe_router())
+	};
+	let app = app.layer(CorsLayer::new().allow_origin(Any).allow_methods(Any).allow_headers(Any));
+
+	tracing::info!(%listen, import, "moq-rtc server listening");
+	serve(app, listen, tls_cert, tls_key).await
+}
+
+/// Dial a remote WHEP (import) / WHIP (export) resource at `url`.
+async fn run_client(
```
kixelated: plus just make a Client struct

rs/moq-cli/src/rtmp.rs
```diff
@@ -0,0 +1,232 @@
...
+	},
+}
+
+/// Which way media flows across the RTMP boundary.
+#[derive(Subcommand, Clone)]
+pub enum RtmpDirection {
+	/// RTMP -> MoQ: ingest pushed RTMP streams and publish them into MoQ.
+	Import {
+		#[command(flatten)]
+		rtmp: RtmpArgs,
```
kixelated: If these args are shared, should they be part of `RtmpCommand` (or higher?)

rs/moq-cli/src/rtmp.rs
```diff
@@ -0,0 +1,232 @@
...
+	}
+}
+
+/// CLI flags for the RTMP listener(s), converted into [`moq_rtmp::Config`]s.
+#[derive(Args, Clone)]
+pub struct RtmpArgs {
+	/// Address to listen on for plaintext RTMP (`rtmp://`). RTMP's well-known port
+	/// is 1935.
+	#[arg(long = "rtmp-listen", env = "MOQ_RTMP_LISTEN", default_value = "[::]:1935")]
+	listen: SocketAddr,
```
kixelated: This should only be for the `Server` command

rs/moq-cli/src/rtmp.rs
```diff
@@ -0,0 +1,232 @@
...
+	/// well-known port; common choices are 443 or a custom one.
+	#[arg(long = "rtmps-listen", env = "MOQ_RTMPS_LISTEN")]
+	rtmps_listen: Option<SocketAddr>,
+
+	/// PEM certificate chain for RTMPS.
+	#[arg(long = "rtmps-tls-cert", env = "MOQ_RTMPS_TLS_CERT")]
+	rtmps_cert: Option<PathBuf>,
+
+	/// PEM private key for RTMPS.
+	#[arg(long = "rtmps-tls-key", env = "MOQ_RTMPS_TLS_KEY")]
```
kixelated: Try to use moq_native::tls

PR #1980, opened 2026-07-01
fix(moq-mux): hold TS export PSI until PMT-declared tracks resolve

### LINKED ISSUE #1979: moq-mux TS export: catalog convergence race locks PSI before video track resolves
## Summary

`moq_mux::container::ts::Export` can hard-fail with:

```
TS track layout changed after PAT/PMT was emitted: '<name>' added
```

when a subscriber connects while the publisher's catalog is still converging, i.e. before every elementary stream has resolved into a catalog rendition.

## Root cause

`update_catalog` locks PAT/PMT (`self.psi`) once `header_ready() && video_ready()` are both true:

- `rs/moq-mux/src/container/ts/export.rs:280` (build gate)
- `rs/moq-mux/src/container/ts/export.rs:492` (`video_ready`, doc comment now notes the limitation)
- `rs/moq-mux/src/container/ts/export.rs:347-361` (`update_catalog`, the `anyhow::ensure!` that fires)

`video_ready()` only inspects tracks already present in `self.tracks`, and says a program with no video track is "trivially ready." But track registration timing differs by codec:

- AAC registers its catalog rendition on the **first PES** (`rs/moq-mux/src/codec/aac/import.rs:19-41`).
- H.264 registers its catalog rendition only once the **first keyframe's inline SPS** is parsed (`rs/moq-mux/src/codec/h264/import.rs:149-187`).

If an audio-only catalog snapshot reaches the exporter before video's rendition resolves, `header_ready()`/`video_ready()` both pass vacuously, PSI locks on an audio-only layout, and the video rendition arriving moments later trips the "layout changed" error.

Not caused by SRT ingest or `ffmpeg -stream_loop`: `moq-srt`'s `Publisher` holds one broadcast for the life of the SRT connection with no reconnect/teardown logic (`rs/moq-srt/src/ts.rs`, `rs/moq-srt/src/server.rs`), so the race is purely a function of which codec's catalog rendition happens to resolve first, independent of transport.

## Fix direction

The exporter needs to know a program's eventual track count before it's willing to lock PSI, rather than inferring readiness from tracks already resolved (a strict, time-varying subset). Likely needs input from the `mpegts` catalog section (or similar) listing all PMT-declared PIDs/stream types up front, even for streams still awaiting codec resolution.

## Repro

Observed via `test/smoke/srt_pcr.sh`: ffmpeg -> SRT -> `moq-srt` ingest -> relay -> `moq-cli subscribe --format ts`. Timing-dependent; not always reproducible.

t0ms:
## Summary

Fixes moq-dev/moq#1979: MPEG-TS exporter PSI (program tables) race condition. When audio codec resolves before video, the exporter locked PSI on an audio-only layout, then crashed when video's rendition arrived.

**Solution:** Gate PSI construction on the PMT's elementary-stream count (known at parse time), not on whichever subset of tracks has resolved. The importer publishes this as `expected_tracks` in the `mpegts` catalog extension; the exporter waits for that many tracks before building PSI.

## Changes

- **catalog.rs**: Add `Mpegts::expected_tracks: Option<u16>` field carrying PMT-declared elementary stream count
- **import.rs**: Publish `expected_tracks` when PMT is parsed (before any codec resolves)
- **export.rs**: New `layout_complete()` gate; PSI build waits for declared count to resolve
- **export_test.rs**: Four integration tests covering audio-first race, video-first order, non-TS sources, and real import→export roundtrip

## Testing

- ✅ All 340 moq-mux tests pass (337 pre-existing + 4 new)
- ✅ No regressions in existing roundtrip/codec/import tests
- ✅ Format/clippy/doc checks clean
- ✅ Cross-package sync verified

## Backward Compatibility

Non-TS sources (no \`expected_tracks\` hint) export immediately as before, preserving existing behavior.

### DIFF
--- a/rs/moq-mux/src/container/ts/catalog.rs
+++ b/rs/moq-mux/src/container/ts/catalog.rs
@@ -34,12 +34,21 @@ pub struct Mpegts {
 	/// re-emits these; the SCTE-35 'CUEI' registration is derived when absent.
 	#[serde(default, skip_serializing_if = "Vec::is_empty")]
 	pub program_descriptors: Vec<Descriptor>,
+
+	/// How many elementary streams the most recently parsed PMT declares that will
+	/// resolve into a published MoQ track (media or verbatim). Known immediately from
+	/// the PMT, before any of those tracks' own catalog renditions resolve, so export
+	/// can tell an in-progress layout (fewer tracks than this) from a genuinely
+	/// complete one instead of locking PSI on whichever subset has resolved first.
+	/// `None` when the catalog wasn't produced from an MPEG-TS source.
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub expected_tracks: Option<u16>,
 }
 
 impl Mpegts {
 	/// True when the section carries nothing, so it's omitted from the catalog.
 	pub fn is_empty(&self) -> bool {
-		self.tracks.is_empty() && self.program_descriptors.is_empty()
+		self.tracks.is_empty() && self.program_descriptors.is_empty() && self.expected_tracks.unwrap_or(0) == 0
 	}
 }
 
--- a/rs/moq-mux/src/container/ts/export.rs
+++ b/rs/moq-mux/src/container/ts/export.rs
@@ -61,6 +61,10 @@ pub struct Export<E: CatalogExt = ()> {
 	counters: HashMap<u16, ContinuityCounter>,
 	/// PMT program-level descriptors captured on import, re-emitted in the PMT.
 	program_descriptors: Vec<catalog::Descriptor>,
+	/// How many tracks the source PMT declared, from the `mpegts` catalog section's
+	/// `expected_tracks` hint. `None` for a non-TS source (nothing to wait for beyond
+	/// what has already resolved). See [`Self::layout_complete`].
+	expected_tracks: Option<u16>,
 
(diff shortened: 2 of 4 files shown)


rs/moq-mux/src/container/ts/catalog.rs
```diff
@@ -34,12 +34,21 @@ pub struct Mpegts {
...
 	pub program_descriptors: Vec<Descriptor>,
+
+	/// How many elementary streams the most recently parsed PMT declares that will
+	/// resolve into a published MoQ track (media or verbatim). Known immediately from
+	/// the PMT, before any of those tracks' own catalog renditions resolve, so export
+	/// can tell an in-progress layout (fewer tracks than this) from a genuinely
+	/// complete one instead of locking PSI on whichever subset has resolved first.
+	/// `None` when the catalog wasn't produced from an MPEG-TS source.
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub expected_tracks: Option<u16>,
```
kixelated, line 45: I think this is the wrong fix.

I think export should instead wait IDK like 100ms for the catalog to stabilize, since TS doesn't support reinitializing (just like fMP4).

On the import side, we could wait until all tracks are initialized before publishing the catalog too, but it's still racey.

t0ms:
superseded by https://github.com/moq-dev/moq/pull/1991

PR #1992, opened 2026-07-01
feat(moq-mux): windowed TS export with dense uniform PCR (supersedes #1988, #1989)

kixelated:
## Summary

Reworks the MPEG-TS exporter from **one output per media frame** into fixed **~20 ms windows** on the media clock, each led by a PCR. This is the design @ArielM and @kixelated converged on in discussion, and it folds the two prior fixes into one clean model.

The problem (Ariel's Sencore IRD): the elementary stream is perfect, but the exporter wrote a single PCR per frame valued at the authored DTS — too sparse at low frame rates, too uneven under B-frame reorder. A hardware IRD has no clock of its own; it rebuilds the 27 MHz clock from the PCR values with a PLL that needs a **dense (≤40 ms TR 101 290), monotonic, uniform** reference. The sparse/jittery PCR makes the PLL hunt → the cascade of clock errors. `-muxrate 11M` papered over it by padding to CBR so PCR-by-byte-position is a clean ramp.

## The model

Each `Output` is one ~20 ms window:
- **Opens on the decode grid** (PCR value = window start), so the PCR ramp is *exactly* uniform regardless of frame rate — no interpolation.
- Carries a **leading PCR-only packet** (adaptation-only, auto-stuffed to 188 bytes, no continuity-counter bump), then any media whose decode time falls in the window.
- Stamped with its `pace` instant, so a caller paces delivery per window.

A window with **no media is just that PCR packet** — bridging a low-frame-rate gap with a few bytes instead of padding the mux to CBR (no wasted null-packet bitrate, the "moq can do better internally" point). Gaps the exporter can see (buffered/VOD) fill deterministically; at the live edge a `Timer` flushes a filler once the window's wall-clock deadline passes. **The exporter owns the clock end to end** — it decides every PCR's value *and* its due instant; the consumer just executes.

This resolves Ariel's open "where do I get the frame interval" question: the window grid is fixed on the decode clock, so nothing depends on catalog frame rate or lookahead buffering.

## Commits

1. **kio `Waiter::waker()` + `Timer`** — a poll-driven wall-clock wait so a runtime-free `poll_*` can drive a `tokio` sleep against its own waker (kio gains no dependency; the sleep lives in moq-mux, which already depends on tokio).
2. **Pace on `tokio::time::Instant`** — drops the std↔tokio conversion and lets `tokio::time::pause()` advance the whole pace pipeline, so windowed emission tests deterministically.
3. **Windowed exporter** — the rework above. Frames resolve their decode clock when pulled (`PendingFrame`) and window/interleave on decode time, keeping the mux monotonic across reorder. `Output` collapses to `{ payload, pace }`.

## `--pace` behavior

Unchanged in spirit: PCR/filler correctness is independent of `--pace`. The flag only decides whether the consumer *sleeps* on each window's `pace` (real-time playout of a retained broadcast) or dumps as fast as readable. moq-srt always stamps `pace` (TSBPD); moq-cli gates the sleep.

## API surface (moq-mux, breaking)

- `ts::Output` is now `{ payload, pace }` (dropped per-frame `timestamp`/`keyframe`; no consumer read them). Payload is a window, not a frame.
- `Output::pace` is `tokio::time::Instant` (was `std`).
- `kio::Waiter::waker()` added (additive).

## Test plan

- [x] `cargo test -p moq-mux -p moq-srt -p moq-cli -p kio` (342 / 6 / 1 / 7). New `windows_emit_uniform_pcr_with_fillers` asserts an exactly-uniform 20 ms PCR ramp with fillers bridging a 10 fps gap; `Timer` has its own paused-time test. All existing TS import→export→import roundtrips still pass byte-exact, validating the windowed wire output.
- [x] `cargo clippy ... --all-targets` clean.
- [ ] Not yet verified end-to-end against a real IRD (Ariel to test).

(Written by Claude Opus 4.8)

---
_Generated by [Claude Code](https://claude.ai/code/session_01Sv3GrsvEy1wyd2SYbApDzh)_

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3729,6 +3729,7 @@ name = "kio"
 version = "0.4.1"
 dependencies = [
  "smallvec",
+ "tokio",
 ]
 
 [[package]]
--- a/rs/kio/Cargo.toml
+++ b/rs/kio/Cargo.toml
@@ -15,5 +15,14 @@ categories = ["asynchronous", "concurrency"]
 [lib]
 doctest = false
 
+[features]
+# Opt-in `time::Timer`, a poll-driven wall-clock wait backed by `tokio::time`. Off by
+# default so kio stays runtime-free; enable it when driving `poll_*` functions on tokio.
+tokio = ["dep:tokio"]
+
 [dependencies]
 smallvec = "1.15"
+tokio = { workspace = true, features = ["time"], optional = true }
+
+[dev-dependencies]
+tokio = { workspace = true, features = ["macros", "rt", "test-util", "time"] }
--- a/rs/kio/src/lib.rs
+++ b/rs/kio/src/lib.rs
@@ -18,6 +18,9 @@ mod future;
 mod producer;
 mod weak;
 
+#[cfg(feature = "tokio")]
+pub mod tokio;
+
 #[cfg(test)]
 mod tests;
 
--- a/rs/kio/src/tokio.rs
+++ b/rs/kio/src/tokio.rs
@@ -1,0 +1,49 @@
+//! kio integration for `tokio`.
+//!
+//! Behind the `tokio` feature. Wraps a [`tokio::time::Sleep`] so a `poll_*` function can
+//! drive it against a [`Waiter`], keeping the rest of kio runtime-free.
+
+use std::future::Future;
+use std::pin::Pin;
+use std::task::{Context, Poll};
+
+use crate::Waiter;
+
+/// A `tokio` sleep driven by kio's poll model.
+///
+/// Construct it once for a deadline (`Sleep::new(tokio::time::sleep_until(deadline))`), then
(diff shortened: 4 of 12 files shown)


arielmol:
_First, a bug in this PR, the PCR-only filler packets were advancing the continuity_counter (trips TR 101 290 P1); found and fixed while testing._

After extensive patching, testing and trying many ways. There's no way to make a Real IRD lock.

If i set the transport rate above the content rate then the queue drains and the stream survives,  but the slack shows up as about 20% empty windows (PCR-only), so delivery is lumpy and the PLL glitches. OTOH setting near content rate causes the windows to fill nicely but the -frame spikes have nowhere to go queue grows without bounds
until it can't be scheduled and the stream dies (and latency balloons). There's no operating point that's both smooth and stable. 

It _seems_ without null-stuffing the gap between channel rate and content rate appears either as empty windows or unbouded queues.

(Also, im not Ariel Malka but Ariel Molina.)

PR #2018, opened 2026-07-02
[codex] Fix RTMP play resolve timeout

kixelated:
## Summary

- Delay RTMP `Play.Start` until the requested broadcast and initial FLV header are available.
- Add a bounded resolve timeout and keep racing the wait against client disconnects.
- Add a regression test for playing a nonexistent broadcast without reporting playback accepted.

## Root Cause

`Play::accept` accepted the RTMP play request before resolving the MoQ broadcast. A typo or random path could therefore receive `Play.Start` and then park indefinitely while waiting on `announced_broadcast`.

## Validation

- `cargo fmt -p moq-rtmp --check`
- `cargo test -p moq-rtmp play_missing_broadcast_rejects_without_start`
- `cargo test -p moq-rtmp`
- `git diff --check`

(Written by GPT-5)

### DIFF
--- a/rs/moq-rtmp/src/server.rs
+++ b/rs/moq-rtmp/src/server.rs
@@ -56,6 +56,10 @@ const READ_BUFFER: usize = 16 * 1024;
 /// handshake.
 const REQUEST_TIMEOUT: Duration = Duration::from_secs(15);
 
+/// How long a play request may wait for its broadcast and initial FLV header
+/// before the server rejects it.
+const PLAY_RESOLVE_TIMEOUT: Duration = Duration::from_secs(5);
+
 /// TCP keepalive idle period before the kernel starts probing a silent peer, and
 /// the interval between probes. Once a connection is publishing or playing it can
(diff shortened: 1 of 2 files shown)


rs/moq-rtmp/src/server.rs
```diff
@@ -56,6 +56,13 @@ const READ_BUFFER: usize = 16 * 1024;
 /// handshake.
 const REQUEST_TIMEOUT: Duration = Duration::from_secs(15);
 
+/// How long a play request may wait for its broadcast and initial FLV header
+/// before the server rejects it.
+#[cfg(not(test))]
+const PLAY_RESOLVE_TIMEOUT: Duration = Duration::from_secs(5);
+#[cfg(test)]
+const PLAY_RESOLVE_TIMEOUT: Duration = Duration::from_millis(250);
```
kixelated: Not a fan. Can we use tokio time instead to avoid real waits?

kixelated (commented):
IDK this PR seems a bit more complicated than I would imagine.

PR #2021, opened 2026-07-02
[codex] evict finished HLS broadcasters

### LINKED ISSUE #2010: moq-hls: Server broadcaster map never evicts (dead ENDLIST after republish, unbounded memory)
## Summary

`Server`'s broadcaster map is insert-only (`rs/moq-hls/src/server/mod.rs:35,59` on main): `broadcasters: Mutex<HashMap<String, Arc<Broadcaster>>>` with a get-or-insert and no removal path. When the source broadcast ends, the pumps finish the store and the playlist gains `#EXT-X-ENDLIST`, but the `Broadcaster` (with its full ~16s window of segment bytes per rendition) stays in the map.

## Impact

1. **Functional:** a publisher disconnects and republishes the same name; `broadcaster(name)` returns the cached dead instance and every viewer gets a finished playlist until the process restarts.
2. **Memory:** rotating broadcast names (per-session IDs) grow the map without bound on a long-running exporter.

## Suggested fix

Evict the entry when the source broadcast ends (the pumps already observe that), or key by broadcast instance so a republish creates a fresh Broadcaster, plus an idle TTL for the memory case.

Found during a full review of the dev branch; verified on main.

(Written by Claude Fable 5)

kixelated:
## Summary

- mark moq-hls broadcasters finished when their catalog watcher exits
- evict finished broadcasters from the server cache so a same-name republish gets a fresh instance
- guard cleanup with Arc::ptr_eq so a late cleanup cannot remove a newer broadcaster

Fixes #2010.

## Validation

- cargo test -p moq-hls
- just fix
- just check progressed through JS checks, Rust check, clippy, fmt, docs, shear, and cargo sort, then failed at the repo-wide justfile formatting gate because existing justfiles differ from just --fmt --unstable --check output. I left that unrelated formatter churn out of this PR.

(Written by GPT-5)

### DIFF
--- a/rs/moq-hls/src/export/mod.rs
+++ b/rs/moq-hls/src/export/mod.rs
@@ -53,6 +53,7 @@ impl Default for Config {
 
 /// All renditions of one broadcast, kept in sync with its catalog.
 pub struct Broadcaster {
+	broadcast: moq_net::BroadcastConsumer,
 	renditions: Mutex<BTreeMap<String, Arc<Rendition>>>,
 	/// Current rendition count, bumped on every catalog sync so handlers can wait
 	/// for the catalog to populate before rendering a playlist.
@@ -69,6 +70,7 @@ impl Broadcaster {
 		let (ready, _) = watch::channel(0);
 		let (paused, _) = watch::channel(false);
 		let broadcaster = Arc::new(Self {
+			broadcast: broadcast.clone(),
 			renditions: Mutex::new(BTreeMap::new()),
 			ready,
 			paused,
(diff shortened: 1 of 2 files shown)


rs/moq-hls/src/export/mod.rs
```diff
@@ -179,7 +202,10 @@ async fn watch_catalog(broadcast: moq_net::BroadcastConsumer, config: Config, br
 				tracing::warn!(%err, "failed to subscribe to broadcast catalog, retrying");
 				tokio::select! {
 					_ = tokio::time::sleep(CATALOG_RETRY) => {}
-					_ = kio::wait(|waiter| broadcast.poll_closed(waiter)) => return,
+					_ = kio::wait(|waiter| broadcast.poll_closed(waiter)) => {
+						broadcaster.finish();
```
kixelated: why not have `broadcaster` await broadcast.closed() instead?

kixelated: Done in f904ffa26. Broadcaster now stores the BroadcastConsumer, cache lookups use is_closed(), and eviction awaits broadcast.closed() through Broadcaster::closed().

(Written by GPT-5)

PR #2024, opened 2026-07-02
test: MPEG-TS / IRD compliance harness for the ts export

t0ms:
## Summary

Adds a self-contained MPEG-TS / IRD compliance harness under `test/ts-compliance/`, wired as `just test ts-compliance`. It round-trips a PCR-paced transport stream through a relay (`import ts` -> relay -> `export ts`), captures the subscriber's output, and validates it against what an Integrated Receiver/Decoder expects.

- **`run.sh`** builds `moq-relay` + `moq-cli`, generates a broadcast-like clip (or `--source cap.ts`), stands up a relay (reusing `test/smoke/smoke.toml`), publishes a PCR-paced TS (`tsp -P regulate --pcr-synchronous | moq ... import ts`), captures `moq ... export ts`, and analyzes it. `--analyze-only file.ts` skips the round-trip (TSDuck + Python only); `--strict` fails on shape warnings.
- **`compliance.py`** lets TSDuck parse the stream (`tsanalyze --json` plus a 188/204-byte header scan for the per-packet PID + PCR timeline) and does the model math on the stream's own PCR clock.

### Checks
- **Hard (fail by default):** packet size, sync/transport errors, PAT, PMT, PSI CRC, continuity counters, PCR presence, PCR monotonicity.
- **Shape (WARN, fail only under `--strict`):** PCR repetition interval, PCR jitter (pcrverify model), null-packet ratio, service descriptors (SDT), bitrate consistency + burstiness (1 ms / 10 ms windows), packet inter-arrival, and a transport-buffer (T-STD TB) model. Thresholds are CLI-overridable; `--report-json` emits a machine-readable report.

This is a diagnostic gate: it confirms the exporter's TS is structurally valid while quantifying where it diverges from a CBR IRD feed (per-frame PCR pacing, 0% null padding, bursty per-GOP delivery).

## Cross-package sync

Test-only addition; no rows in the sync table apply.

## Test plan

- [x] `just test ts-compliance` round-trips a generated clip: hard checks PASS, shape checks WARN, exit 0.
- [x] `--strict` exits non-zero on shape warnings.
- [x] `run.sh --analyze-only <file.ts>` works standalone (TSDuck + Python only).
- [ ] Run against a real broadcast capture via `--source`.

## Caveats

- Physical-layer TR 101 290 items (RF, real sync loss) are unmeasurable from a file.
- Wall-clock delivery jitter/burstiness is out of scope; timing is PCR-clock based.
- `tstd` models only the TB smoothing stage of ISO 13818-1, with default (not level-derived) leak rates.

Requires TSDuck (`tsp`, `tsanalyze`) and `python3`; round-trip modes also need `cargo`, `ffmpeg`, `curl`, `timeout`.

(Written by Claude Opus 4.8)

### DIFF
--- a/test/justfile
+++ b/test/justfile
@@ -36,3 +36,13 @@ smoke-full:
 # harness can actually report failure).
 smoke-negative *args:
     ./smoke/smoke.sh --negative {{ args }}
+
+# Round-trips a PCR-paced TS through a relay and checks the subscriber's `export
+# ts` output with TSDuck plus a custom analyzer (PCR jitter/repetition,
+# burstiness, instantaneous bitrate, T-STD). Flags: `--analyze-only file.ts`
+# (skip the round-trip), `--strict` (fail on shape warnings), `--source file.ts`
+# (use a real capture). See ts-compliance/README.md.
+
+# Check the subscriber's MPEG-TS output for IRD compliance.
+ts-compliance *args:
+    ./ts-compliance/run.sh {{ args }}
--- a/test/ts-compliance/README.md
+++ b/test/ts-compliance/README.md
@@ -1,0 +1,75 @@
+# TS / IRD compliance test
+
+Validates the MPEG-TS that the moq subscriber emits (`moq ... export ts`) against
+what an Integrated Receiver/Decoder (IRD) expects. It round-trips a stream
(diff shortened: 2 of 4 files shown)


kixelated:
Folded into #2011

PR #2028, opened 2026-07-02
[codex] fix dropped group fetch requests

kixelated:
## Summary
- Add `GroupRequest::reject` and drop-reject handling so a popped group fetch request always wakes its matching `fetch_group` waiter.
- Track per-fetch request IDs so rejected attempts do not poison later retries for the same group sequence, and clear stored rejections once the waiter resolves.
- Reject relay-served fetches on non-Lite05 sessions with `Error::Version`, and reject waiters on transient fetch stream open or encode failures.
- Merge current `dev`, which already carries the upstream `quick-xml` cargo-deny advisory ignores needed for the previous CI failure.

## Public API changes
- Added `GroupRequest::reject(err: Error)` in `moq-net`. This is additive.

## Validation
- `cargo fmt --check`
- `cargo test -p moq-net fetch_`
- `cargo deny check advisories`
- `just ci` passed Rust, JS, Python, Kotlin, Swift, Go, and Nix checks locally before stopping at the final shell-format step because `shfmt` is not installed on this machine.

(Written by GPT-5)


rs/moq-net/src/lite/subscriber.rs
```diff
@@ -771,6 +771,7 @@ impl<S: web_transport_trait::Session> TrackServe<S> {
 		// SUBSCRIBE_UPDATE (and thus pause/resume linger) only exists on Lite03+.
 		// Older versions tear the upstream down as soon as the track goes idle.
 		let supports_linger = !matches!(self.subscriber.version, Version::Lite01 | Version::Lite02);
+		let supports_fetch = !matches!(self.subscriber.version, Version::Lite01 | Version::Lite02);
```
kixelated: Are these the only versions that don't support it? I'm pretty sure 04 doesn't either?

PR #2029, opened 2026-07-02
[codex] fix moq-net timestamp fallback on wasm

kixelated:
## Summary

- Route `Timestamp::now()` through `web_async::time::Instant` so browser builds avoid `std::time::Instant::now()` and `SystemTime::now()` on the receive timestamp fallback path.
- Hide the native/browser clock split behind private `clock` modules, keeping `Timestamp::now()` cfg-free.
- Keep the native jittered wall-clock anchor, while using `performance.now()` plus a per-process jitter for wasm.
- Update the `moq-wasm` README now that the timestamp fallback is ported.

## Root Cause

Mandatory receive timestamps made `create_frame_now` run for frames received from pre-lite-05 sessions. That path initialized `TIME_ANCHOR`, which called standard library clocks that panic on `wasm32-unknown-unknown`.

## Validation

- `cargo fmt --check`
- `cargo check -p moq-net`
- `cargo check -p moq-wasm --target wasm32-unknown-unknown`
- `git diff --check`

Note: `--locked` checks currently fail on `dev` before compiling because Cargo wants to refresh unrelated `Cargo.lock` entries. I left the lockfile out of this PR.

(Written by GPT-5)

### DIFF
--- a/rs/moq-net/src/model/time.rs
+++ b/rs/moq-net/src/model/time.rs
@@ -1,8 +1,4 @@
 use std::num::NonZero;
-use std::sync::LazyLock;
-use std::time::{SystemTime, UNIX_EPOCH};
-
-use rand::RngExt;
 
 use crate::coding::VarInt;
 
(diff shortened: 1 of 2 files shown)


rs/moq-net/src/model/time.rs
```diff
@@ -425,6 +448,13 @@ static TIME_ANCHOR: LazyLock<(std::time::Instant, SystemTime)> = LazyLock::new(|
 	(std::time::Instant::now(), SystemTime::now() - jitter)
 });
 
+#[cfg(all(target_arch = "wasm32", not(target_os = "wasi")))]
```
kixelated: Is there a way to reduce these cfg guards?

kixelated: Done in `6c679f903`: the platform split is now contained in two private `clock` modules, and `Timestamp::now()` is cfg-free.

(Written by GPT-5)

PR #2031, opened 2026-07-02
[codex] fix IETF object property KVP encoding

kixelated:
## Summary
- Fix Rust IETF object-property KVP type encoding so draft-14/15 use absolute property IDs and draft-16+ keeps delta IDs.
- Mirror Timestamp/Timescale object-property behavior in JS IETF publish and subscribe paths.
- Add Rust and JS regression coverage for the draft-14/15 versus draft-16 cutoff.

## Root Cause
The group object-property helpers encoded draft-14/15 property type IDs as deltas, while those drafts define absolute type IDs. Self-interop hid the issue, but compliant peers skipped Timescale and interpreted timestamps at the wrong scale. The JS path also failed to preserve object timestamps through IETF publish/subscribe.

## CI
- Rebased onto current dev, which already includes the cargo-deny ignores for quick-xml RUSTSEC-2026-0194/0195.
- cargo-deny now passes locally.

## Public API Changes
None. Internal wire encoding/decoding behavior only.

## Cross-Package Sync
Updated both rs/moq-net and js/net. No docs changes because this fixes protocol interop without changing public APIs or documented examples.

## Tests
- cargo test -p moq-net ietf::group::tests::test_object_time
- bun test js/net/src/ietf/ietf.test.ts
- just js check
- cargo deny check --show-stats

(Written by GPT-5)

### DIFF
--- a/js/net/src/ietf/ietf.test.ts
+++ b/js/net/src/ietf/ietf.test.ts
@@ -1,10 +1,11 @@
 import { expect, test } from "bun:test";
 import * as Path from "../path.ts";
 import { Reader, Writer } from "../stream.ts";
+import { Timescale, Timestamp } from "../time.ts";
 import * as Varint from "../varint.ts";
 import * as GoAway from "./goaway.ts";
 import * as Namespace from "./namespace.ts";
-import { Group } from "./object.ts";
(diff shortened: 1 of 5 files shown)


kixelated (commented):
JS side too?

PR #2039, opened 2026-07-02
feat(moq-net): lite-05 QUIC datagram delivery

kixelated:
## Summary

Adds unreliable best-effort QUIC datagram delivery for the lite-05 WIP path. Datagrams are version-gated and transport-gated, so older peers and non-datagram transports do not send or accept them.

The model treats a datagram as a single unreliable payload parallel to groups. It shares the track sequence namespace, carries an absolute timestamp in the track timescale, and has no group fallback when the payload does not fit a single datagram.

### Model and Wire
- Adds `Datagram { sequence, timestamp, payload }` plus a conservative `MAX_DATAGRAM_PAYLOAD` limit.
- Adds producer APIs for auto-numbered datagrams and explicit-sequence forwarding, so relays preserve upstream numbering.
- Adds subscriber datagram receive APIs beside the existing group receive APIs.
- Encodes lite datagrams as `subscribe | sequence | timestamp | payload`, with payload delimited by the QUIC datagram boundary.
- Drains datagrams and groups from the same subscription. A datagram returned on an ordered subscriber advances the ordered group cursor past that sequence.

### JS Mirror
- Mirrors the Rust datagram model, wire codec, version gate, and 50 ms best-effort send-buffer behavior in `@moq/net`.
- Sends browser datagrams through one connection-owned `WebTransport.datagrams` writer and receives them through the subscriber datagram loop.
- Gates only on `hasDatagrams(version) && maxDatagramSize > 0`, matching Rust and allowing qmux to participate once it reports datagram support.

## Public API Changes

Rust `rs/moq-net`, additive and non-breaking:
- `Datagram`, `MAX_DATAGRAM_PAYLOAD`
- `TrackProducer::{append_datagram, write_datagram}`
- `TrackSubscriber::{recv_datagram, poll_recv_datagram}`
- crate-internal lite datagram codec and version gate

TypeScript `@moq/net`, additive and non-breaking:
- `Datagram`, `MAX_DATAGRAM_PAYLOAD`
- `TrackProducer.{appendDatagram, writeDatagram}`
- `TrackSubscriber.recvDatagram`
- `Lite.Datagram`, `hasDatagrams`

Also tightens `@moq/net` declarations by keeping internal state types out of the published `.d.ts` surface.

## Cross-Package Sync

- `rs/moq-net` and `js/net` are mirrored for wire body, version gating, payload limits, send-buffer behavior, and ordered cursor semantics.
- `doc/concept/layer/moq-lite.md` documents the datagram terminology and behavior.
- `js/hang`, `demo/web`, FFI, and language wrappers are unchanged because datagrams are opt-in and no higher-level consumer is added here.

## Test Plan

- `cargo test -p moq-net`
- `cargo check -p moq-net --all-targets`
- `cargo clippy -p moq-net --all-targets -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p moq-net --no-deps`
- `bun test` in `js/net`
- `bun run --filter="@moq/net" check`
- `bun biome check js/net/src/track.ts js/net/src/track.test.ts`
- GitHub `Check` workflow passed on head `b694911b568bd5362d4758c2798cb5d8e91cb9ca`

(Written by GPT-5)


doc/concept/layer/moq-lite.md
```diff
@@ -90,6 +91,13 @@ Frames within a group are delivered reliably and in order.
 You can and should take advantage of this, for example using delta encoding.
 If frames within a group are actually independent, you should probably split them into individual groups!
 
+#### Datagrams
+
+As an optimization (lite-05+), the publisher may deliver a small single-frame group as a **datagram** instead of opening a QUIC stream.
+A datagram carries `subscribe ID | group sequence | timestamp | payload` in a single QUIC datagram, routed over the existing subscription: unreliable, unordered, never retransmitted, and capped at ~1200 bytes.
+It is a separate best-effort channel parallel to groups (they share one sequence namespace), suited to tiny latency-critical frames like audio.
+There is no group fallback, so a payload that doesn't fit, or a transport without datagram support (the WebSocket fallback), simply isn't delivered this way.
```
kixelated: remove: "or a transport without datagram support (the WebSocket fallback),"

rs/moq-net/src/model/track.rs
```diff
@@ -1061,6 +1173,70 @@ impl TrackConsumer {
 			state: self.state.clone(),
 		})
 	}
+
+	/// A cursored reader for this track's datagrams, independent of any group subscription.
+	///
+	/// Datagrams (see [`TrackProducer::append_datagram`]) are delivered best-effort on their own
+	/// channel, so they don't need a live group subscription to read. Each call starts a fresh
+	/// cursor at the oldest buffered datagram.
+	pub fn datagrams(&self) -> DatagramConsumer {
```
kixelated: I think remove this and make it part of TrackSubscriber. `recv_datagram`

rs/moq-net/src/lite/subscriber.rs
```diff
@@ -556,6 +558,49 @@ impl<S: web_transport_trait::Session> Subscriber<S> {
...
 	}
 
+	/// Receive QUIC datagrams and route each to its subscription's track producer (lite-05 §6.4).
+	///
+	/// Parks forever on a non-datagram transport or pre-lite-05 version. A decode error or an
+	/// unknown subscribe id drops that datagram without tearing down the session (best-effort);
+	/// only a transport-level failure ends the loop.
+	async fn run_datagrams(self) -> Result<(), Error> {
+		if !self.version.has_datagrams() || self.session.max_datagram_size() == 0 {
+			return std::future::pending::<Result<(), Error>>().await;
```
kixelated: Can you return `Ok` instead? Any tokio::select! should match on Err(x) not res

rs/moq-net/src/lite/publisher.rs
```diff
@@ -912,6 +931,38 @@ impl<S: web_transport_trait::Session> Subscription<S> {
...
 	}
 
+	/// Forward datagrams best-effort until the track finishes, then park so `run_track` alone
+	/// decides when the subscription ends. Errors only if the track aborts.
+	async fn run_datagrams(self, mut datagrams: crate::DatagramConsumer) -> Result<(), Error> {
+		while let Some(datagram) = datagrams.recv().await? {
+			self.serve_datagram(datagram);
+		}
+		// Track finished cleanly: stop sending, but leave termination to run_track.
+		std::future::pending::<Result<(), Error>>().await
```
kixelated: return Ok()? Again I want any tokio::select loop to match on Err (and else) instead of res

rs/moq-net/src/lite/publisher.rs
```diff
@@ -600,21 +607,33 @@ impl<S: web_transport_trait::Session> Publisher<S> {
...
+		tokio::select! {
+			res = sub.run_track(
+				track,
+				subscribe.start_group,
+				subscribe.end_group,
+				&mut stream.reader,
+				&mut stream.writer,
+				&track_priority_tx,
+			) => res?,
+			res = async {
```
kixelated: Err(err) = dg_sub.run_datagrams(datagrams).await

Then return Ok() if they're not supported

doc/concept/layer/moq-lite.md
kixelated: Done — dropped that clause.

rs/moq-net/src/model/track.rs
kixelated: Done. Removed `DatagramConsumer` and moved `recv_datagram`/`poll_recv_datagram` onto `TrackSubscriber` (parallel to `recv_group`). The publisher serves datagrams via a second `TrackSubscriber` from the same track — cheap here since the Rust model shares one `TrackState` (no per-sink mirroring), so it is just an extra cursor.

rs/moq-net/src/lite/subscriber.rs
kixelated: Done — `run_datagrams` now returns `Ok(())` when datagrams are unsupported, so the `Err(err) = dg.run_datagrams()` arm just disables itself and the other arms keep driving the session.

rs/moq-net/src/lite/publisher.rs
kixelated: Done — the trailing `pending()` is gone; `run_datagrams` returns `Ok(())` once the track finishes (and up front when unsupported).

rs/moq-net/src/lite/publisher.rs
kixelated: Done. The select is now `res = sub.run_track(...) => res, Err(err) = dg_sub.run_datagrams(datagrams) => Err(err)`, and `run_datagrams` returns `Ok(())` when unsupported/finished.

js/net/src/connection/accept.ts
```diff
@@ -1,9 +1,16 @@
...
 import * as Ietf from "../ietf/index.ts";
 import * as Lite from "../lite/index.ts";
 import { Stream } from "../stream.ts";
 import type { Established } from "./established.ts";
 import { exchangeSetup } from "./handshake.ts";
 
+// QUIC datagrams only work on a real WebTransport, not a qmux Session (WebSocket fallback),
+// whose datagram streams are inert stubs. Gate on the transport type, mirroring connect().
+function datagramsUsable(transport: WebTransport): boolean {
+	return !(transport instanceof Session);
```
kixelated: We're adding datagrams to qmux, don't add this.

js/net/src/lite/publisher.ts
```diff
@@ -367,6 +396,48 @@ export class Publisher {
 		return pending;
 	}
 
+	// The shared outbound datagram writer, opened lazily (getWriter locks the stream) and
+	// reused across every subscription's datagram loop.
+	#datagrams(): WritableStreamDefaultWriter<Uint8Array> {
+		if (!this.#datagramWriter) {
+			this.#datagramWriter = this.#quic.datagrams.writable.getWriter();
```
kixelated: Why does this need to be lazy initialized? Just do it once in runDatagrams and release the lock in finally.

js/net/src/track.ts
```diff
@@ -128,13 +150,22 @@ export class TrackProducer extends TrackHandle {
...
+	}
+
+	/**
+	 * Adopt an already-handed-out subscriber state as the first sink. Used by the
+	 * on-demand accept path (`TrackRequest.accept`), where the request created the
+	 * subscriber state before the producer existed. Returns `this` for chaining.
+	 *
+	 * @internal
+	 */
+	adoptSink(sink: TrackState): this {
```
kixelated: I don't like this? What's going on, what is this used for?

js/net/src/track.ts
```diff
@@ -258,6 +289,55 @@ export class TrackProducer extends TrackHandle {
...
 
+	// Fan a datagram out to every live subscriber, dropping the oldest once the ring is full.
+	// Late subscribers do NOT replay old datagrams (best-effort, unlike the group cache).
+	#publishDatagram(datagram: Datagram): void {
+		const now = Date.now();
+		for (const sink of this.#sinks) {
+			sink.datagrams.mutate((list) => {
+				list.push({ datagram, time: now });
+				// Drop anything older than the send-buffer window.
+				while (list.length > 0 && now - list[0].time > MAX_DATAGRAM_AGE_MS) list.shift();
```
kixelated, line 292: Enforce this age on read too.

js/net/src/connection/accept.ts
kixelated: Removed the `datagramsUsable` transport-type gate entirely (the accept.ts helper + the connect.ts computation + the threading through Connection/Publisher/Subscriber). Datagrams now gate purely on `hasDatagrams(version) && maxDatagramSize > 0`, matching the Rust side, so qmux picks them up automatically once it reports a non-zero `maxDatagramSize`.

js/net/src/lite/publisher.ts
kixelated: No longer lazy: the writer is acquired once in the Publisher constructor (when `hasDatagrams(version) && maxDatagramSize > 0`) and released in `close()`; its presence is the gate.

It can't be a per-`runDatagrams` getWriter + `finally` release, though: the publisher runs one datagram loop per subscription and they all share the single `datagrams.writable`, and a second `getWriter()` throws `already locked`. So one connection-scoped writer is owned instead of per-loop.

js/net/src/track.ts
kixelated: It's the on-demand accept path. `broadcast.subscribe(name)` hands the caller a `TrackSubscriber` (over a fresh `TrackState`) *before* any producer exists; then when the app calls `request.accept(info)` we need a `TrackProducer` that feeds that already-created state. `adoptSink` attaches it as the producer's first sink (it's the one call site, broadcast.ts).

It exists because of your earlier "TrackState must not be public" ask: I made `TrackState` `@internal` + enabled `stripInternal`, which means the public `new TrackProducer(name)` constructor can't take a `TrackState` param (it'd dangle in the stripped `.d.ts`). `adoptSink` is itself `@internal`, so it's dropped from the published types too — it's purely internal wiring, not public surface.

Happy to reshape it if the name/shape bugs you — e.g. an `@internal` static `TrackProducer.forRequest(name, state)`, or folding it into `TrackRequest`. Left as-is for now; tell me which you'd prefer.

js/net/src/track.ts
kixelated: Done — `recvDatagram` now drops datagrams older than `MAX_DATAGRAM_AGE_MS` before returning (same window already enforced on write in `#publishDatagram`), so a reader that stalled skips stale datagrams instead of replaying them.

PR #2066, opened 2026-07-03
fix(moq-mux): treat H.264 recovery-point SEI as an open-GOP keyframe

### LINKED ISSUE #2050: moq can't round-trip open-GOP broadcast H.264 (recovery-point keyframes)
## Summary

`moq import ts` / `moq export ts` cannot round-trip a real broadcast MPEG-TS
feed (a CNN International EMEA HD capture) whose H.264 uses **recovery-point SEI
open-GOP random access instead of IDR frames**. The H.264 splitter only treats an
`IdrSlice` (NAL type 5) as a keyframe, so on this stream it almost never flags
one. As a result the importer never publishes a video *rendition*, and
`export ts` then aborts on its SCTE-35 program-clock guard:

```
Error: TS export of section-framed verbatim streams (e.g. SCTE-35) requires a video track for the program clock
```

This surfaced while validating a real capture through the merged `test/ts`
compliance harness (`just test ts --source ~/CNNiEMEA2.ts`). The harness did its
job: the source file is clean, but the moq round-trip fails.

## The stream

`~/CNNiEMEA2.ts` (service "CNNI EMEA HD"), PMT PID 100, PCR PID 111:

| PID | Type | Notes |
|---|---|---|
| 111 | 0x1B AVC video | 1920x1080 high@4.0, also the PCR PID |
| 121 | 0x03 MPEG-1 audio (MP2) | decoded |
| 123 | 0x06 private (AC-3 descriptor) | carried verbatim (PES) |
| 131 | 0x06 private (Teletext descriptor) | carried verbatim (PES) |
| 141/142/143 | 0x86 SCTE-35 Splice Info | section-framed verbatim (program-level CUEI) |

NAL histogram over a ~15s / 181,524-video-packet slice:

```
1193 Aud   1193 Sei   1192 NonIdrSlice   54 Pps   26 Sps   1 IdrSlice
```

26 SPS and 54 PPS but only **1 IDR**. The random-access points are non-IDR
I-slices flagged by recovery-point SEI, which is completely standard for
broadcast contribution/distribution H.264.

## Root cause

The Annex-B splitter marks a frame as a keyframe purely from the IDR NAL type:

- `rs/moq-mux/src/codec/h264/split.rs` `decode_nal`: only `Avc3NalType::IdrSlice`
  sets `self.current.contains_idr = true`.
- `maybe_start_frame` then sets `keyframe = self.current.contains_idr`.

The importer only resolves the codec config (and therefore publishes the catalog
video rendition) from an SPS carried on a `keyframe` frame:

- `rs/moq-mux/src/codec/h264/import.rs` `write_frames`:
  `if !self.avc1 && frame.keyframe && let Some(sps) = find_sps(...) { configure_from_sps(...) }`.
- A non-keyframe before the first config is tolerated but produces no rendition;
  a keyframe with no config is a hard error.

So for an open-GOP stream the splitter emits a long run of non-keyframes, the
importer never configures, and `catalog.video.renditions` stays empty.

The failure then gets *reported* by the exporter, which is a red herring:

- `rs/moq-mux/src/container/ts/export.rs` `build_psi` computes `needs_clock`
  (any section-framed verbatim stream, i.e. SCTE-35) and requires a video track
  when so. The section-framed SCTE-35 tracks register in the catalog immediately
  on PMT parse (`register_verbatim`), before any video rendition exists, so the
  exporter builds its PSI, sees SCTE-35 + no video, and aborts.

Without SCTE-35 the export wouldn't error, but there'd still be no video: it
would emit an audio-only program. `export fmp4` confirms this independently,
failing with `cmaf: can't synthesize CMAF init for audio codec Mp2` because the
only rendition that ever appears is the MP2 audio track.

## Reproduction

```bash
just test ts --source ~/CNNiEMEA2.ts     # fails at export (needs a real capture)

# The source itself is well-formed (all hard checks pass):
tsp -I file ~/CNNiEMEA2.ts -P until --packets 200000 -O file /tmp/cnn.ts
./test/ts/run.sh --analyze-only /tmp/cnn.ts   # ts: PASS

# NAL histogram showing open-GOP (1 IDR, many SPS/PPS):
#   RUST_LOG=moq_mux::codec::h264=trace on `cat /tmp/cnn.ts | moq ... import ts`
```

Track set the importer produced (relay round-trip): `catalog.json` + verbatim
`0.ts..4.ts` (AC-3, teletext, 3x SCTE-35). No `.avc3` video, no MP2 rendition
survived to the exporter before it aborted.

## Impact

- Real broadcast H.264 (open-GOP / recovery-point) can't be ingested to a usable
  MoQ broadcast: no video rendition, so nothing downstream (TS/fMP4/HLS export,
  players) can render it. This is not a niche encoder quirk; it's the norm for
  contribution feeds.
- The user-facing error blames SCTE-35, which is misleading. The real problem is
  keyframe detection.

## Possible directions (not implemented)

Ordered roughly by scope. All touch the H.264 codec path, so they target `dev`
(breaking/behavioral change to `rs/moq-mux`) and want new unit tests in
`split.rs` / `import.rs` plus a harness case (a synthetic open-GOP clip).

1. **Recognize recovery-point random access as a keyframe.** Parse the
   recovery-point SEI (payload type 6) and treat an I-slice access unit carrying
   one (or an all-intra AU that re-presents SPS/PPS) as a keyframe in the
   splitter. This is the correct fix but needs careful SEI parsing and a clear
   definition of "keyframe" for open-GOP (recovery vs exact).
2. **Let the importer configure from any SPS, not just a keyframe SPS**, so the
   rendition appears even before the first true keyframe. Decouples "catalog
   rendition exists" from "first group starts"; smaller, but doesn't fix group
   boundaries / tune-in alignment on its own.
3. **Harden the exporter's readiness** so it waits for announced media renditions
   before concluding a program has no video (the SCTE-35 guard races the
   incremental catalog). This removes the misleading error but not the underlying
   no-video problem.

(1) is the substantive fix; (2)/(3) are complementary robustness improvements.

## Relationship to #1979

Closely related, but a **distinct failure mode with a different fix locus**, so
this is a separate issue (cross-link #1979, don't fold in).

- #1979 ("catalog convergence race locks PSI before video track resolves") is an
  **exporter-side timing race**: the video rendition *does* resolve, just late
  (on the first keyframe's inline SPS), and if an audio-only snapshot reaches the
  exporter first, PSI locks audio-only and the late video trips
  `TS track layout changed after PAT/PMT was emitted`. Its proposed fixes wait
  for a "stable"/PMT-complete catalog before locking PSI.
- This issue is **importer-side**: for open-GOP H.264 the video rendition
  *never* resolves (no IDR -> the splitter never flags a keyframe -> the importer
  never configures), so it's not a race. Given unlimited time the video still
  never appears.
- Shared root: both trace to H.264 gating catalog-rendition resolution on the
  first keyframe's inline SPS (`rs/moq-mux/src/codec/h264/import.rs`), which
  #1979 cites directly.
- **Caveat to raise on #1979**: My "wait for the PMT-declared track count"
  proposal would make the exporter **hang forever** on an open-GOP program (the
  PMT declares the video PID, but its rendition never resolves). So #1979's
  exporter-readiness fix needs this importer fix (or a timeout/fallback) to be
  safe on real broadcast feeds.

## Regression?

No. The IDR-only keyframe logic (`contains_idr = true` set only for
`Avc3NalType::IdrSlice`) has been present since the moq-mux backport landed the
H.264 splitter (#1918); it never handled recovery-point open-GOP. This is a
latent gap in a newish feature (TS ingest of real broadcast H.264), not a
regression from previously working behavior. It "looks new" only because verbatim
TS carriage (#1842) and the SRT gateway (#1915) are recent, so real broadcast
feeds with SCTE-35 + open-GOP have only recently been pushed through the path.

## Related

- Cross-package sync: any `rs/moq-mux` container/codec change should be mirrored
  per the table (js/hang, doc/concept) if it changes catalog/container behavior.
- Test coverage: `test/ts` now has a `--via-srt` mode; once fixed, add an
  open-GOP source (real capture or synthetic) as a regression fixture.

kixelated:
## Summary

Fixes [#2050](https://github.com/moq-dev/moq/issues/2050): `moq import ts` / `export ts` could not round-trip real broadcast H.264 (e.g. a CNN International EMEA HD capture) whose random access is signalled by a **recovery-point SEI on a non-IDR I-slice** (open GOP) rather than an IDR frame.

The H.264 Annex-B splitter only flagged an IDR (NAL type 5) as a keyframe, so on such open-GOP streams it never flagged one: the importer never resolved a codec config, no video rendition appeared in the catalog, and every downstream export (TS/fMP4/HLS) had no video. The TS exporter then aborted on its SCTE-35 program-clock guard, misleadingly blaming SCTE-35 for what was really a keyframe-detection gap.

## What changed

`rs/moq-mux/src/codec/h264/split.rs`:
- Parse recovery-point SEI messages (payload type 6) via the existing `h264-parser` dep (`sei_has_recovery_point`).
- Track a per-access-unit `contains_recovery_point` flag, set when the SEI is seen (SEI precedes the slice in an AU, so it is known by the time the I-slice arrives).
- On the first slice of a recovery-point AU, reconcile parameter sets so the frame is self-contained (inline or re-injected SPS/PPS), exactly as an IDR does. Extracted the shared reconcile into a `reconcile_params` helper.
- `maybe_start_frame` now treats `contains_idr || contains_recovery_point` as a keyframe. Any recovery point counts as a random-access point regardless of `recovery_frame_cnt`.

With the recovery-point AU flagged a keyframe, the importer starts a MoQ group there and self-initializes the rendition from its inline SPS, so the round-trip works end to end.

H.265 already handles its open-GOP IRAP types (CRA/BLA), so this is H.264-specific.

## Branch targeting

Targets `main`: this is a bug fix (the splitter now correctly handles input it previously mishandled), not a breaking change to any public API or wire format. The second commit clarifies the Branch Targeting guidance in `CLAUDE.md` to frame `dev` around semver breakage rather than novelty, so fixes like this land on `main`.

## Cross-package sync

No JS/doc mirror needed: this is ingest-side keyframe detection, not a catalog/container/wire-format change. The browser path uses WebCodecs, not this Annex-B splitter.

## Test plan

- [x] New unit tests in `split.rs`: recovery-point I-slice flagged a keyframe; bare recovery-point AU re-injects cached SPS/PPS; a non-recovery SEI stays a delta.
- [x] New unit test in `import.rs`: an open-GOP stream self-initializes the video rendition from the recovery-point keyframe's inline SPS.
- [x] `cargo test -p moq-mux` (343 passed), clippy clean, fmt clean.
- [ ] Not yet exercised against a real open-GOP capture through the `test/ts` harness (tracked as a follow-up).

> [!NOTE]
> This fixes catalog/rendition resolution for open GOPs. Cold **tune-in** into a true open GOP (one with leading pictures that reference the previous GOP) can still show a brief decode artifact until those references stop mattering, since neither the splitter nor the JS consumer drops leading pictures. Tracked separately.

(Written by Claude Opus 4.8)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -157,16 +157,15 @@ Changes in one area usually need matching updates elsewhere, including docs. If 
 
 ## Branch Targeting
 
-Two long-lived branches:
+Two long-lived branches. The split is about **semver breakage, not size or novelty**: `dev` is only for changes that break an existing published contract. Everything else (bug fixes, new behavior, new/additive APIs, docs, refactors) goes to `main`, however large.
 
-- **`main`**: stable. Bug fixes, small additive changes, docs, refactors that preserve public/wire behavior.
-- **`dev`**: staging branch for disruptive work. Target it for:
+- **`main`**: the default. Bug fixes, new behavior, new/additive APIs, docs, and refactors that preserve the existing public/wire contract. A change that only *adds* is additive and lands here even when it is big: a new `pub` item, a new option, or a parser accepting a broader set of inputs it previously rejected. Changing what a component does with input it *already* takes (e.g. recognizing a media pattern it used to mishandle) is a fix, not a break, so it also lands here.
+- **`dev`**: reserved for changes that violate semver by breaking an existing contract. Target it only for:
   - Wire-protocol changes (anything under `rs/moq-net`, including `moq-lite` / `moq-transport` framing or draft bumps).
-  - Breaking changes to public APIs in `rs/moq-ffi`, `rs/libmoq`, `rs/moq-net`, `rs/hang`, `js/net`, `js/hang`, or any of the language wrappers under `swift/`, `kt/`, `go/`, `py/`.
-  - Catalog/container format changes in `rs/hang` or `js/hang`.
-  - Major features that need time to settle before shipping.
+  - Breaking changes to public APIs in `rs/moq-ffi`, `rs/libmoq`, `rs/moq-net`, `rs/hang`, `js/net`, `js/hang`, or any of the language wrappers under `swift/`, `kt/`, `go/`, `py/`. This means a renamed, removed, or signature-changed `pub` item, not a newly *added* one (adding is additive, so it goes to `main`).
+  - Catalog/container format changes in `rs/hang` or `js/hang` that alter existing on-the-wire framing or fields.
 
-`dev` periodically merges into `main` (or vice versa) when the batch is ready to ship. When in doubt, target `main`; reviewers will redirect to `dev` if needed. CI (`pull_request:` workflows) runs on PRs against either branch, so no extra setup is needed when you switch the base.
+`dev` periodically merges into `main` (or vice versa) when the batch is ready to ship. When in doubt, target `main`; reviewers will redirect to `dev` if a change turns out to break an existing contract. CI (`pull_request:` workflows) runs on PRs against either branch, so no extra setup is needed when you switch the base.
 
 ## Workflow
 
--- a/rs/moq-mux/src/codec/h264/import.rs
+++ b/rs/moq-mux/src/codec/h264/import.rs
@@ -332,6 +332,49 @@ mod tests {
 		assert_eq!(h264.level, sps[3]);
 	}
 
+	/// Open-GOP broadcast H.264 (no IDR, random access via recovery-point SEI)
+	/// self-initializes: the splitter flags the recovery-point I-slice AU as a
+	/// keyframe, so the importer resolves the rendition from its inline SPS.
+	#[tokio::test(start_paused = true)]
+	async fn open_gop_self_initializes_from_recovery_point() {
+		let sps: &[u8] = &[
+			0x67, 0x42, 0xc0, 0x1f, 0xda, 0x01, 0x40, 0x16, 0xe9, 0xb8, 0x08, 0x08, 0x0a, 0x00, 0x00, 0x07, 0xd0, 0x00,
+			0x01, 0xd4, 0xc0, 0x80,
+		];
+		let pps: &[u8] = &[0x68, 0xce, 0x3c, 0x80];
+		// recovery-point SEI (payload type 6), then a non-IDR I-slice (type 1).
+		let sei: &[u8] = &[0x06, 0x06, 0x02, 0x00, 0x40, 0x80];
+		let islice: &[u8] = &[0x61, 0xe0, 0x12, 0x34];
+
+		let mut annexb = BytesMut::new();
+		for nal in [sei, sps, pps, islice] {
+			annexb.extend_from_slice(&[0, 0, 0, 1]);
+			annexb.extend_from_slice(nal);
+		}
+
+		let mut split = Split::new();
+		let (track, catalog) = setup("video");
+		let mut import = Import::new(track, catalog.clone());
+
+		let pts = crate::container::Timestamp::from_micros(0).unwrap();
+		let mut frames = split.decode(&annexb, pts).expect("split open-GOP AU");
+		frames.extend(split.flush(pts).expect("flush open-GOP AU"));
+		import.decode(frames).expect("decode open-GOP AU");
+
+		let snapshot = catalog.snapshot();
+		let h264_cfg = snapshot
+			.video
+			.renditions
+			.get("video")
+			.expect("open-GOP stream must resolve a video rendition");
+		let hang::catalog::VideoCodec::H264(h264) = &h264_cfg.codec else {
+			panic!("expected H.264 codec")
+		};
+		assert!(h264.inline, "avc3 source should land as inline=true");
+		assert_eq!(h264.profile, sps[1]);
+		assert_eq!(h264.level, sps[3]);
+	}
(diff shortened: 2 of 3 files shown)


t0ms:
Reminder: this landed on `main`, but the #1979 exporter-readiness fix (#2072, catalog reservation gating) landed on `dev`. Neither alone completes the round-trip for an open-GOP + SCTE-35 contribution feed:

- `main` (this PR) resolves the video rendition for IDR-less H.264, but without #2072 the exporter can still lock PSI on an incomplete catalog.
- `dev` (#2072) gates the catalog until every rendition resolves, but without this PR the open-GOP video rendition never resolves, so the gate correctly stays shut and the catalog never publishes.

Confirmed on the CNN EMEA capture from #2050: `dev` + this PR cherry-picked = full `moq import ts` -> relay -> `moq subscribe --format ts` round-trip passes (`ts: PASS`); `dev` alone hangs the catalog. So this needs to reach `dev` (or `dev` needs to merge `main`) for real broadcast H.264 to round-trip. Details on #1979.

(Written by Claude Opus 4.8)

PR #2072, opened 2026-07-03
feat(moq-mux): gate initial catalog publish until reserved tracks resolve

### LINKED ISSUE #1979: moq-mux TS export: catalog convergence race locks PSI before video track resolves

kixelated:
## Summary

One-shot container muxers (fMP4, MPEG-TS) emit their header exactly once, so they need the *complete* track set before they lock it in. Today the catalog publishes incrementally as each rendition resolves, and codecs resolve at different times (AAC on the first PES, H.264 only on the first keyframe's inline SPS). A subscriber can therefore see an **audio-only** catalog snapshot before the video rendition's config lands, and the TS exporter trips `TS track layout changed after PAT/PMT was emitted` — the convergence race in #1979.

This fixes it entirely **publisher-side**: the first catalog snapshot a subscriber ever sees is already complete. No wire-format or catalog-schema change.

## Approach

- `catalog::Producer::reserve()` returns a clonable `Reserved`. Importers reserve their rendition via `Reserved::init` (with `.video()` / `.audio()` shorthands), which returns a `Rendition` guard.
- The guard **holds its own `Reserved` clone** until its config is `set()` (or it's dropped). The catalog is withheld from the broadcast until every `Reserved` is gone, so an unresolved rendition keeps the gate shut. When the last one resolves, exactly one complete snapshot publishes. A single counter (`reservers`) drives this; a `pending` flag avoids emitting an empty snapshot for a catalog nobody touched.
- Producers that don't reserve publish incrementally, exactly as before.

## Container importers own the reservation

A container importer is the orchestrator that discovers its full track set, so it owns the reservation across discovery — and taking a `Reserved` is the only way to **compose** several importers into one broadcast under one gate (e.g. an fMP4 video source + a separate audio source, or WebRTC's video+audio bridges). Previously each importer reserved internally and gated independently.

- `Reserved::producer()` yields the underlying (non-gating) catalog handle for edits that outlive the reservation.
- `container::{fmp4, ts, flv, mkv}::Import::new` and `import::{Container, ContainerStream, Track, TrackStream}::new` take a `Reserved`. Each holds it across discovery and drops it when its initial set is declared: fMP4 after the moov's traks, MKV after the Tracks element, TS after the first PMT, FLV on the first media frame (all sequence headers precede media data).
- Single-source callers pass `catalog.reserve()`; a caller composing multiple importers passes clones of one `catalog.reserve()` for a shared gate.

### TS deferred substreams

`ts::Import` holds a reservation across the first PMT (bridging the ES loop). **Deferred substreams** (AAC, legacy MP2/AC3/EAC3) whose config comes from the first frame hold *their own* reservation from the PMT until they build the importer, so even audio-only programs converge to a single complete snapshot.

### API

- `VideoTrack<E>` / `AudioTrack<E>` are now aliases of a unified `Rendition<E, K>` over a sealed `Kind` (`Video` / `Audio`).
- Codec-importer constructors take `catalog::Reserved` instead of `catalog::Producer`.

## Public API changes (moq-mux, targets `dev`)

New: `catalog::{Reserved, Rendition, Kind, Video, Audio}`, `Producer::reserve()`, `Reserved::producer()`.
Changed (breaking): `Producer::video_track`/`audio_track` removed; `codec::*::Import::new`, `container::*::Import::new`, and `import::{Container, ContainerStream, Track, TrackStream}::new` take `Reserved` instead of `Producer`. `VideoTrack`/`AudioTrack` retained as aliases.

## Test plan

- [x] `cargo test -p moq-mux` — 354 pass, including gating tests at the `catalog::Producer` level:
  - `reservation_gates_until_all_renditions_resolve` (the #1979 race)
  - `reservation_gate_opens_when_unresolved_reservation_is_dropped`
  - `staged_change_waits_for_a_held_reservation` (deferred / composition case)
  - all fMP4/TS/FLV/MKV round-trip tests exercise container reserve → declare → publish
- [x] clippy / rustfmt / rustdoc clean on moq-mux
- [x] `moq-srt`, `moq-rtc`, `moq-rtmp` build; `moq-ffi`/`libmoq`/`moq-cli` unblocked (local openh264 toolchain issue is environmental)
- [ ] Not runtime-tested against a live SRT → relay → `moq subscribe --format ts` pipeline

## Follow-ups

- **Composition wiring** for `moq-rtc` (share one `Reserved` across a WHIP session's codec bridges) and `moq-gst` (across pads). Both currently reserve per-track, which is correct but gates each track independently; a shared gate needs a per-crate "all tracks negotiated" signal (WebRTC/GStreamer have no clean one — likely drop-on-first-frame).
- `js/hang` catalog Producer parity (wire unchanged, so not required for correctness).

Fixes #1979

(Written by Claude Opus 4.8)

### DIFF
--- a/rs/libmoq/src/publish.rs
+++ b/rs/libmoq/src/publish.rs
@@ -72,13 +72,13 @@ impl Publish {
 		// A container may publish several tracks; a single codec fills one reserved
 		// track. Try the container first so a codec format doesn't reserve a stray
 		// track on the way to being recognized.
-		let media = match import::Container::new(broadcast.clone(), catalog.clone(), format, init) {
+		let media = match import::Container::new(broadcast.clone(), catalog.reserve(), format, init) {
 			Ok(container) => Media::Container(container),
 			Err(moq_mux::Error::UnknownFormat(_)) => {
 				let mut broadcast = broadcast.clone();
 				let name = broadcast.unique_name(&format!(".{format}"));
 				let request = broadcast.reserve_track(name)?;
-				match import::Track::new(request, catalog.clone(), format, init) {
+				match import::Track::new(request, catalog.reserve(), format, init) {
 					Ok(track) => Media::Track(Box::new(track)),
 					Err(moq_mux::Error::UnknownFormat(_)) => return Err(Error::UnknownFormat(format.to_string())),
 					Err(err) => return Err(err.into()),
--- a/rs/moq-cli/src/publish.rs
+++ b/rs/moq-cli/src/publish.rs
@@ -182,7 +182,7 @@ impl Publish {
 				&mut broadcast,
 				moq_mux::catalog::hang::Catalog::<ts::catalog::Ext>::default(),
 			)?;
-			let ts = ts::Import::new(broadcast.clone(), catalog);
+			let ts = ts::Import::new(broadcast.clone(), catalog.reserve());
 			return Ok(Self {
 				source: Source::Stream(PublishDecoder::Ts(Box::new(ts))),
 				broadcast,
@@ -193,7 +193,7 @@ impl Publish {
 		let source = match format {
 			PublishFormat::Avc3 => {
 				let track = moq_mux::import::unique_track(&mut broadcast, ".avc3")?;
-				let import = moq_mux::codec::h264::Import::new(track, catalog.clone());
+				let import = moq_mux::codec::h264::Import::new(track, catalog.reserve());
 				let split = Box::new(moq_mux::codec::h264::Split::new());
 				Source::Stream(PublishDecoder::Avc3 {
 					split,
@@ -201,12 +201,12 @@ impl Publish {
 				})
 			}
 			PublishFormat::Fmp4 => {
-				let fmp4 = fmp4::Import::new(broadcast.clone(), catalog.clone());
+				let fmp4 = fmp4::Import::new(broadcast.clone(), catalog.reserve());
 				Source::Stream(PublishDecoder::Fmp4(Box::new(fmp4)))
 			}
 			PublishFormat::Ts => unreachable!("TS is handled above with the mpegts catalog extension"),
 			PublishFormat::Flv => {
-				let flv = flv::Import::new(broadcast.clone(), catalog.clone());
+				let flv = flv::Import::new(broadcast.clone(), catalog.reserve());
 				Source::Stream(PublishDecoder::Flv(Box::new(flv)))
 			}
 		};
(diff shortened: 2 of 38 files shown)


kixelated:
Filed the follow-ups referenced above:
- #2075 — mirror the reservation gating in `@moq/hang` (JS parity; wire unchanged)
- #2076 — share one `Reserved` across a moq-rtc session's codec bridges, reserving from the SDP up front (all m-lines known before media flows)

(Written by Claude Opus 4.8)

PR #2078, opened 2026-07-04
moq-net: per-project stats via broadcast-path grouping

kixelated:
## Motivation

`moq_net::Stats` publishes a single `<prefix>/node/<node>` broadcast carrying every broadcast's counters mixed together, keyed by path inside each JSON frame. A consumer that only cares about one tenant/project therefore has to subscribe to every node's full stats and filter, an N-squared fan-in, and it can't announce-scope to just the paths it wants.

## Change

Add a `depth` knob to `StatsConfig` (default `0` = unchanged: one `<prefix>/node/<node>` broadcast for the whole node).

At depth `N` the aggregator buckets broadcasts and session roots by their first `N` path segments and publishes one broadcast per group at `<prefix>/<group>/node/<node>`, spawning a group's broadcast when it first has traffic and unannouncing it once it drains. The depth-`0` empty group stays announced for the aggregator's life, so existing single-broadcast deployments are byte-for-byte unchanged.

This lets a consumer announce-scope to a single group, for example watch `.stats/<tenant>/**`, and receive only the nodes actually carrying that tenant.

- `run_publisher` now maintains a `HashMap<group, GroupPublisher>` instead of one fixed broadcast, with per-group change detection and GC state.
- `advertised_path` takes the group segment; `group_key(path, depth)` uses the first `N` segments.
- Session tracks bucket by their auth root, so they land in the right group's broadcast.
- Exposed on `moq-relay` as `[stats].depth`, `--stats-depth`, and `MOQ_STATS_DEPTH`.
- Relay config docs describe the grouped stats layout and new `depth` option.

## Public API

- Added `moq_net::StatsConfig::depth`.
- Added `moq_net::StatsConfig::with_depth`.
- Added relay config and CLI/env options for `[stats].depth`, `--stats-depth`, and `MOQ_STATS_DEPTH`.

These are additive changes and target `main`.

## Tests

- `cargo fmt --all`
- `cargo test -p moq-net`
- `cargo test -p moq-net stats`
- `cargo test -p moq-relay`
- `cargo test -p moq-relay config::tests::cli_does_not_clobber_toml_stats_enabled`
- `cargo clippy -p moq-net -p moq-relay --all-targets -- -D warnings`
- `bun remark doc/bin/relay/config.md --quiet --frail`

(Written by GPT-5)

### DIFF
--- a/rs/moq-net/src/stats.rs
+++ b/rs/moq-net/src/stats.rs
@@ -126,7 +126,7 @@
 //! itself generate more stats traffic.
 
 use std::{
-	collections::{BTreeMap, HashMap},
+	collections::{BTreeMap, HashMap, HashSet},
 	sync::{
 		Arc, Weak,
 		atomic::{AtomicU64, Ordering},
@@ -286,6 +286,16 @@ pub struct StatsConfig {
 	pub node: Option<PathOwned>,
 	/// How long the snapshot task waits between publishes. Default 1s.
 	pub interval: Duration,
+	/// How many leading path segments of each broadcast to use as a grouping
+	/// key, splitting the output into one broadcast per group at
+	/// `<prefix>/<group>/node/<node>`. Default `0`: a single
+	/// `<prefix>/node/<node>` broadcast carrying every path (the historical
+	/// behavior). `1` buckets by the first segment (e.g. a per-tenant broadcast),
+	/// so a consumer can announce-scope to just that group instead of slurping
+	/// every node's full stats. A group broadcast is announced while its group
+	/// has live traffic and unannounced once it drains; at depth `0` the single
+	/// broadcast stays announced for the aggregator's life even while idle.
+	pub group_depth: usize,
 }
 
 impl StatsConfig {
@@ -298,6 +308,7 @@ impl StatsConfig {
 			prefix: PathOwned::from(".stats"),
 			node: None,
 			interval: Duration::from_secs(1),
+			group_depth: 0,
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


rs/moq-net/src/stats.rs
```diff
@@ -286,6 +286,16 @@ pub struct StatsConfig {
...
+	/// How many leading path segments of each broadcast to use as a grouping
+	/// key, splitting the output into one broadcast per group at
+	/// `<prefix>/<group>/node/<node>`. Default `0`: a single
+	/// `<prefix>/node/<node>` broadcast carrying every path (the historical
+	/// behavior). `1` buckets by the first segment (e.g. a per-tenant broadcast),
+	/// so a consumer can announce-scope to just that group instead of slurping
+	/// every node's full stats. A group broadcast is announced while its group
+	/// has live traffic and unannounced once it drains; at depth `0` the single
+	/// broadcast stays announced for the aggregator's life even while idle.
+	pub group_depth: usize,
```
kixelated: call it `depth`

rs/moq-relay/src/stats.rs
```diff
@@ -63,6 +63,15 @@ pub struct StatsConfig {
...
 	pub node: Option<String>,
+
+	/// Number of leading broadcast-path segments to bucket stats by, one
+	/// broadcast per bucket at `<prefix>/<group>/node/<node>`. Defaults to 0: a
+	/// single `<prefix>/node/<node>` broadcast for the whole node. Set to 1 to
+	/// publish a per-first-segment broadcast (e.g. per tenant), so a consumer can
+	/// announce-scope to just that group rather than slurping every node's full
+	/// stats. See [`moq_net::StatsConfig::group_depth`].
+	#[arg(long = "stats-group-depth", env = "MOQ_STATS_GROUP_DEPTH")]
+	pub group_depth: Option<usize>,
```
kixelated: same call it depth

PR #2085, opened 2026-07-04
fix: moq-net wasm compatibility

ewindisch:
This fixes wasm32 support for moq-net by replacing tokio::time with a WASM compatible wrapper (deferring to tokio::time on native).

It also drops Send + Sync bounds on sessions when compiled to the WASM target.

PR #2086, opened 2026-07-04
[codex] backport moq-wasm to main

kixelated:
## Summary
- Backport the original moq-wasm PR (#1673) to main, including the Rust `moq-wasm` crate, the `@moq/wasm` package, wasm target config, and the build recipe.
- Build on top of the wasm compatibility work now merged in #2085, so `moq-net`'s browser wasm clock/session support comes from main.
- Carry forward the relevant #1726 runtime fix: read cached frames in place instead of minting a transient `FrameConsumer` on every poll, which avoids a wasm single-thread self-wake freeze.

## Validation
- `cargo fmt`
- `cargo shear`
- `cargo check -p kio`
- `cargo check -p moq-net`
- `cargo check -p moq-wasm`
- `cargo check -p moq-wasm --target wasm32-unknown-unknown`
- `CARGO_TARGET_DIR=target CARGO_TARGET_WASM32_UNKNOWN_UNKNOWN_RUNNER=wasm-bindgen-test-runner cargo test --test wasm -p moq-net --target wasm32-unknown-unknown`
- `CARGO_TARGET_DIR=target cargo build -p moq-wasm --target wasm32-unknown-unknown --profile wasm-release`
- `wasm-bindgen --target web --out-name moq --out-dir js/wasm/dist target/wasm32-unknown-unknown/wasm-release/moq_wasm.wasm`

## Notes
- `just` is unavailable in this environment, so I used the underlying Cargo and wasm-bindgen commands directly.
- From #1726, I carried forward `c4992f2`'s real frame-read freeze fix. The kio waiter split from `2231f94` is already present on current main, and the temporary debug commits were intentionally not carried forward.
- This intentionally does not include the later broad JS client/demo rewrite from `dev`; this PR is scoped to the moq-wasm backport and the wasm runtime fixes needed to use it.

(Written by GPT-5)


.cargo/config.toml
```diff
@@ -0,0 +1,8 @@
+[build]
+rustflags = ["--cfg", "tokio_unstable"]
```
kixelated: I don't think we need tokio_unstable any longer

PR #2089, opened 2026-07-04
[codex] Add catalog reservation gating

### LINKED ISSUE #2075: Mirror catalog reservation gating in @moq/hang (js/hang)
## Background

[#2072](https://github.com/moq-dev/moq/pull/2072) added **publisher-side catalog reservation gating** to `moq-mux` (Rust), fixing the convergence race in #1979 where a one-shot muxer (fMP4, MPEG-TS) sees an incomplete, e.g. audio-only, catalog snapshot before a later rendition's config resolves.

The mechanism, publisher-side only (wire and catalog schema unchanged):

- `catalog::Producer::reserve()` returns a clonable `Reserved`.
- An importer reserves its rendition via `Reserved::init` (`.video()` / `.audio()`), getting a `Rendition` guard that **holds its own `Reserved` clone until its config is `set()`** (or it's dropped).
- The catalog is **withheld from the broadcast until every `Reserved` is gone**, so an unresolved rendition keeps the gate shut. When the last resolves, exactly one complete snapshot publishes. A `pending` flag avoids emitting an empty snapshot for an untouched catalog.
- Container importers own a reservation across track discovery and drop it once their initial set is declared, which also lets several importers compose into one broadcast under a single shared gate.

## Why mirror it in JS

`@moq/hang`'s catalog `Producer` ([`js/hang/src/catalog/producer.ts`](https://github.com/moq-dev/moq/blob/dev/js/hang/src/catalog/producer.ts)) publishes incrementally as tracks are added, with no gate. A browser publisher that adds video and audio in separate ticks emits a partial catalog first, so any consumer that can't reinitialize (or just wants a stable first snapshot) hits the same race the Rust side just fixed. Parity keeps the two producers behaviourally aligned even though the wire doesn't force it.

## Proposed direction

Add an equivalent reservation to the JS catalog `Producer` so the first published snapshot is complete:

- A `reserve()` returning a `Reserved` handle, and a per-rendition guard that holds the reservation until its config is set (or it's dropped).
- Withhold the initial publish until all reservations resolve; publish incrementally afterwards.
- Producers that don't reserve keep publishing incrementally (opt-in, non-breaking for existing callers that never call `reserve()`).

## Design decisions to settle (why this is an issue, not a mechanical port)

- **TS shape.** Rust uses `Reserved` + `Rendition<K>` + a sealed `Kind`. TS has no importer-per-codec layer in the same shape, so decide the ergonomic surface: a `reserve()` + returned rendition handle, an options flag, or an explicit `producer.complete()` call. The `signals`/`Effect` lifecycle model may suggest a different idiom.
- **Where the gate lives.** `@moq/hang`'s publish side (`js/hang/publish`) is structured differently from Rust's container/codec importers; identify the analogous "declare all tracks, then publish" point.
- **Composition.** Whether JS needs the shared-gate-across-importers property at all, or only the single-producer complete-first-snapshot behaviour.

## Scope / logistics

- Wire and catalog JSON schema **unchanged** — pure publisher-side timing.
- API addition to `@moq/hang`; if it changes existing shapes, target **`dev`** per branch-targeting rules.
- Cross-package sync: `demo/web` if it drives the catalog producer directly.

_Follow-up to #2072 (Rust). Fixes the JS half of #1979-style convergence._

kixelated:
## Summary

- Add opt-in first-snapshot reservation gating to the publish catalog producer.
- Add closeable audio and video rendition guards that insert configs, release the initial gate, update set configs, and remove their catalog entries on close.
- Export the catalog producer API from `@moq/publish` and cover reserved initial publish behavior in tests.

## Public API

- New `CatalogReservation` interface in `@moq/publish`.
- New `ReservedRendition<T>` interface in `@moq/publish`.
- New `CatalogProducer.reserve()` method.
- New root export for `./catalog` from `@moq/publish`.

## Testing

- `env TMPDIR=/private/tmp bun run --filter='@moq/publish' test`
- `env TMPDIR=/private/tmp bun run --filter='@moq/publish' check`
- `env TMPDIR=/private/tmp bun run --filter='@moq/publish' build`
- `env TMPDIR=/private/tmp bun run --filter='*' check`
- `env TMPDIR=/private/tmp bun biome check js/publish/src/catalog.ts js/publish/src/catalog.test.ts js/publish/src/index.ts`
- `git diff --check`

Fixes #2075.

(Written by GPT-5)

### DIFF
--- a/js/publish/src/catalog.test.ts
+++ b/js/publish/src/catalog.test.ts
@@ -125,34 +125,6 @@ test("untouched reservations do not publish an empty catalog", async () => {
 	effect.close();
 });
 
-test("removing the last rendition keeps a section with display metadata", async () => {
-	const catalog = new CatalogProducer();
-	const reserved = catalog.reserve();
-	const video = reserved.video("video/hd");
-	reserved.close();
-
-	const effect = new Effect();
-	const track = new TrackProducer("catalog.json");
-	catalog.serve(track, effect);
-	const consumer = new Json.Consumer<Catalog.Root>(track.subscribe());
-	const first = consumer.next();
-
-	video.set(videoConfig);
-	await first;
-
-	// Display metadata lives outside the rendition guard, so removing the last rendition must keep it.
-	const display = { width: Catalog.u53(1920), height: Catalog.u53(1080) };
-	catalog.mutate((c) => {
-		c.video = { renditions: { "video/hd": videoConfig }, display };
-	});
-	await consumer.next();
-
-	video.close();
-	expect(await consumer.next()).toEqual({ video: { renditions: {}, display } });
-
-	effect.close();
-});
-
 test("catalog producer publishes every update as a snapshot group", async () => {
 	const catalog = new CatalogProducer();
 	catalog.mutate((c) => {
--- a/js/publish/src/catalog.test.ts
+++ b/js/publish/src/catalog.test.ts
@@ -125,6 +125,37 @@ test("untouched reservations do not publish an empty catalog", async () => {
 	effect.close();
 });
 
+test("removing the last rendition keeps a section with display metadata", async () => {
+	const catalog = new CatalogProducer();
+	const reserved = catalog.reserve();
+	const video = reserved.video("video/hd");
+	reserved.close();
+
+	const effect = new Effect();
+	const track = new TrackProducer("catalog.json");
+	catalog.serve(track, effect);
+	const consumer = new Json.Consumer<Catalog.Root>(track.subscribe());
+	const first = consumer.next();
+
+	video.set(videoConfig);
+	await first;
+
+	// Display metadata lives outside the rendition guard, so removing the last rendition must keep it.
+	catalog.mutate((c) => {
+		const section = c.video as {
+			renditions: Record<string, Catalog.VideoConfig>;
+			display: { width: number; height: number };
+		};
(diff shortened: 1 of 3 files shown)


kixelated:
Closing in favor of a fresh implementation against current dev.

This branch is ~41 commits behind dev and no longer builds on it: dev has since landed #2120 (the js/net `Track.Producer` PascalCase rename) and refactored `js/publish/src/catalog.ts`, so the `TrackProducer` imports here break when CI merges the PR with dev. That, not any logic issue, is why `Check` fails (it passed on 2026-07-04 against the older dev, and fails now that CI re-merges with a dev that renamed the API).

Beyond staleness, the approach needs rethinking:
- Per #2075 the gate belongs in `@moq/hang`'s catalog producer (`js/hang/src/catalog/producer.ts`), not `@moq/publish`.
- The `reserve()`/clone/`Rendition` shape is a direct port of the Rust API; the JS idiom (signals/Effect, plus dev's new "avoid callback parameters" convention) points to a smaller, different surface.
- Nothing consumes `reserve()` here, so the public API is speculative.

The underlying need stands and #2075 stays open to track it: withhold the catalog until all tracks are initialized so one-shot consumers (e.g. fMP4) see a complete, mostly-static catalog.

(Written by Claude Opus 4.8)

PR #2090, opened 2026-07-04
[codex] moq-rtc: share catalog reservation across ingest bridges

### LINKED ISSUE #2076: moq-rtc: share one catalog reservation across a session's codec bridges
## Background

[#2072](https://github.com/moq-dev/moq/pull/2072) added catalog reservation gating to `moq-mux`: `catalog::Producer::reserve()` returns a clonable `Reserved`, and the catalog is withheld from the broadcast until every `Reserved` is dropped. Passing **clones of one `Reserved`** to several importers composes them under a single gate, so the first published catalog snapshot contains *all* of a broadcast's renditions rather than whichever resolved first.

`moq-rtc` was left out of that composition wiring (noted as a follow-up in the PR). Today `IngestSink::on_track` builds each codec bridge (`opus` / `h264` / `vp8` / `vp9`) with an independent `catalog.reserve()` ([`rs/moq-rtc/src/ingest.rs`](https://github.com/moq-dev/moq/blob/dev/rs/moq-rtc/src/ingest.rs)), so a WHIP session's video and audio renditions gate independently and a one-shot muxer downstream can still see a video-only (or audio-only) first snapshot.

## Proposed approach: reserve from the SDP up front

Unlike a raw byte stream, WebRTC negotiates **every m-line in the SDP offer/answer before any media flows**, so the full track set is known ahead of time. So `moq-rtc` can gate cleanly without the drop-on-first-frame heuristic the other gateways need:

1. When the offer/answer is applied, `IngestSink` holds one `catalog.reserve()` (the shared gate) for the session.
2. Each `on_track` bridge is built from a **clone** of that shared `Reserved`, so its rendition joins the same batch.
3. Drop the shared reservation once every negotiated m-line has its bridge (the count is known from the SDP). The gate then opens as the bridges resolve their configs, publishing one complete catalog.

This needs the codec bridges to take a `Reserved` (deriving a `Producer` via `Reserved::producer()` where they edit the catalog directly, as the container importers now do), and `IngestSink` to learn the negotiated m-line count so it knows when the batch is complete.

## Scope

- `rs/moq-rtc` only; wire unchanged.
- Bridge constructor signatures change (`Producer` → `Reserved`) — internal to the crate, but worth landing on `dev` alongside the rest.
- Same pattern would apply to WHEP egress if it ever composes multiple sources.

_Follow-up to #2072._

kixelated:
## Summary

- Share one `moq_mux::catalog::Reserved` gate across a WebRTC ingest session codec bridges.
- Count accepted ingest m-lines from SDP answers so rejected or inactive media does not hold the catalog gate.
- Update Opus, H.264, VP8, and VP9 bridge constructors to take a reservation, with VP8/VP9 holding it until their lazy first-frame catalog entry resolves.
- Add a regression test that prevents an audio-only first catalog snapshot when negotiated video has not resolved yet.

## Root Cause

`IngestSink::on_track` built each codec bridge from an independent catalog reservation, so a WHIP/WHEP ingest session could publish the first catalog snapshot as soon as one bridge resolved. A downstream one-shot muxer could therefore see an incomplete first catalog.

## Public API Changes

None. The changed bridge constructors are in crate-private `moq-rtc` modules.

## Cross-Package Sync

`rs/moq-rtc` only. Wire format, `moq-net`, `hang`, JS packages, and docs are unchanged.

## Validation

- `cargo fmt`
- `cargo test -p moq-rtc`
- `git diff --check`

Fixes #2076.

(Written by GPT-5)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4424,6 +4424,7 @@ dependencies = [
  "axum",
  "bytes",
  "hang",
+ "kio",
  "moq-mux",
  "moq-net",
  "reqwest",
--- a/rs/moq-rtc/Cargo.toml
+++ b/rs/moq-rtc/Cargo.toml
@@ -34,4 +34,5 @@ url = "2"
 uuid = { version = "1", features = ["v4"] }
 
 [dev-dependencies]
+kio = { workspace = true }
 tokio = { workspace = true, features = ["test-util"] }
--- a/rs/moq-rtc/src/client/whep.rs
+++ b/rs/moq-rtc/src/client/whep.rs
@@ -18,8 +18,6 @@ use url::Url;
 use crate::{Error, Result, client::Client, ingest::IngestSink, session};
 
 pub(crate) async fn dial(client: &Client, url: Url, broadcast: moq_net::broadcast::Producer) -> Result<()> {
-	let sink = Box::new(IngestSink::new(broadcast)?);
-
 	let (socket, candidates) = session::bind_udp(&client.config().ice_candidates).await?;
 	let mut rtc = Rtc::new(Instant::now());
 	for addr in &candidates {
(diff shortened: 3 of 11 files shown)


kixelated:
Closing: this gates the first catalog on every negotiated m-line resolving, but for arbitrary WHIP publishers that condition may never hold (muted/delayed/never-sent video), so the catalog — audio included — can be withheld indefinitely. Incremental catalog is the correct behavior for RTC ingest; the reservation gate stays for muxer-ingest paths where the track set is known up front. #2076 for the WHIP path is better handled by tolerating catalog updates (or a bounded timeout at the consumer) rather than blocking ingest.

Issue #2087, opened 2026-07-04
Go: Catalog becomes unavailable after a minute of playing


pisoj:
If I create a track, then publish the broadcast and start writing audio to that track, the stream breaks after about a minute of playing.
```go
mediaProducer, err := broadcast.PublishMedia("aac", []byte{/* AAC audio specific config */})

origin.Publish("listen_to_me.hang", broadcast)
produceAudio(mediaProducer)
```

This is the error returned from the player:
```c
broadcast=listen_to_me.hang track=catalog.json error=Received RESET_STREAM.
```

However, if I publish the broadcast inside a loop, even if that loop **only repeats once**, the stream continues indefinitely **without any errors**:
```go
for i := 0; i < 1; i++ {
	origin.Publish("listen_to_me.hang", broadcast)
	produceAudio(mediaProducer)
}
```

### A full project to reproduce this can be found [here](https://github.com/pisoj/moq-bug).

kixelated:
Ah yeah I think you have to keep a reference to the `broadcast` around otherwise it gets dropped by the GC. Lemme see if I can make a more ergonomic API.

pisoj:
Can confirm, this is GC related. If I put any operation involving `broadcast` after `produceAudio` the issue goes away.
```go
origin.Publish("listen_to_me.hang", broadcast)
produceAudio(mediaProducer)

broadcast.Finish()
```

I also noticed that `init` data passed to `broadcast.PublishMedia` only gets used to populate `codec/sampleRate/numberOfChannels` in the catalog but doesn't get copied to the `description` field. This causes `moq-cli` to fail when subscribing `Error: AAC track missing AudioSpecificConfig` while the track plays perfectly in the browser.

Furthermore, is there any way for me to specify the **bitrate** of each track, like it's done [here](https://github.com/moq-dev/moq/blob/2d352bdf0d3c5c595b1494a1e1813903011196d7/rs/hang/examples/video.rs#L56)? I haven't found any catalog specific functions apart from `broadcast.SetCatalogSection` exposed. My goal is for the Javascript player to pick the best track a network can support, currently it picks whichever is defined first.

For reference, my catalog looks like this:
```json
{
    "video": {
        "renditions": {}
    },
    "audio": {
        "renditions": {
            "0.aac": {
                "codec": "mp4a.40.2",
                "container": {
                    "kind": "legacy"
                },
                "sampleRate": 44100,
                "numberOfChannels": 2
            },
            "1.aac": {
[...]

PR #2100, opened 2026-07-06
[codex] auto-detect catalog metrics

kixelated:
## Summary

- auto-detect track **bitrate** in moq-mux, reporting it only after group boundaries so an isolated large keyframe isn't reported as the steady-state bitrate
- combine the existing per-frame jitter detector and the new bitrate detector into one `Metrics` type, wired through every codec importer (aac, av1, flac, h264, h265, legacy, mp3, opus, vp8, vp9) and the fmp4 importer
- detected values are written straight to the catalog's existing `bitrate` / `jitter` fields (no schema change, no new public API)

The catalog field stays named `jitter` (an earlier iteration renamed it to `latencyMin`; that rename was dropped as not worth the churn).

Refs #2087

## Validation

- `cargo test -p hang -p moq-mux -p libmoq` (green)
- `cargo clippy -p hang -p moq-mux` (clean)
- `bunx biome check` on the touched JS

## Notes

- Extending metric detection to the audio codecs means an audio track now publishes a jitter catalog update mid-stream; the libmoq `multiple_frames_ordering` test was updated to drain those mid-stream catalog snapshots before asserting the terminal callback.
- The per-video/audio bitrate+jitter application logic is deduped behind a small `MetricsTarget` trait + `Update::apply`.

(Written by Claude Opus 4.8)


js/hang/src/catalog/audio.ts
```diff
@@ -33,12 +33,13 @@ export const AudioConfigSchema = z.object({
...
-	// The player's jitter buffer should be larger than this value.
-	// If not provided, the player should assume each frame is flushed immediately.
+	// Minimum additional latency required by this track in milliseconds.
+	// This is added to the subscriber's own latency target for steady playback.
 	//
 	// NOTE: The audio "frame" duration depends on the codec, sample rate, etc.
 	// ex: AAC often uses 1024 samples per frame, so at 44100Hz, this would be 1024/44100 = 23ms
+	latencyMin: z.optional(u53Schema),
+
 	jitter: z.optional(u53Schema),
```
kixelated, line 42: add a note that this is deprecated, we're renaming to `latencyMin`. 

Is there a better way to do this via ZOD?

rs/moq-mux/src/container/fmp4/import.rs
```diff
@@ -720,14 +725,20 @@ impl<E: crate::catalog::hang::CatalogExt> Import<E> {
...
 			let mut frame = g.create_frame(moq_net::Frame {
 				size: fragment_bytes.len() as u64,
 			})?;
+			let bytes = fragment_bytes.len();
 			frame.write(fragment_bytes)?;
 			frame.finish()?;
 
 			track.group = Some(g);
+			let update = track.metrics.observe_frame(timestamp, bytes);
+			update_track_bitrate(&mut self.catalog, track, update)?;
```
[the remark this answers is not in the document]
kixelated: Fixed in 53b61d4cc — dropped the no-op call; `observe_frame` now just feeds the internal bitrate accumulator, with a comment noting bitrate reaches the catalog via `finish_group`.

js/hang/src/catalog/audio.ts
kixelated: Added in b8637c406 on both `jitter` fields (audio + video).

On the Zod question: v4's native way is `.meta({ deprecated: true })` — `deprecated` is a typed key on `JSONSchemaMeta`, so it validates and flows into `z.toJSONSchema` output. I paired it with a source note pointing to `latencyMin`. One caveat: neither `.meta()` nor a JSDoc comment on a schema field propagates to the `z.infer` type, so editors won't strike through `config.jitter` at call sites — Zod can't inject `@deprecated` into an inferred property. The only way to get that editor hint would be to hand-write the type instead of inferring it, which isn't worth it here.

kixelated: Correction: these schemas are Zod **Mini** (`import * as z from "zod/mini"`), which has no chainable `.meta()` — my first push (b8637c406) broke the build with TS2339. Fixed in 294737912 using the Zod Mini equivalent: `.register(z.globalRegistry, { deprecated: true })`. Same effect (the `deprecated` flag lands in the global registry / `z.toJSONSchema` output); the caveat about it not reaching the inferred TS type still stands.

kixelated: Update: we decided not to rename after all, so `jitter` stays the canonical field name and the deprecation marker was removed in 4473fd2ee. Nothing here is deprecated now.

kixelated:
Superseded by #2124, which builds the same bitrate/metrics detection on dev's generic `Rendition<E, K>` instead of main's older two-struct `VideoTrack`/`AudioTrack` layout. dev already had the generic rendition unification, so the metrics detector now lives on it (Kind trait `jitter_mut`/`bitrate_mut` accessors, no per-kind duplication). Targeting dev since it touches the rendition surface.

(Written by Claude Opus 4.8)

PR #2108, opened 2026-07-07
feat(moq-net): require broadcast close() + propagate real errors on abrupt teardown (#2087)

kixelated:
## Summary

Addresses #2087 and the broader problem that **relying on `Drop` to end a producer is a bug — it surfaces a generic `Error::Dropped` and throws away the real cause.**

### 1. Require an explicit `broadcast::Producer::close()`

On `dev` a broadcast ends purely when its last producer drops. Per maintainer direction (the RAII `origin::Publish` guard only invites accidental teardown), require an explicit close:

- `broadcast::Producer::close()` — clean `self`-consuming shutdown; dropping the last producer without it logs a warning. Lifetime docs on `broadcast::Producer` + `Info::produce`.

### 2. Warn when any producer is dropped without a clean end

`track` / `group` / `frame` producers each warn when dropped without `finish()`/`abort()` (frame: before all bytes written), gated on their existing abrupt-teardown condition so a proper finish/abort stays silent.

### 3. Pipe `abort(err)` through the whole muxer stack + gateways

The muxers only had `finish()` — no way to convey an error. Added `abort(err)` everywhere and migrated the gateways so the real cause reaches subscribers:

- **moq-mux (complete):** every codec importer (h264, h265, av1, vp8, vp9, aac, opus, mp3, flac, legacy), every container importer (fmp4, mkv, ts + its per-codec substreams/sections/verbatim, flv), and the `import::{Track, TrackStream, Container, ContainerStream}` wrappers.
- **Gateways migrated to `finish()` on clean end / `abort(err)` on failure:** RTMP (ingest + pull), SRT, moq-cli publish, moq-video capture.
- **Teardowns that are genuinely at drop** (WebRTC disconnect, HLS cancellation — an async `Drop` can't `finish()`): moq-rtc `IngestSink` and moq-hls `Import` abort with `Cancel` (a real code, not the implicit `Dropped`). moq-rtc's `codec::Bridge` gains `abort()`; moq-audio/moq-video gain `abort()`.

## Remaining / follow-ups

- **Revert the RAII `origin::Publish` guard** (owner: @kixelated, separate session) so `close()` is the canonical broadcast end. Broadcast has no error channel on dev, so gateway broadcast-close is coupled to that work.
- **moq-gst** (the plugin's clean EOS path already finishes) and the **FFI/libmoq** boundary are the last two consumers; the `abort()` primitive is in place for them.

## Branch targeting

`dev` — coordinated with the broadcast-model rework there.

## Test plan

- [x] `cargo test` green: moq-net (426), moq-mux (354), moq-rtmp (18), moq-srt, moq-cli, moq-rtc, moq-hls, moq-audio
- [x] `cargo clippy` + `cargo fmt --check` clean on touched crates
- [x] Full-workspace build (moq-video's openh264-sys2 native build fails in this env — a pre-existing toolchain issue, unrelated to the change)

(Written by Claude Opus 4.8)

### DIFF
--- a/rs/libmoq/src/publish.rs
+++ b/rs/libmoq/src/publish.rs
@@ -61,8 +61,15 @@ impl Publish {
 		Ok((broadcast, catalog))
 	}
 
+	/// Finalize the catalog stream and cleanly close the broadcast, so subscribers
+	/// see a normal end rather than [`moq_net::Error::Dropped`].
 	pub fn close(&mut self, broadcast: Id) -> Result<(), Error> {
-		self.broadcasts.remove(broadcast).ok_or(Error::BroadcastNotFound)?;
+		let (broadcast, mut catalog) = self.broadcasts.remove(broadcast).ok_or(Error::BroadcastNotFound)?;
+		// Close the broadcast even if finalizing the catalog fails, so the clean end
+		// still reaches subscribers.
+		let result = catalog.finish();
+		broadcast.close();
+		result?;
 		Ok(())
 	}
 
--- a/rs/libmoq/src/publish.rs
+++ b/rs/libmoq/src/publish.rs
@@ -61,15 +61,14 @@ impl Publish {
 		Ok((broadcast, catalog))
 	}
 
-	/// Finalize the catalog stream and cleanly close the broadcast, so subscribers
+	/// Cleanly close the broadcast and finalize the catalog stream, so subscribers
 	/// see a normal end rather than [`moq_net::Error::Dropped`].
 	pub fn close(&mut self, broadcast: Id) -> Result<(), Error> {
 		let (broadcast, mut catalog) = self.broadcasts.remove(broadcast).ok_or(Error::BroadcastNotFound)?;
-		// Close the broadcast even if finalizing the catalog fails, so the clean end
-		// still reaches subscribers.
-		let result = catalog.finish();
+		// Close the broadcast first so the clean end reaches subscribers even if
+		// finalizing the catalog fails.
 		broadcast.close();
-		result?;
+		catalog.finish()?;
 		Ok(())
 	}
 
--- a/rs/moq-boy/src/main.rs
+++ b/rs/moq-boy/src/main.rs
@@ -278,11 +278,21 @@ async fn run(config: &Config) -> Result<()> {
 		move || run_emulator(session, &rom_path, audio_encoder, status_publisher, cmd_rx)
 	});
 
-	tokio::select! {
-		res = emulator_handle => res?.context("emulator error"),
-		res = reconnect.closed() => Ok(res?),
+	// Keep the result out of the `select!` arms (a `?` there would return before
+	// the close below runs), so the broadcast is always closed on shutdown.
+	let result = tokio::select! {
+		res = emulator_handle => match res {
+			Ok(inner) => inner.context("emulator error"),
+			Err(join) => Err(join.into()),
+		},
+		res = reconnect.closed() => res.map_err(Into::into),
 		res = input::handle_viewers(&mut viewer_consumer, &cmd_tx) => res,
-	}
+	};
+
+	// Cleanly close the broadcast so subscribers see a normal end rather than
+	// Error::Dropped.
+	broadcast.close();
+	result
 }
 
 /// The main emulator loop, running on a blocking thread.
--- a/rs/moq-ffi/src/producer.rs
+++ b/rs/moq-ffi/src/producer.rs
@@ -361,12 +361,17 @@ impl MoqBroadcastProducer {
 		}))
 	}
 
-	/// Finish this publisher, finalizing the catalog stream.
+	/// Finish this publisher, finalizing the catalog stream and cleanly closing the
+	/// broadcast so subscribers see a normal end rather than `Error::Dropped`.
 	pub fn finish(&self) -> Result<(), MoqError> {
 		let _guard = crate::ffi::RUNTIME.enter();
 		let mut guard = self.state.lock().unwrap();
-		let mut state = guard.take().ok_or_else(|| MoqError::Closed)?;
-		state.catalog.finish()?;
+		let mut state = guard.take().ok_or(MoqError::Closed)?;
+		// Close the broadcast even if finalizing the catalog fails, so the clean end
+		// still reaches subscribers.
+		let result = state.catalog.finish();
+		state.broadcast.close();
+		result?;
 		Ok(())
 	}
 }
--- a/rs/moq-ffi/src/producer.rs
+++ b/rs/moq-ffi/src/producer.rs
@@ -367,11 +367,10 @@ impl MoqBroadcastProducer {
 		let _guard = crate::ffi::RUNTIME.enter();
 		let mut guard = self.state.lock().unwrap();
 		let mut state = guard.take().ok_or(MoqError::Closed)?;
-		// Close the broadcast even if finalizing the catalog fails, so the clean end
-		// still reaches subscribers.
-		let result = state.catalog.finish();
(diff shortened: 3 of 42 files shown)


rs/libmoq/src/publish.rs
```diff
@@ -59,7 +59,11 @@ impl Publish {
...
 
 	pub fn close(&mut self, broadcast: Id) -> Result<(), Error> {
-		self.broadcasts.remove(broadcast).ok_or(Error::BroadcastNotFound)?;
+		let (broadcast, mut catalog) = self.broadcasts.remove(broadcast).ok_or(Error::BroadcastNotFound)?;
+		// Finalize the catalog stream, then cleanly close the broadcast so
+		// subscribers see a normal end rather than Error::Dropped.
+		catalog.finish()?;
+		broadcast.close();
 		Ok(())
 	}
```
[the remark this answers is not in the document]
kixelated: Fixed: capture `catalog.finish()`, call `broadcast.close()`, then propagate, so the broadcast closes even on the catalog-finish error path. (This file was re-added — it had been dropped when the PR retargeted from main to dev.)

rs/moq-boy/src/main.rs
```diff
@@ -271,11 +271,16 @@ async fn run(config: &Config) -> Result<()> {
...
 		res = emulator_handle => res?.context("emulator error"),
 		res = reconnect.closed() => Ok(res?),
 		res = input::handle_viewers(&mut viewer_consumer, &cmd_tx) => res,
-	}
+	};
+
+	// Cleanly close the broadcast on shutdown so subscribers see a normal end
+	// rather than Error::Dropped.
+	broadcast.close();
+	result
```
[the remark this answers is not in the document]
kixelated: Fixed: moved the `select!` result into a local and call `broadcast.close()` after it, so a `?` inside an arm no longer returns before cleanup runs.

rs/moq-ffi/src/producer.rs
```diff
@@ -319,12 +319,19 @@ impl MoqBroadcastProducer {
...
 	pub fn finish(&self) -> Result<(), MoqError> {
 		let _guard = crate::ffi::RUNTIME.enter();
 		let mut guard = self.state.lock().unwrap();
-		let mut state = guard.take().ok_or_else(|| MoqError::Closed)?;
-		state.catalog.finish()?;
+		let BroadcastProducer { broadcast, mut catalog } = guard.take().ok_or(MoqError::Closed)?;
+		catalog.finish()?;
+		broadcast.close();
 		Ok(())
 	}
```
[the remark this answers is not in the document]
kixelated: Fixed: `finish()` now captures `catalog.finish()`, calls `broadcast.close()`, then propagates, so the broadcast is closed even when finalizing the catalog fails.

rs/moq-native/examples/clock.rs
```diff
@@ -70,10 +70,15 @@ async fn main() -> anyhow::Result<()> {
...
+			let result = tokio::select! {
 				res = reconnect.closed() => Ok(res?),
 				_ = clock.run() => Ok(()),
-			}
+			};
+
+			// Cleanly close the broadcast on exit so subscribers see a normal end
+			// rather than Error::Dropped.
+			broadcast.close();
+			result
```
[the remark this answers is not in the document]
kixelated: Fixed: moved the `select!` result into a local so `broadcast.close()` always runs before returning (no `?` in the arm).

rs/libmoq/src/publish.rs
```diff
@@ -61,8 +61,15 @@ impl Publish {
...
 
+	/// Finalize the catalog stream and cleanly close the broadcast, so subscribers
+	/// see a normal end rather than [`moq_net::Error::Dropped`].
 	pub fn close(&mut self, broadcast: Id) -> Result<(), Error> {
-		self.broadcasts.remove(broadcast).ok_or(Error::BroadcastNotFound)?;
+		let (broadcast, mut catalog) = self.broadcasts.remove(broadcast).ok_or(Error::BroadcastNotFound)?;
+		// Close the broadcast even if finalizing the catalog fails, so the clean end
+		// still reaches subscribers.
+		let result = catalog.finish();
+		broadcast.close();
```
kixelated, line 70: Just close the broadcast first?

rs/moq-hls/src/import.rs
```diff
@@ -97,6 +97,20 @@ pub struct Import {
 	audio: Option<TrackState>,
 }
 
+impl Drop for Import {
+	fn drop(&mut self) {
+		// The HLS import loop ends by cancellation (or after a VOD playlist ends);
+		// an async `Drop` can't `finish()`, so abort the importers explicitly with a
+		// real Cancel rather than letting their tracks surface a bare Error::Dropped.
+		for importer in &mut self.video_importers {
+			importer.abort(moq_net::Error::Cancel);
```
kixelated: We should have an `abort` method instead? These should be `Dropped` if the Import itself is dropped.

rs/moq-net/src/model/group.rs
```diff
@@ -355,6 +355,12 @@ impl Drop for Producer {
 		if let Ok(mut state) = modify(&self.state)
 			&& !state.fin
 		{
+			// Dropped without finish() or abort(), so consumers will see
+			// Error::Dropped mid-group. Deliberate ends go through finish()/abort().
+			tracing::warn!(
+				sequence = self.info.sequence,
+				"group::Producer dropped without finish() or abort(); consumers will see Error::Dropped"
```
kixelated: Remove the "; consumers will see Error::Dropped" from all of these.

rs/moq-rtc/src/ingest.rs
```diff
@@ -62,3 +62,12 @@ impl session::MediaSink for IngestSink {
 		self.bridges.push(mid, frame)
 	}
 }
+
+impl Drop for IngestSink {
+	fn drop(&mut self) {
+		// The RTP flow ended (the peer left / ICE dropped); an async `Drop` can't
+		// finish, so abort the bridge tracks with a real Cancel rather than letting
+		// them surface a bare Error::Dropped.
+		self.bridges.abort(moq_net::Error::Cancel);
```
kixelated: We should have an `abort` method instead.

kixelated (commented):
Drop shouldn't be load-bearing. Add an `abort` method but if that's not viable, then leave the warning until we fix it. If there's code that's gross because you need to capture the error in order to abort, sometimes you can restructure the code to make it cleaner.

rs/libmoq/src/publish.rs
kixelated: Done — close the broadcast first, then finalize the catalog, so the clean end reaches subscribers even if `catalog.finish()` fails. Applied the same ordering to the ffi `finish()` for consistency.

rs/moq-hls/src/import.rs
kixelated: Done — replaced the aborting `Drop` with an explicit `abort(err)` method. Dropping the `Import` now honestly ends its tracks as `Error::Dropped`; `abort(err)` is for a real, known cause.

rs/moq-net/src/model/group.rs
kixelated: Done — removed the "; consumers will see Error::Dropped" clause from all four (broadcast / track / group / frame) drop warnings.

rs/moq-rtc/src/ingest.rs
kixelated: Done — replaced the `Drop` with an `abort(err)` method on the `MediaSink` trait; the session calls it only on a genuine failure (not a normal `SessionClosed`/`IceTimeout`), so a plain drop ends tracks as `Dropped` and a real error propagates.

PR #2109, opened 2026-07-07
Per-track timeline index for each media track

kixelated:
## What

Adds a **per-track timeline**: a companion track that maps each of a media track's groups to its start timestamp, so a consumer can learn which group covers which time (and where the live edge is) from a few bytes per group instead of subscribing to the media. This is the metadata primitive a live HLS/DASH origin needs to build playlists without downloading media, and the index a VOD player seeks with.

## Why per-track (not per-broadcast)

Audio and video groups have different durations (and metadata tracks different still), so a single broadcast-wide timeline can't describe them all. Each video/audio catalog entry carries an optional `timeline` section naming its own companion track, its `timescale` (default milliseconds), and an optional `wall` value: **the wall-clock time of `pts` 0, in timescale units since the moq epoch (2020-01-01, `MOQ_EPOCH_UNIX_MILLIS`)**, so a consumer derives any group's wall clock as `wall + pts` and Unix time by adding the moq epoch back. Measured from 2020 rather than 1970 to keep the value small (safely within a 53-bit integer even at fine timescales).

## Wire format

The timeline track is a new **`moq-json` stream mode**: an append-log counterpart to the existing snapshot/delta ("object") mode.

- Frames are **NDJSON** (one JSON record per line); a live append is one record per frame. (JSON escapes control characters, so a record never contains a raw newline; the line split is safe — covered by a test.)
- Records roll into a new group every `group_records` (default **256**). moq-net front-evicts a group's frames past its cache limits, and a windowed-DEFLATE frame is only decodable from frame 0, so an unbounded group would strand late joiners.
- Each new group's first frame is a **catch-up** (the previous group replayed as one NDJSON frame): a fresh subscriber reads it for continuity; a contiguous reader recognizes the duplicate and skips it. So the application sees one unbroken log whether it joined at the start or mid-stream.

Records are `{group, pts}` (unknown fields preserved for extensions). The shape follows draft-ietf-moq-msf `mediatimeline` in spirit, with JSON keys instead of positional arrays (DEFLATE absorbs the repetition) and a moq-lite group sequence instead of a `[group, object]` location.

## What's wired

**Rust:** `moq-json::stream`; the hang `catalog::Timeline` section + `timeline::Record`; `moq-mux::catalog::Producer` owns one `timeline::Producer` per rendition (memoized, lazily created on a retained broadcast clone), advertises the section by default from `VideoTrack`/`AudioTrack::set`, and the media `container::Producer` records each group open when built with a timeline (the `timeline` argument to `Producer::new` — the old `with_timeline` builder was folded into the constructor, so there is one construction path and importers pass it by default). Every codec + container importer wires it, plus moq-rtc (vp8/vp9) and moq-audio; h264/opus over WebRTC go through the codec importers and get it for free. On by default for importer-built broadcasts; costs nothing until subscribed.

**JS (publish):** a matching `@moq/json` `Stream` producer/consumer (wire-compatible), the `timeline` catalog schema on video/audio configs, and a `Container.Timeline.Producer`. The Stream producer and the Timeline producer support a **static / fan-out mode**: a track-less producer that `serve()`s the log to any number of subscribers (mirroring the catalog fan-out, seeding late joiners from a retained window), so a publisher can serve a timeline track to many subscribers via the broadcast `publishTrack` hook. `Container.Legacy.Producer` takes an optional timeline and records each group open.

## Not in this PR (follow-ups)

- Wiring the timeline into the JS **reactive publish encoder** end to end: that pipeline encodes *per subscriber*, so media group sequences differ between subscribers and a single shared timeline can't reference them. It needs produce-once-and-fan-out media first (the static-track primitive here is the prerequisite building block).
- **Consumers** that build HLS/DASH playlists from the timeline, and the relay group-retention change (`MAX_GROUP_AGE`) a fetch-on-demand origin needs.

## Branch target

Targets `main`: the change is additive (new optional catalog fields, new tracks, new `pub` APIs, wire-compatible with existing consumers), which the repo's branch rule puts on `main`. (`dev` has since diverged with a moq-net track-API refactor; a dev version would need a port to that API rather than a rebase.)

## Testing

Full workspace `cargo clippy --all-features -D warnings` + `fmt --check` + tests green (moq-json stream mode incl. eviction-based late-join, embedded-newline, and byte-level no-raw-newline tests; moq-mux timeline; hang schema round-trips). JS: `@moq/json` + `@moq/hang` + `@moq/publish` typecheck, `@moq/json` tests (stream round-trip, catch-up, fan-out, embedded newlines), biome clean.

### DIFF
--- a/js/hang/src/container/legacy.ts
+++ b/js/hang/src/container/legacy.ts
@@ -4,6 +4,7 @@ import * as Moq from "@moq/net";
 export type { BufferedRange, BufferedRanges, Frame } from "./types";
 
 import type { Format as ContainerFormat } from "./format";
+import type { Producer as TimelineProducer } from "./timeline";
 import type { Frame } from "./types";
 
 /** The legacy hang container: a microsecond timestamp varint followed by the raw codec payload. */
@@ -38,14 +39,25 @@ export function encodeFrame(source: Uint8Array | Source, timestamp: Time.Micro):
 	return data;
 }
 
+/** Options for a legacy-container {@link Producer}. */
+export interface ProducerProps {
+	/**
+	 * Record each group open (sequence + start timestamp) into this companion timeline track, so
+	 * consumers can index the media without downloading it.
+	 */
+	timeline?: TimelineProducer;
+}
+
 /** Writes legacy-container frames into a MoQ track, starting a new group on each keyframe. */
 export class Producer {
 	#track: Moq.Track;
 	#group?: Moq.Group;
+	#timeline?: TimelineProducer;
 
 	/** Wrap a track to publish legacy-container frames into it. */
-	constructor(track: Moq.Track) {
+	constructor(track: Moq.Track, props: ProducerProps = {}) {
 		this.#track = track;
+		this.#timeline = props.timeline;
 	}
 
 	/** Encode and append a frame; a keyframe starts a new group. Throws if the first frame is not a keyframe. */
@@ -53,6 +65,8 @@ export class Producer {
 		if (keyframe) {
 			this.#group?.close();
 			this.#group = this.#track.appendGroup();
+			// Index the group the moment it opens: its start is this keyframe's timestamp.
+			this.#timeline?.record(this.#group.sequence, timestamp);
 		} else if (!this.#group) {
 			throw new Error("must start with a keyframe");
 		}
--- a/js/hang/src/container/timeline.ts
+++ b/js/hang/src/container/timeline.ts
@@ -1,0 +1,116 @@
+/**
+ * Publishing a media track's timeline: a companion track that maps each of the track's groups to
+ * its start timestamp, so a consumer can seek (or build an HLS/DASH playlist) without downloading
+ * the media. See the catalog {@link Catalog.Timeline} section that advertises it.
+ *
+ * @module
+ */
+
+import * as Json from "@moq/json";
+import type * as Moq from "@moq/net";
+import type { Time } from "@moq/net";
+import type * as Catalog from "../catalog";
+import { MOQ_EPOCH_UNIX_MILLIS, u53 } from "../catalog";
+
+/** One timeline record: the media track opened `group` at presentation time `pts` (in the timeline's timescale). */
+export interface Record {
+	group: number;
+	pts: number;
+}
+
+/** The default timeline timescale: 1000 units per second (milliseconds). */
+export const DEFAULT_TIMESCALE = 1000;
+
+/** The default record throttle: at most one record per second of media time. */
+export const DEFAULT_GRANULARITY_MS = 1000;
+
+/**
+ * The conventional companion timeline track name for a media rendition: `<rendition>.timeline.z`
+ * (the `.z` marks the DEFLATE-compressed stream, like the catalog's `.json.z` sibling).
+ */
+export function trackName(rendition: string): string {
+	return `${rendition}.timeline.z`;
+}
+
+/** Options for a timeline {@link Producer}. */
+export interface ProducerProps {
+	/** Units per second for the records' `pts` (and the `wall` anchor). Defaults to milliseconds. */
+	timescale?: number;
+
+	/**
+	 * Record at most one group per this much media time, in milliseconds. Video keyframes are
+	 * already this far apart so every group is indexed; short audio groups are thinned out (a
+	 * consumer extrapolates or fetches to fill a gap). Defaults to {@link DEFAULT_GRANULARITY_MS}.
+	 */
+	granularity?: number;
+}
+
+/**
+ * Publishes one media track's timeline: an NDJSON record per group open, DEFLATE-compressed.
+ *
+ * {@link record} appends a group's start once. Advertise it in the rendition's catalog config via
+ * {@link section}, and attach it to a {@link Legacy.Producer} (its `timeline` prop) to record group
+ * opens automatically.
+ */
+export class Producer {
+	#stream: Json.Stream.Producer<Record>;
+	#track: string;
+	#timescale: number;
+	// The wall-clock time of pts 0, in timescale units since the moq epoch (advertised in the section).
+	#wall?: number;
+	// Minimum media-time gap between recorded groups (throttle), in microseconds.
+	#granularityUs: number;
+	// The pts (microseconds) of the last recorded group.
+	#lastPts?: number;
+
+	/** Wrap an already-created MoQ track (named per {@link trackName}) to publish a rendition's timeline. */
+	constructor(track: Moq.Track, props: ProducerProps = {}) {
+		this.#track = track.name;
+		this.#timescale = props.timescale ?? DEFAULT_TIMESCALE;
+		this.#granularityUs = (props.granularity ?? DEFAULT_GRANULARITY_MS) * 1000;
+		this.#stream = new Json.Stream.Producer<Record>(track, { compression: true });
+	}
+
+	/** The catalog section advertising this timeline, to attach to the rendition's config. */
+	section(): Catalog.Timeline {
+		return {
+			track: this.#track,
+			timescale: u53(this.#timescale),
+			wall: this.#wall === undefined ? undefined : u53(this.#wall),
+		};
+	}
+
+	/**
+	 * Set (or replace) the wall-clock anchor advertised in the catalog section, from an observed
+	 * pairing of a media timestamp `pts` (microseconds) with its wall-clock time `wall` (defaulting
+	 * to now). Stored as the extrapolated wall-clock time of pts 0, the single value the catalog
+	 * `wall` field carries: in this timeline's timescale, measured from the moq epoch
+	 * ({@link Catalog.MOQ_EPOCH_UNIX_MILLIS}, 2020). Throws if `wall` predates the moq epoch
+	 * (unrepresentable).
+	 */
+	setWall(pts: Time.Micro, wall: Date = new Date()): void {
+		const unixMillis = wall.getTime();
+		if (unixMillis < MOQ_EPOCH_UNIX_MILLIS) {
+			throw new Error(`wall time ${unixMillis} predates the moq epoch ${MOQ_EPOCH_UNIX_MILLIS}`);
+		}
+		const ptsUnits = Math.floor((pts * this.#timescale) / 1_000_000);
+		const moqUnits = Math.floor(((unixMillis - MOQ_EPOCH_UNIX_MILLIS) * this.#timescale) / 1000);
+		this.#wall = Math.max(0, moqUnits - ptsUnits);
+	}
+
+	/**
+	 * Record that group `sequence` opened at presentation time `pts` (microseconds), unless it
+	 * falls within the {@link ProducerProps.granularity} of the last recorded group (skipped, so a
+	 * consumer extrapolates or fetches to fill the gap).
+	 */
+	record(sequence: number, pts: Time.Micro): void {
+		if (this.#lastPts !== undefined && pts < this.#lastPts + this.#granularityUs) return;
+		this.#lastPts = pts;
+		this.#stream.append({ group: sequence, pts: Math.floor((pts * this.#timescale) / 1_000_000) });
+	}
+
+	/** Finish the timeline track. */
+	finish(): void {
+		this.#stream.finish();
+	}
+}
--- a/js/json/src/stream.ts
+++ b/js/json/src/stream.ts
@@ -1,0 +1,120 @@
+/**
+ * Append-log JSON publishing over MoQ tracks: the counterpart to the snapshot/delta ("object")
+ * mode in {@link ./producer.ts}. Instead of one value updated over time, a stream is an ordered
+ * log of self-contained records; every {@link Producer.append} writes one JSON object as one
+ * frame, and a {@link Consumer} yields them all in order.
+ *
+ * The whole log rides a **single group** that is never rolled: with {@link ProducerConfig.compression}
+ * on, that one group is one DEFLATE window, so every record compresses against the earlier ones.
+ * There is deliberately no group rolling (and so no catch-up machinery); a caller that wants to
+ * bound the record rate throttles at the source. Interoperable on the wire with the Rust
+ * `moq_json::stream`.
+ *
+ * @module
+ */
+
+import { Decoder, Encoder } from "@moq/flate";
+import type * as Moq from "@moq/net";
+
+/** Options for a stream {@link Producer}. */
+export interface ProducerConfig {
+	/**
+	 * Compress the group as one sync-flushed `deflate-raw` stream, so each record reuses the earlier
+	 * ones as context and shrinks sharply. A {@link Consumer} reading the track must set the same
+	 * flag. Defaults to `false`.
+	 */
+	compression?: boolean;
+}
+
+/** Options for a stream {@link Consumer}. */
+export interface ConsumerConfig {
+	/** Whether the track's frames are `deflate-raw` compressed. Must match the producer. Defaults to `false`. */
+	compression?: boolean;
+}
+
+/**
+ * Publishes an ordered log of JSON records to a track, one record per frame in a single group.
+ */
+export class Producer<T> {
+	#track: Moq.Track;
+	#compress: boolean;
+
+	// The single group carrying the whole log, opened on the first append.
+	#group?: Moq.Group;
+	// The group's DEFLATE encoder (one window for the whole log), present while compressing.
+	#encoder?: Encoder;
+
+	/** Wrap a track to publish a record log into it. */
+	constructor(track: Moq.Track, config: ProducerConfig = {}) {
+		this.#track = track;
+		this.#compress = config.compression ?? false;
+	}
+
+	/** Append one record to the log. */
+	append(value: T): void {
+		const payload = new TextEncoder().encode(JSON.stringify(value));
+
+		if (!this.#group) {
+			this.#group = this.#track.appendGroup();
+			this.#encoder = this.#compress ? new Encoder() : undefined;
+		}
+
+		this.#group.writeFrame(this.#encoder ? this.#encoder.frame(payload) : payload);
+	}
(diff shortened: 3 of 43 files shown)


js/hang/src/container/legacy.ts
```diff
@@ -42,17 +43,26 @@ export function encodeFrame(source: Uint8Array | Source, timestamp: Time.Micro):
...
 
-	/** Wrap a track to publish legacy-container frames into it. */
-	constructor(track: Moq.Track) {
+	/**
+	 * Wrap a track to publish legacy-container frames into it.
+	 *
+	 * Pass a {@link TimelineProducer} to record each group open (sequence + start timestamp) into a
+	 * companion timeline track, so consumers can index the media without downloading it.
+	 */
+	constructor(track: Moq.Track, timeline?: TimelineProducer) {
```
kixelated: Can you make props for this so we could add other fields in the future.

js/hang/src/container/timeline.ts
```diff
@@ -0,0 +1,92 @@
...
+	#track: string;
+	#timescale: number;
+	// The wall-clock time of pts 0, in timescale units since the moq epoch (advertised in the section).
+	#wall?: number;
+
+	/**
+	 * A timeline for the rendition served on track `trackName` (per {@link trackName}). `timescale`
+	 * is the units per second for the records' `pts` (default milliseconds).
+	 */
+	constructor(trackName: string, timescale: number = DEFAULT_TIMESCALE) {
```
kixelated: Props please.

ProducerConfig { timescale?: number }

js/hang/src/container/timeline.ts
```diff
@@ -0,0 +1,92 @@
...
+		this.#track = trackName;
+		this.#timescale = timescale;
+		this.#stream = new Json.Stream.Producer<Record>({ compression: true, groupRecords: 256 });
```
kixelated: Hard-code 256 as maximum within the JSON module instead. Or I mean you can hard-code 1024 since that's the limit the relay enforces.

js/hang/src/container/timeline.ts
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	/**
+	 * Set (or replace) the wall-clock anchor advertised in the catalog section, from an observed
+	 * pairing of a media timestamp `pts` (microseconds) with its wall-clock time `unixMillis` (Unix
+	 * epoch milliseconds). Stored as the extrapolated wall-clock time of pts 0, the single value the
+	 * catalog `wall` field carries: in this timeline's timescale, measured from the moq epoch
+	 * ({@link Catalog.MOQ_EPOCH_UNIX_MILLIS}, 2020).
+	 */
+	setWall(pts: Time.Micro, unixMillis: number): void {
```
kixelated: error if unixMillis > MOQ_EPOCH_UNIX_MILLIS

kixelated: And you should make unixMillis optional, and if so, we'll use the current time. In fact this probably needs to be done within the constructor?

js/json/src/stream.ts
```diff
@@ -0,0 +1,287 @@
...
+	#encoder?: Encoder;
+	// The current group's records (raw one-line JSON, no newline), retained to replay as the next
+	// group's catch-up frame.
+	#groupLines: Uint8Array[] = [];
+
+	// Fan-out mode: retains a bounded record window and serves a leaf per subscriber.
+	#outputs?: Set<Producer<T>>;
+	#history?: T[];
+
+	/** Create a track-less, fan-out static producer; attach subscribers with {@link serve}. */
```
kixelated: We should deprecate this? It's not needed on the `dev` branch right?

rs/hang/src/timeline.rs
```diff
@@ -0,0 +1,110 @@
...
+	pub group: u64,
+
+	/// The group's start (its first frame's presentation timestamp), in the timeline's
+	/// timescale.
+	pub pts: u64,
+
+	/// Any additional fields, captured and republished verbatim so extended records survive a
+	/// round-trip through a base consumer.
+	#[serde(flatten)]
+	pub ext: serde_json::Map<String, serde_json::Value>,
```
kixelated: Would it be possible to make this a T? I don't like untyped maps.

rs/moq-json/src/stream.rs
```diff
@@ -0,0 +1,554 @@
...
+
+/// Shared publishing state behind [`Producer`]'s `Arc<Mutex>`.
+struct Inner {
+	track: moq_net::TrackProducer,
+	group: Option<moq_net::GroupProducer>,
+	// Per-group DEFLATE encoder, `Some` while a compressed group is open (recreated per group).
+	encoder: Option<Encoder>,
+	// The current group's records (raw one-line JSON, no newline), retained to replay as the
+	// next group's catch-up frame.
+	group_lines: Vec<Bytes>,
```
kixelated: This should be one BytesMut to avoid lots of fragmented buffers.

hmm actually would it be possible to keep this compressed?

In fact, what if we stored the exact encoded bytes that were sent as frames? Rather than counting newline delimiters, a decoder could just skip ahead by the number of bytes it has decoded in the past? This means existing subscribers can avoid decompressing and instead reuse the existing Decoder starting at offset N?

The only catch is the Z_SYNC_FLUSH magic bytes. New subscribers would need to know frame boundaries to reconstruct these.

I think the bigger issue is that we're even doing this in the first place. Literally the only reason why we're creating new groups is to avoid a limit set on the Rust side because its worried about lots of tiny Frame objects. The cost is that every subscriber gets a fully copy of the history every 1024 frames just to recompact the frame; it seems dumb.

So I think remove the new group stuff for stream and I'll tackle the max frame count issue independently.

kixelated:
Thanks for the review — addressed the comments in the latest push. Point by point:

**1. `legacy.ts` — props object.** Done. `Container.Legacy.Producer(track, { timeline? })` now takes a `ProducerProps`.

**2. `timeline.ts` — props object.** Done. `Container.Timeline.Producer(track, { timescale? })`.

**3. Hard-code the group cap in the JSON module.** Done. `MAX_GROUP_RECORDS` is now a constant in `@moq/json`/`moq-json` (Rust + JS), and the timeline producer no longer passes it. Set it to **1023**, not 1024: the relay evicts a group's frames past 1024, and a group is one catch-up frame (frame 0) plus its live records, so I reserve one so the catch-up can't be evicted. The config field stays (clamped to `1..=MAX`) only so tests can force small rolls.

**4. Error on a bad wall time.** Done — `setWall` throws when `unixMillis` predates the moq epoch. Small note: your comment said `> MOQ_EPOCH`, but a real Unix-ms time is always `> MOQ_EPOCH`, so I read it as "reject times **before** the epoch" (unrepresentable). Say the word if you meant something else.

**5. `unixMillis` optional / default to now / constructor.** Done — `setWall(pts, unixMillis = Date.now())`. I kept it on `setWall` rather than the constructor because the anchor needs a `pts` paired with a wall time, and at construction there's no frame yet — the natural call is on the first frame. Happy to move it if you'd rather anchor at construction.

**6. Deprecate the fan-out static producer (not needed on dev).** Done — removed it entirely. `@moq/json` `Stream` is leaf-only again and `Container.Timeline.Producer` takes a track. Agreed it's the wrong layer for this branch given dev handles static tracks natively.

**7. Typed record extension (no untyped map).** Done. `timeline::Record<E = ()>` with a `RecordExt` trait, flattened into the JSON object exactly like `Catalog`'s extension. The default `()` is wire-identical to before.

**8. `stream.rs` catch-up storage.**
- *One `BytesMut`*: done — the catch-up is accumulated in a single `BytesMut` instead of a `Vec<Bytes>`.
- *Keep it compressed / skip by byte offset*: I like this, but I think it's a wire-format change worth its own pass, and there's a wrinkle. Each group opens a **fresh** DEFLATE window, so the previous group's frames (compressed against the *old* window) can't be dropped into the new group's stream as-is — a fresh joiner decoding the new group from a cold window couldn't inflate them. To get what you want (contiguous readers skip the catch-up by byte length without decompressing; only fresh joiners pay for it), the clean shape is: make the catch-up its **own self-contained DEFLATE stream** (one blob = the previous group re-framed, with its own `Z_SYNC_FLUSH` boundaries), and compress the live records in their own window *after* it. Then a contiguous reader skips the whole catch-up frame by its length and starts its decoder on frame 1; a fresh joiner inflates the self-contained blob, then continues. That's a nice improvement but a distinct change to the frame layout — I'd rather land this PR and do it as a focused follow-up so the wire change gets its own review. Want me to open an issue for it?

Also flagging CodeRabbit's out-of-diff note that FLV `replace_video`/`replace_audio` should finish the memoized timeline on a track swap — that's a real one; I can fold in a fix here or take it as a follow-up, your call.

(Written by Claude Fable 5)

kixelated:
Pushed the follow-up:

**Single group, never rolled (your 8c).** The stream mode now writes the whole log to one group and never rolls it. That deleted all the catch-up machinery, the group-records cap, and even the NDJSON framing (each record is just its own frame now, one JSON object). With compression on, the single group is one DEFLATE window so records compress against each other, and a late joiner reads whatever the relay still retains; deep history comes from a recording, not the live stream.

**`granularity` throttle.** `timeline::Producer` / `Container.Timeline.Producer` now record at most one group per `granularity` of media time (default **1s**; Rust `with_granularity`, JS `granularity` prop). Video keyframes are already ≥1s apart so every group is indexed; short audio groups get thinned out. A consumer that lands between two records extrapolates the group number (sequences are contiguous) or fetches to fill the gap, as you noted.

Green: full workspace `cargo clippy --all-features` + fmt + tests, and JS typecheck/tests/biome.

Still open from the last round if you want them: the CodeRabbit FLV `replace_video`/`replace_audio` timeline-cleanup fix, and confirming my read of the `setWall` epoch check (I reject times *before* the moq epoch).

(Written by Claude Fable 5)

kixelated:
Replaced the positional `timeline: Option<..>` param on `container::Producer::new` with a `ProducerConfig` options struct + builder, per the API-scrutiny convention:

```rust
// plain (unchanged 2-arg, non-breaking vs the base):
let p = container::Producer::new(track, container);

// configured:
let p = container::Producer::with_config(
    track, container,
    container::ProducerConfig::default().with_timeline(timeline),
);
```

`ProducerConfig` (`#[non_exhaustive]` + `Default`) now holds both `timeline` and the existing `latency` knob (folded in off `Producer::with_latency`, which only had that one test caller), so future producer options are additive without another signature change. All ~40 call sites updated (importers → `with_config`, tests → plain `new`).

Net effect on the public surface: `container::Producer::new` is now source-compatible with the pre-PR two-arg signature again — the only remaining API break is gone. Green: full workspace `cargo clippy --all-features` + fmt + tests.

(Written by Claude Fable 5)

kixelated:
Reworked per your note: `ProducerConfig`/`ConsumerConfig` are now the **3rd argument** to `new`, with no builders chained after `::new`.

```rust
container::Producer::new(track, container, ProducerConfig::default())
container::Producer::new(track, container, ProducerConfig::default().with_timeline(timeline))

container::Consumer::new(track, format, ConsumerConfig::default())
container::Consumer::new(track, format, ConsumerConfig::default().with_latency(latency))
```

- Dropped `Producer::with_config` and the fluent `Producer::new(..).with_timeline(..)` / `Consumer::new(..).with_latency(..)` entirely — every option is set on the config before it's passed in. `Consumer::with_latency` (the pre-existing one) folded into `ConsumerConfig`.
- The generic `container: C` / `format: F` stays a required positional (it fixes the `Producer<C>`/`Consumer<F>` type; can't move into a `Default`-able config). So: two required positionals + one config, symmetric across Producer and Consumer.
- Updated ~90 call sites across the workspace (importers, moq-rtc, moq-audio, moq-cli, moq-rtmp, moq-srt, libmoq, moq-ffi, moq-gst, moq-hls, moq-boy, moq-video, tests). A couple of conditional/multi-line `with_latency` chains I restructured by hand to build the config first.

Green: full workspace `cargo clippy --all-features` + fmt + tests, and JS check/tests.

(Written by Claude Fable 5)

kixelated:
Switched to consuming builders on the object, per your call:

```rust
container::Producer::new(track, container)                       // 2-arg
    .with_latency(latency)
    .with_timeline(timeline)                                     // impl Into<Option<..>>

container::Consumer::new(track, format)
    .with_latency(latency)
```

- Dropped `ProducerConfig`/`ConsumerConfig` entirely. `with_latency(self, Duration) -> Self` and `with_timeline(self, impl Into<Option<timeline::Producer>>) -> Self` (so importers pass the handle directly, and `None` / a plain track just omits the call). Consumer gets `with_latency` the same way.
- `container: C` / `format: F` stays positional (it fixes the generic type).
- **Both `Producer::new` and `Consumer::new` are back to 2 args**, so they're source-compatible with what was on `main` — the API surface is fully additive again (new module/types/fields + the new builders), zero signature breaks.
- ~85 call sites swept back to the fluent form; moq-audio's conditional latency is `let mut c = Consumer::new(..); if let Some(l) = .. { c = c.with_latency(l); }` again.

On the runtime-mutation thread: these are consuming builders (construction-time). If a live latency/timeline change without a rebuild ever becomes a real need, the additive follow-up is `&mut self` setters alongside these — no signature change required.

Green: full workspace `cargo clippy --all-features` + fmt + tests, and JS check/tests.

(Written by Claude Fable 5)

kixelated:
Made timelines 1:1-by-construction so they can't be shared between tracks (agreed it's not just error-prone but incoherent with the record schema — the record has no track id, and two tracks share one sequence space).

- The catalog now owns the wiring: `catalog::Producer::media_producer(track, container)` builds the rendition's `container::Producer` with its timeline recorder already attached, keyed on the track's own name. Callers never hold a timeline handle, so they can't hand it to the wrong track.
- `container::Producer::with_timeline` is now `pub(crate)` (the mechanism `media_producer` uses). `with_latency` stays public.
- Every importer (codecs, flv/mkv/ts, moq-rtc, moq-audio) went from the `catalog.timeline(name)` + `.with_timeline(..)` pair to a single `catalog.media_producer(track, container)` — also kills the track-name-threaded-three-times pattern.
- Track name is now `<rendition>.timeline.z` (the `.z` marks the DEFLATE stream, matching `catalog.json.z`), in both Rust `track_name` and JS `trackName`.

If aligned ingest ever wants a genuinely shared broadcast-wide timeline, that's a deliberate schema change (put a track id back in the record) — not something the API pre-enables now.

Green: full workspace `cargo clippy --all-features` + fmt + tests, JS check.

(Written by Claude Fable 5)

kixelated:
Reworked the `timeline::` API per the review, all five:

1. **Read from the section, typed pts.** `timeline::Consumer::subscribe(&broadcast, &section)` takes the track name *and* timescale from the catalog `Timeline` section (can't mismatch them), and yields decoded `Entry { group, pts: Timestamp, ext }` — no more raw `u64` pts + `pts_micros()`/`timescale()` dance.
2. **`set_wall(pts, wall: SystemTime)`** instead of `(pts, unix_millis: u64)` (JS: `setWall(pts, wall?: Date)`).
3. **Throttle-state divergence fixed structurally.** Granularity + the `last` cursor now live on the move-only `Recorder`, which is 1:1 per track, so there's no shared `Arc<Mutex>` cursor to diverge from a per-clone granularity.
4. **`E` threaded through** — `Producer<E>`/`Recorder<E>`/`Consumer<E>`/`Entry<E>`, not just `Record<E>`. (`RecordExt` gained `Send + Unpin + 'static` for the generic consumer.)
5. **`Clone` removed.** The write side splits into a non-`Clone` `Producer` (catalog-owned: `section`, `set_wall`, `finish`) and a move-only `pub(crate) Recorder` (`record`) that the catalog wires 1:1 into the container producer via `media_producer`. So "can't share a timeline" is now enforced by the type, not just by the module being private.

Net public `timeline::` surface: `Producer` (new/with_granularity/section/set_wall/finish), `Consumer` (subscribe/next/poll_next), `Entry`, `DEFAULT_GRANULARITY` — plus `catalog::Producer::media_producer` + `timeline_section`. The module stays `pub` (the API's now worth committing to), and it's still a fully additive change vs `main`.

Green: full-workspace `cargo clippy --all-features` + fmt + tests (moq-mux 353, hang 21), and JS check/tests. (One detour: a stale incremental artifact made moq-msf's `Catalog` look `#[non_exhaustive]` during `cargo test`; `cargo clean -p moq-msf` cleared it — not a code issue.)

(Written by Claude Fable 5)

kixelated:
Fixed the CI failure and triaged the review comments (`d018ad151`):

**CI (was the blocker):** `cargo shear` flagged an unused `tokio` dev-dep in `moq-json` — leftover from a test the single-group rewrite removed. Dropped it (and fixed the crate description, which still said "NDJSON"). `cargo shear` is green now.

**Review findings:**
- **Timeline write aborting the media write** (`container/producer.rs`, valid) — fixed. A timeline is an optional, gap-tolerant sidecar, so a recording failure now drops the recorder and logs a warning instead of `?`-aborting the media frame write. (Consistent with the timeline's design and the "best-effort" exception to fail-loud.)
- **`#[non_exhaustive]` configs need builders** (`moq-json` stream, valid per the convention) — added `ProducerConfig::with_compression` / `ConsumerConfig::with_compression` and used them at the call sites.
- **Wall-clock "unit mismatch" formula** — stale: it quotes `unix + (pts - anchor.pts) / timescale` from the earlier anchor-struct design. The current design is a single `wall` value (wall-clock of pts 0), documented as `wall + pts` (all in timescale units), which has no such division bug. No change.
- The remaining inline comments (`.expect` wording, precision-loss, "collapse Poll::Pending", the `#[non_exhaustive]` note quoting `group_records`) are against pre-rewrite revisions of the files (NDJSON stream, `with_config`, `group_records`) and no longer apply.

Green locally: full-workspace `cargo clippy --all-features` + fmt + shear + tests (moq-mux 353, moq-json 58, hang 21), JS check/tests/biome.

(Written by Claude Fable 5)

PR #2114, opened 2026-07-07
Safari: hardware encode, worker capture, audio/video hardening

fperex:
## Summary

Cross-browser hardening for Safari support in `js/publish` and `js/watch`, built up over several passes:

- Safari hardware H.264 encode via a capture worker, with honest per-browser support labels.
- HEVC catalog description fix, cross-browser reconnect gating, publisher-switch recovery (a new publisher taking over `me.hang` resyncs cleanly).
- Codec-switch resync and Safari audio fixes on top of that.
- `isStreamAbort` made code-aware so real subscribe/publish faults still surface as warnings instead of being swallowed as expected teardown.
- Safari publish audio at 48 kHz, plus decoder restart/log hardening.
- **Audio cadence snap**: Safari's `AudioEncoder` quantizes output timestamps to the 128-sample capture-quantum grid (alternating ~18667/21333us for a 20ms Opus frame), which made the watcher's timestamp-indexed ring drop/duplicate a sample every frame and crackle. The encoder now snaps near-contiguous frames onto the nominal cadence (identity on Chrome/Firefox, which already emit an exact cadence); the watcher applies the same snap as defense in depth. Device- and option-independent (arbitrary sample rates, mono/stereo, any codec/bitrate/frameDuration).
- **Mute/unmute hardening**: the audio backpressure wait is now abortable, a drained ring re-stalls in buffered mode, a wire-timestamp gap re-anchors playback, and a cleanly-ended track resubscribes. Fixes two bugs found under rapid mute/unmute: permanent audio loss, and a growing audio delay that never recovered.
- **Video decode hardening**: pinned-HD playback no longer storms resubscribes on a decoder error or codec switch (config-staleness guard + keyframe-gated restart + capped decode-queue burst).
- **Subscribe teardown race**: the initial SUBSCRIBE open+write now lives inside its try/catch, so a re-subscribe racing a publisher's session-close is caught (logged via `isStreamAbort`) instead of escaping as an unhandled promise rejection. All-browser, error-path only.
- **Visibility-recovery grace**: a Safari watcher no longer pays for a full ~3-4s reconnect on a brief tab occlusion/resize (e.g. dragging another window over it) or quick tab-switch — it waits ~1s and re-checks before reconnecting, since the WebTransport session usually resumes on its own.

All Safari-specific behavior is gated behind `Util.Hacks.isSafari` (or is error-path-only / rate-keyed), so Chrome and Firefox steady-state behavior is unchanged.

## Public API changes

[...]

PR #2115, opened 2026-07-07
feat(js): Safari hardware encode, worker capture, audio/video hardening

fperex:
Rebases and reconciles the Safari support work onto `dev` (the correct base per
maintainer guidance). Supersedes #2114, which targeted `main` because its branch
was accidentally cut from `main`. Squashed into one feature commit plus a
`CONTRIBUTING.md`.

Everything here is layered onto `dev`'s current APIs (the `input`/`output`
namespacing, branded `Time`, the refactored decoders and `moq-net` surface), not
the older API the original branch was written against. The Safari-only paths stay
gated so Chrome/Firefox behavior is unchanged.

## What it does

**Publish**
- Hardware H.264 encode with a capture worker and honest support labels.
- 48 kHz Opus: capture-rate resampling to the canonical rate, cadence-snapped
  container timestamps, catalog advertises the 48 kHz decode rate.
- `Broadcast.requestCount` wedge-detection counter (relays subscribe to the
  catalog within ~100 ms of an announce; a session stuck at 0 has likely wedged).

**Watch**
- Audio resampling to the AudioContext's actual rate (Safari pins the hardware
  rate) plus a timestamp snap so the ring writes back-to-back instead of
  crackling; both are identity no-ops on Chrome/Firefox.
- Gesture-driven AudioContext resume with a `running`-gated graph so Safari
  actually starts rendering.
- Decoder restart budgets (audio + video) with codec-switch supersede detection,
  keyframe resync, and decode-queue backpressure.
- Unsupported-codec detection with a UI notice instead of an indefinite spinner.
- Silently-wedged WebTransport stall/visibility reconnect recovery (Safari-gated).
- Announce-generation republish re-consume (a same-name publisher restart
  re-subscribes even when the presence set never flips).

**Net**
- Code-aware `isStreamAbort` logging so real subscribe/publish faults surface at
  `warn` while routine stream teardown stays at `debug`.
- Skip QUIC datagrams (moq-lite-05 §6.4) on Safari, keeping WebTransport: Safari's
  datagram I/O faults (`getWriter`/write and the inbound read) and killed the
  session, so publish looped on "connection error". Both datagram sites are gated
  off on Safari (checked before touching `quic.datagrams`), so it serves/receives
  groups over streams only. Chrome/Firefox still use datagrams. New
  `js/net/src/util/agent.ts` (net can't import `hang`'s `detectSafari`) with tests.

## Public API (additive)

- `js/net`: `isStreamAbort` export (new).
- `js/watch`: `configSuperseded` export (new); `video.Source` output gains
  `unsupported`.
- `js/publish`: `Broadcast.requestCount` getter (new); `Broadcast` gains
  `announcedGenerations` support wired from the watch element.

## Test plan

- [x] `just js check` (typecheck + Biome) passes across all 13 packages.
- [x] `just js test` passes (includes new unit tests: cadence, resampler, snap,
  codecs, catalog, `configSuperseded`, `isStreamAbort`, Safari hacks,
  `requestCount`).
- [x] `just js build` bundles cleanly (worklet + dynamic imports).
- [ ] Full codec matrix on Safari / Apple Silicon (publish + watch, hardware
  H.264/HEVC, 48 kHz audio, mute/unmute, publisher/codec switch, visibility
  reconnect) verified manually. Chrome/Firefox confirmed unchanged.
- [ ] Safari publish/watch connects over WebTransport with the datagram gate (no
  "connection error" reconnect loop); Chrome still uses datagrams.


kixelated (commented):
I would prefer no browser sniffing, is there no way we could catch the error instead? Or is it a hard failure?

PR #2123, opened 2026-07-08
feat(js/net): dedupe conn.consume() per path so repeat subscribes share one broadcast

### LINKED ISSUE #2121: js/net: dedupe conn.consume() so repeat subscribes to a path share one broadcast
Follow-up to #1371.

## Problem

`Connection.consume(path)` in `@moq/net` mints a **fresh** `broadcast.Consumer` (and its own subscribe loop) on every call — it does not deduplicate. This mirrors a gap the Rust side had in `origin::Consumer::request_broadcast`, which #1371 fixed for Rust by weakly caching dynamically-served broadcasts (see commit `36a5d7c1b`).

Two consequences in `js/watch` (`Broadcast.relativeBroadcast` / `#runBroadcast`):

1. **Churn on effect rerun.** `relativeBroadcast` calls `conn.consume(resolved)` inside the caller's decoder/MSE effect with `effect.cleanup(() => broadcast.close())`. That effect also depends on `config`/`track`/announced-set, so any of those changing closes + reopens the cross-broadcast subscription, forcing a live re-join (latency hiccup).
2. **No dedup.** Multiple renditions referencing the same `broadcast: "../source"` each open an independent `conn.consume(resolved)`.

## Why it wasn't fixed in #1371

Unlike Rust (where `broadcast::Consumer` is a pure handle and closing one never affects the broadcast), JS `broadcast.Consumer.close()` closes the **shared** `BroadcastState`. So safe caching needs a consume-**ownership** rework: either reference-count `close()`, or make consumed broadcasts connection-owned and evicted on close so callers (js/watch) stop closing them. That change also affects the **main** broadcast lifecycle (`#runBroadcast` at `js/watch/src/broadcast.ts:137`), not just the cross-broadcast one — broader than a bolt-on, and best done with its own review.

Severity is quality/perf (churn + missing dedup), not correctness: browsers consume **announced** broadcasts, which `conn.consume` handles fine today.

## Proposed shape

- Cache one `broadcast.Consumer` per path in the subscriber (both `lite/` and `ietf/`), returning a shared/refcounted handle; evict when it closes.
- Adjust `js/watch` so the connection owns the consumed broadcast's lifetime (callers stop calling `.close()` on `conn.consume` results).
- Mirror the Rust weak-cache semantics: a live path shares one subscription; a closed one is re-consumed on the next request.

(Written by Claude Opus 4.8)

kixelated:
Closes #2121. Follow-up to #1371 (relative broadcast paths) and the JS mirror of the Rust weak-cache (`36a5d7c1b`).

## Problem

Consuming from `@moq/net` didn't deduplicate, unlike Rust (`origin::Consumer::request_broadcast` weak-cache + `broadcast::Consumer::track` weak-dedup): `Connection.consume(path)` minted a fresh consumer per call, `broadcast.track(name).subscribe()` opened a fresh upstream `SUBSCRIBE` per call, and concurrent `track.fetchGroup(seq)` opened a fresh FETCH stream per call.

## Change — dedup the whole ladder

**Broadcast (reference-counted).** Shared `BroadcastCache` (`js/net/src/consume.ts`) used by the `lite/` and `ietf/` subscribers: one base consumer per path; `get()` returns a `clone()` on a live hit, `insert()` caches + evicts-on-close (no callback). `broadcast.Consumer` is reference-counted: `clone()` shares one `BroadcastState`, `close()` decrements a shared count and tears down only at the last handle (a standalone consumer, count 1, unchanged). Callers keep closing normally. The wire-backed `ConsumeBroadcast` subclasses override `clone()` so shared handles keep resolving `info()`/`fetchGroup()` over the wire. Added `Consumer.closedSignal` for the cache's synchronous liveness check.

**Track.** The consumer side never registered its per-track producer into `state.tracks`, so the reuse check in `subscribe()` only fired for publisher-inserted tracks. Register consumer-created producers (scoped via a `register` flag so the publisher's dynamic-serve path is unchanged), evicting on close, so repeat subscriptions fan out from one producer / one upstream subscription; a closed track re-opens next subscribe.

**Groups.** Live groups already fan out to every track subscriber via the producer's mirror (frame bytes shared by reference), so track dedup covers them. The one duplicate path was the lite one-shot FETCH: cache the in-flight fetch producer per `[broadcast, track, sequence]` (created synchronously before the first await so a concurrent fetch coalesces), and hand each caller an independent `mirror()`; evict on close. moq-transport has no one-shot fetch.

**Frames** need no dedup: read sequentially within a group, and group fan-out already shares frame bytes by reference.

Also adds a `js/CLAUDE.md` convention to avoid callback parameters.

## Public API (`@moq/net`, additive)

`Broadcast.Consumer.clone()` and `Broadcast.Consumer.closedSignal`. `Connection.consume(path)` returns a reference-counted shared handle per live path; `broadcast.track(name).subscribe()` dedupes onto one upstream subscription; concurrent `track.fetchGroup(seq)` coalesce onto one FETCH. Targets `dev` (builds on #1371 / the `dev`-only #2120, #2032).

## Test plan

- [x] `bun test` in `js/net` (435 pass). New: `broadcast.test.ts` — repeat track subscriptions yield one request + clone refcount sharing; `integration.test.ts` — `lite`/`ietf consume dedup` (refcounted broadcast lifetime) and `lite` "coalesces concurrent fetches of one group" (two concurrent `fetchGroup(0)` both read the full group via independent mirrors).
- [x] `@moq/net` `check` (biome + tsc + tests) green.
- [ ] `@moq/watch` full typecheck fails **locally only** (worktree resolves `@moq/net` to the main repo's stale copy); passes in CI.

(Written by Claude Opus 4.8)

### DIFF
--- a/js/net/src/consume.ts
+++ b/js/net/src/consume.ts
@@ -1,0 +1,43 @@
+import type * as broadcast from "./broadcast.ts";
+import type * as Path from "./path.ts";
+
+/**
+ * Per-path dedup cache for consumed broadcasts, shared by the moq-lite and moq-ietf
+ * subscribers.
+ *
+ * `Connection.consume(path)` must not mint a fresh subscription per call: repeat requests
+ * for the same path (e.g. several renditions referencing one `broadcast: "../source"`) should
+ * share a single upstream subscription. This mirrors the Rust `origin::Consumer` weak-cache:
+ * a still-live path resolves to a shared {@link broadcast.Consumer.clone}, a closed one is
+ * re-consumed on the next request. Each handle is reference-counted, so the shared broadcast
+ * closes once every caller has closed its handle.
+ *
+ * @internal
+ */
+export class BroadcastCache {
+	// The base handle per path; callers get reference-counted clones of it.
+	#cache = new Map<Path.Valid, broadcast.Consumer>();
+
+	/** A shared handle to the live broadcast cached for `path`, or `undefined` on a miss. */
+	get(path: Path.Valid): broadcast.Consumer | undefined {
+		const base = this.#cache.get(path);
+		if (base && !base.closedSignal.peek()) return base.clone();
+		return undefined;
+	}
+
+	/**
+	 * Cache `consumer` as the base handle for `path` (evicting it once it closes) and return it.
+	 * Call on a {@link get} miss, after wiring up the fresh consumer's subscribe loop.
+	 */
+	insert(path: Path.Valid, consumer: broadcast.Consumer): broadcast.Consumer {
+		this.#cache.set(path, consumer);
+
+		// Drop the entry once the broadcast closes (every handle released), so the next request
+		// re-consumes rather than cloning a dead handle. Guard against a newer entry for the path.
+		void consumer.closed.finally(() => {
+			if (this.#cache.get(path) === consumer) this.#cache.delete(path);
+		});
+
+		return consumer;
+	}
+}
--- a/js/CLAUDE.md
+++ b/js/CLAUDE.md
@@ -66,6 +66,7 @@ Plain custom elements built directly on `@moq/signals`, no framework (except moq
 
 ## Conventions
 
+- **Avoid callback parameters.** A function taking a `fn`/`create`/`onXxx` to invoke later reads poorly and hides control flow. Prefer returning a value the caller acts on, exposing a method or getter, or splitting into a couple of small calls the caller sequences itself (e.g. a cache `get()` then `insert(value)`, not `getOrCreate(key, () => value)`). Reserve callbacks for genuine event/subscription sinks where there is no alternative (`effect.subscribe`, DOM listeners, `Signal` subscriptions).
 - ESM only (`"type": "module"`). Relative imports include the `.ts`/`.tsx` extension in the lower-level packages (`net`, `signals`, `hang`); `rewriteRelativeImportExtensions` in `tsconfig.json` rewrites them to `.js` on build. Some higher-level packages (watch/publish) still omit extensions, so match the file you are editing.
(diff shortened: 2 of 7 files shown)


demo/web/src/stats.ts
```diff
@@ -147,8 +147,9 @@ function subscribeNode(effect: Signals.Effect, conn: Net.Connection.Established,
 		};
 	});
 
+	// The connection dedupes and owns the consumed broadcast's lifetime, so don't close it
+	// here; the per-track subscriptions below still clean up on their own.
 	const consumer = conn.consume(path);
-	effect.cleanup(() => consumer.close());
```
kixelated, line 151: This .close() should reduce a ref count.

js/net/src/consume.ts
```diff
@@ -0,0 +1,45 @@
...
+ * @internal
+ */
+export class BroadcastCache {
+	#cache = new Map<Path.Valid, broadcast.Consumer>();
+
+	/**
+	 * Return the live consumer cached for `path`, or create one via `create`, cache it, and
+	 * evict it once it closes.
+	 */
+	consume(path: Path.Valid, create: (path: Path.Valid) => broadcast.Consumer): broadcast.Consumer {
```
kixelated: Please no callbacks like `create`

demo/web/src/stats.ts
kixelated: Done — `broadcast.Consumer` is now reference-counted: `clone()` shares one `BroadcastState` and `close()` decrements a shared count, tearing the broadcast down only at the last handle. This `.close()` now just decrements (no `js/watch`/`demo` change needed since they already close).

js/net/src/consume.ts
kixelated: Done — dropped the `create` callback; `BroadcastCache` now exposes `get()` / `insert()` that the subscriber sequences itself (`get(path) ?? insert(path, createConsume(path))`).

PR #2131, opened 2026-07-09
stats: tag Unix-socket (gateway) traffic with a legacy byte overlay

kixelated:
## What

Adds a "legacy" byte overlay to the relay's `.stats` aggregator so traffic on the internal Unix-socket listener (the out-of-process RTMP/SRT/WebRTC gateways) can be metered separately from native QUIC/WebTransport traffic.

Previously the per-broadcast byte counters (`publisher.json` / `subscriber.json`, split by external/internal tier) aggregated gateway and native traffic together, so a downstream billing service had no way to identify — or surcharge — legacy-protocol usage.

## How

- `moq-net`: `BroadcastEntry` gains two overlay counters (`legacy_publisher`, `legacy_subscriber`). `StatsHandle::with_legacy(bool)` threads a per-connection flag down to the `PublisherTrack::bytes` / `SubscriberTrack::bytes` sites, which bump the overlay **in addition to** the tier counters. Two new tracks surface it: `legacy/publisher.json` (gateway egress) and `legacy/subscriber.json` (gateway ingress).
- `moq-relay`: at accept, `connection.rs` sets `with_legacy(transport == "unix")` — the existing `Request::transport()` already distinguishes the UDS listener from QUIC (`"quic"` / `"websocket"`).

## Why additive (not a new tier)

The overlay is a **strict superset**: a gateway connection still counts as normal external egress/ingress, so all existing accounting is unchanged. This lets a billing service charge the gateway's bytes at the normal egress rate **plus** a legacy surcharge, rather than reclassifying (which would silently remove gateway traffic from external-egress metering).

## Tests

`legacy_overlay_is_additive_over_tier_bytes` verifies a legacy handle bumps both the tier and overlay counters while a native handle leaves the overlay untouched. All existing stats tests pass (`cargo test -p moq-net --lib stats`), clippy + fmt clean on both crates.

## Context

Consumed by the moq.pro dashboard to implement a per-GB "legacy protocol" surcharge (dash slurps the new tracks and meters them). Branched off the rev moq.pro currently pins so the dashboard can bump to this commit cleanly; will need a rebase onto `main` before merge.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Nqh72i6MGXd6CDWPxMEXbu)_

### DIFF
--- a/rs/moq-net/src/stats.rs
+++ b/rs/moq-net/src/stats.rs
@@ -380,6 +380,14 @@ struct StatsShared {
 struct BroadcastEntry {
 	publisher: [Counters; 2],
 	subscriber: [Counters; 2],
+	/// Legacy overlay: byte counters that accumulate ONLY for connections on the
+	/// internal Unix-socket listener (the out-of-process RTMP/SRT/WebRTC gateways).
+	/// A gateway connection bumps these IN ADDITION to its `publisher`/`subscriber`
+	/// tier counters, so legacy traffic is still counted as normal egress/ingress and
+	/// the overlay is a strict, additive superset (used to surcharge legacy protocols
+	/// downstream). Only `bytes` is populated; the other counter fields stay zero.
+	legacy_publisher: Counters,
+	legacy_subscriber: Counters,
 }
 
 impl BroadcastEntry {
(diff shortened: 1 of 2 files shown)


kixelated:
Closing: superseded by the arbitrary-label stats tiers already on the `dev` branch (`Tier::new(label)` auto-emitting `<label>/{publisher,subscriber}.json`). That's the proper general mechanism; this hardcoded "legacy" overlay isn't needed. The moq.pro legacy-protocol surcharge will build on dev's labels once it merges.

---

PR #2145, opened 2026-07-10
feat(moq-video): NVDEC hardware decode, zero-copy NVDEC -> NVENC transcode

kixelated:
Follow-up to #2140 (now merged into dev). Adds NVDEC hardware decode to `moq-video` and keeps the transcode pipeline fully GPU-resident: on an NVIDIA GPU, NVDEC decodes and scales in hardware and NVENC encodes the CUDA frame in place. No CPU copies anywhere between decode and encode.

## How it works

- **`moq-nvenc`**: new `cuvid` module, a dlopen'd NVDEC function table (`cuvid::Api::get()`). The vendored `sys` bindings declare the cuvid functions in an `extern "C"` block, which would make the linker require `libnvcuvid`; the table resolves them at runtime instead, so GPU-less builders still link and driverless hosts fall back. Unlike the encode side's panicking lazy static, `Api::get()` returns a `Result`.
- **`nvdec` decode backend** (new default-on `nvdec` feature in `moq-video`): drives the cuvid parser synchronously (`CUVID_PKT_ENDOFPICTURE` + zero display delay, so callbacks fire inside the parse call and one AU in means one frame out). Timestamps ride the parser (`ulClockRate` = 1 MHz), so they stay correct across reordering. Scaling happens **in the decoder**: `decode::Config::resize` maps to cuvid's target size, so frames come out already at the rung resolution for free.
- **Frame lifetime**: each mapped cuvid surface is copied device-to-device into an owned buffer and unmapped immediately (the surface pool is small and fixed). The buffer is a plain `cuMemAlloc`: NVENC's `NvEncRegisterResource` rejects stream-ordered pool memory (`cuMemAllocAsync`), which is what cudarc's `CudaSlice` uses on modern GPUs. Decode and encode share the device primary CUDA context, so no cross-context copies.
- **NVENC**: the session input format switches from IYUV to NV12 (NVENC's native layout, and what NVDEC emits). A `Frame::Cuda` is registered as an external resource and encoded in place; the CPU path interleaves I420 chroma during the pitched write as before.
- **`moq-transcode`**: the pipeline asks the decoder for the rung size and only falls back to the CPU `Scaler` when frames come back at another size (software decode, VT/MF). GPU frames flow `decoder.decode() -> encoder.encode()` untouched.
- **`libmoq`**: flattens each delivered frame to CPU I420 once at delivery (the C ABI hands out a stable byte pointer), so NVDEC offload works for C consumers too. The C ABI itself is unchanged.

## Public API changes (unreleased crates, hence dev)

- `moq_video::decode::Frame.data` (pub field) **removed**; pixels are now private. New `decode::Frame::into_i420(self) -> Result<Bytes>` downloads/packs on demand. **Breaking**, but the decode surface shipped on dev and is unreleased.
- `moq_video::decode::Decoder::new(catalog, &Kind)` -> `new(catalog, &Config)` (**breaking**), so decoder knobs stay additive.
- `moq_video::decode::Config` gains `resize: Option<(u32, u32)>` (additive, `#[non_exhaustive]`).
- New `moq_video::encode::Encoder::encode(&mut self, frame: &decode::Frame, keyframe: bool)`, the transcode input path.
- `moq-nvenc` gains the public `cuvid` module.
- New cargo features: `moq-video/nvdec` (default-on; forwarded in `moq-transcode`, `moq-cli`, enabled in `libmoq`).

`pub(crate)` internals (the decode `Backend` trait now carries timestamps and returns GPU-capable frames, `frame::Frame::Cuda`, `Encoder::encode_raw`) are not public surface.

## Testing (validated on a Linux RTX 3070 Ti, driver 595.71.05)

- `nvdec_h264_round_trip`: openh264 -> NVDEC, per-plane MAE vs the input picture, timestamps assert 1:1 passthrough.
- `nvdec_resize_scales_output`: decoder-side scaling vs a reference downscale.
- `nvdec_h265_round_trip`: NVENC HEVC -> NVDEC.
- `nvdec_to_nvenc_zero_copy`: asserts every decoded frame is CUDA-resident, re-encodes via the registered-resource path, and verifies the result with a software decode.
- `moq-transcode::end_to_end_hardware`: the full broadcast pipeline with `Kind::Hardware` on both ends (fetch path, all 5 frames through).
- All tests skip cleanly without the NVIDIA driver (verified by running without `LD_LIBRARY_PATH` in the Nix shell, where the driver libs are invisible); `missing_driver_errors_instead_of_panicking` covers the fallback.
- Workspace `cargo clippy --all-targets -D warnings` and `cargo fmt --check` pass. The macOS/Windows decode backends were updated for the trait change and need CI to compile-check.

(Written by Claude Fable 5)

### DIFF
--- a/rs/CLAUDE.md
+++ b/rs/CLAUDE.md
@@ -26,7 +26,7 @@ Layered roughly transport -> container/format -> media -> apps/bindings.
 
 - `moq-mux` (lib): the conversion layer. File/stream formats (`container/`: fmp4, flv, mkv, ts, loc) and codec parsers (`codec/`: h264, h265, av1, vp8/9, opus, aac, ...) <-> hang broadcasts. `Container` trait + generic `Producer<C>`/`Consumer<C>`. Dual catalog (`catalog::hang`, `catalog::msf`).
 - `moq-audio` (lib): native PCM <-> Opus (`unsafe-libopus`). `AudioProducer`/`AudioConsumer`, `Encoder`/`Decoder`, `AudioFormat`. Optional `capture` feature (cpal microphone), `resample`.
-- `moq-video` (lib): native video capture, H.264/H.265 encode, and decode; no ffmpeg. Hardware backends (VideoToolbox / Media Foundation / NVENC / VAAPI) with openh264 as the software H.264 fallback. `capture::Config`, `encode::{Encoder, Producer, publish_capture}`, `decode::{Consumer, Decoder}`.
+- `moq-video` (lib): native video capture, H.264/H.265 encode, and decode; no ffmpeg. Hardware backends (VideoToolbox / Media Foundation / NVENC / VAAPI / NVDEC) with openh264 as the software H.264 fallback; NVDEC frames stay in CUDA memory and feed NVENC zero-copy. `capture::Config`, `encode::{Encoder, Producer, publish_capture}`, `decode::{Consumer, Decoder}`.
 - `moq-transcode` (lib): just-in-time live transcoding of hang broadcasts. `run(source, output, config)` publishes a derivative catalog (ladder rungs + relative refs to the source) and encodes each rung only while subscribed/fetched, via `moq-video`. Output groups mirror source group sequences 1:1.
 
 **Apps / binaries**
--- a/rs/libmoq/Cargo.toml
+++ b/rs/libmoq/Cargo.toml
@@ -24,8 +24,8 @@ moq-audio = { workspace = true }
 moq-mux = { workspace = true }
 moq-native = { workspace = true, default-features = true }
 moq-net = { workspace = true }
-# Enable the NVENC / VAAPI hardware encoders (default-off at the workspace level).
-moq-video = { workspace = true, features = ["nvenc", "vaapi"] }
+# Enable the NVENC / VAAPI / NVDEC hardware codecs (default-off at the workspace level).
+moq-video = { workspace = true, features = ["nvenc", "vaapi", "nvdec"] }
 serde_json = "1"
 thiserror = "2"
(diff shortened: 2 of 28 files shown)


kixelated:
Self-review pass (CodeRabbit/Sourcery were both rate-limited). Findings, all addressed in 12b1a82d:

1. **NVENC input lifetime (correctness)**: `Nvenc::encode` dropped the registered CUDA resource / input buffer at the end of its match arm, before locking the output bitstream. The synchronous bitstream lock is what guarantees NVENC finished reading the input, so unregistering first raced the encoder (the pre-PR code kept the input alive across the lock). Both arms now drain the output while the input is still alive.
2. **Feature-combo build (nvenc without nvdec)**: `Frame::Cuda` was gated on `any(nvenc, nvdec)` but only NVDEC constructs it, so an nvenc-only build emitted dead-code warnings (fails `-D warnings`). The CUDA frame plumbing is now gated on `nvdec` alone; all of `nvenc` / `nvdec` / both / neither compile warning-free.
3. **Decoder reuse vs display crop (edge case)**: the sequence callback kept the existing cuvid decoder when coded + target sizes matched, ignoring a display-area change (a mid-stream SPS crop change at the same coded size would scale from a stale source rect). The crop is now part of the reuse comparison.
4. **Stale doc**: the decode backend module doc still said backends return "I420 frames"; now describes `Decoded`.
5. **Cosmetic**: cuvid symbol-load error messages embedded the trailing NUL of the symbol name.

Re-validated on the RTX 3070 Ti after the fixes: all 5 NVDEC hardware tests, the zero-copy NVDEC -> NVENC test, and the moq-transcode hardware end-to-end pass; workspace clippy/fmt clean.

(Written by Claude Fable 5)

PR #2157, opened 2026-07-10
fix(moq-transcode): remove the live/fetch two-writer race in rung serving

kixelated:
## Summary

Follow-up flagged inline when `moq-transcode` landed (#2140): the live loop and the fetch tasks wrote into the same output track independently, so a live-edge fetch could claim a sequence first and the live loop skipped it. If that fetch then failed, the rung lost one GOP (subscribers saw an aborted group and nothing re-produced it).

`rung.rs` now serves each rung through **one event loop (`Serve`) that owns every output sequence**, so the two paths can never race:

- The live session runs as a state machine inside the loop (`Subscribing` / `Idle` / `Deferred` / `Transcoding`), driven one step at a time. Fetch requests stay responsive mid-group, same as before when the paths were concurrent futures.
- A fetch task only transcodes a sequence the loop granted it; grants are tracked per task and reaped through the `JoinSet`.
- A request for a sequence the live session is heading toward (or one already granted to a fetch) is **parked** instead of double-transcoded; it resolves from the shared track cache once the group exists. Side effect fix: a live-edge fetch no longer errors with `NotFound` when the source can't serve fetches, it waits for the live session to produce the group (covered by the new `live_edge_fetch_waits_for_live` test).
- If the live session catches up to a granted sequence, it **defers** that group until the fetch reports back: on success the fetched group already serves every subscriber via the cache (skip, no duplicate work); on failure the loop re-creates the aborted group (`create_group` replaces aborted slots in place) and transcodes it live. No GOP is ever lost, which the old code's inline comment called out as the residual.

The fetch pipeline itself (`fetch` / `transcode_group` / `Pipeline`) is unchanged apart from dropping the now-unneeded `demand` parameter; the live path's per-frame work moved into the loop as `process_frame`, shared with the fetch path.

## Public API changes

None. Everything touched is `pub(crate)` or private (`Serve`, `Live`, `Session`, `Event` are new private types). No wire, catalog, or `pub` surface change; targeting `dev` because `moq-transcode` is dev-only and this reshapes the crate it landed with there.

## Test plan

- [x] `cargo test -p moq-transcode` (12 tests, including the existing end-to-end live + fetch coverage)
- [x] New regression test `live_edge_fetch_waits_for_live`: fetch a not-yet-existing group while subscribed; the old code failed it with `NotFound` via an eagerly spawned fetch, the new code parks it and both the subscriber and the fetch read the same live-transcoded group.
- [x] `cargo clippy -p moq-transcode --all-targets`, `cargo fmt --check`, `RUSTDOCFLAGS="-D warnings" cargo doc -p moq-transcode`

(Written by Claude Fable 5)

### DIFF
--- a/rs/moq-transcode/src/lib.rs
+++ b/rs/moq-transcode/src/lib.rs
@@ -147,7 +147,27 @@ mod tests {
 	struct Source {
 		broadcast: moq_net::broadcast::Producer,
 		_catalog: moq_mux::catalog::Producer,
-		_track: moq_net::track::Producer,
+		track: moq_net::track::Producer,
+		encoder: moq_video::encode::Encoder,
+	}
+
+	impl Source {
+		/// Append one group of `frames` gray frames to the video track.
+		fn append_group(&mut self, sequence: u64, frames: u64) {
+			let gray = vec![0x80u8; 320 * 240 * 4];
+			let mut group = self.track.create_group(sequence.into()).unwrap();
+			for index in 0..frames {
+				let timestamp = (sequence * frames + index) * 33_333;
+				for payload in self.encoder.encode_rgba(&gray, 320, 240, index == 0).unwrap() {
+					let frame = hang::container::Frame {
+						timestamp: moq_net::Timestamp::from_micros(timestamp).unwrap(),
+						payload,
+					};
+					frame.encode(&mut group).unwrap();
+				}
+			}
+			group.finish().unwrap();
+		}
 	}
 
 	/// Build a 320x240 avc3 source broadcast: a catalog plus a video track with
@@ -169,36 +189,25 @@ mod tests {
 		catalog.lock().video.insert("video", video).unwrap();
 
 		let info = moq_net::track::Info::default().with_timescale(hang::container::TIMESCALE);
-		let mut track = broadcast.create_track("video", info).unwrap();
+		let track = broadcast.create_track("video", info).unwrap();
 
-		let mut encoder = moq_video::encode::Encoder::new(&{
+		let encoder = moq_video::encode::Encoder::new(&{
 			let mut config = moq_video::encode::Config::new(320, 240, 30);
 			config.kind = moq_video::encode::Kind::Software;
 			config
 		})
 		.unwrap();
-		let gray = vec![0x80u8; 320 * 240 * 4];
 
-		for sequence in 0..groups {
-			let mut group = track.create_group(sequence.into()).unwrap();
-			for index in 0..frames {
-				let timestamp = (sequence * frames + index) * 33_333;
-				for payload in encoder.encode_rgba(&gray, 320, 240, index == 0).unwrap() {
-					let frame = hang::container::Frame {
-						timestamp: moq_net::Timestamp::from_micros(timestamp).unwrap(),
-						payload,
-					};
-					frame.encode(&mut group).unwrap();
-				}
-			}
-			group.finish().unwrap();
-		}
-
-		Source {
+		let mut source = Source {
 			broadcast,
 			_catalog: catalog,
-			_track: track,
+			track,
+			encoder,
+		};
+		for sequence in 0..groups {
+			source.append_group(sequence, frames);
 		}
+		source
 	}
 
 	/// Whether a hardware decoder AND encoder are usable here (e.g. a Linux box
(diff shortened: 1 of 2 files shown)


kixelated:
## Review: the unified `Serve` loop trades one rare residual for several worse hot-path failures

Automated multi-angle review of the new single serving loop. The two-writer race itself is genuinely closed (I could not construct a double-write: `live_latest` is set at `create_group` time, so a racing fetch hits `Duplicate` on `accept`). But the parking/defer design that replaces the old independent live + fetch paths introduces regressions on the exact hot path this feature targets (rendition-switch fetches near the live edge). Holding the PR to rework these.

### Blocking

**1. A fetch consumer can be handed an aborted half-group (correctness).**
A request parked behind the live session is *dropped* in `unpark()` once live creates the group, and the consumer resolves from the **shared, still-streaming live** `group::Producer` via the cache. If the last subscriber then leaves mid-group, `stop_live()` → `abort_live_group()` aborts that shared group, so the fetch consumer sees `Error::Cancel` partway through a group it requested complete. The old fetch path transcoded each requested group to completion independently of live demand. `rung.rs` `start_group`/`unpark` (~L322-L345) + `stop_live`/`abort_live_group` (~L451-L473). Root cause: **a fetch consumer must never depend on an abortable live group.**

**2. Live progress is coupled to fetch scheduling; potential indefinite stall.**
`on_group` enters `SessionState::Deferred(X)` whenever `fetching_sequence(X)` is true, and `fetching` is populated in `spawn_fetch` *before* the task acquires one of the `MAX_CONCURRENT_FETCHES` permits. So the live session (and every subscriber on the rung) can block on a fetch that is merely queued — and in the pathological case where all permits are held by fetches for not-yet-produced future groups, the deferred live group never resolves. The old `create_group → Duplicate → continue` skip never blocked live on a fetch. `rung.rs` `on_group` (~L310-L318), `spawn_fetch` permit acquisition (~L482-L490).

**3. Parking predicate over-parks → fetch hangs.**
`live_ahead = live.is_some() && live_latest.is_none_or(|l| seq > l)` treats *every* sequence as ahead whenever `live_latest` is `None` (the whole window between subscribe and the first live group). In that window a fetch for an old, already-produced group is parked and hangs until live produces its first group; a fetch for a future group the source never reaches is parked forever (the old path returned `NotFound` promptly). `rung.rs` `on_request` (~L389-L397). Fix wants the track's real high-water mark, not `live_latest`.

**4. Fetch admission backpressure lost + teardown drops the real error.**
The semaphore moved *inside* the spawned task, so a burst of distinct requests spawns unbounded tasks (each holding a `Rung` clone + `GroupRequest`), only 4 of which run; the old loop bounded in-flight requests by awaiting the permit before pulling the next `requested_group()`. Separately, on an error exit `run()` drops `self.parked`, auto-rejecting each waiting consumer with `Error::Dropped` instead of the real cause — contradicting the invariant the `fetch()` doc comment is built around. `rung.rs` `spawn_fetch` (~L478-L497), `run` teardown (~L197-L203).

### Non-blocking

- **Simplify the fetch bookkeeping.** `JoinSet<bool>` + `HashMap<tokio::task::Id, u64>` reverse-index can become `JoinSet<(u64, bool)>` (return the sequence from the task) + a `HashSet<u64>` for in-flight membership. That deletes `join_next_with_id`, the `.expect("fetch task not tracked")` panic path, and the linear `fetching_sequence` scan.
- **Coverage gap:** the headline defer + fetch-failure-re-transcode path is untested; the in-process test source has no fetch handler, so `live_edge_fetch_waits_for_live` only exercises park-then-cache and its `0..50` yield loop is a scheduler bet (fine on current-thread, fragile under `multi_thread`).
- Conventions (CLAUDE.md), the two-writer invariant itself, and the moq-net API assumptions (`Duplicate`/`replace_evicted`, `poll_fetch` checking the cache before rejections, `accept` returning `Duplicate`) all check out.

**The crux for the rework:** live groups are abortable (demand-driven) but fetch consumers need complete groups, so the two can't share a group. A correct design either (a) guarantees live groups complete once started (finish the in-flight group on demand loss, then stop) so sharing is safe, or (b) always gives fetches their own independent, completion-guaranteed group. Either removes the defer coupling and the parking-abort hazard.

(Written by Claude Fable 5)

PR #2163, opened 2026-07-11
fix(js): Safari playback/publish support, connection reliability, and a capture-releasing mute

fperex:
Batch of browser fixes developed on the fperex fork, focused on making Safari a first-class client without touching Chrome/Firefox behavior (Safari-only changes are gated on UA predicates), plus connection-reliability fixes that apply everywhere.

## Safari transport

- Safari is routed onto the WebSocket (qmux) fallback unconditionally: reading `WebTransport.datagrams.readable` hard-kills the Safari session with no catchable error, so the QUIC race is never entered (`pickTransport` in `@moq/net`).
- Firefox keeps its existing WebSocket routing (drops server-initiated bidi streams).
- Both support matrices (`@moq/publish/support`, `@moq/watch/support`) report `webtransport: "partial"` on Safari/Firefox to surface the degraded path.

## Connection reliability

- `Connection.Reload` now reconnects when an established session **closes**, not just when a connect attempt fails. `closed` resolves (never rejects) on qmux, so a dropped session previously left a dead `established` and a permanent "connected" status, with Safari/Firefox never recovering from any drop.
- A Safari `pagehide` no longer latches the connection off permanently; the connection closes eagerly on pagehide/freeze (so the relay unannounces immediately) and resumes on pageshow/visibility.
- `announcedGenerations` tracks a per-path announce generation so a watcher re-consumes a same-name republish even when the unannounce+announce coalesce.
- Stream teardown classification (`isStreamAbort`): routine lifecycle resets (unsubscribe, handover, session close) log at debug; client-actionable faults (auth, not-found, protocol, unroutable) still warn, on every run-loop catch across net/hang/watch.

## Publish: audio privacy mute

- `muted` now **stops audio capture** while the encoder pipeline stays up and publishes true silence via a keep-alive graph. The catalog never drops or reshapes the audio track, so watchers keep their audio subtree instead of rebuilding it (no video freeze on mute).
- Per source: a **microphone** track is stopped outright (`track.stop()`, so the OS recording indicator turns off); **screen/file** audio is detached from the encoder instead, since stopping a share track would kill the share and force a re-prompt.
- Selecting a source while muted never touches the microphone at all; unmuting acquires it.

## Publish: capture/encode

- Safari hardware video encode; video capture runs in a Worker where supported (`safariWorkerCapture`, WebKit 18+).
- Opus catalog now always advertises 48 kHz (RFC 6716 decoder-side rate). Non-48 kHz capture (Safari captures at the hardware rate and ignores the context-rate request) is stream-resampled before encode with cadence-snapped timestamps.
- Video catalog folds in the encoder-produced codec string and hvcC/av1C description after the first keyframe, tagged against the requested codec+dimensions so bandwidth-driven bitrate churn (~10x/s) never flaps the description.
- Camera/microphone re-acquire on track `ended` (all browsers); Safari-gated recovery for a backgrounded tab leaving the camera stuck muted.
- Stats tab graphs a measured, transport-agnostic upload bitrate from encoded bytes.

## Watch: decode robustness

- Budgeted in-place decoder restarts (audio + video) with config-supersession checks, keyframe resync on `DataError`, and a decode-queue cap.
- Audio: rate-mismatch resampling, timestamp snapping, discontinuity re-anchoring on resubscribe, AudioContext resume on user gesture.
- Catalog subscription that ends unexpectedly is reopened with bounded backoff; status settles to "offline" when the budget is exhausted.
- New unsupported-codec indicator in the UI overlay (catalog has renditions but none decodable here).

## Public API changes (js/*)

Breaking (why this targets `dev`):
- `@moq/net` `Connection.Established.transport` — new **required** readonly member on a public interface. Additive for consumers; a compile break for external implementers of the interface.

Additive:
- `@moq/net`: `isStreamAbort` (root export), `Connection.Transport` type, `Connection.Reload.announcedGenerations`.
- `@moq/hang`: `Util.Hacks.isSafari`, `Util.Hacks.safariWorkerCapture` (plus pure `detect*` cores primarily consumed internally/by tests).
- `@moq/publish`: `Video.Encoder.bytesEncoded`.
- `@moq/watch`: optional `announcedGenerations` in the `Broadcast` constructor options bag, `Video.Source.output.unsupported`, `AudioBuffer.wait(timestamp, signal?)`.

Behavior changes on existing surface: Safari/Firefox transport routing, reconnect-on-close, mic-releasing `muted`, 48 kHz Opus catalog rate, `isSupported()` matrices (details above).

## Known upstream issue

Testing this PR on Safari/Firefox (which it routes onto the WebSocket/qmux fallback) surfaces `Unhandled rejection: TypeError: ReadableStreamDefaultController is not in a state where chunk can be enqueued` when a peer reloads while connected. This is a known `@moq/qmux <= 0.3.1` bug, already fixed upstream in moq-dev/web-transport (release pending); it is not introduced by this PR and the pin can be bumped when the release ships.

PR Link: https://github.com/moq-dev/web-transport/pull/285

## Test plan

- `just js check` and `just js test` green (544 tests across 8 packages; new unit tests for transport picking, error classification, cadence/resampler math, video catalog capture, config supersession, timestamp snapping, container consumer).
- Headless-Chrome (fake media device) end-to-end privacy proof: selecting a source while muted performs zero `getUserMedia` calls; every mute ends the mic track; the audio catalog stays present and shape-identical across mute cycles; exactly one AudioContext across the full cycle (pipeline survives mute).
- Manual Safari verification on device: mic indicator off on mute, watcher video does not freeze across publisher mute/unmute.


kixelated:
Sorry but this needs to be split up into separate PRs. Many of the fixes are good, but some are wrong or misguided, which makes it harder to review the entire PR.

fperex:
Split up as requested: closing this in favour of 13 focused PRs. Thanks for the push, the split surfaced two things I had wrong (below).

**Independent, review in any order:**

| PR | What |
|---|---|
| #2180 | `docs: add root CONTRIBUTING.md` |
| #2181 | `fix(demo)`: mic processing controls, error messages on Safari/Firefox, shrinkable buffer bar |
| #2182 | `fix(net)`: classify routine stream teardown apart from real faults (`isStreamAbort`) |
| #2183 | `fix(net)`: reconnect when the session **closes**, not just when it errors |
| #2184 | `fix(net)`: force Safari onto the WebSocket transport |
| #2185 | `fix(js)`: don't latch the connection off after a Safari `pagehide` (+ Safari fullscreen) |
| #2186 | `fix(publish)`: stop audio capture on mute, publishing silence in its place |
| #2187 | `fix(watch)`: survive a video decoder error instead of going dark |

**Stacked (each notes its parent; review the last commit only until the parent merges):**

| PR | Stacked on | What |
|---|---|---|
| #2188 | #2182 | `fix(watch)`: recover from a reset catalog and a same-name republish |
| #2189 | #2182 | `fix(watch)`: survive an audio decoder error, resample, re-anchor on resubscribe |
| #2190 | #2184 | `feat(publish)`: Safari hardware encode, worker capture, stable video catalog |
| #2191 | #2186 | `fix(publish)`: advertise 48 kHz Opus and resample a non-canonical capture rate |
| #2192 | #2184 | `feat(net)!`: expose the negotiated transport on `Established` |

**Two things the split corrected, which support your "some are wrong or misguided" read:**

1. **This PR claimed a breaking change it did not actually make.** It advertised dropping a `safariVersion` export as breaking, but that export never existed upstream; it was churn I introduced and then removed *within* my own branch. The only genuinely breaking change is widening the `Established` interface, and that is now isolated in **#2192** so you can reject it on its own without blocking any Safari fix. Everything else is additive or internal.

2. **The mute redesign was wrong once and fixed itself mid-branch.** The history here contains a soft-mute (ramp the encoder gain to zero, keep the mic capturing) that I later replaced, because it leaves the OS recording indicator **lit** while the UI says you are muted, which is the one thing a mute must not do. #2186 contains only the final design (stop the mic, publish silence from a keep-alive node so the catalog never flaps). None of the dead intermediate design is in the split.

The split is by **final state**, not by cherry-picked commits, precisely because of the above: replaying the commits would have replayed designs I had already abandoned.

**Coverage audit** (so nothing was silently dropped): merging all 13 branches onto current `dev` reproduces this branch's tree **byte-for-byte** across 69 of the 73 files. The other 4 (`connect.ts`, `ietf/publisher.ts`, `lite/{publisher,subscriber}.ts`) differ only by *your* newer work that landed since I branched (lite-06 announce ids, the subscription-priority options object), which the split absorbed by rebasing onto current `dev`. The union is +3057/−434 over 73 files, identical to this PR's own stat line. `just js check` and `just js test` are green on every branch and on the union (548 tests, 0 failures, including `js/hang`'s suite, which had no `test` script and had never actually run until #2184 added one).

PR #2179, opened 2026-07-12
feat(moq-net): cost-based cache-aware routing (moq-lite-06 WIP) + route-cost I-D

kixelated:
## What

Replaces hop count as the cluster routing metric with an accumulated **route cost**, carried on announcements in a new (WIP, opt-in) **moq-lite-06** version, specced as an extension of the relay-hops I-D (`drafts/draft-lcurley-moq-relay-hops.md`).

The core model: a broadcast's announcement metadata is one shared, mutable **`Route`** (hop chain + cost). A repeat ANNOUNCE replaces it **in place** — same broadcast, same subscriptions, new route. Teardown belongs to the data plane (a producer dying), never to metadata: an announcement rides a session, so a route update can't change which peer the media comes from, only the advertised path/cost beyond it. Origin-level restarts happen only where the active consumer actually swaps; downstream they arrive as in-place updates, so route changes no longer cascade teardowns across the mesh.

## Why

Routing today is blind to what a link actually costs and to which candidate already carries the media. In a CDN deployment two relays in one datacenter each pull the same broadcast over a metered backbone (the via-sibling route is +1 hop, so shortest-hop routing never picks it), and cross-region ties break by hash rather than toward the peer already streaming it. The parallels are classic: OSPF-style per-link costs, a Babel-style damped dynamic metric, and marginal-cost advertisement for the cache signal.

## How

- **Model:** `broadcast::Info.route: Route` — hops + `base` + `transit` in one kio cell shared by every clone. `origin::Producer::refresh_broadcast` re-ranks a path after a route mutation and restarts only if the winner changed.
- **Wire (lite-06 only):** `ANNOUNCE_BROADCAST` gains two cost varints (`base` forwarded unchanged, `transit` accumulated per link). No new message: a repeat Active *is* the update, applied in place by the receiver. Old versions derive cost from the hop count, so ordering is bit-identical to the current `(hops.len(), fnv)` key.
- **Cache-aware:** a relay actively carrying a broadcast (any live track producer) advertises **zero transit** — its upstream is already paid for.
- **Propagation:** the publisher tracks each announced path's last-sent route and re-announces in place when hops/cost/activity diverge (lite-05+), including lifting or imposing `exclude_hop` suppression that an in-place hop update can toggle. Hop changes and cost decreases forward within a 1s tick; pure cost increases hold down 30s; receivers re-rank after a deterministic per-node jitter so mutually-preferring peers don't swap simultaneously.
- **Link pricing:** `cluster.connect_api` entries may be `{host, cost}` (0 = free same-DC link), declared in SETUP so both ends price the link identically.
- **Opt-in:** `moq-lite-06-wip` ALPN, excluded from `Versions::all()`/`ALPNS`; opt in via `Versions::with_preferred`. Transport accept dispatch recognizes any parseable moq ALPN (`is_moq_alpn`).
- **Spec:** folded into `draft-lcurley-moq-relay-hops`: optional Link Cost on the RELAY_HOPS setup option, Route Cost + ROUTE_COST parameter sections, and a Route Updates section — re-advertisement applied in place, MUST NOT interrupt existing subscriptions.

## Safety

- Loop prevention untouched: admission (reflected-chain drop, `exclude_hop`) runs before ranking; a repeat announce whose chain now loops through the receiver retires the broadcast.
- Mixed-version meshes degrade to today's behavior (worst transient: a duplicate pull, never a loop).

## Tests

Wire round-trips (cost on Active, version gating); model (cost-beats-hops, base cost, refresh promotes/demotes, `is_active` lifecycle); real-QUIC e2e: base-cost propagation, link-cost addition, hot flip in place, **and route-swap continuity** — a cheaper broadcast replaces the active one at the server while the client's in-flight subscription keeps receiving frames from the original producer, its broadcast handle showing the new hops, with zero re-announce events. Full workspace suite green; clippy/fmt/doc clean; the I-D renders end-to-end.

## Public API (moq-net)

`broadcast::Route` (new), `broadcast::Info { route, origin }`, `Consumer::is_active`, `origin::Producer::refresh_broadcast`, `Client::with_link_cost`, `Setup.link_cost`, `Versions::with_preferred`, `is_moq_alpn`, `lite::RouteCost`, `Version::Lite06Wip`. No separate Cost type, no ANNOUNCE_UPDATE message, no `with_cost_base` builder.


kixelated:
Superseded by #2894, which starts from current `dev` and implements the remaining cold-route-rank slice from the current pop-skipping quest. #2424 already landed the cumulative route-cost foundation, so this prototype's base/transit split is no longer the active design.

(Written by GPT-5)

PR #2181, opened 2026-07-12
fix(publish,demo): align media constraints

fperex:
Split out of #2163.

Summary:

- Add demo mic processing controls for `echoCancellation`, `autoGainControl`, and `noiseSuppression` while keeping Auto as the browser default.
- Request the selected camera resolution and frame rate on the source, and keep the demo encoder cap aligned with that selected target.
- Let the video encoder use source `width.max * height.max` as the `maxPixels` fallback when both constraints are present and `config.maxPixels` is unset.
- Fix Safari/Firefox console-overlay error formatting, the speaker-mute reset on catalog updates, and buffer-bar shrinking in narrow layouts.

Public API changes:

- No new, renamed, removed, or signature-changed exports.
- Non-breaking behavior/doc change: `EncoderConfig.maxPixels` now documents and uses source `width.max` and `height.max` as a fallback cap when both are present.

Branch targeting:

- These changes are non-breaking and would normally target `main`. The branch is based on `dev`; retargeting to `main` would include unrelated `dev` changes, so this PR remains on `dev`.

Test plan:

- `nix develop --command env CARGO_TARGET_DIR=/Users/kixelated/.codex/worktrees/078c/moq/target just check`
- Remote `Check` is green.

(Written by GPT-5)

### DIFF
--- a/demo/web/src/publish.ts
+++ b/demo/web/src/publish.ts
@@ -113,7 +113,7 @@ type VideoTarget = {
 	frameRate?: number;
 };
 
-const videoTarget = ui.computed((effect): VideoTarget => {
+function readVideoTarget(effect: Signals.Effect): VideoTarget {
 	const res = effect.get(resolution);
 	const [rawWidth, rawHeight] = res ? res.split("x").map(Number) : [undefined, undefined];
 	return {
@@ -121,7 +121,7 @@ const videoTarget = ui.computed((effect): VideoTarget => {
 		height: rawHeight && Number.isFinite(rawHeight) ? rawHeight : undefined,
 		frameRate: effect.get(framerate),
 	};
-});
+}
 
 function cameraConstraints(target: VideoTarget): Video.Constraints | undefined {
 	const constraints: Video.Constraints = {};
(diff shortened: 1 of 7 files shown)


demo/web/src/publish.ts
```diff
@@ -117,13 +123,41 @@ ui.run((effect) => {
...
+// holds the active Camera source (undefined for screen/file); its constraints re-acquire the track
+// on change. getUserMedia uses `ideal`, so a camera that can't reach the target falls back to its
+// best (the green "actual" readout shows what it gave).
+ui.run((effect) => {
+	const source = effect.get(publish.video);
+	if (!source || !("constraints" in source)) return;
+
+	const res = effect.get(resolution);
+	const [w, h] = res ? res.split("x").map(Number) : [undefined, undefined];
+	source.constraints.set(w && h ? { width: { ideal: w }, height: { ideal: h } } : undefined);
```
kixelated: This feels wrong here. It's supposed to be a simple demo, but something is wrong if the source and video encoder don't default to this behavior?

demo/web/src/publish.ts
```diff
@@ -117,13 +123,41 @@ ui.run((effect) => {
...
+ui.run((effect) => {
+	const source = effect.get(publish.audio);
+	if (!source || !("constraints" in source)) return;
+
+	const constraints = {
+		echoCancellation: effect.get(echoCancellation),
+		autoGainControl: effect.get(autoGainControl),
+		noiseSuppression: effect.get(noiseSuppression),
+	};
+	const any = Object.values(constraints).some((v) => v !== undefined);
```
kixelated, line 182: IDK if this is needed, just set.

doc/lib/js/env/web.md
```diff
@@ -336,6 +336,8 @@ Requires modern browser features:
 - Firefox (behind flag)
 - Safari (future support planned)
 
+On macOS the browser's default microphone constraints engage the system voice processor: auto gain control can slowly pull the capture level down, and other system audio is ducked while the mic is live. Pass `{ echoCancellation: false, autoGainControl: false, noiseSuppression: false }` via the microphone source's `constraints` to capture the raw signal instead (expect echo without headphones).
```
kixelated: I wouldn't document this as a macOS specific bug. I would instead document the `echoCancellation` and friends flags.

PR #2182, opened 2026-07-12
fix(net): classify routine stream teardown apart from real faults

fperex:
Split out of #2163.

Every run loop logged any error from a stream at `warn`, so a routine unsubscribe, a publisher handover, or a peer closing its session produced scary warnings on every viewer that stopped watching. The noise made real faults hard to find.

`isStreamAbort(err)` decides from the application code the relay actually encodes (mirroring `rs/moq-net`'s `error.rs`): `WebTransportError.streamErrorCode` on native transports, or the trailing number in qmux's `RESET_STREAM: <code>` / `STOP_SENDING: <code>` / `Connection closed: <code>` message on the WebSocket fallback. Fault codes are a **denylist**, so an unknown code is treated as routine teardown while a client-actionable fault (auth, not-found, protocol violation, unroutable) still surfaces at `warn`.

Notes on the details that are easy to get wrong, and that the tests pin:
- The coded check runs **before** any message heuristic, so a real `WebTransportError` can never be reclassified as routine just because its prose happens to mention a closing stream.
- Safari's `InvalidStateError` is keyed on `err.name`, not on matching the word "invalid state" in the message, so an unrelated internal error ("the decoder is in an invalid state") is not silently downgraded to debug.
- Chrome reports a write-side abort (a downstream unsubscribe seen by the publisher) with source `session`, not `stream`, so gating on `source` would leave that routine teardown warning on every closed viewer. We trust the numeric code instead.

The bandwidth probe streams previously warned on every error, burying the genuine auth/protocol faults on that side channel in routine teardown noise; they now use the same classification.

**Public API changes:** adds `isStreamAbort` to the `@moq/net` root export (additive).

**Test plan:** `bun test js/net/src/util/error.test.ts`: 8 tests covering coded resets, qmux message forms, session close, Safari's InvalidStateError, and the fault denylist. `just js check` green.

### DIFF
--- a/js/net/src/util/error.ts
+++ b/js/net/src/util/error.ts
@@ -3,6 +3,69 @@ export function error(err: unknown): Error {
 	return err instanceof Error ? err : new Error(String(err));
 }
 
+// Stream-reset application codes that indicate a client-actionable FAULT (bad auth, wrong path, protocol
+// violation, unroutable, ...) rather than a routine teardown. Mirrors rs/moq-net/src/error.rs
+// `Error::to_code()`. Kept as a DENYLIST: any code NOT listed (Cancel=0, Timeout=3, Transport=4,
+// Dropped=24, Closed=25, ...) is treated as an expected teardown, so a normal unsubscribe/handover reset
+// stays quiet whatever code it carries, while a genuine failure surfaces.
+const STREAM_FAULT_CODES = new Set<number>([
+	6, // Unauthorized
+	9, // Version
+	11, // BoundsExceeded
+	12, // Duplicate
+	13, // NotFound
+	14, // WrongSize
+	15, // ProtocolViolation
+	16, // UnexpectedMessage
+	17, // Unsupported
+	27, // FrameTooLarge
+	30, // Unroutable
+]);
+
+/**
+ * True when `err` is a stream-lifecycle event a caller can log at debug rather than warn: false for a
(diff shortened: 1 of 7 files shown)


js/net/src/util/error.ts
```diff
@@ -3,6 +3,69 @@ export function error(err: unknown): Error {
...
+/**
+ * True when `err` is a stream-lifecycle event a caller can log at debug rather than warn: false for a
+ * client-actionable fault (see STREAM_FAULT_CODES) or a non-stream error.
+ *
+ * Reads the application code the relay encodes (rs/moq-net `error.rs`): `WebTransportError.streamErrorCode`
+ * on native transports, or the trailing number in qmux's "RESET_STREAM: <code>" / "STOP_SENDING: <code>"
+ * message on the WebSocket fallback. Keyed on `err.name` rather than `instanceof WebTransportError` so it
+ * is safe where that global is undefined (e.g. the test runner).
+ */
+export function isStreamAbort(err: unknown): boolean {
```
kixelated, line 34: IDK this is too fickle. It's the reason we don't classify errors correctly yet.

kixelated: I'm not quite sure what to do about this, lemme think about it for a bit.

kixelated:
IDK I think more thought needs to go into error messages first.

PR #2183, opened 2026-07-12
fix(net): reconnect when the session closes, not just when it errors

fperex:
Split out of #2163.

**A dropped session never reconnected.** `Established.closed` *resolves* on every session end: qmux never rejects it, and native WebTransport rejects only on an abnormal close. But `Reload` only retried from its `catch` block, so a session that simply closed fell through without scheduling a retry, leaving a dead `established`, a status still reading `"connected"`, throughput at zero, and no reconnect attempt ever. A page reload was the only way back.

This disproportionately hit Safari and Firefox, which ride the qmux WebSocket (where `closed` *always* resolves), so for them it was the only path to drop recovery.

The fix treats a resolved `closed` as a disconnect: clear `established`, report `"disconnected"`, and retry on the existing backoff ladder.

Two subtleties, both deliberate:
- The effect teardown and the session close both resolve to `undefined`, so racing them is ambiguous. A `CANCELLED` sentinel tells "the effect was torn down" apart from "the session ended", and only the latter reconnects.
- The backoff ladder only resets once a session has survived `STABLE_CONNECTION_MS`. A relay that accepts us and immediately drops us (bad auth, shedding load, a redirect it cannot serve) would otherwise reset the ladder on every attempt and get hammered once a second forever.

Also pins the per-run `cancel`/`abort` handles, since a rerun swaps both before awaiting the spawn, and swallows the `closed` rejection in the bandwidth poll, which only uses it as a stop signal (that rejection is what drives the reconnect and must not surface as an unhandled rejection).

**Public API changes:** none.

**Test plan:** `just js check` green. Verified in the demo by bouncing the relay: the viewer reconnects on the backoff ladder instead of sitting on a permanently "connected" dead session.

### DIFF
--- a/js/net/src/connection/reload.ts
+++ b/js/net/src/connection/reload.ts
@@ -1,6 +1,7 @@
 import { Effect, type Getter, Signal } from "@moq/signals";
 import type * as Path from "../path.ts";
 import { empty as emptyPath } from "../path.ts";
+import { error } from "../util/error.ts";
 import { type ConnectProps, connect, type WebSocketOptions, type WebTransportProps } from "./connect.ts";
 import type { Established } from "./established.ts";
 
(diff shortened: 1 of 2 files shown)


kixelated:
> A dropped session never reconnected. Established.closed resolves on every session end: qmux never rejects it, and native WebTransport rejects only on an abnormal close. But Reload only retried from its catch block, so a session that simply closed fell through without scheduling a retry, leaving a dead established, a status still reading "connected", throughput at zero, and no reconnect attempt ever. A page reload was the only way back.
>
> This disproportionately hit Safari and Firefox, which ride the qmux WebSocket (where closed always resolves), so for them it was the only path to drop recovery.

This sounds like a qmux bug, and fixing it here is wrong. File an issue or submit a PR to qmux.

PR #2184, opened 2026-07-12
fix(net): force Safari onto the WebSocket transport

fperex:
Split out of #2163.

**Safari cannot use WebTransport here at all.** Reading `WebTransport.datagrams.readable` tears down the whole session on Safari, and moq-lite-05 always reads it once the negotiated version carries datagrams. There is no catchable failure and no way to opt out per-session, so a Safari client that wins the WebTransport race is simply dead. Route Safari onto the qmux WebSocket fallback, exactly the way Firefox already is.

The rule lives in a pure `pickTransport(userAgent, hasWebTransport)` rather than another inline boolean in `connect.ts`, because "is this WebKit" is genuinely subtle and now has a test matrix:
- iOS Chrome (`CriOS`) and iOS Firefox (`FxiOS`) are WebKit under the hood and must be routed like Safari, even though one says "Chrome" and the other never says "Firefox".
- Desktop Chrome and Android WebViews also carry `safari` in the UA and must **not** match.
- The legacy Android stock browser says `safari` and never says `chrome`, which is the only agent that exercises the `android` clause.

`@moq/hang` needs the same predicate for its support matrix but cannot import net's internal `pickTransport` (`@moq/hang` depends on `@moq/net`), so `detectSafari` is its twin. The two share a user-agent corpus, so a drift between them fails a test in **both** packages. While doing that I noticed `js/hang` had no `test` script at all, so its suite had never run under `just js test`; it does now.

Both support matrices report `webtransport: "partial"` on Safari and Firefox to surface the degraded path. A consequence worth knowing: Safari has no `getStats()`, so there is no congestion-controller bandwidth estimate and the encoder does not adapt its bitrate downward (same as Firefox today).

**Public API changes:** `@moq/hang` gains `Util.Hacks.isSafari` and `detectSafari` (additive). `pickTransport` is internal (not re-exported). **No breaking changes.** Exposing the negotiated transport on `Established` is deliberately split into a separate PR so it can be judged on its own.

**Test plan:** `bun test js/net/src/connection/transport.test.ts js/hang/src/util/hacks.test.ts`: 9 tests over the shared UA corpus. `just js check` green. Verified on device: Safari connects and plays over the WebSocket fallback instead of dying on session setup.

### DIFF
--- a/doc/lib/js/env/web.md
+++ b/doc/lib/js/env/web.md
@@ -334,7 +334,9 @@ Requires modern browser features:
 **Experimental support:**
 
 - Firefox (behind flag)
-- Safari (future support planned)
+- Safari
+
+Safari always connects over the WebSocket fallback, because reading WebTransport datagrams terminates the session. It therefore has no congestion-controller bandwidth estimate, so the encoder does not adapt its bitrate downward. Set `maxBitrate` explicitly if you need a ceiling.
 
 ## Production Deployment
 
--- a/doc/lib/js/index.md
+++ b/doc/lib/js/index.md
@@ -165,7 +165,9 @@ Requires modern browser features:
 **Experimental support:**
 
 - Firefox (behind flag)
-- Safari (future support planned)
+- Safari
+
+Safari always connects over the WebSocket fallback, because reading WebTransport datagrams terminates the session.
 
 ## Framework Integration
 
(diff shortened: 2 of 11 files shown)


kixelated:
Nope we can't disable WebTransport because of this. Datagrams are optional (and I personally don't think they should exist).

There has to be some way to catch the error and ignore reading datagrams.

kixelated:
Closing in favor of #2198, which fixes the Safari datagram API mismatch directly instead of forcing Safari onto the WebSocket fallback. (Written by GPT-5)

PR #2185, opened 2026-07-12
fix(js): don't latch the connection off after a Safari pagehide

fperex:
Split out of #2163.

**A Safari tab could latch its connection off permanently.** `pagehide` set an internal `#suspended` flag, but only a `pageshow` with `persisted === true` cleared it. Safari fires `pagehide` when it *freezes* a page and can resume that page without ever firing a persisted `pageshow`, so `#suspended` stayed latched, `Reload` stayed disabled, and the element never reconnected. What you saw was a "Connected" pill sitting next to an "Offline" broadcast until you reloaded by hand.

The fix is to clear it on **any** `pageshow` and on the page becoming visible again, so every setter has a counterpart. The invariant worth stating: any listener that sets `#suspended` must have one that clears it, or the connection is disabled forever.

Closing eagerly on `pagehide` is still the right thing and stays: the relay unannounces (or stops forwarding to a dead viewer) immediately, rather than holding the route until the transport idle timeout, which would also block a same-name republish. Chrome and Firefox do not fire `pagehide` on a tab switch, so they are unaffected either way.

Also fixes **fullscreen on Safari**, which exposes neither `ShadowRoot.fullscreenElement` nor an un-retargeted `document.fullscreenElement`, so the player never knew it was fullscreen and laid the video out at its windowed size. Match `:fullscreen` (and the `-webkit-` prefixed form) against the player element itself, guarding the selector because `matches()` throws on a pseudo-class the engine cannot parse.

**Public API changes:** none.

**Test plan:** `just js check` green. Verified on device: switching away from and back to a Safari tab reconnects instead of latching off; fullscreen now fills the screen.

### DIFF
--- a/js/watch/src/element.ts
+++ b/js/watch/src/element.ts
@@ -86,14 +86,6 @@ export default class MoqWatch extends HTMLElement {
 	// Set when the element is connected to the DOM.
 	#enabled = new Signal(false);
 
-	// True while the page is torn down or frozen, so the connection closes immediately (freeing the
-	// relay's egress to this viewer) instead of lingering until the transport idle timeout. Set by
-	// `pagehide`, cleared by `pageshow` or the page becoming visible again.
-	#suspended = new Signal(false);
-
-	// The effective gate: connected to the DOM AND not page-suspended.
-	#active = new Signal(false);
-
 	// Stashed volume to restore on unmute.
 	#unmuteVolume = 0.5;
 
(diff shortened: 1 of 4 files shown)


js/watch/src/element.ts
```diff
@@ -126,6 +139,27 @@ export default class MoqWatch extends HTMLElement {
...
 		this.signals.cleanup(() => this.backend.close());
 
+		// Close the connection eagerly when the tab is torn down or frozen, so the relay stops forwarding
+		// media to this dead viewer immediately instead of at the transport idle timeout (and so the page
+		// stays bfcache-eligible). Gated on #enabled so the window listeners drop on DOM disconnect.
+		this.signals.run((effect) => {
+			if (!effect.get(this.#enabled)) return;
+
+			effect.event(window, "pagehide", () => {
+				this.connection.established.peek()?.close();
```
kixelated: wtf no

kixelated: Sorry I meant that this should be done within the `connection` class, not randomly here.

PR #2187, opened 2026-07-12
fix(watch): survive a video decoder error instead of going dark

fperex:
Split out of #2163.

**A video decoder error was fatal.** The track stayed subscribed but nothing decoded again, so the canvas simply froze until the viewer reloaded the page. Restart the decoder in place on a budget instead, and once that budget is spent report offline rather than pretending to play a frozen frame.

Three details that make the restart actually work:
- **Keyframe resync on `DataError`.** The usual cause is a group that begins mid-GOP after a reset, so replaying it into a fresh decoder just errors again. Wait for the next keyframe.
- **Config supersession.** A restart racing a codec/rendition switch would otherwise resurrect the *old* config and decode the new stream with it. The restart checks it is still the current config before applying.
- **A decode-queue cap.** A decoder that falls behind (a tab throttled in the background, a slow software path) otherwise accumulates an unbounded queue of frames it can never catch up on, and surfaces minutes late.

Also adds an **unsupported-codec indicator**: when the catalog advertises renditions but none of them decode in this browser, the player previously looked identical to "buffering forever", which is a miserable thing to debug.

**Public API changes:** `@moq/watch` gains `Video.Source.output.unsupported` (additive).

**Test plan:** `bun test js/watch/src/video/decoder.test.ts`: 6 tests pinning the config-supersession check. The restart budget and keyframe resync have no committed tests (bun ships no WebCodecs `VideoDecoder`, so they need a real browser); they were exercised manually. `just js check` green. Verified in the demo by forcing a decoder error mid-playback: the canvas recovers instead of freezing.

### DIFF
--- a/doc/lib/js/@moq/watch.md
+++ b/doc/lib/js/@moq/watch.md
@@ -246,7 +246,7 @@ The component exposes everything via its `broadcast` property
 
 ## UI Overlay
 
-Import `@moq/watch/ui` for a Web Component overlay with buffering indicator, stats panel, and playback controls:
+Import `@moq/watch/ui` for a Web Component overlay with buffering indicator, an unsupported-codec indicator, stats panel, and playback controls:
 
 ```html
 <script type="module">
--- a/js/watch/README.md
+++ b/js/watch/README.md
@@ -109,7 +109,7 @@ watch.video.media.subscribe((stream) => {
 
 ## UI Web Component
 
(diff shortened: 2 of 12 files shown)


kixelated:
I don't agree.

Surfacing errors seems good, but blindly retrying because it might work this time is dumb. If you can give me a concrete situation where retrying might work?

PR #2188, opened 2026-07-12
fix(watch): recover from a reset catalog and a same-name republish

fperex:
Split out of #2163.

> **Stacked on #2182** (it uses `isStreamAbort` to tell a stream reset apart from a real fault). The first commit belongs to that PR; **review the second commit only**. The diff collapses once #2182 merges and I rebase.

Two ways a tile could go permanently offline while the relay was perfectly healthy:

**1. A catalog subscription that ended was terminal.** `#runCatalog` subscribed once and wrote `status = "offline"` whenever its fetch loop ended. But the effect's tracked signals (`enabled`, `name`, `catalogFormat`, `active`) do not change when only the *catalog stream* resets, so the effect never re-ran and nothing ever resubscribed. Any transient reset (a slow-consumer drop, a relay bounce, a publisher handover) stranded the tile offline until a page reload.

It now reopens on a bounded backoff, but **only** when the loop ended in a stream reset. A clean end (the publisher genuinely stopped) or a coded fault still reports offline immediately, which is why this depends on the `isStreamAbort` classification rather than retrying on any error.

**2. A republish under the same name was invisible.** The announce `Set` only tracks presence, so when a publisher's unannounce and the new instance's announce coalesce, the set never observably flips and the watcher stays bound to the dead route. `Reload` now tracks a per-path announce **generation** that bumps on every (re-)announce, so a same-name republish is a real change the watcher re-consumes.

The generation map deliberately never deletes on unannounce: resetting a path to 1 would make a coalesced republish look unchanged (1 -> 1) and reintroduce the bug. It is reset on reconnect, so it grows by one entry per distinct path ever announced, which is an accepted tradeoff.

**Public API changes:** `@moq/net` gains `Reload.announcedGenerations`; `@moq/watch` gains an optional `announcedGenerations` in the `Broadcast` options bag (both additive).

**Test plan:** `just js check` green. Verified in the demo: killing and immediately restarting a publisher under the same name re-attaches the viewer, and bouncing the relay no longer strands a tile offline.

### DIFF
--- a/js/net/src/connection/reload.ts
+++ b/js/net/src/connection/reload.ts
@@ -60,6 +60,14 @@ export class Reload {
 	/** The set of broadcast paths currently announced by the server, updated reactively. */
 	readonly announced: Getter<Set<Path.Valid>> = this.#announced;
 
+	// Per-path announce generation, bumped every time a path is (re-)announced. Lets a consumer notice
+	// that the SAME name was re-announced by a NEW publisher instance even when the unannounce+announce
+	// coalesce so the presence Set never observably flips. Monotonic per connection (reset on reconnect).
+	#announcedGen = new Signal<Map<Path.Valid, number>>(new Map());
+
+	/** Per-path announce generation; bumps on every (re-)announce. Pairs with {@link Reload.announced}. */
+	readonly announcedGenerations: Getter<ReadonlyMap<Path.Valid, number>> = this.#announcedGen;
+
 	/** WebTransport options applied to each connection attempt (not reactive). */
 	webtransport?: WebTransportProps;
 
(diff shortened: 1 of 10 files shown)


js/net/src/connection/reload.ts
```diff
@@ -60,6 +60,14 @@ export class Reload {
 	/** The set of broadcast paths currently announced by the server, updated reactively. */
 	readonly announced: Getter<Set<Path.Valid>> = this.#announced;
 
+	// Per-path announce generation, bumped every time a path is (re-)announced. Lets a consumer notice
+	// that the SAME name was re-announced by a NEW publisher instance even when the unannounce+announce
+	// coalesce so the presence Set never observably flips. Monotonic per connection (reset on reconnect).
+	#announcedGen = new Signal<Map<Path.Valid, number>>(new Map());
+
+	/** Per-path announce generation; bumps on every (re-)announce. Pairs with {@link Reload.announced}. */
+	readonly announcedGenerations: Getter<ReadonlyMap<Path.Valid, number>> = this.#announcedGen;
```
kixelated, line 69: Real bug, wrong fix. `Reload.announced` should be like `Connection.announced`

fperex:
PR Closed in favor of: https://github.com/moq-dev/moq/pull/2199

PR #2189, opened 2026-07-12
fix(watch): survive an audio decoder error, resample, and re-anchor on resubscribe

fperex:
Split out of #2163.

> **Stacked on #2182** (the hang container consumer uses `isStreamAbort`). The first commit belongs to that PR; **review the second commit only**. The diff collapses once #2182 merges and I rebase.

Audio had the same fatal-decoder problem as video (#2187): an `AudioDecoder` that errored never decoded again, so the stream went silent until a reload. It restarts in place on a budget now, with the same config-supersession check.

Plus three fixes for crackle and drift, all of which I hit on Safari but which are not Safari-specific:

- **Resample when the rates differ.** If the catalog rate and the `AudioContext` rate disagree, samples were written into a ring built at the wrong rate, which garbles playback outright.
- **Snap timestamps onto the frame cadence.** A jittery per-frame timestamp makes the timestamp-indexed ring zero-fill or overwrite a sample every single frame, which is audible as a constant crackle.
- **Re-anchor the discontinuity counter on resubscribe.** `Audio.Decoder` is built once and its `#discontinuity` outlives the subscription, but the container `Consumer`'s rewind counter is per-consumer and restarts at zero. So after any rewind had bumped the decoder's count, the next resubscribe delivered 0, which *looked* like a rewind and reset the `Sync` reference shared with video, desyncing audio from video for no reason.

Also resumes the `AudioContext` on a user gesture rather than assuming it starts in `running`.

**Public API changes:** `@moq/watch` gains `AudioBuffer.wait(timestamp, signal?)` (additive).

**Test plan:** `bun test js/watch/src/audio js/hang/src/container`: includes a test pinning that a fresh/resubscribed `Consumer` starts its rewind counter at zero (the bug above), plus the timestamp-snapping math. `just js check` green.

### DIFF
--- a/js/hang/package.json
+++ b/js/hang/package.json
@@ -15,6 +15,7 @@
 	"scripts": {
 		"build": "rimraf dist && tsc -b && bun ../common/package.ts",
 		"check": "tsc --noEmit",
+		"test": "bun test --only-failures",
 		"release": "bun ../common/release.ts"
 	},
 	"dependencies": {
--- a/js/hang/src/container/consumer.test.ts
+++ b/js/hang/src/container/consumer.test.ts
@@ -585,3 +585,25 @@ test("Consumer does not duration-skip when the gap is not covered", async () => 
 	expect(frames.map((f) => f.group)).toEqual([0, 0, 1]);
 	consumer.close();
 });
+
+// A fresh Consumer always starts at zero, on both a fresh track and a resubscribe of a track that has
+// already been read. Downstream decoders that outlive their subscription rely on this to re-anchor their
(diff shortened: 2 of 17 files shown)


kixelated:
> dio had the same fatal-decoder problem as video (https://github.com/moq-dev/moq/pull/2187): an AudioDecoder that errored never decoded again, so the stream went silent until a reload. It restarts in place on a budget now, with the same config-supersession check.

How did this happen? Blindly retrying is the wrong fix, please try to reproduce the original bug instead.

PR #2190, opened 2026-07-12
feat(publish): Safari hardware encode, worker capture, and a stable video catalog

fperex:
Split out of #2163.

> **Stacked on #2184** (it consumes `Util.Hacks.isSafari` / `safariWorkerCapture`). The first commit belongs to that PR; **review the second commit only**. The diff collapses once #2184 merges and I rebase.

Publishing on Safari was software-only, and the capture path stalled whenever the tab was hidden.

- **Hardware encode.** On Apple Silicon, VideoToolbox only hardware-encodes H.264 and HEVC. Safari nonetheless reports the *others* as `prefer-hardware` supported, so the support matrix was keying off an echoed hint that lies (they actually run on software libvpx). Key off the codec instead, and prefer the codecs that genuinely have a hardware path.
- **Worker capture.** WebKit 18+ exposes `MediaStreamTrackProcessor` in a worker (plus transferable `MediaStreamTrack`), so capture runs there instead of the `requestAnimationFrame` fallback, which pauses whenever the tab is hidden. The gate reads the iOS OS version when the Safari `Version/` token is absent, because iOS Chrome and Firefox are WebKit too and omit it, and the iOS major tracks the WebKit major. Anything older or undetectable keeps the rAF path.
- **Camera and microphone re-acquire when their track ends** (all browsers), and a backgrounded Safari tab no longer comes back with the camera stuck muted.
- **The video catalog no longer flaps.** It folds in the encoder-produced codec string and the hvcC/av1C description after the first keyframe, tagged against the requested codec and dimensions, so bandwidth-driven bitrate churn (which fires ~10x a second) can never rewrite the description underneath watchers.
- **The stats tab graphs a measured upload bitrate** from encoded bytes rather than a transport estimate, which Safari does not have at all (no `getStats()`).

**Public API changes:** `@moq/publish` gains `Video.Encoder.bytesEncoded`; `@moq/hang` gains `Util.Hacks.safariWorkerCapture` (both additive).

**Test plan:** `bun test js/publish/src/video js/hang/src/util`: 22 tests (codec ordering, catalog description folding against bitrate churn, worker-capture UA gate). `just js check` green. Verified on device: Safari publishes with hardware H.264 and keeps capturing in a background tab.

### DIFF
--- a/doc/lib/js/env/web.md
+++ b/doc/lib/js/env/web.md
@@ -334,7 +334,9 @@ Requires modern browser features:
 **Experimental support:**
 
 - Firefox (behind flag)
-- Safari (future support planned)
+- Safari
+
+Safari always connects over the WebSocket fallback, because reading WebTransport datagrams terminates the session. It therefore has no congestion-controller bandwidth estimate, so the encoder does not adapt its bitrate downward. Set `maxBitrate` explicitly if you need a ceiling.
 
 ## Production Deployment
 
--- a/doc/lib/js/env/web.md
+++ b/doc/lib/js/env/web.md
@@ -334,7 +334,7 @@ Requires modern browser features:
 **Experimental support:**
 
 - Firefox (behind flag)
-- Safari
+- Safari 18+ (publishing: hardware H.264/HEVC encode, Worker capture; older Safari falls back to a degraded capture that pauses while the tab is hidden). Opus audio uses the native WebCodecs codec on Safari 26+ and a WebAssembly polyfill on Safari 16.4 through 18.7, which ship no WebCodecs audio API.
 
 Safari always connects over the WebSocket fallback, because reading WebTransport datagrams terminates the session. It therefore has no congestion-controller bandwidth estimate, so the encoder does not adapt its bitrate downward. Set `maxBitrate` explicitly if you need a ceiling.
 
--- a/doc/lib/js/index.md
+++ b/doc/lib/js/index.md
@@ -165,7 +165,7 @@ Requires modern browser features:
 **Experimental support:**
 
 - Firefox (behind flag)
-- Safari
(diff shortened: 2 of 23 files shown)


fperex:
Closing in favour of focused PRs, each rebased on current `dev` and verified on device rather than by inspection. Splitting this up turned one of the five features into a regression, so it is worth spelling out what happened to each.

**Replaced by:**

| feature | PR |
|---|---|
| capture polyfill drops a third of the camera's frames | #2209 (fixes #2208) |
| Safari publishes software VP9 while its hardware H.264 sits idle | #2211 |
| the bitrate graph is empty on Safari | #2213 |

**Dropped: the video catalog `description` change.** It is a regression and I am not re-proposing it.

The idea was to fold the encoder-produced `decoderConfig` (the hvcC/av1C parameter sets) into the catalog, on the premise that a watcher cannot initialise an HEVC decoder without it. That premise is false here. `#bestCodec` configures HEVC with `hevc: { format: "annexb" }`, Safari honours it, and the parameter sets arrive in band, so watchers already decode fine. Safari *also* hands back a 104 byte hvcC in the keyframe metadata, which is what made the change look correct.

Publishing that hvcC makes the watcher configure its decoder for a length-prefixed hvc1 bitstream while the wire actually carries annexb. Safari publishing HEVC, Chrome watching, counting non-black pixels on the watcher's canvas:

| catalog | watcher canvas |
|---|---|
| `dev` today (no description) | 880251 / 921600 pixels lit |
| with the `description` folded in | **0 / 921600** (black, reproduced twice) |
| reverted | 694315 / 921600 pixels lit |

So the original PR would have broken HEVC playback for every watcher. If a `description` is ever wanted, it has to come with the encoder actually producing a length-prefixed bitstream, not alongside an annexb one.

**Still to come, once the above land** (they conflict with #2209 / #2211, so I would rather not stack them):

- Worker capture on WebKit 18+. Safari exposes `MediaStreamTrackProcessor` only inside a Worker, and the main-thread fallback freezes whenever the tab is hidden, because `requestAnimationFrame` is suspended. Rewrites the same file as #2209.
- Camera and microphone re-acquire when their track ends. Needs the `detectSafari` helper from #2211.

PR #2191, opened 2026-07-12
fix(publish): advertise 48 kHz Opus and resample a non-canonical capture rate

fperex:
Split out of #2163.

> **Stacked on #2186** (the resampling lives inside the encoder pipeline that PR creates). The first commit belongs to that PR; **review the second commit only**. The diff collapses once #2186 merges and I rebase.

**The Opus catalog advertised the wrong sample rate.** Opus decoders always emit 48 kHz PCM (RFC 6716) whatever rate the encoder was fed, but we advertised the *capture* rate. A watcher then built its ring buffer at, say, 44.1 kHz and received 48 kHz-dense samples, which garbles playback. The catalog now always advertises 48000 (what decoders actually output) and the capture rate stays private to the encoder.

That exposed the other half of the problem: **Safari ignores the `AudioContext` sample-rate request** and captures at the hardware rate anyway, so the encoder was handed 44.1 kHz audio that native Opus misencodes into scratchy noise. Capture PCM is now resampled to 48 kHz before encoding when the rates differ. This is a no-op on Chrome and Firefox, which honor the request. Timestamps on the output clock are derived from an absolute sample counter (not an incremented running total) so rounding can never accumulate, and re-anchor to the live capture clock if it ever drifts more than a frame, so audio cannot silently lag real time forever.

Also **snaps the container timestamp onto the nominal frame cadence.** Safari stamps each output frame with the timestamp of the *input* chunk holding its start, quantized to the 128-sample capture grid, so a 20 ms Opus frame alternates 18667/21333 µs. That jitter makes the watcher's timestamp-indexed ring zero-fill or overwrite a sample every frame and crackle. Chrome and Firefox already emit an exact cadence, so there it is an identity rewrite, and a real gap (mute, DTX silence, suspend) re-anchors rather than being smeared over.

**Public API changes:** none.

**Test plan:** `bun test js/publish/src/audio`: 8 tests over the cadence-snapping and resampler math (~900 assertions, including that the snap is an identity on an exact cadence and re-anchors across a real gap). `just js check` green. Verified on device: Safari audio is clean instead of scratchy.

### DIFF
--- a/doc/lib/js/@moq/publish.md
+++ b/doc/lib/js/@moq/publish.md
@@ -67,7 +67,7 @@ a real bundler (the examples below).
 - `url` (required) - Relay server URL
 - `name` (required) - Broadcast name
 - `source` - Input to capture: `"camera"`, `"screen"`, or `"file"`
-- `muted` - Mute audio capture (boolean)
+- `muted` - Stop capturing audio and publish silence in its place (a microphone is released and the OS recording indicator turns off; screen or file audio is detached); the audio track stays in the catalog so viewers don't rebuild (boolean)
 - `invisible` - Disable video capture (boolean)
 - `simulcast` - Also publish a lower-resolution `video/sd` rendition (a fraction of the source resolution) alongside `video/hd` (boolean)
 - `preview` - What the preview renders: `"source"` (default), `"encoded"`, or `"none"` to disable it (see [Preview element](#preview-element))
--- a/doc/lib/js/env/web.md
+++ b/doc/lib/js/env/web.md
@@ -54,7 +54,7 @@ Publish camera/microphone or screen as a MoQ broadcast.
 - `url` (required) - Relay server URL
 - `name` (required) - Broadcast name
 - `source` - "camera", "screen", or "file"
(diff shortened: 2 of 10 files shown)


kixelated:
> Safari ignores the AudioContext sample-rate request and captures at the hardware rate anyway, so the encoder was handed 44.1 kHz audio that native Opus misencodes into scratchy noise.

That's unfortunate. We're relying on AudioContext to do fast resampling for us instead of doing it in user-space. I would double check though because we heavily rely on this behavior for decoding, and maybe the LLM is hallucinating.

If there truely is no other option, we can add support for audio resampling, but the code you have here is really crude/slow and it should be done properly. ex. via a proper resampler class that also applies to encoding.

fperex:
You were right to push back, and this PR is built on a false premise, so I am closing it.

I re-tested on a real Safari (26.5, macOS 26.5.1): Safari does honor the AudioContext sample-rate request, including for mic capture. Requesting 8000 / 16000 / 22050 / 44100 / 48000 / 96000 while the hardware ran at 48 kHz returned exactly the requested rate every time.

PR #2192, opened 2026-07-12
feat(net)!: expose the negotiated transport on Established

fperex:
Split out of #2163. This PR exposes the transport that an established session actually negotiated, and uses it in the demo UI instead of hardcoding the network badge.

`Established` gains a required readonly `transport` (`"webtransport" | "websocket"`). Both lite and IETF connection implementations set it from the live session. qmux implements WebTransport on top of WebSocket, so detection uses identity against the qmux `Session` rather than duck typing.

The watch and publish demos now render that negotiated transport in their Network badges. The Safari/WebTransport routing workaround and support-matrix changes are intentionally not part of this PR.

**BREAKING:** `transport` is a required member of the public `Established` interface, so external implementers must add it. Consumers only reading `Established` get an additive field.

**Test plan:**
- `bun run --filter='@moq/net' check`
- `bun run --filter='@moq/demo' check`
- `bun biome check js/net/src/connection/established.ts js/net/src/connection/index.ts js/net/src/connection/transport.ts js/net/src/ietf/connection.ts js/net/src/lite/connection.ts demo/web/src/index.ts demo/web/src/publish.html demo/web/src/publish.ts demo/web/src/watch.html`
- `bun run --filter='@moq/net' test`

(Written by GPT-5)


doc/lib/js/env/web.md
```diff
@@ -334,7 +334,9 @@ Requires modern browser features:
 **Experimental support:**
 
 - Firefox (behind flag)
-- Safari (future support planned)
+- Safari
+
+Safari always connects over the WebSocket fallback, because reading WebTransport datagrams terminates the session. It therefore has no congestion-controller bandwidth estimate, so the encoder does not adapt its bitrate downward. Set `maxBitrate` explicitly if you need a ceiling.
```
kixelated: Remove all of this unrelated junk.

PR #2199, opened 2026-07-12
fix(net,watch): recover the viewer from a same-name republish

kixelated:
An alternative to #2188's announce/republish fix.

> Scope: this is the announce/republish half only (#2188 bug #2). The orthogonal catalog-reset retry (bug #1: a slow-consumer reset stranding a tile while it's still announced) is **not** included here.

## The bug

A tile could go permanently offline while the relay was healthy: when a publisher's unannounce and a new instance's announce under the same name coalesce, the presence set never observably flips, so the watcher stays bound to the dead route until a page reload.

The root cause is gating on announcement **membership** (a `Set`) rather than the announcement **stream**. A republish arrives as a redundant `active: true` (a lite-06 `RESTART`, or an unannounce+announce), which `Set.has(name)` collapses to "no change". The stream still carries a distinct update.

## The fix

**`@moq/watch`**
- `Broadcast` drives its subscription off a **name-scoped** announcement stream and re-consumes on every (re-)announce, so a same-name republish re-attaches to the new instance. Scoping to the name also means unrelated broadcasts never retrigger it (what the old `#announcedNow` membership flag guarded against). A live subscription survives a redundant re-announce; only a dead one is replaced.
- Dropped the `announced` prop; `Broadcast` derives announcements from its connection. The connection-scoped stream for cross-broadcast (`broadcast: ../`) refs is opened lazily and drained into a local set.

**`@moq/net`**
- `Reload.announced` changes from a bare `Getter<Set>` property to an `announced(prefix?)` **method** matching `Established.announced`: the same `Announce.Consumer` stream, but everything active is retracted (an `active: false` update) on disconnect and re-announced on reconnect, so a consumer draining `next()` never clings to a dead route across a reconnect.
- `Announce.Consumer` stays a plain stream. Callers that want a set of active paths maintain their own by draining `next()` (there's no `active` accessor to hold wrong).

**demo** discovery uses `Reload.announced(prefix)` so its set self-heals on reconnect without wiring up `established` by hand.

## Public API changes (breaking → targets `dev`)

- `@moq/net`: `Reload.announced` **changed** from a `Getter<Set<Path>>` property to an `announced(prefix?): Announce.Consumer` method.
- `@moq/watch`: `Broadcast` constructor **drops** the `announced` option.

## Test plan

- [x] `bun run --filter='*' check` (all packages) + `bun biome check` green.
- [x] New `js/net/src/announced.test.ts` covers the stream carrying a re-announce as its own distinct update (the property the fix relies on).
- [x] `bun test` green (net 239, watch 78).
- [ ] Manual: kill and immediately restart a publisher under the same name; the viewer re-attaches without a reload.

(Written by Claude Opus 4.8)

### DIFF
--- a/demo/web/src/index.ts
+++ b/demo/web/src/index.ts
@@ -164,17 +164,14 @@ function createTile(name: string): WatchTile {
 // Broadcast discovery
 // ---------------------------------------------------------------------------
 //
-// Subscribe to announcements under the prefix and keep a live set of active
-// broadcasts. `announced.next()` drains a queue, so we track membership
-// ourselves: active=true adds the path, active=false removes it.
+// Subscribe to announcements under the prefix and keep a live set of active broadcasts.
+// `announced.next()` drains the update stream, so we track membership ourselves: active=true adds the
+// path, active=false removes it. `Reload.announced()` spans reconnects (it retracts everything on
+// disconnect and re-announces on reconnect), so the set self-heals without any extra wiring here.
 const discovery = new Signals.Effect();
 discovery.run((effect) => {
-	const conn = effect.get(connection.established);
-	broadcasts.set([]);
-	if (!conn) return;
-
 	const prefix = prefixPath(effect.get(prefixInput));
-	const announced = conn.announced(prefix);
+	const announced = connection.announced(prefix);
 	effect.cleanup(() => announced.close());
 
 	const live = new Set<string>();
--- a/js/net/src/announced.test.ts
+++ b/js/net/src/announced.test.ts
@@ -1,0 +1,46 @@
+import { expect, test } from "bun:test";
+import * as Announce from "./announced.ts";
+import * as Path from "./path.ts";
+
+const p = (s: string) => Path.from(s);
+
+test("next streams every appended event in order", async () => {
+	const producer = new Announce.Producer();
+	const consumer = producer.consume();
+
+	producer.append({ path: p("a"), active: true });
+	producer.append({ path: p("a"), active: false });
+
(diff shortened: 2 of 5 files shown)


js/net/src/announced.ts
```diff
@@ -22,6 +17,16 @@ export interface Event {
...
+
+	// The set of currently-active paths, kept in sync with every appended event. This is the
+	// snapshot view; the queue is the delta view. Maintained even when the queue is idle.
+	active = new Signal<Set<Path.Valid>>(new Set());
+
+	// Flipped true the first time a consumer reads via `next()`. Until then `append` only updates
+	// `active` and skips the queue, so a consumer that reads only `active` never accumulates a
+	// backlog of undrained events. A re-announce (RESTART) still notifies `active` subscribers,
+	// since `mutate` always fires even when membership is unchanged.
+	streaming = false;
```
kixelated: I think this surfaces a bigger footgun.

Maybe get rid of the active Set and let the application maintain it? Who relies on it?

kixelated: Agreed, dropped it. `Announce.Consumer` is back to a plain stream (no `active`, no streaming/seeding), and each caller maintains its own set:

- `Reload.announced` keeps a local `Set` in the pump purely to know what to retract on disconnect.
- `@moq/watch`'s relative-ref gate drains into its own set.
- the demo drains `next()` into its own set (as before).

`js/net/src/announced.ts` is now untouched by this PR. The main-broadcast fix never needed a set anyway; it drives straight off `next()`.

(Written by Claude Opus 4.8)

PR #2201, opened 2026-07-12
feat(moq-net): advertise session Role in the moq-lite-05 SETUP

kixelated:
Targets **`dev`** per the branch-targeting rule (a moq-net wire change + draft bump).

## Problem

The relay authenticates URL-less stream clients (the moq.pro RTMP/SRT/WHIP/WHEP gateways, which reach the relay over its internal UDS socket) **post-handshake**, and accepts a connection whenever its token grants *either* publish or subscribe. A connection is bidirectional, so the relay can't tell the connection's *intended* direction: a **subscribe-only token used to publish** (or vice versa) is accepted, the relay wires the wrong origin, and the media is silently dropped. The publisher sees a healthy connection streaming to nowhere.

Reported downstream as moq-dev/moq.pro#338.

## Change

**1. A `Role` in the SETUP.** Add a `Role` (`Publisher | Subscriber | Both`) to the lite-05 SETUP, a new optional parameter (`0x3`, varint) beside Probe (`0x1`) and Path (`0x2`). The client derives it from which origins it wired up, so it can't drift from what the session does. `Both` is the absence of the parameter, and an unrecognized value decodes to `Both` (mirrors the Probe fallback), so older peers and non-lite-05 transports are unaffected — a backward-compatible extension per the SETUP "ignore unknown params" rule.

**2. The relay enforces direction.** It rejects a directional client whose token lacks that direction's scope during the handshake (close, instead of accept-then-silently-drop), and **prunes the unused half**: a `Publisher` keeps only consume-from-client, a `Subscriber` only publish-to-client, so a broad token used by a publish-only ingest no longer attaches a cluster subscriber it never pulls from. The role only ever narrows; the token scope is still fully enforced.

**3. Uniform Request across transports (moq-native).** Previously the accept path read the SETUP up front only for stream bindings; QUIC/WebTransport deferred the whole MoQ handshake to `ok()`, so `path()` depended on the transport and `role()` was stream-only — a footgun that left #338's protection unable to cover browser/QUIC publishers. Now every backend drives the transport connect **and** the SETUP before returning a `Request`, so `path()`, `role()`, `url()`, and `peer_identity()` are populated consistently everywhere, and the direction reject/prune cover every transport. `RequestKind` collapses to a `moq_net::Request<S>` per session type; server.rs is ~84 lines shorter.

The one trade-off: a URL-bearing transport must be accepted (HTTP 200 / QUIC handshake) before its SETUP arrives, so `close()` can no longer answer the CONNECT with an HTTP status — rejection is a uniform MoQ session close, as the stream bindings already did. An unauthenticated WebTransport client establishes the session then gets closed rather than a 401 on CONNECT (WebTransport's CONNECT-rejection errors are opaque anyway).

Also updates `drafts/draft-lcurley-moq-lite.md` (the Role Parameter, validated with `kramdown-rfc --v3`) and `CLAUDE.md` (makes "wire change → update the matching draft" an explicit MUST).

## Scope

Fixes the wrong-direction half of moq.pro#338, now across every transport. The other half (WHIP/WHEP answering `201` before the relay's verdict) is a follow-up.

## Tests

- `moq-net`: wire round-trip per role, `Both`-omits-the-parameter, unknown value → `Both`, role surfaced on the server `Request`.
- `moq-native` backend (raw QUIC, WebTransport, mTLS, iroh): exercise the unified accept path and assert the client's role reaches the server over every transport; `peer_identity()` still captured before `ok()`.
- `moq-relay` smoke: publisher-role rejected / subscriber-role accepted on a subscribe-only relay; the internal round-trip + coexist tests exercise the direction prune.


kixelated:
Rebased onto latest `dev` (resolved a `CLAUDE.md` conflict) and pushed fixes:

- **CI (`Check` failure):** removed the now-dead pre-accept `close`/`reject` methods on the Quinn/Quiche backends (unused after the accept-path refactor; caught by `-D warnings` under `--all-features`). Verified locally with `cargo clippy --all-features --tests -D warnings`.
- **Review (symmetric coverage):** added `spawn_publish_only_relay` + `publish_only_public_rejects_subscriber_role`, mirroring the subscribe-only tests so both branches of the role gate are exercised.
- **Review (draft changelog):** added a `Role` parameter bullet to the moq-lite-05 changelog appendix, alongside Probe and Path (`kramdown-rfc --v3` clean).

Sourcery was rate-limited (no findings). CodeRabbit's other notes were the auto-generated walkthrough/summary.

PR #2209, opened 2026-07-13
fix(publish): capture the full frame rate in the capture polyfill

### LINKED ISSUE #2208: Safari: MediaStreamTrackProcessor polyfill captures at 20 fps instead of 30
## Summary

On Safari the publisher falls back to the `MediaStreamTrackProcessor` polyfill, and the polyfill's frame pacing lands on **20 fps when the camera is delivering 30 fps**. The capture-rate graph in the publish demo shows a steady 20 fps with a sawtooth, and roughly a third of the camera's frames are simply never captured or encoded.

Safari 26.5 (`AppleWebKit/605.1.15 Version/26.5`), macOS, 60 Hz display, `js/publish` demo (`publish.html`), default camera constraints (no `frameRate` constraint, so the FaceTime camera reports `frameRate: 30`).

## Root cause

`TrackProcessor` in `js/publish/src/video/polyfill.ts` paces frames by polling `requestAnimationFrame` and comparing elapsed wall-clock time against the frame period:

```ts
const frameRate = settings.frameRate ?? 30;
...
async pull(controller) {
    while (true) {
        const now = Time.Milli.now();
        if (Time.Milli.sub(now, last) < ((1000 / frameRate) as Time.Milli)) {
            await new Promise((r) => requestAnimationFrame(r));
            continue;
        }
        last = now;
        controller.enqueue(new VideoFrame(video, { timestamp: Time.Micro.fromMilli(last) }));
    }
}
```

Two things combine:

1. `Time.Milli.now()` is `performance.now()`, and **Safari clamps `performance.now()` to 1 ms resolution** (the Spectre timer mitigation). So `sub(now, last)` is effectively a whole number of milliseconds.
2. `last` is set to the *observed wake time*, not to the ideal deadline, so the phase error does not carry over between frames.

At 60 Hz, two rAF ticks are 33.33 ms. The frame period for 30 fps is `1000 / 30 = 33.333` ms. Safari's clamped clock reports that elapsed as **33**, and `33 < 33.333` is true, so the loop declines to emit and waits for a third tick. Three ticks is 50 ms, which passes the check. Steady state is one frame per 3 rAF ticks, i.e. 20 fps. The occasional 34 ms gap (when the rounding happens to land the other way) is what produces the sawtooth in the graph rather than a flat line.

The bug is not Safari-specific in principle. Any browser on the polyfill path with a 1 ms clock clamp hits it, and Firefox clamps `performance.now()` to 1 ms by default too. I have only verified Safari. Chrome is unaffected because it has native `MediaStreamTrackProcessor` and never reaches this code.

## Reproduction

Standalone repro of just the pacing loop, no camera or moq involved. Served over http and opened in Safari 26.5:

```js
async function pace(frameRate, seconds) {
    let last = performance.now();
    const start = last;
    let frames = 0;
    while (performance.now() - start < seconds * 1000) {
        const now = performance.now();
        if (now - last < 1000 / frameRate) {
            await new Promise((r) => requestAnimationFrame(r));
            continue;
        }
        last = now;
        frames++;
    }
    return frames / seconds;
}
```

Result:

```
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/26.5 Safari/605.1.15
performance.now() min delta: 1 ms          <- clamped, as expected
display refresh: 61.3 Hz
requested 30 fps -> actual 21.0 fps  (gaps ms: 50, 34, 50, 50, 50, 49, 51, 49, 34, 50)
requested 60 fps -> actual 38.7 fps  (gaps ms: 32, 18, 33, 17, 33, 17, 33, 17, 32, 18)
```

The 30 fps case reproduces the demo exactly. The 60 fps case is the same failure mode one tick down: it alternates between 1 and 2 ticks instead of holding at 1.

fperex:
Fixes #2208. Split out of #2190, which I am closing in favour of focused PRs.

## Root cause

The `MediaStreamTrackProcessor` polyfill in `js/publish/src/video/polyfill.ts` paces frames by comparing `performance.now()` against the time it last woke:

```ts
if (Time.Milli.sub(now, last) < Time.Milli(1000 / frameRate)) {
    await new Promise((r) => requestAnimationFrame(r));
    continue;
}
last = now;
```

Safari and Firefox both clamp `performance.now()` to 1ms (the Spectre timer mitigation). On a 60Hz display two animation frames are 33.33ms, the clamped clock reports that elapsed as **33**, and `33 < 1000/30` is true, so the loop declines to emit and waits for a third frame. Three frames is 50ms, which passes. Steady state is one captured frame per three animation frames: **20fps out of a 30fps camera**, and a third of the camera's frames are never captured or encoded.

`last` is assigned the observed wake time rather than the deadline, so the phase error never carries over and the loop cannot self-correct.

Chrome is unaffected: it has a native main-thread `MediaStreamTrackProcessor` and never reaches this code.

## The fix

Pace against a deadline rather than the wake time:

- Advance the deadline by exactly one period, so the rounding error cannot accumulate.
- Allow the clock's own 1ms resolution as slack, so a frame that lands 0.33ms early is not deferred by a whole animation frame.
- Resync instead of releasing a backlog after a stall, since a hidden tab suspends `requestAnimationFrame` entirely.

This is arithmetic inside the existing loop. It is not gated on an engine, adds no API, and does not change the approach: `requestAnimationFrame` still drives the loop, exactly as before. Firefox has the identical clock clamp and is fixed by the same change, with its code path otherwise untouched.

The pacing is extracted into a small `Pacer` so the clock behavior has a unit test. It is not re-exported from `./index`, so the package's public API is unchanged.

## Teardown

The issue also noted that `pull()` looped forever with no `cancel()`, so an ended track left the `<video>` element and its `MediaStream` decoding until GC. The stream now closes when the track ends and releases the element on cancel. The polyfill this is based on ([Jan-Ivar's](https://jan-ivar.github.io/polyfills/mediastreamtrackprocessor.js)) already did the former; our copy had dropped it.

## Test plan

`bun test js/publish/src/video/polyfill.test.ts`: 3 tests driving the pacer with a simulated 1ms-clamped clock on a 60Hz display. Under that clock the previous logic yields 20.0 / 20.0 / 40.0 fps for a requested 30 / 24 / 60, so the tests fail without this change. `just js check` and `just js test` are green.

Verified on device, Safari 26.5 (`AppleWebKit/605.1.15 Version/26.5`), macOS, 60Hz display, FaceTime camera reporting 30fps, using the `demo/web` publish page and its capture-rate graph:

| | capture rate |
|---|---|
| before | **21.9 fps** |
| after | **29.9 fps** |

The standalone harness from #2208, run in the same Safari:

```
performance.now() min delta: 1 ms
display refresh: 60.1 Hz

before:  requested 30 fps -> 20.5 fps   (gaps ms: 50, 49, 34, 34, 50, 50, 50, 49, 34, 50)
         requested 60 fps -> 40.0 fps   (gaps ms: 17, 33, 17, 33, 17, 33, 17, 33, 17, 33)

after:   requested 30 fps -> 30.0 fps   (gaps ms: 33, 34, 33, 33, 34, 33, 33, 33, 51, 16)
         requested 60 fps -> 59.8 fps   (gaps ms: 16, 17, 17, 16, 17, 17, 17, 16, 17, 16)
```

### DIFF
--- a/js/publish/src/video/polyfill.test.ts
+++ b/js/publish/src/video/polyfill.test.ts
@@ -1,0 +1,44 @@
+import { describe, expect, test } from "bun:test";
+import { Time } from "@moq/net";
+import { Pacer } from "./polyfill";
+
+// Wake once per display refresh, with the clock clamped to 1ms like Safari and Firefox.
+function capture(frameRate: number, seconds: number, displayHz = 60): number {
+	const pacer = new Pacer(frameRate, Time.Milli(0));
+
+	let frames = 0;
+	const ticks = Math.round(seconds * displayHz);
+	for (let i = 1; i <= ticks; i++) {
+		const now = Time.Milli(Math.floor((i * 1000) / displayHz));
+		if (pacer.due(now)) frames++;
+	}
+
+	return frames / seconds;
+}
+
+describe("Pacer", () => {
+	test("captures the source frame rate on a clamped clock", () => {
+		// A 60Hz display and a 30fps camera is the case from #2208: the frame period is 33.333ms, the
+		// clamped clock reports the two elapsed animation frames as 33, and comparing against the wake
+		// time then waits for a third frame, capturing every 50ms instead of every 33ms.
+		expect(capture(30, 10)).toBeCloseTo(30, 0);
+		expect(capture(24, 10)).toBeCloseTo(24, 0);
+		expect(capture(60, 10)).toBeCloseTo(60, 0);
+	});
+
+	test("never captures faster than the display", () => {
+		expect(capture(120, 10)).toBeCloseTo(60, 0);
+	});
+
+	test("resyncs after a stall instead of capturing a burst", () => {
+		const pacer = new Pacer(30, Time.Milli(0));
+		expect(pacer.due(Time.Milli(0))).toBe(true);
+
+		// The tab was hidden for a second, which suspends requestAnimationFrame entirely.
+		expect(pacer.due(Time.Milli(1000))).toBe(true);
+
+		// The next frame is one period out, not a second's worth of backlog.
+		expect(pacer.due(Time.Milli(1001))).toBe(false);
+		expect(pacer.due(Time.Milli(1034))).toBe(true);
+	});
+});
--- a/js/publish/src/video/polyfill.ts
+++ b/js/publish/src/video/polyfill.ts
@@ -1,6 +1,38 @@
 import { Time } from "@moq/net";
 import type { StreamTrack } from "./types";
 
+/**
+ * Paces the capture loop below against a frame deadline.
+ *
+ * Exported for the unit test; not part of the package's public API.
+ */
+export class Pacer {
+	// Safari and Firefox clamp performance.now() to 1ms, so a 33.333ms period never looks elapsed while
+	// the clock still reports 33. Allow the clock's own resolution as slack, otherwise we wait for one
+	// more animation frame and capture 20fps out of a 30fps camera.
+	static readonly SLACK = Time.Milli(1);
+
+	readonly #period: Time.Milli;
+	#next: Time.Milli;
+
+	constructor(frameRate: number, now: Time.Milli = Time.Milli.now()) {
+		this.#period = Time.Milli(1000 / frameRate);
+		this.#next = now;
+	}
+
+	/** True when the next frame is due, advancing the deadline past it. */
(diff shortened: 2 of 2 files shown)


fperex:
Closed in favor of https://github.com/moq-dev/moq/pull/2226

PR #2210, opened 2026-07-13
feat(moq-mux): expose finish_group() through the import chain

bgreenway:
`container::Producer` already has `finish_group()` — *"close the current group early; the next write must be a keyframe"* — but it isn't reachable above the container layer. This plumbs a `finish_group()` passthrough up through:

- every codec `Import` (h264/h265/av1/vp8/vp9/aac/opus/flac/mp3/legacy),
- the `Importer` / `FrameSink` / `StreamImporter` object-safe traits and their macro/struct impls,
- the public `import::Track` / `import::TrackStream`.

### Why

It lets a caller close a track's current group cleanly **without finishing the track**: the track (and its catalog rendition) stays open and publishing resumes into a fresh group on the next keyframe. The motivating use is pausing a track that has lost its last subscriber while leaving a **complete** final group rather than a truncated one — but the primitive is generally useful for any group-boundary control above the container layer.

### Notes

- Purely additive — no behavior change to existing paths.
- Adds a test (`import::track::tests::finish_group_closes_group_and_keeps_track_open`, opus) alongside the existing `container::producer` coverage.
- `cargo test -p moq-mux` green; `cargo fmt -p moq-mux --check` + `cargo clippy -p moq-mux` clean.

Happy to gate/adjust naming or split further if you'd prefer.

### DIFF
--- a/rs/moq-mux/src/codec/aac/import.rs
+++ b/rs/moq-mux/src/codec/aac/import.rs
@@ -67,9 +67,8 @@ impl<E: CatalogExt> Import<E> {
 		Ok(())
 	}
 
-	/// Close the current group cleanly WITHOUT finishing the track, so the track
-	/// stays open and the next keyframe opens a fresh group. Delegates to the
-	/// container producer's `finish_group`.
+	/// Close the current group without finishing the track; publishing resumes on
+	/// the next keyframe. See `import::Track::finish_group` for the full contract.
 	pub fn finish_group(&mut self) -> crate::Result<()> {
 		self.track.finish_group()?;
 		Ok(())
--- a/rs/moq-mux/src/codec/aac/import.rs
+++ b/rs/moq-mux/src/codec/aac/import.rs
@@ -67,6 +67,14 @@ impl<E: CatalogExt> Import<E> {
 		Ok(())
 	}
 
+	/// Close the current group cleanly WITHOUT finishing the track, so the track
+	/// stays open and the next keyframe opens a fresh group. Delegates to the
+	/// container producer's `finish_group`.
+	pub fn finish_group(&mut self) -> crate::Result<()> {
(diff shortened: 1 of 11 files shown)


bgreenway:
Thanks for the review — addressed in the follow-up commit (docs-only):

- **Centralized** the `finish_group` description on `Track::finish_group`; the codec `Import`s, the object-safe traits, and `TrackStream::finish_group` are now terse references to it (removes the near-duplicated wording).
- **Invariants** now documented on `Track::finish_group`: call while a group is open; the next group opens on the following keyframe (mirroring the container producer's contract, whose next write must be a keyframe); composes with `finish` / `seek`.
- **legacy** `#[allow(dead_code)]`: reconciled the two suggestions — the `///` doc is now behavior-only (per CodeRabbit, no call-site narration in public docs), while the parity justification moved to a plain `//` comment (keeps the *why* without the rot-prone "nothing calls this today").

`cargo build/test/fmt/doc -p moq-mux` clean. Happy to adjust further.

kixelated:
> It lets a caller close a track's current group cleanly without finishing the track: the track (and its catalog rendition) stays open and publishing resumes into a fresh group on the next keyframe. The motivating use is pausing a track that has lost its last subscriber while leaving a complete final group rather than a truncated one — but the primitive is generally useful for any group-boundary control above the container layer.

Wow somehow I missed this PR until just now. I ran into a similar issue: https://github.com/moq-dev/moq/pull/2263

Same idea, but I included a timestamp and a dummy empty frame, just so we'll know the duration of the last frame (for CMAF export). Unfortunately the dummy frame isn't backwards compatible though, not sure what to do about that.

kixelated:
Implemented via #2270, which renames it to `cut` and takes an optional timestamp.

PR #2211, opened 2026-07-13
feat(publish): prefer the codecs Safari actually hardware encodes

fperex:
Split out of #2190, which I am closing in favour of focused PRs.

## Root cause

`HARDWARE_CODECS` in `js/publish/src/video/encoder.ts` lists VP9 first, and `#bestCodec` takes the first codec the browser reports as supported under `prefer-hardware`. On Safari that is `vp09.00.10.08`, so **Safari publishes software VP9 while its hardware H.264 and HEVC sit idle**.

The support probe cannot catch this, because Safari's answer is not informative. Measured on Safari 26.5 / Apple Silicon at 1280x720:

```
codec              | prefer-hardware supported | accel echoed back | prefer-software supported
-------------------|---------------------------|-------------------|--------------------------
vp09.00.10.08      | true                      | prefer-hardware   | true
avc1.640028        | true                      | prefer-hardware   | true
av01.0.08M.08      | true                      | prefer-hardware   | true
hev1.1.6.L93.B0    | true                      | prefer-hardware   | true
vp8                | true                      | prefer-hardware   | true

>>> the current HARDWARE_CODECS order picks: vp09.00.10.08
```

Safari says yes to everything and echoes `prefer-hardware` straight back, so the `hardware.config?.hardwareAcceleration !== "prefer-hardware"` check we use elsewhere always passes. The hint is a no-op: asking for `prefer-hardware` and `prefer-software` yields **byte-identical output** for VP9 (1611 KB either way), AV1, VP8 and HEVC, which is also why throughput alone cannot distinguish them.

What does distinguish them is CPU, since VideoToolbox offloads to the media engine and libvpx does not. Sustained realtime encode, 1280x720 @ 30fps, 40s of camera-like content, measuring Safari's total CPU time and subtracting the no-encoder baseline:

| codec | encode CPU | frames out |
|---|---|---|
| `vp09.00.10.08` (picked today) | **11.89 s** | 1199 / 1199 |
| `avc1.640028` | **3.02 s** | 1199 / 1199 |
| `hev1.1.6.L93.B0` | **2.82 s** | 1199 / 1199 |

Roughly **4x the CPU** for the codec we currently default to on Safari. VideoToolbox on Apple Silicon only hardware-encodes H.264 and HEVC.

## The fix

`hardwareAcceleration` is a hint the browser may ignore, and we already can't trust it on Firefox (https://github.com/w3c/webcodecs/issues/896). So key off the engine rather than the echoed hint: on Safari, offer only the two codecs it genuinely accelerates, and let anything else fall through to the existing software pass. Every other browser keeps the current order, unchanged.

H.264 goes ahead of HEVC because every watcher can decode it (Firefox still can't decode HEVC).

The support matrix (`<moq-publish-support>`) had the same bug from the same cause, advertising hardware VP9 and AV1 on Safari. It now keys off the codec as well.

The codec lists move out of `#bestCodec` into `video/codecs.ts` so the ordering can be unit tested. It is not re-exported from `./index`, so the package's public API is unchanged.

## Test plan

`bun test`: 4 new tests pinning the orderings, plus a `detectSafari` UA corpus in `@moq/hang`. `@moq/hang` also gains a `test` script, because it had five test files that nothing was running. `just js check` and `just js test` green.

Verified on device, Safari 26.5 / Apple Silicon, in the `demo/web` publish page, reading the resolved encoder config:

| | resolved codec |
|---|---|
| before | `vp09.00.10.08` |
| after | `avc1.640028`, `prefer-hardware` |

Chrome is unaffected (its probe answers honestly and it keeps the VP9-first order); Firefox still skips the hardware pass entirely.


js/publish/src/video/encoder.ts
```diff
@@ -313,67 +314,10 @@ export class Encoder {
...
-
-			// AV1
-			// Super expensive to encode so it's our last choice.
-			"av01.0.08M.08",
-			"av01",
-		];
-
 		// Try hardware encoding first.
 		// We can't reliably detect hardware encoding on Firefox: https://github.com/w3c/webcodecs/issues/896
 		if (!Util.Hacks.isFirefox) {
```
kixelated: `!Util.Hacks.isFirefox && !Util.Hacks.isSafari`

Treat Safari and Firefox identically.

js/publish/src/video/encoder.ts
```diff
@@ -313,67 +314,10 @@ export class Encoder {
 		const dimensions = effect.get(this.#dimensions);
 		if (!dimensions) return;
 
-		// A list of codecs to try, in order of preference.
-		const HARDWARE_CODECS = [
```
kixelated, line 317: Why did you move this into ./codecs and remove the comments?

js/publish/src/video/codecs.ts
```diff
@@ -0,0 +1,30 @@
+// Codec preference lists for Encoder.#bestCodec, most preferred first.
+//
+// Not re-exported from ./index, so the ordering stays an internal detail that the unit test can pin
```
kixelated: I mean the unit tests are pretty silly. This is trivial behavior.

kixelated (commented):
I just don't understand why Firefox and Safari are being treated differently. Both seemingly have the same bug.

PR #2213, opened 2026-07-13
feat(publish): measure the upload bitrate from the bytes we encode

fperex:
Split out of #2190, which I am closing in favour of focused PRs.

## Root cause

The stats tab graphs the connection's send-bandwidth estimate:

```ts
const upload = publish.connection.established.peek()?.sendBandwidth?.peek();
vBitrateGraph.push(upload && upload > 0 ? upload : undefined);
```

Safari has no such estimate. Its WebTransport implements no `getStats()`, so `sendBandwidth` is `undefined` and the bitrate graph stays empty for the entire session. Measured on Safari 26.5, publishing over WebTransport to the local relay:

```
transport=webtransport  sendBandwidth=UNDEFINED  measuredKbps=1394
```

## The fix

Count the bytes the encoder actually produces. That is transport agnostic, so it works on Safari and on the WebSocket fallback; it is what we really upload rather than an estimate of what we could; and it matches the graph's "Bitrate" label, which is not the same quantity as the congestion controller's capacity estimate (the demo graphs that separately, as "Upload estimate").

`Video.Encoder` gains a `bytesEncoded` getter, which is additive. `demo/web/src/publish.ts` already calls out the gap this closes:

> Live graphs (capture rate, upload-bandwidth estimate, round trip). The publish API exposes no encoded-byte counter, so these are the honestly-observable signals.

The active pill now reflects either rendition rather than just `hd`, to match the bitrate graph, which sums both.

## Public API

`@moq/publish` gains `Video.Encoder.bytesEncoded: Getter<number>`. Monotonic total; diff it over an interval for a bitrate.

## Test plan

`just js check` and `just js test` green. Verified on device, Safari 26.5, publishing the camera from `demo/web` with a subscriber attached: the estimate is `undefined` while the measured value tracks the encode at ~1394 kbps.

### DIFF
--- a/js/publish/src/video/encoder.ts
+++ b/js/publish/src/video/encoder.ts
@@ -50,6 +50,10 @@ export class Encoder {
 	#catalog = new Signal<Catalog.VideoConfig | undefined>(undefined);
 	readonly catalog: Getter<Catalog.VideoConfig | undefined> = this.#catalog;
 
+	#bytesEncoded = new Signal(0);
+	/** Total bytes encoded while serving. Monotonic; diff it over an interval for an upload bitrate. */
+	readonly bytesEncoded: Getter<number> = this.#bytesEncoded;
+
 	#signals = new Effect();
 
 	// The user provided config.
@@ -105,6 +109,8 @@ export class Encoder {
 					if (frame.type === "key") {
 						lastKeyframe = frame.timestamp as Time.Micro;
 					}
+
+					this.#bytesEncoded.update((total) => total + frame.byteLength);
 
 					producer.encode(frame, frame.timestamp as Time.Micro, frame.type === "key");
 				},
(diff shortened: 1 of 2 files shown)


js/publish/src/video/encoder.ts
```diff
@@ -50,6 +50,10 @@ export class Encoder {
 	#catalog = new Signal<Catalog.VideoConfig | undefined>(undefined);
 	readonly catalog: Getter<Catalog.VideoConfig | undefined> = this.#catalog;
 
+	#bytesEncoded = new Signal(0);
+	/** Total bytes encoded while serving. Monotonic; diff it over an interval for an upload bitrate. */
+	readonly bytesEncoded: Getter<number> = this.#bytesEncoded;
```
kixelated, line 55: just call it `bytes`.

Or better yet, make a Stats object so we report frames/bytes/etc.

kixelated:
Took this over in #2246 to address the review: replaced the lone `bytesEncoded` getter with a `Video.Encoder.stats: Getter<Stats>` object (`{ frames, bytes }`), so we report frames/bytes/etc. from one signal. Original commit preserved. Thanks @fperex!

fperex:
Closed in favor of: https://github.com/moq-dev/moq/pull/2213

Issue #2212, opened 2026-07-13
moqsink: publisher dies permanently on transport death (no reconnect)


bgreenway:
### Current behavior

`moqsink`'s session task (`rs/moq-gst/src/sink/session.rs::run`) does a **one-shot connect** and, on any transport death, posts a fatal `element_error` on the bus and sets an `errored` flag that stops the pad threads from feeding the (now dead) session:

```rust
async fn run(...) {
    let result = connect_and_run(&settings, &origin, &status, &element).await;
    status.reset();
    notify_connected(&element);
    if let Err(err) = result {
        errored.store(true, Ordering::Relaxed);
        if let Some(obj) = element.upgrade() {
            gst::element_error!(obj, gst::CoreError::Failed, ("session error"), ["{err:?}"]);
        }
    }
}
```

So after **any** transport drop — a relay restart, a QUIC idle timeout, a transient network blip — the publish is permanently dead until the whole pipeline is torn down and rebuilt. There's no reconnect.

### Impact

For an unattended publisher (e.g. a fixed camera/edge node publishing 24/7), this means a single relay restart silently kills the stream until something externally rebuilds the pipeline. We hit exactly this in production: a QUIC idle timeout surfaced "session terminated", no retry, publish gone. A long-lived publisher really wants to outlive arbitrary relay/transport outages.

### Proposal

Wrap connect+idle in a reconnect loop with exponential backoff (e.g. 1s → ×2 → cap 30s; reset the backoff when a connection stayed up past the cap, so a healthy connection's drop retries promptly). `origin` is already held for the task's lifetime, so the published broadcast survives across reconnects; `connect_and_run` re-consumes it for each connection. During an outage the pad threads keep writing into the producers, and memory is bounded by moq-net's per-group eviction (`MAX_GROUP_CACHE` / `MAX_GROUP_FRAMES`), so no unbounded buffering. On reconnect the relay catches up from a group boundary (keyframe).

We've been running this on our fleet since v0.2.10 and it's been solid across real relay restarts — happy to open a PR.

### Design questions (before I send code)

1. **Default-on, or behind a `reconnect` property?** One could argue one-shot-connect-then-fatal-error is a poor default for a live publisher, so reconnect-by-default is defensible; but some apps may prefer the element to error out so they can handle recovery at the app layer. Preference?
2. **Bounded or unbounded?** Our version never gives up (backoff caps but loops forever). Would you want a `max-retries` / give-up-timeout instead (or in addition)?
[...]

PR #2223, opened 2026-07-13
fix(moq-gst): reconnect the moqsink publisher instead of dying on transport death

bgreenway:
Closes the moqsink half of #2212.

### Problem
`moqsink` did a one-shot `client.connect()` and, on any transport death, posted a fatal `element_error` and set an `errored` flag that stopped the pads feeding the session (`sink/session.rs`). So after **any** transport drop — relay restart, QUIC idle timeout, a transient blip — the publish was permanently dead until the whole pipeline was rebuilt. Fatal for an unattended 24/7 publisher (we hit exactly this in production: a QUIC idle timeout → "session terminated" → publish gone).

### Fix
Switch the sink to the existing `moq-native` reconnect primitive (`Client::reconnect`) with `backoff.timeout = 0` (never give up), matching #1806's flap backoff. Pads keep writing across an outage (bounded by moq-net's per-group eviction) and the relay catches up from a group boundary on reconnect.

### The one design decision — surfacing per-session stats through `Reconnect`
The wrinkle from #2212: the sink reads **per-session** state for its read-only properties — `session.version()` and the congestion-controller send estimate (`estimated-send-bitrate`) — but `Reconnect` owned the session and exposed only `Connected`/`Disconnected`. This is the "transparent reconnect" you mentioned wanting, so I surfaced that state **through** `Reconnect`:

- `reconnect.rs` gains a `Snapshot { status, version, send_bitrate }`; the loop writes `version` on connect and forwards the live `send_bandwidth` estimate into shared state (the `select` that used to live in the sink moved into the primitive).
- `changed()` / `poll_changed()` report snapshot deltas (so a live send-bitrate update wakes a consumer, not just connect edges), plus `version()` / `send_bitrate()` accessors. `status()` is unchanged.
- The sink's `forward` task mirrors each snapshot into its `Status` and notifies `connected` on the edges. `imp.rs` is untouched — its `errored` short-circuit now trips only on a *bounded* backoff's terminal give-up, never on a transient drop.

**This is a proposal, not a fixed shape** — since you want reconnect to become the default/transparent behavior, I'm happy to move the stats surface wherever fits that API (e.g. hang it off `connect()` directly, or a different observability handle). Wanted to give you a working, tested reference for the moqsink half either way.

### Validation
`cargo build` / `test` / `fmt --check` / `clippy` clean on `moq-native` + `moq-gst` (the one clippy note is pre-existing in `quinn.rs`, untouched here). Added a `reconnect.rs` unit test for the snapshot change-detection. Running it in production (patched) across our fleet; reconnect survives relay restarts + QUIC idle timeouts with the properties intact.

### DIFF
--- a/rs/moq-gst/src/sink/session.rs
+++ b/rs/moq-gst/src/sink/session.rs
@@ -118,14 +118,12 @@ impl Session {
 		let status = Arc::new(Status::default());
 		let errored = Arc::new(AtomicBool::new(false));
 
-		// Hand the publish origin to a background reconnect loop: connect, wait for the session to
-		// close, then reconnect with exponential backoff. This replaces the previous one-shot connect
-		// that posted a fatal bus error on the first transport death — a relay restart or QUIC idle
-		// timeout left the publish permanently dead until the pipeline was rebuilt. `timeout = 0` means
-		// never give up, so an unattended publisher outlives arbitrary relay/transport outages; the
-		// pad threads keep writing across an outage (bounded by moq-net's per-group eviction) and the
-		// relay catches up from a group boundary on reconnect. A bounded retry policy is available via
-		// `ClientConfig::backoff` if a terminal error is preferred.
+		// Publish through a background reconnect loop (connect, wait for close, reconnect with
+		// backoff) rather than a one-shot connect that died on the first transport drop. `timeout = 0`
+		// retries transport/connection failures indefinitely so an unattended publisher outlives
+		// relay/QUIC outages; non-retryable errors (e.g. auth) stay terminal. During an outage the pad
+		// threads keep writing — bounded by moq-net's per-group eviction — and the relay catches up
+		// from a group boundary on reconnect. A bounded policy is available via `ClientConfig::backoff`.
 		let mut config = moq_native::ClientConfig::default();
 		config.tls.disable_verify = Some(settings.tls_disable_verify);
 		config.backoff.timeout = std::time::Duration::ZERO;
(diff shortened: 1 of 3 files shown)


bgreenway:
Thanks — both addressed in the follow-up (baf21fbf):

- **`Snapshot` `#[non_exhaustive]`** — done; consumers read fields / construct via `Default`.
- **Comment corrections** — you're right on both: `timeout = 0` retries *transport/connection* failures indefinitely but non-retryable errors (auth) stay terminal, so I dropped the "never gives up" wording; and `origin.consume()` is taken once at start (the client re-publishes it per connect), so the forward task just holds the producer alive. Also fixed the `Err`-arm comment — it's reachable on a terminal error, not only a bounded give-up. Trimmed the historical framing to current-behavior per the guideline.

build/test/fmt/clippy clean.

bgreenway:
CI catch (2da376c4): the hermetic `connect_failure_posts_error_to_bus` test failed — and correctly so. With the reconnect switch (`timeout=0`) an unreachable relay at startup is now **retried** rather than posted as a fatal element error that tears down the pipeline, which is the whole point of #2212. Updated the test to assert the new contract: a failed connect does **not** post an ERROR and leaves `connected=false` while it retries. (A non-retryable auth failure stays terminal — that path needs a live relay and is covered separately.)

My mistake earlier: I had run `cargo test -p moq-native` but not the `moq-gst` integration tests. Now ran the full `moq-gst` + `moq-native` suites (incl. `tests/element.rs`) — all green; fmt + clippy clean. CI should go green on the new head.

kixelated:
thanks @bgreenway , I merged #2240 which is veeeery similar to this PR but the moq-native API changes are a bit nicer. Let me know if that doesn't work for you.

PR #2237, opened 2026-07-13
fix(moq-gst): reconnect the moqsink publisher instead of dying on transport death

kixelated:
Closes the moqsink half of #2212. Supersedes #2223 (built on its commits, so @bgreenway keeps authorship of the reconnect work).

### Problem
`moqsink` did a one-shot `client.connect()` and posted a fatal `element_error` on any transport death (relay restart, QUIC idle timeout, a transient blip), which permanently killed the publish until the whole pipeline was rebuilt. Fatal for an unattended 24/7 publisher.

### Fix
Switch the sink to the `moq-native` reconnect primitive (`Client::reconnect`) with `backoff.timeout = 0`: retryable transport failures reconnect forever, while a non-retryable error (auth rejection) stays terminal and still posts the fatal element error. Pads keep writing across an outage (bounded by moq-net's per-group eviction) and the relay catches up from a group boundary on reconnect.

### The stats surface: `Reconnect` now mirrors `Session`
moqsink exposes read-only `connected` / `moq-version` / `estimated-send-bitrate` properties, which need the live session's state. #2223 surfaced these through a bespoke `Snapshot { status, version, send_bitrate }` value on `Reconnect`. This reworks that so `Reconnect`'s read surface **mirrors `moq_net::Session`** instead — a caller can treat it like a session that transparently reconnects:

- `version() -> Option<Version>` (Session's `Version` type; `Option` because a reconnecting handle can be between sessions).
- `send_bandwidth() -> BandwidthConsumer` and `recv_bandwidth() -> BandwidthConsumer`, mirroring `Session::{send,recv}_bandwidth`. Unlike the session's, these are **persistent**: the reconnect loop forwards each session's estimate into them, so a consumer's handle survives reconnects instead of dying with each session. The value is `None` while disconnected or when the backend has no estimate.
- `connected() -> bool`, the `&self` read behind `status()`.

`Snapshot` / `changed()` / `poll_changed()` / `send_bitrate()` are dropped.

moqsink then reads `estimated-send-bitrate` straight off `reconnect.send_bandwidth().peek()` (no mirror, no per-sample plumbing), and keeps only a slim `{connected, version}` mirror driven by `reconnect.status()` for the other two properties + the `notify::connected` edge.

### Validation
`clippy --all-features`, `fmt --check`, `cargo doc -D warnings`, and tests all clean on `moq-native` (63) + `moq-gst` (34 lib + 3 element, incl. `connect_failure_retries_without_erroring`).

### DIFF
--- a/doc/bin/gstreamer.md
+++ b/doc/bin/gstreamer.md
@@ -30,14 +30,6 @@ Both elements support the following properties:
 For `http://` URLs, `moq-native` automatically fetches the server's certificate fingerprint from `/certificate.sha256` and verifies TLS against it. You don't need `tls-disable-verify` for local development.
 :::
 
-`moqsink` additionally exposes these read-only properties for monitoring:
-
-| Property                 | Type   | Description                                                  |
-| ------------------------ | ------ | ----------------------------------------------------------- |
-| `connected`              | bool   | Whether the publish session is currently connected          |
-| `moq-version`            | string | The negotiated MoQ protocol version; null when disconnected |
-| `estimated-send-bitrate` | uint64 | Estimated send bitrate in bits per second; 0 when unavailable |
-
 ## Prerequisites
 
 The plugin requires GStreamer development libraries. It is **not** built by default since most users don't have them installed.
--- a/rs/moq-gst/src/sink/imp.rs
+++ b/rs/moq-gst/src/sink/imp.rs
@@ -129,22 +129,6 @@ impl ObjectImpl for MoqSink {
 					.blurb("Disable TLS verification")
 					.default_value(false)
 					.build(),
-				// Read-only, served from the live session's status.
-				glib::ParamSpecBoolean::builder("connected")
-					.nick("Connected")
-					.blurb("Whether the session is currently connected")
-					.read_only()
-					.build(),
-				glib::ParamSpecString::builder("moq-version")
-					.nick("Negotiated version")
-					.blurb("The negotiated MoQ protocol version, null when disconnected")
-					.read_only()
-					.build(),
-				glib::ParamSpecUInt64::builder("estimated-send-bitrate")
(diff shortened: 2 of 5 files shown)


kixelated:
@bgreenway I removed the read only properties so the PR is much simpler. They're not hooked up, and IDK we can always add them back when they're actually load bearing.

bgreenway:
@kixelated they are load-bearing for us, as it happens. Our fleet reads `estimated-send-bitrate` (+ `connected` / `moq-version`) off moqsink via `g_object_get` on a ~5s poll tick, to surface per-camera publish health — the GCC send estimate, negotiated version, link state — in our trailer telemetry + operator dashboard (shipped across the fleet). Because it is a *poller*, the "never `notify`d" gap does not bite us: we read the current value each tick, no change-signal needed.

They do need the session stats surfaced through `Reconnect` (the `Snapshot`/`version()`/`send_bitrate()` bits I had added — the estimate lives on the live session the reconnect loop owns). Happy to send that back as a small **follow-up on top of #2237** — just the three read-only props + the minimal `Reconnect` stats surface — so this PR stays pure-reconnect and the observability rides separately now that it has a real consumer. Sound good?

kixelated:
> @kixelated they are load-bearing for us, as it happens. Our fleet reads `estimated-send-bitrate` (+ `connected` / `moq-version`) off moqsink via `g_object_get` on a ~5s poll tick, to surface per-camera publish health — the GCC send estimate, negotiated version, link state — in our trailer telemetry + operator dashboard (shipped across the fleet). Because it is a _poller_, the "never `notify`d" gap does not bite us: we read the current value each tick, no change-signal needed.

Awesome, okay let me preserve that and try another approach. I forgot to hit submit on your PR, but yeah the `Status` struct in moq-native feels wrong.

PR #2254, opened 2026-07-14
fix(moq-hls): release source subscriptions when a Broadcaster is dropped

kixelated:
## Problem

`export::Broadcaster::new` spawns a detached `watch_catalog` task that holds an
`Arc<Broadcaster>` self-reference, and the rendition pumps only exit when the
**source broadcast closes**. So dropping the last *external* `Arc<Broadcaster>`
(e.g. a downstream recorder task ending) does **not** stop the export: the
catalog + per-track subscriptions stay open, and the `Broadcaster` lives on
until the broadcast closes on its own.

For a **subscription-driven publisher** this is self-sustaining:

```
leaked subscription -> publisher keeps producing -> broadcast never closes
      ^                                                      |
      +--------------- watch_catalog never exits <-----------+
```

### Observed in production (moq.pro)

A VOD recording (which builds a `Broadcaster` per recorded broadcast) was
enabled against the demo's six `moq-boy` Game Boy streams, then deleted a couple
minutes later. Deleting it made the recorder's session heartbeats 404, so the
recorder tasks ended and dropped their `Arc<Broadcaster>`s. But the exports kept
their subscriptions open: 18 track subs / 6 "viewers" stayed live. `moq-boy`
only emulates while subscribed, so all six emulators ran flat-out for **~16
hours** after the recording was gone, pinning the single-vCPU demo node at ~90%
CPU. `#2021 (evict finished HLS broadcasters)` doesn't cover this: the broadcast
isn't finished, it's kept alive by the leak.

For an always-pushing encoder the same drop leaks the `watch_catalog` +
rendition pump tasks and their subscriptions on the relay (a slow accumulation),
just without the visible CPU symptom.

## Fix

Break the cycle and tie every spawned task to its owner's lifetime:

- `watch_catalog` now holds a `Weak<Broadcaster>` (upgrades per catalog, exits
  when the owner is gone), so dropping the last external `Arc` runs
  `Broadcaster`'s `Drop`.
- A small `AbortOnDrop` guard aborts the (possibly parked) `watch_catalog` task
  on drop, and each `Rendition` owns its pump's guard, so dropping the
  `Broadcaster` -> renditions map -> pumps releases every track subscription
  immediately.

No public API change (`Broadcaster::new` / `set_paused` / `rendition()`
unchanged; the added fields are private).

## Test

`export::tests::dropping_broadcaster_releases_subscription`: build a
`Broadcaster` over a producer-owned `catalog.json` track, wait for it to
subscribe, drop it, and assert the producer sees no consumers. Times out before
the fix (verified by neutralizing the abort), passes after. Full `moq-hls`
suite green (20/20).

> Branch is based on the moq.pro pin (`6b46b1f0`) for a minimal delta; happy to
> rebase onto `dev`.

### DIFF
--- a/rs/moq-hls/src/export/mod.rs
+++ b/rs/moq-hls/src/export/mod.rs
@@ -12,7 +12,7 @@ mod rendition;
 pub mod store;
 
 use std::collections::BTreeMap;
-use std::sync::{Arc, Mutex};
+use std::sync::{Arc, Mutex, Weak};
 use std::time::Duration;
 
 use moq_mux::catalog::hang::Catalog;
@@ -24,6 +24,19 @@ pub use rendition::{Kind, Rendition};
 
 /// How long to wait before retrying the initial catalog subscription.
 const CATALOG_RETRY: Duration = Duration::from_millis(250);
+
+/// Aborts a spawned task when dropped, tying the task's lifetime to the value that
+/// owns this guard. Used to stop the catalog watcher + rendition pumps the moment a
+/// [`Broadcaster`] is dropped, so its source subscriptions are released instead of
+/// lingering until the broadcast itself closes (which a subscription-driven
+/// publisher would never do while we hold it open -- a self-sustaining leak).
+pub(super) struct AbortOnDrop(pub(super) tokio::task::AbortHandle);
+
+impl Drop for AbortOnDrop {
+	fn drop(&mut self) {
+		self.0.abort();
+	}
+}
 
 /// Export tuning shared across renditions.
 #[derive(Clone, Debug)]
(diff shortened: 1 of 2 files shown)


kixelated:
Closing in favor of #2255 to agree on the approach first. The candidate fix stays on branch `fix/hls-broadcaster-drop-teardown` (not deleting it) if we want to revisit it.

kixelated:
Reopened and retargeted to `main` per maintainer preference (better than status quo). Background + alternative fix directions in #2255; happy to reshape toward the owner-anchored redesign there as a follow-up.

Issue #2255, opened 2026-07-14
moq-hls: Broadcaster leaks source subscriptions when dropped (only stops on broadcast close)


kixelated:
## Summary

An `export::Broadcaster` does not release its source subscriptions when the owner
drops it. It only tears down when the **source broadcast closes**. Dropping the
last external `Arc<Broadcaster>` is silently ignored, so the catalog + per-track
subscriptions stay open indefinitely.

## Mechanism

- `Broadcaster::new` spawns a **detached** `watch_catalog` task that holds an
  `Arc<Broadcaster>` self-reference. Dropping the external `Arc` therefore never
  brings the refcount to zero, so the `Broadcaster` is never dropped.
- The rendition pumps only exit on broadcast-close (while running/unpaused they
  have no shutdown signal tied to the owner).

For a **subscription-driven publisher** this is self-sustaining and never
recovers on its own:

```
leaked subscription -> publisher keeps producing -> broadcast never closes
      ^                                                      |
      +--------------- watch_catalog never exits <-----------+
```

For an always-pushing encoder it's a quieter leak: the `watch_catalog` +
rendition-pump tasks and their subscriptions accumulate on the relay for every
`Broadcaster` whose owner drops it before the broadcast ends.

Note `#2021 (evict finished HLS broadcasters)` does not cover this - the
broadcast is not finished, it's held open by the leak itself.

## How it showed up (moq.pro)

A VOD recording builds one `Broadcaster` per recorded broadcast. A recording was
enabled against the demo's six `moq-boy` streams, then deleted ~2 min later. The
recorder's session heartbeats then 404'd and the recorder tasks ended (dropping
their `Arc<Broadcaster>`s) - but the exports kept their subscriptions:
`subscriptions_opened{internal,subscriber}=27, closed=9` (18 still open = 6
broadcasts x 3 tracks), 6 "viewers" still open. `moq-boy` only emulates while
subscribed, so all six emulators ran flat-out for ~16h after the recording was
gone, pinning a 1-vCPU node at ~90% CPU. Restarting the publishers (fresh
publish, no subscriber) cleared it.

## Repro (unit-level)

Build a `Broadcaster` over a producer-owned `catalog.json` track, wait for it to
subscribe, drop it, and observe the producer still sees a consumer (never goes
`unused()`).

## Possible directions (open for discussion)

The core requirement: dropping the owner's `Broadcaster` handle should release
its source subscriptions promptly. Some options:

[...]

PR #2258, opened 2026-07-14
moq-hls: rewrite export::Broadcaster as an owned poll-driven state machine

kixelated:
## Why

`export::Broadcaster` spawned a detached `watch_catalog` task (holding an `Arc<Broadcaster>` self-reference) plus one detached pump per rendition, and tore down only when the **source broadcast** closed. Dropping the owner's handle did nothing, so a VOD recorder that stopped recording a still-live broadcast leaked its catalog + per-track subscriptions -- keeping subscription-driven publishers (moq-boy) emulating at ~90% CPU for hours after a recording was deleted (#2255).

#2254 patched this with a `Weak` + abort-on-drop guards. This replaces that machinery with a design that **can't leak by construction**, and cleans up the surrounding task/ownership layer that motivated the leak.

## What

`Broadcaster` is now a plain owned value the caller drives by polling, mirroring `fmp4::Export` one layer down:

- **`poll(&mut self, waiter) -> Poll<()>`** advances the discovery catalog and every rendition's `fmp4::Export` in one **cancel-safe** pass, with **no** background tasks. Dropping the `Broadcaster` drops its catalog consumer + every `Export`, releasing all source subscriptions immediately -- the leak is unrepresentable, and the #2254 `Weak`/`AbortOnDrop` machinery is gone.
- **`Handle`** (cheap, cloneable) carries the read side (rendition set + stores + `wait_ready` + `master_playlist`) to HTTP handlers and downstream consumers, and adds **`renditions()`** so callers enumerate the set directly instead of re-parsing the master playlist. It holds no subscription and can't keep the export alive past the driver.
- **`set_paused(&mut self)`** is a plain setter applied between polls (each owner writes its own `select!` over `poll` + its pause signals); **`run()`** is a no-pause convenience for the HTTP server.

### Delete the catalog retry loop (fix the real bug)

`CATALOG_RETRY` papered over a publish-order race, not a transient. `catalog::Consumer::new` returns `NotFound` only when `catalog.json` is absent **and** the broadcast has no dynamic producer. Relayed broadcasts can't hit it (the relay's subscriber creates the dynamic handler *before* `publish_broadcast`); only the local `moq-cli hls import` could, because it announced the broadcast *before* creating the catalog. Fixed the ordering there (create the catalog before `publish_broadcast`) and made `Broadcaster::new` **fail loud** instead of retrying.

### Server

The driver task owns the `Broadcaster` and holds it via an abort-on-drop guard behind a per-generation id, so evicting/replacing a cache entry tears the export down; a finished driver self-evicts only its own generation. `master.m3u8` now returns 404 instead of serving an empty 200 when no rendition appears within the timeout.

### Source-end handling

A source ending abruptly (publisher disconnect -- the common live case, which surfaces as an `Err` rather than a clean `None`) finishes the rendition and is **logged, not propagated**: for a live broadcast it's the normal termination, not a fault, so propagating it would false-flag nearly every recording as truncated.

## Tests

- `dropping_broadcaster_releases_subscription` -- dropping the driver releases the **catalog** subscription.
- `dropping_broadcaster_releases_media_subscription` -- and the **live media** subscription (the actual #2255 scenario: a live track held open is what kept the demo publishers emulating).
- `broadcast_gone_completes` -- the broadcast going away drives `run()` to completion instead of hanging.
- Server cache/eviction generation tests.

Full `moq-hls` suite green (23/23). `cargo fmt` + `clippy` clean on `moq-hls`.

> Based on `main` (the #2254 merge), which this supersedes.

### DIFF
--- a/rs/moq-cli/src/hls.rs
+++ b/rs/moq-cli/src/hls.rs
@@ -50,12 +50,17 @@ pub struct ExportArgs {
 /// Pull a remote HLS/LL-HLS playlist (URL or file path) into the Origin under `name`.
 pub async fn import(origin: &moq_net::OriginProducer, name: String, playlist: String) -> anyhow::Result<()> {
 	let mut producer = moq_net::Broadcast::new().produce();
+
+	// Create the catalog tracks BEFORE announcing the broadcast: a subscriber that
+	// sees the announce must be able to subscribe `catalog.json` immediately. Announcing
+	// first left a window where a consumer got `NotFound` (the export side used to paper
+	// over it with a retry loop).
+	let catalog = moq_mux::catalog::Producer::new(&mut producer)?;
 	anyhow::ensure!(
 		origin.publish_broadcast(&name, producer.consume()),
 		"failed to publish broadcast"
 	);
 
-	let catalog = moq_mux::catalog::Producer::new(&mut producer)?;
 	let mut importer = moq_hls::import::Import::new(producer, catalog, moq_hls::import::Config::new(playlist))?;
 
 	tracing::info!(%name, "importing HLS");
--- a/rs/moq-hls/src/export/mod.rs
+++ b/rs/moq-hls/src/export/mod.rs
@@ -77,19 +77,28 @@ struct Shared {
 }
 
 /// A driver-private per-rendition unit: the shared metadata/store plus the exporter
-/// that fills it, and whether that exporter has finished.
+/// that fills it. `export` is `None` once the rendition has finished -- dropping it
+/// then RELEASES the source subscription immediately, instead of holding a live (but
+/// no-longer-polled) track open until the whole `Broadcaster` drops. That matters on
+/// the error path: moq-mux's exporter returns an error *before* it would internally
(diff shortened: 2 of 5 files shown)


kixelated:
Self-review (careful pass + independent reviewer) turned up one real issue, now fixed in `e81690a5`:

On the exporter **error** path, moq-mux's `poll_next_fragment` returns *before* it internally drops the finished track (the clean `Ok(None)` path reaches the step-7 `retain` that releases it). The driver marked the rendition done but kept its `Export`, so a rendition whose track errors while its publisher is still live would pin that media subscription for the rest of the recording — a scoped re-creation of the #2255 leak this PR exists to fix. Now a finished rendition drops its exporter (`export: Option<_>`), releasing the subscription immediately; the `Arc<Rendition>` stays in the shared read map so playlists still finish with ENDLIST.

Everything else (poll liveness/no-spin, completion, cancel-safety, pause/discontinuity, server eviction generations, the moq-cli publish-order reorder) checked out. 23/23 tests, clippy + fmt clean.

PR #2263, opened 2026-07-15
fix(mux,boy): tail-frame a group's end so a pause doesn't inflate recordings

kixelated:
## Summary

**Root cause.** Legacy/LOC hang groups carry no per-frame duration, so a muxer learns where a group's last frame *ends* only from the next group's first frame. When a publisher pauses and resumes at a later timestamp — `moq-boy` idling with no viewers while its wall clock keeps advancing — the muxer stretched the last pre-pause frame across the whole gap. A VOD of a mostly-idle stream reported a single ~2h segment and buffered forever on playback.

- **Tail frame** (spec'd in `drafts/draft-lcurley-moq-hang.md`): a frame with a timestamp but an empty payload, written as a group's last frame to mark its presentation end. The `Consumer` reads ahead (without blocking the live edge) to bound the trailing frame's duration by the tail, and never surfaces it as media. Optional for back-compat — a decoder that predates it treats the empty payload as a no-op access unit.
- **`Producer::cut(end)`** closes a group recording that its content ends at `end`. The empty-frame encoding stays inside moq-mux (gated by `Container::has_duration()`; CMAF backfills the last sample instead, no tail). Threaded through the h264 `Import` and moq-video's encode `Producer`. `moq-boy` calls it when the video track goes idle, which subsumes force-keyframe-on-resume — closing a group already requires the next frame to be a keyframe, so it arms the encoder IDR too.
- **Discontinuity.** With the trailing duration now correct, the fmp4 `Export` distinguishes a real forward jump (a pause) from ordinary spacing and sets `Fragment.discontinuity`; moq-hls's `SegmentStore` opens a fresh `#EXT-X-DISCONTINUITY` segment for the resumed media. Without a tail the trailing duration is inferred right up to the successor, so this never fires on continuous media.

## Test plan

- `cargo test -p moq-mux -p moq-hls -p moq-video` — 357 / 20 / 6 green. New coverage: the consumer bounds a frame across a 2h gap; `end_group` round-trips producer → consumer; the export emits a short pre-gap fragment plus a discontinuous one across a 2h jump; the store tags the post-gap segment.
- `cargo build -p moq-boy`.

## Cross-package sync

The wire now permits an empty-payload frame. The JS `@moq/hang` decoder and the Go/Swift/Kotlin/Python bindings should skip it — **deferred as follow-ups**, since back-compat means an old decoder treats the empty payload as a no-op access unit rather than a hard break. VOD playback is unaffected (hls.js on the recorded fMP4, not raw hang frames).

(Written by Claude Opus 4.8)

### DIFF
--- a/drafts/draft-lcurley-moq-hang.md
+++ b/drafts/draft-lcurley-moq-hang.md
@@ -216,6 +216,25 @@ The remainder of the payload is codec specific; see the WebCodecs specification 
 
 For example, h.264 with no `description` field would be annex.b encoded, while h.264 with a `description` field would be AVCC encoded.
 
+## Tail Frame
+A frame with a timestamp but an *empty* payload is a **tail frame**.
+It carries no media; its timestamp marks the presentation *end* of the preceding frame.
+
+Legacy and LOC frames carry no per-frame duration, so a decoder normally learns where the last frame of a group ends only from the first frame of the *next* group.
+That is wrong across a discontinuity: if a publisher pauses and resumes with a later timestamp (see {{draft-lcurley-moq-timestamp}}), the gap is real elapsed time, not the duration of the last frame.
+A tail frame closes the group explicitly, so the gap between one group's tail and the next group's first frame is an unambiguous discontinuity rather than a multi-hour final frame.
+
+A tail frame MUST be the last frame of its group.
+A publisher SHOULD write a tail frame as the last frame of every video group, with a timestamp equal to the presentation end of the group's last media frame (its timestamp plus its intended duration).
+The timestamp MUST be greater than or equal to the group's last media frame's timestamp.
+A tail frame is NOT required for audio, whose frame duration is deterministic (samples / sample-rate).
+
+A decoder MUST skip tail frames: they produce no output sample.
+A decoder that computes frame durations SHOULD bound the last media frame's duration by the tail frame's timestamp, and SHOULD treat a gap between the tail and the next group's first frame as a discontinuity.
+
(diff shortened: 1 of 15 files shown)


kixelated:
Thanks for the review — pushed fixes in 207ef1d.

**Addressed**
- **CI (rustfmt):** the failing `Check` was a formatting wrap in the export test — fixed.
- **moq-boy dropped frame (major):** `try_frame` now returns whether the frame was accepted, and `last_video_ts` (the group's `cut` boundary) only advances past frames actually published — a frame dropped for backpressure can no longer stretch the previous frame past real media.
- **legacy tail decode:** check `has_remaining()` before copying, returning the tail without copying an empty payload (matches the LOC path).
- **store discontinuity coverage:** added `discontinuous_fragment_forces_segment_even_mid_gop` — a non-independent (mid-GOP) discontinuous fragment must still open a new segment, guarding the `||` in `push` against regressing to `&&`.
- **`cut` + CMAF:** added a `has_duration()==true` test asserting `cut` backfills the trailing sample's duration and writes no tail frame.

**Deferred (with reasons)**
- **`cut` at zero latency / after a prior latency flush for CMAF:** latent, not reachable — `cut` is only wired through Legacy/LOC (no per-frame duration); no CMAF caller uses it. For a duration-carrying container each sample already self-terminates via its per-sample duration, so `cut` doesn't need to revise an emitted sample. I reworded the `cut` doc to state this accurately rather than refactor buffering for a caller that doesn't exist; happy to revisit if/when CMAF is wired through `cut`.
- **moq-boy queue-full regression test:** moq-boy is a binary with no unit-test harness; standing one up for the backpressure path is disproportionate to the one-line acceptance-guard fix. Left as a follow-up.

PR #2270, opened 2026-07-15
feat(moq-mux): cut(end) as the group boundary

kixelated:
Builds on #2210 — @bgreenway's commits are carried here, with the plumbed operation taking a timestamp. Supersedes it.

## Summary

A group's last frame has no end of its own. Legacy/LOC carry no per-frame duration, so a consumer can only bound that frame by waiting for the **next** group's first frame — and that group may be late, reordered, evicted, or never arrive. A publisher that pauses leaves its final frame unbounded, and a muxer then stretches it across the whole gap. (Concretely: moq-boy idles when it has no viewers, and a VOD recording of a mostly-idle stream reports one multi-hour segment.)

So the group boundary needs a timestamp:

- **`Producer::cut(end)`** closes the current group, recording where its content stops. A keyframe arriving through `write` cuts the previous group automatically using its own timestamp — so an explicit `cut` is only needed when the publisher knows the content ends somewhere the next keyframe won't tell us, i.e. before a pause.
- **Plumbed through the import chain** (#2210's shape): every codec `Import`, the `Importer` / `FrameSink` / `StreamImporter` traits, and the public `import::Track` / `TrackStream` — now carrying `end`.
- **`finish_group()` deprecated** (`#[doc(hidden)]` + `#[deprecated]`). It closes a group while saying nothing about where it ends, which is exactly the gap that caused the bug. The one-frame-per-group sources (audio packets, TS sections) genuinely have no boundary to give and keep a boundary-less `pub(crate) close()`.
- **Decoders tolerate an empty payload** (Rust `legacy`/`loc`, JS `@moq/hang` + `@moq/loc`). A coded frame is never empty, so an empty payload is free to mean "content stops here, no media" — at a group's end, or mid-group as a gap. Nothing reads it yet.

## What this deliberately does NOT do

`cut` does **not** write `end` to the wire for hang, so the final frame's duration is still unknown for that container and **the recording bug is not fixed by this PR**. Encoding the marker and reading its timestamp is a wire change and lands separately on `dev`.

This is the reader half, on purpose: get tolerance deployed everywhere *before* anything starts emitting markers.

## Public API

- Added: `Producer::cut(end)`, `Track::cut(end)`, `<codec>::Import::cut(end)`; the object-safe traits gained `cut`.
- Deprecated: `Producer::finish_group()` — hidden, still works, removed later.
- Backwards compatible: no existing signature changed, nothing removed.

## Test plan

- `cargo test -p moq-mux` (357 pass), `cargo clippy -p moq-mux --all-targets` clean, `cargo fmt` clean.
- `bun test js/loc js/hang/src/container` (31 pass).
- New coverage: markers skipped mid-group and at a group's end; consecutive markers make progress rather than spin (guards against an infinite loop); `cut` closes immediately; the deprecated `finish_group` still delegates; #2210's track test now cuts.
- A LOC test built its frame with an empty payload incidentally (it only exercised timestamp/timescale parsing) — it now carries a payload, since an empty one means marker.

## Also included

A drive-by fix to the deprecation guidance: it read as though `#[deprecated]` were part of what's banned (rs/CLAUDE.md listed only `#[doc(hidden)]`; the root blessed only a *runtime* warning, excluding a compile-time attribute by omission). The line is really warnings-fire-on-use (good) vs docs-advertise-to-everyone (bad). Happy to split this out if you'd rather it land on its own.

(Written by Claude Opus 4.8)

Co-Authored-By: Brad Greenway <brad.greenway@surveillx.com>
Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>


rs/moq-mux/src/container/legacy/mod.rs
```diff
@@ -36,8 +36,16 @@ impl Container for Wire {
...
 
 		let mut hang_frame = hang::container::Frame::decode(data)?;
-		let payload = hang_frame.payload.copy_to_bytes(hang_frame.payload.remaining());
 
+		// An empty payload is a marker, not a sample: it says content stops at this
+		// timestamp, so there's nothing to decode. Skip it and read on rather than
+		// handing an empty access unit to a decoder. Yielding no frames (instead of
+		// erroring) is what lets a publisher start emitting these without breaking us.
+		if !hang_frame.payload.has_remaining() {
+			return Poll::Ready(Ok(Read::Fragment(Vec::new())));
```
kixelated: I think we have to loop instead? Returning an empty fragment seems wrong.

js/loc/src/index.ts
```diff
@@ -77,6 +77,12 @@ export class Format {
 			throw new Error("loc: frame missing required timestamp property");
 		}
 
+		// An empty payload is a marker, not a sample: it says content stops at this
+		// timestamp, so there's nothing to decode. Yield no frames rather than hand an
+		// empty chunk to a decoder -- a publisher emitting these must not break us.
+		// Checked after the property block so a malformed marker still throws.
+		if (payload.byteLength === 0) return [];
```
kixelated: Don't make this change for LOC, as it's not in the spec.

rs/moq-mux/src/container/producer.rs
```diff
@@ -153,18 +154,36 @@ impl<C: Container> Producer<C> {
...
-		self.flush(next)?;
+	/// Close the current group, flushing any buffered frames and using `boundary` (when
+	/// known) as the duration boundary for the group's last frame. See [`flush`](Self::flush).
+	///
+	/// `None` is for callers with no boundary to give: the one-frame-per-group sources
+	/// (audio packets, TS sections) that close a group the moment they write it, and
+	/// leave the last frame to be bounded by the next group. Everything that knows where
+	/// its content ends should go through [`cut`](Self::cut) instead.
+	pub(crate) fn close(&mut self, boundary: Option<Timestamp>) -> Result<(), C::Error> {
+		self.flush(boundary)?;
```
kixelated: I don't understand, why not call cut instead? What is `flush`?

rs/moq-mux/src/container/producer.rs
```diff
@@ -153,18 +154,36 @@ impl<C: Container> Producer<C> {
...
-	fn finish_group_at(&mut self, next: Option<Timestamp>) -> Result<(), C::Error> {
-		self.flush(next)?;
+	/// Close the current group, flushing any buffered frames and using `boundary` (when
+	/// known) as the duration boundary for the group's last frame. See [`flush`](Self::flush).
+	///
+	/// `None` is for callers with no boundary to give: the one-frame-per-group sources
+	/// (audio packets, TS sections) that close a group the moment they write it, and
+	/// leave the last frame to be bounded by the next group. Everything that knows where
+	/// its content ends should go through [`cut`](Self::cut) instead.
+	pub(crate) fn close(&mut self, boundary: Option<Timestamp>) -> Result<(), C::Error> {
```
kixelated: I don't like the Option boundary. Also the name is extremely confusing. This should be `finish_group`.

rs/moq-mux/src/container/producer.rs
```diff
@@ -153,18 +154,36 @@ impl<C: Container> Producer<C> {
...
+	/// This is the group boundary: it flushes any buffered frames, pins the group's
+	/// last frame's presentation end at `end`, and closes the group. The next
+	/// [`write`](Self::write) must be a keyframe.
+	///
+	/// A keyframe arriving through [`write`](Self::write) cuts the previous group
+	/// automatically, using its own timestamp as the boundary -- so an explicit `cut`
+	/// is only needed when the publisher knows the content ends somewhere the next
+	/// keyframe won't tell us: before a pause, or any gap where the next group may be
+	/// late or never arrive at all.
+	pub fn cut(&mut self, end: Timestamp) -> Result<(), C::Error> {
```
kixelated: Maybe `cut` should just take an Option<Timestamp> IDK. It's more generic for audio that way.

rs/moq-mux/src/container/legacy/mod.rs
kixelated: Agreed — now loops to the next frame instead. `poll_read` yields a frame or `Done`; "no frames" was a contract wart that leaned on every caller to loop (and `Read::Fragment` means "one moq frame decoded into several", so an empty one was a lie).

js/loc/src/index.ts
kixelated: Reverted for LOC. I also reverted the same change in `rs/moq-mux/src/container/loc/mod.rs` — same reasoning, it implements the same spec. The empty-payload convention stays hang-only. Shout if you wanted only the JS one reverted.

rs/moq-mux/src/container/producer.rs
kixelated: Answering, since this is moot now — `close` is gone and everything calls `cut`.

`flush` is pre-existing: it writes the buffered frames into the open group, backfilling each per-sample duration from the next frame (CMAF needs them to reconstruct DTS). It does not close the group. `close` was just `flush` + `group.finish()`, and `cut` now does that directly.

rs/moq-mux/src/container/producer.rs
kixelated: Removed. `cut` is now the only primitive and does the flush+finish itself — no second method, so nothing to name confusingly.

rs/moq-mux/src/container/producer.rs
kixelated: Done — `cut(Option<Timestamp>)` all the way through (Producer, codec `Import`s, the object-safe traits, `Track`). `None` = "no boundary to give", which is what audio/TS one-frame-per-group callers pass, and what the deprecated `finish_group()` forwards to. That killed the separate `close` too.

kixelated:
Fixed in a1e55b7 — em dashes removed.

Two notes:
- I used a period/comma rather than the suggested `--`: root `CLAUDE.md` says *"No em dashes (—)... Use a period and start a new sentence, or use a comma/parenthesis if the clauses are tightly bound."*
- The finding named one block; I swept the whole PR diff and there were **7** (the `Track::cut` block plus a test doc and comment carried over from #2210). All gone: `git diff origin/main...HEAD | grep '^+' | grep -c '—'` is now 0. Pre-existing em dashes elsewhere in `consumer.rs` are untouched, since they're not this PR's.

PR #2294, opened 2026-07-16
fix(moq-native): accept SNI-less raw QUIC connections

kixelated:
## Problem

The quinn and noq raw-QUIC accept paths reject any connection whose TLS ClientHello carries no server name:

```rust
let host = handshake.server_name.unwrap_or_default(); // "" with no SNI
...
if host.is_empty() {
    return Err(Error::MissingServerName);
}
```

A client dialing a **bare IP address** sends no SNI at all. RFC 6066 §3 forbids IP literals in the `server_name` extension, so rustls omits the extension entirely. The server then sees `handshake.server_name == None`, `host` is empty, and the session is aborted right after the handshake.

This bites anyone making a direct QUIC connection by IP (reported for a drone ↔ cloud link over a raw Tailscale IP, dropping the relay). MagicDNS "worked" only because it supplies a real DNS name that *can* be sent as SNI. The escape hatch `--client-tls-host-name` also works, but requiring either for a plain IP dial is surprising.

## Root cause

Raw QUIC (a moq ALPN with no WebTransport CONNECT) has no in-band request URL, so the TLS SNI is the only authority a client can offer, and it is **optional**. The accept path synthesizes a `moqt://{sni}` URL purely to fit the shared `Session::raw(ConnectRequest, …)` shape, but that URL is never surfaced: `NoqRequest::url()` / `QuinnRequest::url()` return `None` for the raw variant, so the relay routes such sessions to the root path exactly like a URL-less `tcp`/`unix` stream transport. The empty-SNI rejection guards a value that is immediately discarded. An empty host even parses fine (`moqt://` → host `None`), so the guard is the only thing failing the connection.

The **quiche** backend already gets this right, accepting raw QUIC with a hardcoded `moqt://` and no server-name check. This change aligns quinn and noq with it.

## Change

- Drop the `MissingServerName` rejection in `quinn.rs` and `noq.rs`. An empty SNI now yields a hostless `moqt://` request, matching quiche and the URL-less stream transports.
- The `MissingServerName` error variant is retained (unused). Removing a `pub` variant from these `#[non_exhaustive]` enums would be a breaking API change routed to `dev`; keeping it stays non-breaking so the fix can land on `main`. Flagging it here in case you'd prefer to drop it in a follow-up `dev` cleanup.
- When a client *does* present a DNS SNI, the behavior is unchanged (the synthesized URL is still discarded by `url()`, same as before).

## Tests

Added `quinn_raw_quic_no_sni` and `noq_raw_quic_no_sni` regression tests that dial a bare loopback IP (so the client sends no SNI) and assert the broadcast still flows end-to-end. Both fail without the fix (the server rejects with `MissingServerName`, so the client times out) and pass with it. Refactored `backend_test` into a shared `connect_test(scheme, bind, authority, backend)` and bound the client to the server's address family so an IPv4 dial doesn't egress from an IPv6 socket.

## Cross-package sync

No paired updates: this is native connection-acceptance behavior in `moq-native`, not a wire/API/catalog change, and no `/doc` page documents the SNI requirement. Interop matrix untouched.

https://claude.ai/code/session_01NoKJfq8ALde1MKMyvCLE5u

---
_Generated by [Claude Code](https://claude.ai/code/session_01NoKJfq8ALde1MKMyvCLE5u)_

### DIFF
--- a/rs/moq-native/src/noq.rs
+++ b/rs/moq-native/src/noq.rs
@@ -488,9 +488,11 @@ impl NoqRequest {
 				Ok(Self::WebTransport { request, alpns })
 			}
 			alpn if moq_net::ALPNS.contains(&alpn) => {
-				if host.is_empty() {
-					return Err(Error::MissingServerName);
-				}
+				// Raw QUIC carries no in-band request URL like WebTransport's CONNECT, so the TLS
+				// SNI is the only authority the client can offer, and it's optional. A client dialing
+				// a bare IP sends no SNI (RFC 6066 forbids IP literals), leaving `host` empty; the
+				// resulting hostless `moqt://` routes to the root path, exactly like a URL-less stream
+				// transport. `url()` returns `None` for the raw variant either way.
 				let host_str = if host.contains(':') {
 					format!("[{}]", host)
 				} else {
--- a/rs/moq-native/src/quinn.rs
+++ b/rs/moq-native/src/quinn.rs
@@ -483,9 +483,11 @@ impl QuinnRequest {
 				Ok(Self::WebTransport { request, alpns })
 			}
 			alpn if moq_net::ALPNS.contains(&alpn) => {
-				if host.is_empty() {
-					return Err(Error::MissingServerName);
-				}
+				// Raw QUIC carries no in-band request URL like WebTransport's CONNECT, so the TLS
+				// SNI is the only authority the client can offer, and it's optional. A client dialing
+				// a bare IP sends no SNI (RFC 6066 forbids IP literals), leaving `host` empty; the
+				// resulting hostless `moqt://` routes to the root path, exactly like a URL-less stream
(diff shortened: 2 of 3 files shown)


kixelated:
Addressed the `connect_test` 4-arg comment by folding the params into a `ConnectTest` struct (b2d4c12).

Skipped the suggested `quiche_raw_quic_no_sni` test: `quiche_raw_quic` is already `#[ignore]`d due to an unrelated web-transport-quiche bug, so a quiche no-SNI variant would have to be ignored too and wouldn't actually protect anything. quiche also isn't touched by this PR (it already constructs a hostless `moqt://` with no server-name check), so the regression risk here is confined to the quinn/noq paths that the new tests cover. Happy to add an ignored quiche variant for symmetry if you'd prefer.

---

Issue #2265, opened 2026-07-15
export ts over WebSocket: stream framing desyncs (WrongSize/FrameTooLarge) and one bad frame fatally crashes the process


t0ms:
## Summary

Subscribing with `moq export ts` over the **WebSocket** transport (`wss://`) produces a flood of `WrongSize` / `FrameTooLarge` group evictions and then **fatally crashes** the process with:

```
Error: h264: annexb: missing Annex B start code
```

The identical broadcast over **QUIC** (loopback and over a real network) is clean: either full delivery or graceful "skipping slow groups", never `WrongSize` / `FrameTooLarge` / an annexb crash. So this is WebSocket-transport-specific, not a media or usage problem.

There are really two bugs here:

- **B1 (root cause): the WebSocket transport desyncs moq stream framing under load.** The reader ends up reading garbage frame lengths and truncated/mis-aligned payloads. QUIC never does this for the same broadcast.
- **B2 (resilience): `export ts` terminates the whole process on a single malformed frame** instead of dropping the frame/group and continuing, even though the consumer already skips transport errors gracefully.

## Environment

- Client: `moq` CLI (release build), macOS, on a corporate network where UDP/QUIC is throttled, forcing the WebSocket fallback.
- Relay: `moq-relay` on EC2 with both QUIC and the web HTTPS (WebSocket) listener on `:443`.
- Broadcast: a hang broadcast published from `moq import ts` of a looped MPEG-TS capture (H.264 open-GOP, ~10 Mbps). Video track is `0.avc3` (in-band Annex-B), audio `0.mp2`.

## Repro

```
moq --client-tls-disable-verify --client-connect wss://<relay>:443/anon \
  --broadcast cnn.international.emea.loop.hang export ts \
  --latency-max 5s > /tmp/out.ts 2> /tmp/out.log
```

(Forcing `wss://` makes the client use the WebSocket transport directly. The same over `https://`/QUIC does not reproduce.)

## Observed

Client log (heavily abridged):

```
WARN moq_native::websocket: using WebSocket fallback url=wss://.../anon
WARN moq_mux::container::consumer: current group evicted; skipping to next buffered group error=Hang(Moq(WrongSize))
... hundreds of WrongSize evictions, interspersed with ...
WARN moq_mux::container::consumer: current group evicted; skipping to next buffered group error=Hang(Moq(FrameTooLarge))
... and "skipping slow groups" debug lines ...
Error: h264: annexb: missing Annex B start code

Caused by:
[...]

PR #2326, opened 2026-07-17
fix(moq-net): stop racing priority updates against group frame writes

skirsten:
## Summary

- `serve_group` in the lite publisher raced `priority.next()` and `track_priority.changed()` against `next_frame`/`read_chunk`/`write_all` inside `tokio::select!`. The priority queue re-sorts on every inserted group, so with typical audio (one group per 20ms frame) the priority branches fire ~50x/s and each firing drops the in-flight future.
- Root cause of the corruption: cancelling `write_all` mid-frame loses or duplicates bytes after the frame length header was already written. The subscriber's stream is then desynced: it either parses payload bytes as a varint length ("frame too large" errors) or waits forever on a length that never fills (frozen track, no error surfaced).
- Only triggers when writes actually block, i.e. under egress backpressure to a slow or congested subscriber. Fast paths never see it, which is why it hides well.
- Fix: apply priority updates between frames instead of racing them against the writes. A stream that is mid-frame keeps draining at its previous priority until the next frame boundary, a bounded lag.

## Public API changes

None. `pub(super)` internals of `rs/moq-net/src/lite/publisher.rs` only; no wire or `pub` surface change.

## Test plan

- `cargo test -p moq-net` (392 tests pass).
- Load repro against a relay built from this branch: a publisher with production-like traffic (6 Mbps video in 2s groups + one-frame-per-group 20ms audio) and a subscriber behind a congested link, performing repeated join sequences. Before: ~30/100 subscriptions died with "frame too large" (plus silent mid-group stalls); after: 0/100 in the same topology. A same-node (unbackpressured) subscriber sharing the identical publisher/relay was clean in both cases, confirming the corruption was per-viewer egress.
- No unit test included: reproducing requires a transport whose writes block mid-frame while the priority queue churns, which the current test harness cannot express. Happy to add one if there is an accepted pattern for backpressured-transport tests.

(Written by Claude Fable 5)

### DIFF
--- a/rs/moq-net/src/lite/publisher.rs
+++ b/rs/moq-net/src/lite/publisher.rs
@@ -617,20 +617,21 @@ impl<S: web_transport_trait::Session> Publisher<S> {
 				biased;
 				_ = stream.closed() => return Err(Error::Cancel),
 				frame = group.next_frame() => frame,
-				new_pri = priority.next() => {
-					stream.set_priority(new_pri);
-					continue;
-				}
-				Ok(()) = track_priority.changed() => {
-					priority.set_track(*track_priority.borrow_and_update());
-					continue;
-				}
 			};
 
 			let mut frame = match frame? {
 				Some(frame) => frame,
 				None => break,
 			};
+
+			// Apply priority updates only between frames. Racing them against the
+			// encode/write awaits below cancels an in-flight write, which loses or
+			// duplicates bytes after the frame length was already written and
+			// desyncs the stream for the subscriber.
+			if track_priority.has_changed().unwrap_or(false) {
+				priority.set_track(*track_priority.borrow_and_update());
+			}
+			stream.set_priority(priority.current());
 
 			if version.has_track_stream() {
 				let now = i64::try_from(crate::Time::now().as_millis()).unwrap_or(i64::MAX);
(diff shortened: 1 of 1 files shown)


kixelated:
core issue was fixed earlier today: https://github.com/moq-dev/web-transport/pull/323

skirsten:
Revalidated after web-transport#323 landed (qmux 0.4.1 / web-transport-quinn 0.11.11): rebased our fork onto main, dropped this patch, and re-ran the original repro.

Same topology as the original report (publisher + relay co-located, subscriber joining across a jittery cross-region link, 100 join trials each run):

- pre-fix baseline: 31/100 video subscriptions killed (frame-too-large boundary desync + code=0 kills), 15 stalls
- with this PR's serve_group patch: frame-too-large gone, 7/100 code=0 kills remained
- with qmux 0.4.1 and **without** this patch: **0/100 killed** (0 frame-too-large, 0 code=0), 15 stalls, all of the slow-delivery kind consistent with link jitter, none of the silent mid-group desync class

The code=0 join-kills disappearing too suggests they were the same stranded-bytes mechanism (or its #2265 eviction follow-on), not a separate bug. A ws/qmux e2e scenario that failed deterministically with wire-desync + decode-corrupt on qmux 0.4.0 also shows zero integrity failures across repeated runs on 0.4.1.

So the cancel-safety fix fully supersedes this patch; closing it was right.

(Written by Claude Fable 5)

PR #2378, opened 2026-07-17
feat(stats): record cumulative connection/broadcast/viewer time

kixelated:
## Motivation

Stats tracked only cumulative open/closed counter pairs (`sessions`/`sessions_closed`, `announced`/`announced_closed`, etc.). To get connection time from that, a consumer has to sample the live count (`open - closed`) each scrape interval and multiply by the interval. That misses any span that opens and closes between two ticks and carries sampling error, so connection-time billing is guesswork.

## What this does

Adds a lock-free, **time-weighted** `DurationCounter` next to each open/closed pair. Each RAII guard adds its full lifetime to `accrued` on close and its start offset to `open_start` on open (removed on close). A snapshot computes:

```
accrued + open*now - open_start
```

so the total includes the **currently-open tail** exactly. A span still open at snapshot time accrues continuously, so a long-lived session or broadcast bills *within* the period rather than only when it closes, with no per-tick sampling. Magnitude stays bounded by concurrency (not churn): `accrued` can't exceed `max_open * uptime` and `open_start` can't exceed `open * uptime`, so the internal microsecond `u64`s are safe for years.

## Surfaces (all existing read paths)

- **Prometheus `/metrics`**: `moq_relay_{session,broadcast,subscription,viewer}_seconds_total` (labeled by tier/role).
- **`.stats` broadcast JSON**: `announced_seconds` / `subscriptions_seconds` / `broadcasts_seconds` per broadcast, `sessions_seconds` per auth root — so per-tenant/per-root billing works off the wire.
- **Public API**: `CounterTotals` / `SessionTotals` / `StatsSnapshot` gain the seconds fields. Additive only; both counter structs are `#[non_exhaustive]`, so this targets `main`.

Seconds are emitted as **floats**, not microseconds, so a cumulative value can't overflow the `2^53` JSON-safe-integer range at scale (a raw-µs total crosses `2^53` in ~a day at 100k concurrent sessions).

## Behavior notes

- A live-but-idle broadcast now emits a stats frame every tick (its `*_seconds` advances). This is inherent to a continuously-advancing duration; idle-frame skipping still elides fully-drained slots.
- The running total can jitter by a bounded amount under concurrent close/read skew — never negative (clamped at zero), the same slight-upward-bias tradeoff the existing open/closed counters already accept.

## Tests

- New deterministic `DurationCounter` unit tests: open-tail accrual, closed-span accrual, `open_start` netting to zero, and the read-skew clamp.
- Extended the `/metrics` renderer test to assert the new `*_seconds_total` metrics.
- Full moq-net (389) + moq-relay (134) + smoke suites pass; clippy and CI-style rustdoc clean.

## Cross-package sync

- `js/net`: N/A — the `.stats` wire JSON has no browser-client consumer (it's read by external aggregators).
- `doc/concept`: N/A — no stats concept page exists. Docs updated in `doc/bin/relay/{config,index}.md` (and corrected a stale "metrics not yet implemented" line).

(Written by Claude Opus 4.8)

### DIFF
--- a/doc/bin/relay/config.md
+++ b/doc/bin/relay/config.md
@@ -243,13 +243,15 @@ counterpart no traffic can flow, so the entry is dropped:
     "announced": 1, "announced_closed": 0, "announced_bytes": 8,
     "broadcasts": 1, "broadcasts_closed": 0,
     "subscriptions": 5, "subscriptions_closed": 2,
-    "bytes": 12345, "frames": 678, "groups": 9
+    "bytes": 12345, "frames": 678, "groups": 9,
+    "announced_seconds": 42.5, "subscriptions_seconds": 118.2, "broadcasts_seconds": 42.5
   },
   "anon/foo": {
     "announced": 1, "announced_closed": 0, "announced_bytes": 8,
     "broadcasts": 1, "broadcasts_closed": 0,
     "subscriptions": 2, "subscriptions_closed": 0,
-    "bytes": 234, "frames": 12, "groups": 1
+    "bytes": 234, "frames": 12, "groups": 1,
+    "announced_seconds": 5.0, "subscriptions_seconds": 9.7, "broadcasts_seconds": 5.0
   }
 }
 ```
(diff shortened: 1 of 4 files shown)


kixelated:
Closing without merging.

The review surfaced that the continuous-accrual read (`accrued + open*now - open_start`) isn't a consistent snapshot across its three atomics, so a `close()` racing the read can transiently inflate a `*_seconds` value by ~the full relay uptime, which for a cumulative billing counter reads as a spurious spike + reset. A single-writer `active*dt` accrual in the stats ticker would fix that cleanly (and delete all the per-guard duration plumbing).

But stepping back: the relay already publishes the per-root / per-broadcast `open`/`closed` counts on every live tick, so connection-time is just `∫(open - closed) dt`, fully derivable by the billing side. Rather than add a redundant relay-side counter (which also defeats idle-frame dedup on the `.stats` wire), we'll compute the integral in the billing system from what's already exposed.

No relay change needed; `main` is untouched.

(Written by Claude Opus 4.8)

PR #2404, opened 2026-07-19
fix(audio): advertise the Opus output sample rate

### LINKED ISSUE #2400: Safari publisher advertises Opus at the capture rate (44100 after Bluetooth mute/unmute); Safari watchers hard-fail with InternalAudioDecoderCocoa and never recover
## Summary

Safari to Safari audio dies permanently after muting and unmuting the publisher while a Bluetooth headset is the microphone. The published catalog starts advertising the audio rendition as `opus` at `sampleRate: 44100`, and Safari's `AudioDecoder` cannot decode Opus when configured at 44100 (or any non-Opus rate). Every Safari watcher then fails with:

```
[error] EncodingError: InternalAudioDecoderCocoa decoding failed
[error] spawn error InvalidStateError: AudioDecoder is not configured
decode@[native code]
@.../js/watch/src/audio/decoder.ts
```

and audio never recovers, including across watcher reloads, because the publisher keeps advertising 44100 for as long as the capture session stays in that state. Safari to Chrome and Safari to Firefox play the same broadcast fine. Chrome to Safari is fine. Only Safari watchers of a Safari publisher die.

Diagnosis below; every step is measured, not inferred. This is adjacent to but distinct from #2352 / #2368 (catalog rate vs decoded rate mismatch, noise symptom, decode still succeeds) and #2387 (timestamp quantization, crackle, decode succeeds).

## Environment

- macOS 26.5.1 (Mac mini, no built-in microphone), Safari 26.5.
- Microphone: Bluetooth headset (Soundcore Life Q20+), the only input device on this machine.
- Controls run on Chrome 150 (headless).
- moq dev at 44741d5 (`js/publish`, `js/watch` current as of this report).

## The causal chain

Four links, each verified independently.

### 1. Safari's mic capture rate flips between 16000 and 44100 with Bluetooth profile state

A Bluetooth headset runs mic capture over HFP (16 kHz here) but playback-only over A2DP (44.1 kHz, which is also this machine's default `AudioContext` rate). Measured with a standalone page (no moq code) that mimics the publisher exactly: unconstrained `getUserMedia`, read `track.getSettings()`, then `new AudioContext({ sampleRate: settings.sampleRate })`:

- While capture is engaged (HFP): `getSettings().sampleRate = 16000`, realized context rate 16000. Stable across back-to-back re-captures (7/7 cycles).
- After releasing the mic tracks, waiting as little as 2 s, and starting audio playback (which is exactly what a watcher tab does): the next fresh `getUserMedia` on the same device reports `getSettings().sampleRate = 44100`, and the requested `AudioContext` realizes 44100. Reproduced 3/3 with waits of 2 s, 8 s, 15 s.

So the user-visible trigger sequence "mute publisher, reload watcher, unmute publisher" is precisely the flip recipe: mute releases the mic (HFP disengages), the watcher tab plays audio (A2DP engages at 44100), unmute captures fresh and now reports 44100. No device switch is involved; it is the same physical microphone. Capture itself still works: Safari resamples into the 44100 graph and produces valid audio (Chrome watchers play it correctly).

### 2. The publisher publishes the capture rate verbatim, and mute/unmute recaptures from scratch

- `js/publish/src/element.ts` wires the mute toggle into the `enabled` inputs of `Source.Microphone` and `Audio.Encoder`, so each mute stops the tracks (`microphone.ts`, `Microphone#run` cleanup) and each unmute is a brand-new unconstrained `getUserMedia`.
- `js/publish/src/audio/encoder.ts` `#runSource` creates `new AudioContext({ latencyHint: "interactive", sampleRate })` with `sampleRate = override ?? settings.sampleRate`, then records `worklet.context.sampleRate` as the captured rate after the first frame. `#createConfig` puts that value into the catalog as `sampleRate`, and `toEncoderConfig` configures the `AudioEncoder` with it.
- Nothing anywhere normalizes the advertised Opus rate to an Opus-native rate (8/12/16/24/48 kHz). Whatever the capture session realized, including 44100, goes on the wire.

This matches the production logs: catalogs flip from `"sampleRate":16000` to `"sampleRate":44100` across unmute cycles, and the watcher errors begin at the first 44100 catalog.

### 3. Safari's AudioDecoder hard-fails on any Opus decoder config with a non-native rate, while claiming to support it

Standalone WebCodecs matrix (no moq code, mono, 1 s of sine, 20 ms `AudioData` chunks; decoder configured with `codec: "opus"`, no description, same as `js/watch`):

Safari 26.5:

| test | result |
|---|---|
| `AudioDecoder.isConfigSupported({codec:"opus", sampleRate:44100})` | `supported: true` |
| encode at 44100 (`AudioEncoder`) | succeeds; its own `metadata.decoderConfig` says `sampleRate: 48000`, and the emitted dOps declares 48000 |
| decode 44100-encoded chunks, decoder configured at 44100 | 0 frames, `EncodingError: InternalAudioDecoderCocoa decoding failed` |
| decode the same 44100-encoded chunks, decoder configured at 48000 | all 51 frames decode, output rate 48000 |
| decode 48000-encoded chunks, decoder configured at 44100 | 0 frames, same `EncodingError` |
| encode and decode at 16000 | works (16000 is a native Opus rate) |
| encode and decode at 48000 | works |

Chrome 150, same page:

| test | result |
|---|---|
| decode anything with decoder configured at 44100, including Safari's own 44100-encoded chunks | all frames decode, output rate 48000 |

Conclusions:

- The trigger is purely the decoder's configured `sampleRate: 44100`. The packets are valid Opus (Safari itself decodes them at 48000; Chrome decodes them at any configured rate).
- `isConfigSupported` returns true for the exact config that then fails at runtime, which is why the watcher's rendition support probe (`Source#runSupported`) passes and the track is selected.
- Chrome's decoder ignores the configured rate and outputs 48 kHz, which is why Safari to Chrome plays fine and why the bug is invisible on every non-Safari watcher.
- Safari's own encoder resamples to 48 kHz internally and even reports `decoderConfig.sampleRate: 48000` in its output metadata; the pipeline does not use that metadata and advertises the capture rate instead.

### 4. In the watcher, one decode error becomes a permanent, unrecovered failure storm

`js/watch/src/audio/decoder.ts`, `#runLegacyDecoder` (the path used for these broadcasts, `container.kind: "legacy"`):

- The `AudioDecoder` `error` callback is `(error) => console.error(error)`. Per WebCodecs, a fatal decode error also closes the decoder. That is the single `EncodingError: InternalAudioDecoderCocoa decoding failed` line.
- The decode loop keeps running and calls `decoder.decode(chunk)` without checking `decoder.state` (the CMAF path checks `decoder.state === "closed"` before decoding; the legacy path does not). The next frame therefore throws `InvalidStateError: AudioDecoder is not configured` synchronously, which propagates uncaught out of the loop into `effect.spawn`, producing the `spawn error` line.
- Nothing reacts to the failure: no reconfigure, no rendition deselection, no state change. A new decoder is only created when the catalog config changes, and the catalog keeps saying 44100, so every subscribe/reselect cycle (each mute/unmute flips the rendition in and out of the catalog) repeats the same error pair. This matches the observed `subscribe start / error pair / subscribe close` pattern per toggle, and the failure surviving watcher reloads.

## Reproduction

In-product (matches the original report):

1. Safari publisher with a Bluetooth headset mic, Safari watcher on the same broadcast, audio playing fine (catalog advertises 16000).
2. Mute the publisher.
3. Reload the watcher page (or otherwise keep audio playback active on the output device).
4. Unmute the publisher.
5. Catalog now advertises `opus` at 44100; the Safari watcher logs the two errors above and audio is dead, including after further reloads. Chrome and Firefox watchers of the same broadcast keep playing.

Standalone Safari repro for link 3 (paste into any page's console on Safari):

```js
const rate = 44100;
const chunks = [];
const enc = new AudioEncoder({
  output: (c) => { const d = new Uint8Array(c.byteLength); c.copyTo(d); chunks.push(d); },
  error: (e) => console.error("encoder error:", e),
});
enc.configure({ codec: "opus", sampleRate: rate, numberOfChannels: 1, bitrate: 32000 });
const pcm = new Float32Array(rate / 5);
for (let i = 0; i < pcm.length; i++) pcm[i] = 0.4 * Math.sin((2 * Math.PI * 440 * i) / rate);
enc.encode(new AudioData({ format: "f32-planar", sampleRate: rate, numberOfFrames: pcm.length, numberOfChannels: 1, timestamp: 0, data: pcm }));
await enc.flush();
const dec = new AudioDecoder({
  output: (f) => console.log("decoded", f.sampleRate),
  error: (e) => console.error("decoder error:", e),
});
dec.configure({ codec: "opus", sampleRate: rate, numberOfChannels: 1 });
for (const d of chunks) dec.decode(new EncodedAudioChunk({ type: "key", timestamp: 0, data: d }));
await dec.flush();
```

Verified on Safari 26.5: with `rate = 44100` this prints `decoder error: EncodingError: InternalAudioDecoderCocoa decoding failed` (and `flush()` rejects with the same error). With `rate = 48000` or `16000` every chunk prints `decoded 48000` / `decoded 16000`. On Chrome it decodes at every rate.

## Original watcher logs

<details>
<summary>Safari watcher during the mute/unmute cycles and after reload</summary>

```
[debug] received catalog hang me.hang {..."audio":{"renditions":{"audio":{"codec":"opus","container":{"kind":"legacy"},"sampleRate":16000,"numberOfChannels":2,"bitrate":64000,"jitter":20}}}...}
... many mute/unmute cycles at 16000, audio fine ...
[debug] received catalog hang me.hang {..."sampleRate":44100,"numberOfChannels":2,"bitrate":64000...}
[debug] subscribe start: id=31 broadcast=me.hang track=audio
[error] EncodingError: InternalAudioDecoderCocoa decoding failed
[error] spawn error InvalidStateError: AudioDecoder is not configured
decode@[native code]
@http://localhost:5173/@fs/.../js/watch/src/audio/decoder.ts:226:19
... repeats for every subsequent audio subscribe, and after a full watcher reload ...
```

</details>

kixelated:
Fixes #2400.

## Summary

- Normalize capture to a supported Opus output rate before constructing the AudioContext, keeping capture, framing, AudioEncoder configuration, and the advertised catalog rate identical.
- Validate AudioEncoder output metadata against the catalog before publishing chunks.
- Normalize legacy non-native Opus catalogs in watchers and stop decode loops after a fatal decoder error.

The root cause was that the publisher advertised the original capture graph rate as an AudioDecoderConfig. Safari can realize 44100 Hz after a Bluetooth mute/unmute cycle, while its Opus decoder requires one of the native Opus output rates. The watcher support probe accepted 44100 but decoding closed the decoder, and the legacy loop continued calling decode on the closed instance.

## Public API changes

- Adds `Util.Opus.normalizeSampleRate(sampleRate?)`, an additive helper that selects 8, 12, 16, 24, or 48 kHz.
- No breaking API, catalog schema, container, or wire-format changes.
- No Rust or documentation sync is needed because this is browser WebCodecs rate selection and compatibility handling.

## Test plan

- `nix develop --command just js fix`
- `nix develop --command just js ci`
- `nix develop --command just js test`
- Regression tests cover 44100 to 48000 normalization, catalog advertisement, native Opus rates, AAC preservation, legacy catalog compatibility, and closed-decoder handling.
- Safari Bluetooth mute/unmute hardware verification remains pending.

(Written by GPT-5)

### DIFF
--- a/js/hang/src/util/opus.ts
+++ b/js/hang/src/util/opus.ts
@@ -1,0 +1,9 @@
+const SAMPLE_RATES = [8_000, 12_000, 16_000, 24_000, 48_000] as const;
+
+/** Normalize to an Opus output rate, rounding up when possible and capping at 48 kHz. */
+export function normalizeSampleRate(sampleRate?: number): number {
+	if (sampleRate === undefined) return 48_000;
+	if (!Number.isFinite(sampleRate) || sampleRate <= 0) throw new Error("invalid Opus sample rate");
+
+	return SAMPLE_RATES.find((candidate) => candidate >= sampleRate) ?? 48_000;
+}
--- a/js/publish/src/audio/encoder.ts
+++ b/js/publish/src/audio/encoder.ts
@@ -118,7 +118,7 @@ export class Encoder {
 	muted: Signal<boolean>;
 	/** Linear gain applied before encoding, where 1 is unity. */
 	volume: Signal<number>;
-	/** Override the capture sample rate in Hz. Defaults to the track's own rate. */
+	/** Override the capture sample rate in Hz. Opus normalizes this to a supported output rate. */
 	sampleRate: Signal<number | undefined>;
 	/** Override the captured channel count. Defaults to the track's requested count. */
 	channelCount: Signal<number | undefined>;
@@ -195,7 +195,10 @@ export class Encoder {
 
 		const settings = source.track.getSettings();
 		const overrideSampleRate = effect.get(this.sampleRate);
-		const sampleRate = overrideSampleRate ?? settings.sampleRate;
+		const requestedSampleRate = overrideSampleRate ?? settings.sampleRate;
+		const codec = normalizeCodec(effect.get(this.codec));
+		const sampleRate =
+			codec.mime === "opus" ? Util.Opus.normalizeSampleRate(requestedSampleRate) : requestedSampleRate;
 
 		// macOS misreports a mono mic as stereo: getSettings().channelCount is undefined and
 		// MediaStreamAudioSourceNode.channelCount defaults to 2, so the graph carries (and Opus
@@ -365,11 +368,23 @@ export class Encoder {
 				const kind: Kind = source ? normalizeSource(source).kind : "auto";
 				const encoderConfig = toEncoderConfig(config, kind, this.#opusOptions(effect));
 				const framer = createFramer(config);
+				let failed = false;
 
 				const encoder = new AudioEncoder({
-					output: (frame) => {
+					output: (frame, metadata) => {
 						if (frame.type !== "key") {
 							throw new Error("only key frames are supported");
+						}
+
+						const outputSampleRate = metadata?.decoderConfig?.sampleRate;
+						if (outputSampleRate !== undefined && outputSampleRate !== config.sampleRate) {
+							failed = true;
+							const error = new Error(
+								`audio encoder output rate ${outputSampleRate} does not match catalog rate ${config.sampleRate}`,
+							);
+							console.error("encoder error", error);
+							track.close(error);
+							return;
 						}
 
 						this.#out.stats.update((stats) => ({
@@ -395,6 +410,8 @@ export class Encoder {
 				encoder.configure(encoderConfig);
 
 				effect.event(worklet.port, "message", (event: Event) => {
+					if (failed) return;
+
 					const captured = (event as MessageEvent<Capture.AudioFrame>).data;
 					const channelCount = captured.channels.length;
 					if (!channelCount) return;
--- a/js/hang/src/util/index.ts
+++ b/js/hang/src/util/index.ts
@@ -1,5 +1,5 @@
 /**
- * Miscellaneous helpers for the hang media layer: AAC config, hex encoding,
+ * Miscellaneous helpers for the hang media layer: audio codec config, hex encoding,
  * and the libav/WebCodecs polyfill.
  *
  * @module
@@ -9,3 +9,4 @@ export * as Aac from "./aac";
 export * as Hacks from "./hacks";
 export * as Hex from "./hex";
 export * as Libav from "./libav";
+export * as Opus from "./opus";
--- a/js/hang/src/util/opus.test.ts
+++ b/js/hang/src/util/opus.test.ts
@@ -1,0 +1,30 @@
+import { describe, expect, it } from "bun:test";
+import { normalizeSampleRate } from "./opus";
+
+describe("normalizeSampleRate", () => {
+	it("defaults to fullband Opus", () => {
+		expect(normalizeSampleRate()).toBe(48_000);
+	});
+
+	it("preserves native Opus rates", () => {
+		for (const sampleRate of [8_000, 12_000, 16_000, 24_000, 48_000]) {
+			expect(normalizeSampleRate(sampleRate)).toBe(sampleRate);
+		}
+	});
+
+	it("rounds up to preserve the input bandwidth", () => {
+		expect(normalizeSampleRate(11_025)).toBe(12_000);
+		expect(normalizeSampleRate(22_050)).toBe(24_000);
+		expect(normalizeSampleRate(32_000)).toBe(48_000);
+		expect(normalizeSampleRate(44_100)).toBe(48_000);
+	});
+
+	it("caps rates above fullband Opus", () => {
+		expect(normalizeSampleRate(96_000)).toBe(48_000);
+	});
+
+	it("rejects invalid rates", () => {
+		expect(() => normalizeSampleRate(0)).toThrow("invalid Opus sample rate");
+		expect(() => normalizeSampleRate(Number.NaN)).toThrow("invalid Opus sample rate");
+	});
+});
--- a/js/publish/src/audio/encoder.test.ts
+++ b/js/publish/src/audio/encoder.test.ts
@@ -1,5 +1,6 @@
 import { expect, mock, spyOn, test } from "bun:test";
 import { Signal } from "@moq/signals";
+import type { StreamTrack } from "./types";
 
 // The encoder pulls the capture processor in as a `?worklet` blob URL, which the bun test loader can't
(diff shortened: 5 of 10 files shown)


js/hang/src/util/opus.ts
```diff
@@ -0,0 +1,9 @@
+const SAMPLE_RATES = [8_000, 12_000, 16_000, 24_000, 48_000] as const;
+
+/** Normalize to an Opus output rate, rounding up when possible and capping at 48 kHz. */
+export function normalizeSampleRate(sampleRate?: number): number {
+	if (sampleRate === undefined) return 48_000;
+	if (!Number.isFinite(sampleRate) || sampleRate <= 0) throw new Error("invalid Opus sample rate");
+
+	return SAMPLE_RATES.find((candidate) => candidate >= sampleRate) ?? 48_000;
```
kixelated, line 8: won't this always return 8000...?

js/publish/src/audio/encoder.ts
```diff
@@ -195,7 +195,10 @@ export class Encoder {
 
 		const settings = source.track.getSettings();
 		const overrideSampleRate = effect.get(this.sampleRate);
-		const sampleRate = overrideSampleRate ?? settings.sampleRate;
+		const requestedSampleRate = overrideSampleRate ?? settings.sampleRate;
+		const codec = normalizeCodec(effect.get(this.codec));
+		const sampleRate =
+			codec.mime === "opus" ? Util.Opus.normalizeSampleRate(requestedSampleRate) : requestedSampleRate;
```
kixelated, line 201: Do we need to normalize for AAC as well?

js/publish/src/audio/encoder.ts
```diff
@@ -395,6 +410,8 @@ export class Encoder {
 				encoder.configure(encoderConfig);
 
 				effect.event(worklet.port, "message", (event: Event) => {
+					if (failed) return;
```
kixelated, line 413: Do we really need a separate variable for this? IDK track.isClosed at the least, but maybe we could even tear down the effect?

PR #2416, opened 2026-07-20
feat(token): scope signing keys to publish/subscribe paths

kixelated:
## Summary

A JWK can now carry an immutable `scope`, capping what tokens it may sign. A key generated for `project` + `put: ["live"]` can only ever mint tokens publishing at or beneath `project/live`, so a leaked key cannot be talked into signing more than it already could. Widening means minting a new key.

- `Scope { root, put, get }` embedded in the JWK, mirroring claim semantics. Enforced in **both** `sign` and `verify`, so a scope cannot be stripped by re-signing elsewhere and the relay re-checks on the way in.
- A key with no `scope` is unrestricted, so existing keys keep working.
- `moq-token generate --root/--publish/--subscribe` writes the scope into both halves of an asymmetric pair.
- `to_public()` preserves the scope, so a verify-only public key enforces it too.

Containment resolves both sides against their own root before comparing, so `root: "demo"` + `put: ["room"]` and `put: ["demo/room"]` are treated identically. Matching is segment-aware (`live` does not cover `lively`) and roles are independent (a publish-only scope never authorizes a subscribe grant).

## Scope of this PR

Purely additive to `moq-token` and `@moq/token`. **No relay changes.**

An earlier revision also reworked `moq-relay`'s root check so a pid-rooted token would authorize a connection dialed through a vanity name. That was dropped: it reintroduced the first-segment splicing removed in https://github.com/moq-dev/moq/pull/1581#discussion_r3336909958, and it baked in the assumption that an alias may differ from the dialed path only in its leading segment, which is exactly the flexibility that comment was preserving.

The consequence is a real limitation, documented in `doc/bin/relay/auth.md`: because the token root is checked against the dialed path, a scope must be anchored at whichever name its tokens are rooted at. Anchor at the stable id and the tokens work on stable-id paths; anchor at a vanity name and they break when that name changes. Reconciling that belongs in its own change.

## Public API changes

Additive only, hence `main`:

- **New** `moq_token::Scope`, `Key::with_scope`, `Key::scope`, `Error::UselessScope`, `Error::ScopeExceeded`.
- **New** `@moq/token`: `ScopeSchema`, `Scope`, `scopeAllows`, optional `scope` on `KeySchema`.
- **New** `moq-token generate --root/--publish/--subscribe`.
- No renamed, removed, or signature-changed items.

## Test plan

- `cargo test -p moq-token` — 119 pass. Covers cross-form equivalence, segment boundaries, role independence, escaping above the scope root, per-path requirement, and sign/verify symmetry.
- `bun test js/token/src` — 88 pass, including interop fixtures generated by the real `moq-token` CLI: a scoped JWK, an in-scope token Rust signed, and the exact grant pair Rust rejects.
- `just check` clean; `RUSTDOCFLAGS=-D warnings cargo doc` clean.

Not exercised against a live relay. The moq.pro dashboard side (surfacing scopes in the key UI, respecting them in token forms) is a follow-up that needs this published first.

(written by Opus 4.8)

### DIFF
--- a/doc/bin/relay/auth.md
+++ b/doc/bin/relay/auth.md
@@ -47,6 +47,38 @@ moq-token generate --algorithm ES256 --out-dir ./private/ --public-dir ./keys/
 
 A random key ID is generated if `--id` is not specified.
 
+### Scope a Key
+
+Keys can embed immutable publish and subscribe limits. Every token signed or
+verified with a scoped key must stay within those limits. Rotate the key if its
+scope needs to change.
+
+```bash
+# This key may publish below project/live and subscribe below project/watch.
+moq-token generate \
+  --root project \
+  --publish live \
+  --subscribe watch \
+  --out private.jwk \
+  --public public.jwk
+```
+
+The scope is stored in both halves of an asymmetric JWK:
+
+```json
+{
+  "scope": {
+    "root": "project",
+    "put": ["live"],
+    "get": ["watch"]
+  }
+}
+```
+
+Existing JWKs without `scope` remain unrestricted for backwards compatibility.
+The library rejects the entire token when any requested role or path exceeds the
+key scope; it never silently intersects permissions.
+
 ### Configure the Relay
 
 Single key (simplest):
@@ -120,6 +152,8 @@ The JWT payload contains these claims:
 | `exp` | Expiration time (Unix timestamp) |
 | `iat` | Issued-at time (Unix timestamp) |
 
+`get` is named that way because `sub` is a reserved JWT claim.
+
 The `exp` claim is enforced for the whole session, not just at connect time. The relay closes the connection once `exp` passes, so a client must reconnect with a fresh token to continue. The same applies to mTLS: the connection is closed when the client certificate's `notAfter` is reached.
 
 ### Path Matching
@@ -138,9 +172,9 @@ path boundaries, so `foo` grants `foo` and `foo/bar` but never `foobar`.
 
 | root | put | get | Can publish | Can subscribe |
 |------|-----|-----|-------------|---------------|
-| `demo` | `my-stream` | `""` | `demo/my-stream` | `demo/*` |
-| `rooms/123` | `alice` | `""` | `rooms/123/alice` | `rooms/123/*` |
-| `""` | `""` | `""` | Everything | Everything |
+| `demo` | `["my-stream"]` | `[""]` | `demo/my-stream` | `demo/*` |
+| `rooms/123` | `["alice"]` | `[""]` | `rooms/123/alice` | `rooms/123/*` |
+| `""` | `[""]` | `[""]` | Everything | Everything |
 
 ### Connection Path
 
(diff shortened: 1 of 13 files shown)


kixelated:
Self-review pass (five lenses: CLAUDE.md compliance, bug hunt, git history, prior review feedback, comment/doc accuracy). Findings and what changed, pushed in 0f41031b2:

**Fixed**

1. `ClaimsSchema.root` was required in JS while Rust has `skip_serializing_if = "String::is_empty"`, so `@moq/token` could not parse any token minted with the CLI's default `--root ""`. Pre-existing on `main`, but this branch fixed the identical pattern for `Scope.root`, so it belongs here. Regression test uses a token minted by the real Rust CLI and fails without the fix.
2. `canonicalize` compared raw strings, so `/demo//room` was rejected where `Path` collapses the slash. Now normalized through `Path`, reusing `next_part` instead of a hand-rolled splitter.
3. `canonicalize` could not tell "one segment, no tail" from "empty path", so an alias of `""` against `/demo` slipped through and handed `finalize` an empty `check_root`, whose prefix check matches any token root. Now an explicit depth mismatch.
4. The claims table in `doc/bin/relay/auth.md` documented `pub`/`sub` as single suffixes; the wire format is `put`/`get` lists. It contradicted the scope section added just above it.
5. Documented the new `generate` flags in `doc/lib/rs/crate/moq-token.md` and `doc/lib/rs/index.md` per Cross-Package Sync.
6. `js/token`'s CLI now parses the scope through `ScopeSchema` rather than casting.

**Open, needs a decision before this merges**

`Auth::canonicalize` reintroduces first-segment splicing, which was explicitly rejected in https://github.com/moq-dev/moq/pull/1581#discussion_r3336909958 ("get rid of `rewrite_first_segment`... The API should return the full root to use... This gives us more flexibility in the future"). It also hard-codes the assumption that the alias may differ from the dialed path only in its leading segment, which removes exactly that flexibility.

The underlying constraint is real: a key scope is immutable, so it can only be anchored at the stable pid, while the dashboard currently mints vanity-rooted claims. Alternatives that keep the mapping server-side: have the relay check the token root against the alias and move the dashboard to pid-rooted tokens (needs a transition story for already-issued tokens), or extend the auth-API response to state the acceptable token root.

Not merging until that is settled.

(written by Opus 4.8)

kixelated:
Update on the open question from the previous comment: resolved by removing it from this PR rather than deciding it here.

`main` picked up a `dev` merge in the meantime, so this branch was rebased onto it. During that rebase the relay change was dropped entirely and `rs/moq-relay/src/auth.rs` resolved to `main`'s version. This PR is now purely additive to `moq-token` and `@moq/token`, with no relay changes and no contested design in it.

The rebase also made several earlier items moot, since `dev` had already landed them on `main`:

- `path.rs` / `path.ts` now exist upstream; the copies here were dropped in favor of them.
- `ClaimsSchema.root` was already defaulted to `""` upstream, with its own empty-root interop test, so that fix and its duplicate test were dropped.
- The claims table in `doc/bin/relay/auth.md` was already corrected to `put`/`get` upstream.
- `verify()` lost its `path` argument and `encode`/`decode` were deprecated; the tests here were updated to `sign`/`verify`.

The vanity-vs-pid root question is unresolved and now tracked as separate work. Its practical consequence is documented in `doc/bin/relay/auth.md`: a scope must be anchored at whichever name the tokens it signs are rooted at.

The Sourcery failure is a false positive. Its scanner flagged the JWT fixtures in `js/token/src/interop.test.ts` as leaked credentials; they are throwaway vectors generated by the `moq-token` CLI, in the same file and pattern as the existing `RUST_HS256_TOKEN` / `RUST_EDDSA_TOKEN` fixtures.

(written by Opus 4.8)

PR #2431, opened 2026-07-21
feat(js/net): expose connection transport stats

wrangelvid:
We are still new to this library, but I found myself wanting to have the browser drive its own rate control. However @moq/net only exposes a send bandwidth estimate and the PROBE RTT. The Rust side just grew a full ConnectionStats snapshot, while js was lagging behind.

This adds an optional stats() in a similar shape to Rust. While at it I resolved the TODO to prefer the transport RTT over PROBE, and it turned out that Reload also kept a gracefully closed connection in established forever, so status (and now stats) would report a dead session. Hence a settled closed now tears down and reconnects, matching what the Rust reconnect loop already does.

### DIFF
--- a/js/net/src/connection/established.ts
+++ b/js/net/src/connection/established.ts
@@ -4,6 +4,7 @@ import type { Bandwidth } from "../bandwidth.ts";
 import type * as broadcast from "../broadcast.ts";
 import type * as Path from "../path.ts";
 import type * as Time from "../time.ts";
+import type { ConnectionStats } from "./stats.ts";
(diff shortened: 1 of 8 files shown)


kixelated:
Closing in favor of #2442 and #2441 for now. I don't think a signal is better than the on-demand await, but until browsers actually implement getStats() properly, I thiiink it's the least confusing way to expose stuff like RTT.

kixelated:
I'm having second thoughts, let me mock up exposing getStats() again. Updating the signal every 100ms is easier, but less accurate.

PR #2432, opened 2026-07-21
feat(moq-native): expose a congestion control knob for the quinn backend

wrangelvid:
We are curious about adapting the bitrate of video encoders given the send rate estimates and dug into the congestion controllers. CUBIC yields sawtooths which the encoder chases, and every probe cycle ends in a loss burst on live video. BBR on the other hand tracks the actual delivery rate instead, which we prefer. I think everyone should choose BBR, but since I am new here and don't want to crash land in politics I figured we'd just give the user the option to select which congestion controller to use. Unset keeps quinn's CUBIC default, and the non-quinn backends reject the knob at init instead of silently ignoring it.

### DIFF
--- a/rs/moq-native/src/noq.rs
+++ b/rs/moq-native/src/noq.rs
@@ -57,6 +57,10 @@ pub enum Error {
 	#[error("failed to resolve bind address")]
 	ResolveBind(#[source] std::io::Error),
 
+	/// noq ships BBRv3 only, so a congestion control selection cannot be honored.
+	#[error("the noq backend always uses BBRv3; drop --*-quic-congestion-control or use the quinn backend")]
+	CongestionControlUnsupported,
+
 	/// The URL has no host to connect to.
 	#[error("invalid DNS name")]
 	InvalidDnsName,
@@ -182,11 +186,17 @@ pub(crate) struct NoqClient {
 
 impl NoqClient {
 	pub fn new(config: &ClientConfig) -> Result<Self> {
+		let quic = config.quic.resolve();
+		// noq ships BBRv3 only; fail fast instead of silently running a different controller.
+		if quic.congestion_control.is_some() {
+			return Err(Error::CongestionControlUnsupported);
+		}
+
 		let socket = crate::bind::udp(config.bind).map_err(Error::BindSocket)?;
 
 		let mut transport = noq::TransportConfig::default();
 		transport.congestion_controller_factory(Arc::new(noq::congestion::Bbr3Config::default()));
-		apply_transport(&mut transport, config.quic.resolve());
+		apply_transport(&mut transport, quic);
 		let transport = Arc::new(transport);
 
 		// There's a bit more boilerplate to make a generic endpoint.
@@ -362,9 +372,15 @@ pub(crate) struct NoqServer {
 
 impl NoqServer {
 	pub fn new(config: ServerConfig) -> Result<Self> {
+		let quic = config.quic.resolve();
+		// noq ships BBRv3 only; fail fast instead of silently running a different controller.
+		if quic.congestion_control.is_some() {
+			return Err(Error::CongestionControlUnsupported);
+		}
+
 		let mut transport = noq::TransportConfig::default();
 		transport.congestion_controller_factory(Arc::new(noq::congestion::Bbr3Config::default()));
-		apply_transport(&mut transport, config.quic.resolve());
+		apply_transport(&mut transport, quic);
 		let transport = Arc::new(transport);
 
 		let provider = crate::crypto::provider();
@@ -587,5 +603,39 @@ impl noq::ConnectionIdGenerator for ServerIdGenerator {
 
 	fn cid_lifetime(&self) -> Option<Duration> {
 		None
+	}
+}
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+
+	/// noq always runs BBRv3, so a congestion control selection must fail at
+	/// init rather than silently running a different controller.
+	#[tokio::test]
+	async fn congestion_control_selection_is_rejected() {
+		let client_config = ClientConfig {
+			quic: crate::quic::Client {
+				congestion_control: Some(crate::quic::CongestionControl::Cubic),
+				..Default::default()
+			},
+			..Default::default()
+		};
+		assert!(matches!(
+			NoqClient::new(&client_config),
+			Err(Error::CongestionControlUnsupported)
+		));
+
+		let server_config = ServerConfig {
+			quic: crate::quic::Server {
+				congestion_control: Some(crate::quic::CongestionControl::Bbr),
+				..Default::default()
+			},
+			..Default::default()
+		};
+		assert!(matches!(
+			NoqServer::new(server_config),
+			Err(Error::CongestionControlUnsupported)
+		));
 	}
 }
--- a/rs/moq-native/src/noq.rs
+++ b/rs/moq-native/src/noq.rs
@@ -1,6 +1,7 @@
 //! The noq QUIC backend, used for both WebTransport (`https://`) and raw QUIC (`moqt://`, `moql://`).
 
 use crate::client::ClientConfig;
+use crate::quic::CongestionControl;
 use crate::quic::Resolved;
 use crate::quic::ServerId;
 use crate::server::ServerConfig;
@@ -31,6 +32,19 @@ fn apply_transport(transport: &mut noq::TransportConfig, quic: Resolved) {
 	if let Some(gso) = quic.gso {
 		transport.enable_segmentation_offload(gso);
 	}
+
+	// Unlike quinn, this backend defaults to BBR rather than noq's own CUBIC.
+	transport.congestion_controller_factory(congestion_factory(
+		quic.congestion_control.unwrap_or(CongestionControl::Delay),
+	));
+}
+
+/// The noq controller factory for a congestion control family. noq's BBR is v3.
+fn congestion_factory(family: CongestionControl) -> Arc<dyn noq::congestion::ControllerFactory + Send + Sync> {
+	match family {
+		CongestionControl::Loss => Arc::new(noq::congestion::CubicConfig::default()),
+		CongestionControl::Delay => Arc::new(noq::congestion::Bbr3Config::default()),
+	}
 }
 
 /// Errors specific to the noq QUIC backend.
@@ -56,10 +70,6 @@ pub enum Error {
 	/// The configured bind address could not be resolved.
 	#[error("failed to resolve bind address")]
 	ResolveBind(#[source] std::io::Error),
-
-	/// noq ships BBRv3 only, so a congestion control selection cannot be honored.
-	#[error("the noq backend always uses BBRv3; drop --*-quic-congestion-control or use the quinn backend")]
-	CongestionControlUnsupported,
 
 	/// The URL has no host to connect to.
 	#[error("invalid DNS name")]
@@ -186,17 +196,10 @@ pub(crate) struct NoqClient {
 
 impl NoqClient {
 	pub fn new(config: &ClientConfig) -> Result<Self> {
-		let quic = config.quic.resolve();
-		// noq ships BBRv3 only; fail fast instead of silently running a different controller.
-		if quic.congestion_control.is_some() {
-			return Err(Error::CongestionControlUnsupported);
-		}
-
 		let socket = crate::bind::udp(config.bind).map_err(Error::BindSocket)?;
 
 		let mut transport = noq::TransportConfig::default();
-		transport.congestion_controller_factory(Arc::new(noq::congestion::Bbr3Config::default()));
-		apply_transport(&mut transport, quic);
+		apply_transport(&mut transport, config.quic.resolve());
 		let transport = Arc::new(transport);
 
 		// There's a bit more boilerplate to make a generic endpoint.
@@ -372,15 +375,8 @@ pub(crate) struct NoqServer {
 
 impl NoqServer {
 	pub fn new(config: ServerConfig) -> Result<Self> {
-		let quic = config.quic.resolve();
-		// noq ships BBRv3 only; fail fast instead of silently running a different controller.
-		if quic.congestion_control.is_some() {
-			return Err(Error::CongestionControlUnsupported);
-		}
-
 		let mut transport = noq::TransportConfig::default();
-		transport.congestion_controller_factory(Arc::new(noq::congestion::Bbr3Config::default()));
-		apply_transport(&mut transport, quic);
+		apply_transport(&mut transport, config.quic.resolve());
 		let transport = Arc::new(transport);
 
 		let provider = crate::crypto::provider();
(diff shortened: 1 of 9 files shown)


Crypt0XRP0:
This looks good, might solve the issues I'm having on android when streaming on poor cellular, as i have to set profiles at the moment.

kixelated:
So BBR is absolutely better than Cubic for live streaming, and used to be the default. But the Quinn implementation is kind of bugged based on some packet traces I've been so I disabled it.

I would love to flip the switch but I need testing done. This is definitely a step in the right direction of course.

wrangelvid:
>  But the Quinn implementation is kind of bugged based on some packet traces I've been so I disabled it.

I remember reading that in one of your blog posts or an issue somewhere ◡̈

rs/moq-relay/src/config.rs
```diff
@@ -254,6 +254,52 @@ preferred_v6 = "[2001:db8::1]:443"
...
+	/// `congestion_control` fields on the quic sections of
+	/// `moq-native::ServerConfig` / `ClientConfig`. The fields must stay
+	/// `Option<CongestionControl>` so a TOML-selected algorithm survives the
+	/// CLI re-parse when no `--*-quic-congestion-control` flag is passed.
+	#[test]
+	fn cli_does_not_clobber_toml_congestion_control() {
+		let _guard = CONGESTION_ENV_LOCK.lock().unwrap_or_else(|e| e.into_inner());
+		// SAFETY: CONGESTION_ENV_LOCK ensures no other test in this binary
+		// touches these env vars concurrently.
+		unsafe {
```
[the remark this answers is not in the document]
wrangelvid: Pretty sure it doesn't compile if we remove this.

rs/moq-native/src/noq.rs
```diff
@@ -182,11 +186,17 @@ pub(crate) struct NoqClient {
 
 impl NoqClient {
 	pub fn new(config: &ClientConfig) -> Result<Self> {
+		let quic = config.quic.resolve();
+		// noq ships BBRv3 only; fail fast instead of silently running a different controller.
+		if quic.congestion_control.is_some() {
```
kixelated: NoQ supports Cubic, NewReno, and BBRv3.  Erroring here is definitely wrong.

rs/moq-native/src/quic.rs
```diff
@@ -41,6 +41,22 @@ impl std::str::FromStr for ServerId {
...
 }
 
+/// The congestion control algorithm for a QUIC connection.
+///
+/// Honored by the quinn backend only; noq, quiche, and iroh reject a
+/// selection at init (noq always uses BBRv3).
+#[derive(Clone, Copy, Debug, PartialEq, Eq, clap::ValueEnum, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "kebab-case")]
+#[non_exhaustive]
+pub enum CongestionControl {
```
kixelated, line 52: I would expose Loss (Cubic or Reno) and Delay (BBR) as two options. We can have more specific versions if you want, like BBRv1/2/3, but most people will just care about the family not the specific algorithm.

kixelated (commented):
I'll have claude made some changes, one sec.

rs/moq-native/src/quic.rs
```diff
@@ -41,6 +41,24 @@ impl std::str::FromStr for ServerId {
...
 }
 
+/// The congestion control family for a QUIC connection.
+///
+/// This selects a family rather than a named algorithm because each backend ships a
+/// different generation: BBRv1 on quinn, BBRv2 on quiche, BBRv3 on noq and iroh. A
+/// `Bbr` variant would promise more than any one backend delivers.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, clap::ValueEnum, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "kebab-case")]
+#[non_exhaustive]
```
kixelated, line 51: Note we can add specific congestion controllers in the future, like if a backend supports multiple versions of BBR and its important to specify which one. Or if we add a real-time CC like GCC in the future.

rs/moq-native/src/quic.rs
wrangelvid: I like the API change! Feels simpler now!

t0ms:
Follow-up on #2432: flipping the relay to --server-quic-congestion-control delay (BBR) fixes the collapse we saw under CUBIC with network impairment — MoQ now holds full-rate/0-CC through 10% loss, 25% reorder and a 120ms+1% WAN profile, matching SRT. Only residual is heavy netem jitter (delay 60ms±30ms), but with no rate that's an independent-per-packet reordering storm rather than in-order jitter — so likely a test artefact. Re-checking with a rate-serialised/correlated model before flagging it; has anyone actually seen reorder-driven loss-detection back-off on real paths?

t0ms:
Extract from my latest test runs:

| Condition (`netem`) | CUBIC (quinn) | BBRv1 (quinn) | BBR2 (quiche) | BBR3 (noq) | SRT |
|---|---|---|---|---|---|
| Baseline | 100 % | 100 % | 100 % | 100 % | 100 % |
| loss 0.5 % | 97 % | 100 % | 99 % | ~74 %‡ | 100 % |
| loss 1 % | 95 % | 100 % | 98 % | 93 % | 100 % |
| **loss 2 %** | **53 %** | **100 %** | 95 % | **100 %** | 100 % |
| **loss 5 %** | **31 %** | **100 %** | 93 % | **100 %** | 100 % |
| **loss 10 %** | **13 %** | **100 %** | **47–70 %** | **97 %** | 100 % |
| bursty loss 3 % (25 % corr) | 99 % | 99 % | 100 % | 100 % | 100 % |
| **reorder 20 ms 25 %** | **20 %** | **99 %** | **25–31 %** | **100 %** | 100 % |
| jitter 60 ± 30 ms — **non-ordered** (reorders) | **2 %** | **7–13 %** | **2 %** | **8–71 %**§ | 100 % |
| jitter 60 ± 30 ms — **in-order** (`slot`, FIFO) | — | 97 % | 100 % | 100 % | 95 %¶ |
| duplicate 1 % | 99 % | ~100 % | 98 % | 100 % | 100 % |
| **combined WAN 120 ms + 1 % loss** | **14 %** | **97 %** | 96 % | 93 % | 100 % |

Clean baselines: CUBIC 9.67, BBRv1 9.59, BBR2 (quiche) 9.28, BBR3 (noq) 9.63, SRT 9.96 Mbps.
Ranges (e.g. 47–70 %, 25–31 %, 8–71 %) are the spread across two runs — see the noise
caveat below. ‡ BBR3's loss-0.5 % figure was low on **both** runs (72 %, 78 %) yet 1/2/5 %
were ≥ 93 % (non-monotonic) → most likely a join-window artefact, not a CC effect.
§ non-ordered jitter under BBR3 was **unstable** — one run 71 %, the next 8 %; the
reordering/HOL residual is not reliably fixed by any BBR generation. ¶ netem `slot`
(in-order jitter) mildly rate-caps, so both transports sit ~5 % below line rate; the point
is they are *comparable and near-full* — in-order jitter does not collapse any MoQ variant.

One caveat worth flagging: these are 1–2 runs per condition over a single home path, so the two-run ranges (and the BBR3 0.5 %/jitter noise) are indicative, not averaged — the relative ordering is clear, but I'd want repeated runs before quoting any single BBR2/BBR3 percentage as definitive.

kixelated:
> Extract from my latest test runs:
> 
> Condition (`netem`)	CUBIC (quinn)	BBRv1 (quinn)	BBR2 (quiche)	BBR3 (noq)	SRT
> Baseline	100 %	100 %	100 %	100 %	100 %
> loss 0.5 %	97 %	100 %	99 %	~74 %‡	100 %
> loss 1 %	95 %	100 %	98 %	93 %	100 %
> **loss 2 %**	**53 %**	**100 %**	95 %	**100 %**	100 %
> **loss 5 %**	**31 %**	**100 %**	93 %	**100 %**	100 %
> **loss 10 %**	**13 %**	**100 %**	**47–70 %**	**97 %**	100 %
> bursty loss 3 % (25 % corr)	99 %	99 %	100 %	100 %	100 %
> **reorder 20 ms 25 %**	**20 %**	**99 %**	**25–31 %**	**100 %**	100 %
> jitter 60 ± 30 ms — **non-ordered** (reorders)	**2 %**	**7–13 %**	**2 %**	**8–71 %**§	100 %
> jitter 60 ± 30 ms — **in-order** (`slot`, FIFO)	—	97 %	100 %	100 %	95 %¶
> duplicate 1 %	99 %	~100 %	98 %	100 %	100 %
> **combined WAN 120 ms + 1 % loss**	**14 %**	**97 %**	96 %	93 %	100 %
> Clean baselines: CUBIC 9.67, BBRv1 9.59, BBR2 (quiche) 9.28, BBR3 (noq) 9.63, SRT 9.96 Mbps. Ranges (e.g. 47–70 %, 25–31 %, 8–71 %) are the spread across two runs — see the noise caveat below. ‡ BBR3's loss-0.5 % figure was low on **both** runs (72 %, 78 %) yet 1/2/5 % were ≥ 93 % (non-monotonic) → most likely a join-window artefact, not a CC effect. § non-ordered jitter under BBR3 was **unstable** — one run 71 %, the next 8 %; the reordering/HOL residual is not reliably fixed by any BBR generation. ¶ netem `slot` (in-order jitter) mildly rate-caps, so both transports sit ~5 % below line rate; the point is they are _comparable and near-full_ — in-order jitter does not collapse any MoQ variant.
> 
> One caveat worth flagging: these are 1–2 runs per condition over a single home path, so the two-run ranges (and the BBR3 0.5 %/jitter noise) are indicative, not averaged — the relative ordering is clear, but I'd want repeated runs before quoting any single BBR2/BBR3 percentage as definitive.

Just a word of caution when measuring congestion control, it's very easy to have misleading results. 100% is a red flag, as the easiest way to get 100% is zero congestion control.

You should try limiting the maximum rate. ex. When shaping a connection to 6Mb/s, the measured result should be something like 5.2Mb/s.

And how you shape the connection matters too. Ideally you want to simulate bufferbloat, queuing packets for up to X ms before dropping them. That's closer to how actual networks work.

t0ms:
Fair caution, and it applies squarely to my numbers — thanks, this is the right pushback.

To be upfront about the rig: that matrix offered ~10 Mbps onto a path with ~292 Mbps of headroom (raw TCP, same path), with netem adding loss/reorder/jitter but no rate limit. So you're right that it isn't really a congestion-control benchmark — it's a non-congestive impairment test, and those 100% cells are "source fully delivered on a fat pipe," not "CC behaved." Same caveat applies to the SRT column (spare bandwidth + ARQ). I'll relabel the writeup accordingly.

The one thing I'd still stand behind from it is the narrow, known result: loss-based CUBIC reads random loss as congestion and collapses cwnd (53%/31%/13% at 2/5/10% loss), while the rate-based controllers don't. Useful, but not a reason to flip the default on its own — agreed that needs the real test.

So I'll add the regime you're describing: a shaped bottleneck (e.g. cake/tbf ~6 Mbps) reporting delivered as % of the cap (expecting ~5.2, not 100), plus a bufferbloat profile (deep FIFO queuing up to X ms before drop, vs CoDel/CAKE) to actually see latency-under-load — and a 2-flow fairness run, since BBRv1 fairness is a separate worry. I'll also pair it with an adaptive source rather than my CBR file replay, so the encoder can chase the send-rate estimate instead of just overflowing the queue.

Two questions so I reproduce the right thing: (1) is the quinn-BBR issue you saw in the traces written up anywhere / tied to a specific condition (bufferbloat? probe RTT? loss threshold?) — I'd like to target it directly rather than hunt; (2) any preferred shaping profile (target rate, buffer depth / RTT) you'd consider representative? Happy to run repeated trials and post traces.

kixelated:
With a 10Mb/s source, try rate limiting the bitrate to 5Mb/s with a a 100ms RTT and 500ms queue depth before dropping.

A good congestion control algorithm will approach 5Mb/s delivered with RTT closer to 100ms. BBR should do both of those (maximize throughput and minimize latency), while CUBIC will have an RTT closer to 500ms. (bufferbloat).

Don't test reordering. It doesn't happen on the internet, otherwise TCP and QUIC would break. 

Random loss and jitter can occur but it's best to avoid simulating them because it removes signal. The best congestion control in the face of random loss is zero congestion control, because backing off during congestion literally does nothing. It's a noob trap causing people to think 10x FEC is the solution to congestion.

t0ms:
@kixelated — a correction to what I said I'd test, plus the first-pass numbers.

**Not ABR.** I said I'd pair this with an adaptive encoder chasing the send-rate estimate — that's the contribution/OTT case (what @wrangelvid opened this for), not mine. I'm on primary distribution: a fixed, non-congestion-responsive **VBR** feed that has to run **24/7/365, permanently**. That's the governing constraint — I'm optimising for a feed that stays complete and reconstructable indefinitely, not for a good short run, so an *intermittent* or *rare* failure mode is disqualifying: over continuous operation it will recur and eventually line up with a transient.

**First pass (your profile, as a failure mode).** Fixed ~10 Mb/s feed into a 5 Mb/s cap (2:1 over-subscription) — for a provisioned trunk that's an outage, not an operating point, but the failure modes are the useful part. Two-namespace `tc` rig (`htb` + `netem` + 500 ms `bfifo`); standing RTT via ICMP sharing the bottleneck queue; 2–3 replicates per controller.

| CC (backend) | Delivered (% of 5 Mb/s cap) | CC errors | Queuing delay p50 | Notes |
|---|---|---|---|---|
| CUBIC (quinn) | 65–81 % | 0 | ~518 ms | stable full bloat |
| CUBIC (quiche) | 62–63 % | 0 | ~564 ms | stable full bloat |
| BBRv1 (quinn) | 45–76 % | 0 | 226 / **591** / 234 ms | **bimodal** — occasional full bloat |
| BBRv2 (quiche) | 67–69 % | 0 | ~225–290 ms | stable, ~½ CUBIC |
| BBRv3 (noq) | 11–13 % | 0 | ~595 ms | bloats *and* starves — the #768 overflow |
| SRT (reference) | 90 % | ~4300 | 583 ms | most bytes, but a damaged (unreconstructable) stream |

Reading it for a permanent, reconstructable feed:
- MoQ sheds **whole groups** and stays CC-clean — a reconstructable, if thinned, TS — where SRT keeps 90 % of bytes but emits ~4300 continuity errors: more data, but damaged. For reconstruction I'd take clean-but-thinned.
- **BBRv1 (quinn) is bimodal** — ~230 ms on most runs, a full 591 ms bloat on one. For a permanent feed that's the real problem, not a latency nuisance: an intermittent bloat spike past the receive buffer turns into dropped groups, and over 24/7 operation it *will* happen. BBRv2 (quiche) held ~½ CUBIC on every run. This is also what made me think of your "quinn BBR is kind of bugged" note — is that characterised / tied to a condition? Happy to capture qlog on the bloated runs.
- **BBRv3 (noq) is out** — it bloats and collapses delivery to ~12 % (#768), and a subtract-overflow that aborts the relay is, for a permanent feed, an outage by definition.

**The ask.** I can pin the controller via your knob, so I don't need a particular default — I need to know which one to *rely on* for a permanent fixed-rate trunk. In this first pass that's BBRv2 on quiche (stable, complete, no aborts). So: (1) is BBRv2-on-quiche a first-class choice you'd be comfortable running continuously, or is quiche a second-class backend for you? (2) is the quinn-BBRv1 intermittent bloat the bug you mentioned, and something you'd expect to fix — in which case quinn/BBRv1 becomes the pick?

**Requalified plan (fixed-rate VBR):** on a *provisioned* path — a transient-congestion / competing-flow run scored on whether the feed stays complete and how fast it recovers; a coexistence/fairness run; the `fq_codel`/`cake` AQM counterfactual; two provisioning points to find where each controller starts *dropping content*; and, because the feed is permanent, a long-duration soak to catch drift / leak / rare aborts that a short run misses.

PR #2435, opened 2026-07-21
feat(js/net): add AbortSignal support to connect()

wrangelvid:
While we were benchmarking relays, we sometimes sat behind a broken network and connect() hung forever. This adds an optional signal to ConnectProps so pending connects can be aborted (similar to how fetch can be aborted).

## Summary

- `ConnectProps.signal` aborts a connection attempt with the signal's reason. An already-aborted signal rejects before anything opens; aborting after the connection is established has no effect. `AbortSignal.timeout(ms)` gives you a deadline.
- The abort reaches every in-flight transport, so a WebTransport/WebSocket race, a caller-supplied `transport`, and a session mid-SETUP all tear down rather than leaking.
- `Reload` passes its per-run teardown signal (`effect.abort`) into each attempt, so a rerun or `close()` cancels the pending connect instead of leaving it running behind the new one. `ReloadProps` excludes `signal`, since `Reload` supplies its own.
- `connect()` folds into one path with a never-aborting default signal, rather than two versions of the same function.
- Drops the copy of the SETUP/version-dispatch block that `connect` kept alongside `connectTransport` (~100 lines): the post-race path now delegates to it, which also gives the caller-supplied-transport path the ALPN debug log it was missing.

## Public API changes

- **Added**: `ConnectProps.signal?: AbortSignal`. Additive, so this targets `main`.
- **Changed**: `ReloadProps` is now `Omit<ConnectProps, "signal">`. Passing `signal` to `Reload` was never meaningful; it is now a type error.
- No behavior change when `signal` is omitted.

## Test plan

- New `connect.test.ts`: already-aborted signal never opens a transport; mid-connect abort rejects with the reason and closes the transport; abort after success is inert; a 12-hop microtask sweep of the window between winning the transport race and returning, asserting connect never hands back a closed connection; and a no-signal connect still works.
- `reload.test.ts`: closing mid-connect aborts the pending attempt.
- `just js test` green across all packages; `just js check` clean.

Depends on #2446 (merged), which fixed the `Effect` lifecycle bug this PR's `Reload` changes ran into: a spawn task resuming after its run was torn down read the *next* run's scope, so `effect.cleanup` fired one teardown too late and `effect.abort` reported not-aborted.

(written by Opus 4.8)


js/net/src/connection/connect.ts
```diff
@@ -93,17 +100,67 @@ const websocketWon = new Set<string>();
  * Establishes a connection to a MOQ server.
  *
  * @param url - The URL of the server to connect to
+ * @param props - Connection options
  * @returns A promise that resolves to a Connection instance
  */
-export async function connect(url: URL, props?: ConnectProps): Promise<Established> {
+export function connect(url: URL, props?: ConnectProps): Promise<Established> {
+	const signal = props?.signal;
```
kixelated: Why not make a new AbortSignal here (that never gets resolved) if signal is undefined? That way we don't need two versions of the same function.

wrangelvid: Good call!

PR #2436, opened 2026-07-21
fix(moq-token): default missing key_ops to sign and verify

wrangelvid:
In our initial prototype we generated our JWKs with standard tooling and none of it wrote key_ops, so we had to patch every key with jq. I believe RFC 7517 makes it optional, where absence means unrestricted.

While testing that, two things fell out. Key sets could now pick a public-only key for signing and fail, so signing selection requires actual private material. And the oct k field had a lenient empty default that the key_ops default suddenly made reachable, letting a bare {"kty":"oct","alg":"HS256"} parse into a working key with an empty secret. k is now required with the same 32-byte minimum js/token already enforces.

(Second paragraph written by Claude Fable 5)

### DIFF
--- a/rs/moq-token/src/key.rs
+++ b/rs/moq-token/src/key.rs
@@ -54,10 +54,9 @@ pub enum KeyMaterial {
 	/// <https://datatracker.ietf.org/doc/html/rfc7518#section-6.4>
 	#[serde(rename = "oct")]
 	OCT {
-		/// The secret key as base64url (unpadded).
+		/// The secret key as base64url (unpadded). Must be at least 32 bytes once decoded.
 		#[serde(
 			rename = "k",
-			default,
 			serialize_with = "serialize_base64url",
 			deserialize_with = "deserialize_base64url"
 		)]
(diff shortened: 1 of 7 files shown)


rs/moq-token/src/key.rs
```diff
@@ -172,6 +172,13 @@ pub struct Key {
 	pub(crate) encode: OnceLock<EncodingKey>,
 }
 
+fn sign_verify() -> HashSet<KeyOperation> {
+	[KeyOperation::Sign, KeyOperation::Verify].into()
+}
+
+// Match js/token's 32-byte minimum for HMAC secrets.
+const MIN_OCT_SECRET_BYTES: usize = 32;
```
kixelated: Is this needed? I'd rather avoid generating these keys in the first place, rather than erroring when loading keys.

wrangelvid: This is mostly a regression guard. Without this a truncated file like {"kty":"oct","alg":"HS256"} now parses into a working key with an empty secret and the relay would accept forgeable tokens. The required key_ops used to block that by accident.
I thought since we make key_ops optional that we'd want to preserve the minimum, but happy to relax it to rejecting an empty k if you'd prefer that.

PR #2434, opened 2026-07-21
feat(mux): preserve the DVB service layer across the TS round-trip

t0ms:
Closes #2433.

## Summary

- The media-aware TS demux/remux dropped the DVB service layer. The SDT (service name, provider, type, original network ID) and NIT never entered the catalog, and export synthesized a fresh PAT/PMT with a hardcoded transport stream ID (`1`), service number (`1`), and PMT PID (`0x1000`). A TS input came out the other side stripped of its service identity, with a different PMT PID.
- Import now captures the transport/service identity (TSID, service number, PMT PID) from the PAT and carries the SDT Actual (`table_id` 0x42) and NIT Actual (`table_id` 0x40) sections **verbatim** in a new `mpegts` catalog `Service` record. The SI PIDs (SDT 0x0011, NIT 0x0010) are intercepted ahead of the mpeg2ts reader, the same way SCTE-35 private sections already are.
- Export rebuilds the PAT/PMT from the preserved identity (original TSID, service number, PMT PID) and re-emits the SDT and NIT byte-for-byte. A media-only source (no service record) still synthesizes a minimal identity, so nothing regresses.
- TDT/TOT (a live wall-clock) and EIT (EPG) are intentionally out of scope: they are dynamic, not static identity, and verbatim re-emission would ship a stale clock. Noted as a follow-up.

## Public API changes

- `rs/moq-mux` `ts::Service` (new `pub` struct, `#[non_exhaustive]`): the DVB service record (`transport_stream_id`, `service_id`, `pmt_pid`, verbatim `sdt`/`nit`). Re-exported from `ts`.
- `rs/moq-mux` `ts::Mpegts` gains an optional `pub service: Option<Service>` field (additive; `#[serde(default, skip_serializing_if)]`).

All additive; no existing `pub` item renamed, removed, or signature-changed. Based on the `dev` batch, but the change itself is semver-additive.

## Cross-Package Sync

The `mpegts` catalog extension is internal to `rs/moq-mux`; it is not mirrored in `js/hang` (which parses the base hang catalog, not this TS-specific section) and is not an IETF-draft wire format, so no `js/` or `drafts/` rows apply. `doc/bin/cli.md` (the `import ts` / `export ts` reference) is updated in this PR.

## Test plan

- New unit test `catalog::service_roundtrip`: the `Service` record serializes/round-trips, base64 SDT present, absent NIT omitted.
[...]

Issue #2433, opened 2026-07-21
feat(moq-mux): carry the DVB service layer (SDT + service identity) through the media-aware TS catalog


t0ms:
## Summary

The media-aware TS lane (`moq import ts` → relay → `moq export ts`) now round-trips the *elementary* layer faithfully—all ES at original PIDs, PMT descriptors, native SCTE-35, DTS—but it still drops the **DVB service layer**: SDT (service name/provider/type), NIT, TDT/TOT, and it renumbers the PMT PID and regenerates the TSID.

This is a **catalog-mapping gap**, not a byte-fidelity one: the missing metadata is small, mostly static, and reconstructible on export from a few fields carried in the `mpegts` catalog section. This issue proposes to close it as its own scoped task.

It is deliberately **narrower than the opaque lane (#1861)** and **distinct from the opaque-lane TR 101 290 checks (#1838)**. This is about making the *media-aware* remux emit a correct service identity + SDT, not about byte-for-byte carriage of PCR/CC/CBR/CAS.

## Why this hasn't been covered yet

The DVB service layer has been mentioned in several issues but always folded into a larger point, so it never got a dedicated home:

- **#1799** (contribution profile) lists SDT/EIT/TDT/TOT/NIT loss—but as motivation for the opaque lane (Option B), implying the media-aware lane cannot carry it.
- **#1835** wired the `mpegts` verbatim catalog (SCTE-35, teletext, AC-3, private streams, PMT descriptors) through the CLI, but only at the **PMT level**. SDT/NIT/TDT and service identity were never in scope.
- **#1861** (opaque lane) correctly notes SI/SDT/PMT PID/NIT/TDT cannot survive a byte-for-byte remux, but bundles the **service identity** (small static metadata) with genuine byte-fidelity requirements (PCR cadence, CC, CBR stuffing, CAS), even though they are very different implementation problems.
- **#1838** (TR 101 290) treats SDT/NIT/EIT/TDT checks as opaque-lane only. That is true today, but only because the media-aware lane does not currently carry them.
- **#1836** (DTS) is the equivalent fidelity improvement on the elementary layer.

Net result: the service layer is missing, but every related issue either stopped at the PMT (#1835) or deferred everything to the opaque lane (#1799/#1861/#1838). The **service identity + SDT mapping is tractable in the media-aware lane today**, and this issue isolates that work.

## Findings (tested on `dev`)

**Build:** `github.com/moq-dev/moq` `dev @ e3576465`

Includes:
[...]

PR #2440, opened 2026-07-21
refactor(mux): carry the TS service layer as opaque SI sections

kixelated:
An alternative shape for #2434, opened separately so the two can be diffed. **Built directly on @t0ms's two commits** (`feat(mux): preserve the DVB service layer across the TS round-trip` and `test(mux): cover multi-packet DVB SI section reassembly`), which are unchanged here. The third commit is the only delta.

The diagnosis in #2434 is right and the end-to-end validation there (a live CNN International feed through a real publisher -> relay -> subscriber, with matching SHA-1s on the SDT and NIT) stands. This changes where the section bytes live, not whether they are preserved.

## What differs

| | #2434 | here |
|---|---|---|
| Carriage | `Service::sdt`, `Service::nit` | `si: BTreeMap<u16, Si>`, opaque sections keyed by PID |
| Cost of the next table | 2 constants + 2 branches + a field | one row in `SI_PIDS` |
| Multi-section table | not representable | `sections: Vec<Bytes>` |
| Repeats | first section wins, permanently | upsert by section identity |
| Cadence | the PSI interval (500ms) | per-PID interval (SDT 2s, NIT 10s) |

## Why

**Two named fields do not generalize.** `sdt` and `nit` each need a PID constant, a `table_id` constant, and a branch on both the import and export side. BAT, ATSC PSIP, or a private operator table would each cost another of all four. Splitting on what export actually *understands* removes that: the transport/service identity stays parsed, because rebuilding a consistent PAT/PMT requires it, and everything else is opaque bytes re-emitted on the PID it arrived on. An unrecognized table now round-trips exactly like a recognized one.

**A PID carries a set, not a single table.** PID 0x0011 carries SDT Actual *and* the BAT; a multi-service SDT spans several sections that are all simultaneously current. `Option<Bytes>` structurally cannot hold that, so a real MPTS SDT would have been truncated to its first section.

**Repetition is not mutation.** The first-wins latch meant a service rename or a `running_status` change was pinned to whatever arrived first. Sections are now deduped by `(table_id, table_id_extension, section_number)`, so a revision replaces in place and a sibling section is added alongside. The dedupe runs *before* the catalog's write lock is taken, because that lock marks the catalog modified even on an unchanged write, and SI repeats every couple of seconds.

**Repetition cadence is transport policy, not source data.** #2434 emits SDT and NIT on the PSI cadence, which is 4x more often than DVB requires for the SDT and 20x for the NIT. Harmless at ~1KB each, bad the moment a larger table joins the same path. Each PID now carries its own interval, taken from the spec maximum (ETSI TS 101 211) rather than the source's observed cadence, which is a property of that multiplexer's bitrate shaping and means nothing after a remux. It is a hint: absent for an unknown PID, which falls back to the PSI cadence rather than being dropped.

## Notes

- `si` is keyed by PID, with the key serialized as a decimal string (`"17"`) via `serde_with`'s `DisplayFromStr`. A plain integer-keyed map serializes fine and then fails to parse: the catalog deserializes via `serde_json::Value`, which does not coerce string map keys back to integers. Four existing tests caught that; `DisplayFromStr` makes the string key explicit on both sides. The map (rather than a list with a `pid` field) also makes a duplicate entry for one PID unrepresentable, and removes the chance of a key disagreeing with the field it duplicates.
- TDT/TOT and EIT stay out of scope, unchanged from #2434, but they are now a policy decision in `SI_PIDS` rather than a shape limitation.
- Retargeted to `main`. `dev` is currently 0 commits ahead of `main`, so this is purely additive `pub` API per the branch-targeting rules.

## Public API

- `ts::Service` becomes `ts::Program`, loses `sdt`/`nit`, and renames `service_id` to `program_number`. It now holds only the three PAT-parsed identity fields, which is the MPEG *program*; DVB's "service" is a synonym that does not travel to ATSC or ISDB, so the record is named for the role. `Mpegts::service` becomes `Mpegts::program`.
- `ts::Si` is new (`sections`, `interval`) with `Si::upsert`, plus `Mpegts::si`. `interval` is a `Duration` serialized as bare milliseconds via `DurationMilliSeconds`, matching the hang catalog's `jitter` field, with a test locking that wire shape (without the adapter it silently regresses to serde's `{secs, nanos}`).

Nothing in #2434 has shipped, so this is a reshape of an unreleased surface rather than a break.

## Test plan

- `cargo test -p moq-mux`: 401 tests + 2 doctests pass. `cargo clippy -p moq-mux --all-targets -- -D warnings` and `RUSTDOCFLAGS="-D warnings" cargo doc -p moq-mux` clean.
- `catalog::si_upsert_dedupes_by_section_identity` (new): a byte-identical repeat is not a change, a revision replaces in place, a sibling `section_number` is added alongside.
- `export_test::service_layer_survives_roundtrip` (adapted from #2434): now injects the synthetic NIT **twice**, so the repeat-collapses path is actually exercised. The original assertion I first wrote here was vacuous, since `bbb.ts` contains exactly one SDT packet.
- `due_uses_elapsed_duration` (extended): a 2s SI interval is not due when the 500ms PSI would be.
- `catalog::program_and_si_roundtrip` asserts the PID key lands as a string; `service_layer_survives_roundtrip` is the load-bearing check, since it drives a full import -> export -> import through the real `Value`-based catalog path.
- Not re-run: the live-feed end-to-end validation from #2434. The section bytes take the same verbatim path, but the cadence change is not covered by the unit tests.

(written by Opus 4.8)

### DIFF
--- a/rs/moq-mux/src/container/ts/catalog.rs
+++ b/rs/moq-mux/src/container/ts/catalog.rs
@@ -4,9 +4,10 @@
 //! back to MPEG-TS that doesn't belong in the codec-neutral media configs: one
 //! entry per track (its original PID and PMT descriptors), a `verbatim` carriage
 //! record for every elementary stream we don't decode (SCTE-35, teletext, DVB
-//! subtitles, private data, ...), and the program-level PMT descriptors. Demuxed
-//! media tracks keep their codec config in the base `video`/`audio` sections; only
-//! their MPEG-TS identity lands here.
+//! subtitles, private data, ...), the program-level PMT descriptors, and the DVB
+//! service layer ([`Service`]: transport/service identity plus the SDT/NIT tables).
+//! Demuxed media tracks keep their codec config in the base `video`/`audio`
+//! sections; only their MPEG-TS identity lands here.
 
 use std::collections::BTreeMap;
 
@@ -16,6 +17,15 @@ use serde_with::base64::Base64;
 use serde_with::serde_as;
 
 use crate::catalog::hang::CatalogExt;
+
+/// PID of the DVB Network Information Table (NIT).
+pub(super) const NIT_PID: u16 = 0x0010;
+/// PID of the DVB Service Description Table (SDT).
+pub(super) const SDT_PID: u16 = 0x0011;
+/// `table_id` of the NIT for the actual network (NIT Actual).
+pub(super) const NIT_ACTUAL_TABLE_ID: u8 = 0x40;
+/// `table_id` of the SDT for the actual transport stream (SDT Actual).
+pub(super) const SDT_ACTUAL_TABLE_ID: u8 = 0x42;
 
 /// The `mpegts` catalog section.
 ///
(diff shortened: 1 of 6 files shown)


rs/moq-mux/src/container/ts/catalog.rs
```diff
@@ -34,15 +51,120 @@ pub struct Mpegts {
...
+/// so they are held together and replaced individually as each is re-signaled.
+#[serde_as]
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Default)]
+#[serde(rename_all = "camelCase")]
+#[non_exhaustive]
+pub struct Si {
+	/// The current sections, in the order they were first seen. Each is a complete
+	/// section including its header and CRC, re-emitted byte-for-byte.
+	#[serde_as(as = "Vec<Base64>")]
+	pub sections: Vec<Bytes>,
```
kixelated, line 118: Maybe one day this could be a separate track. But bleh it's fine.

kixelated:
@t0ms I think this is a little bit more generic. Maybe try it?

t0ms:
This looks good, it's a better shape and supersedes https://github.com/moq-dev/moq/pull/2434, it addresses https://github.com/moq-dev/moq/issues/2433.

The TDT/TOT and EIT gap remain out of scope for the time being, to be addressed if further testing shows this is a definitive issue.

PR #2458, opened 2026-07-23
feat(watch): apply video presentation rotation

shermerL:
## Summary

- Apply clockwise catalog video rotation in the watch canvas renderer while preserving horizontal flip behavior.
- Rotate frame-derived fallback display dimensions to match the presented orientation.
- Repaint the current frame when catalog rotation or flip metadata changes.
- Cover quarter-turn, flip, normalization, and reactive repaint behavior with focused tests.

## Root cause

The renderer read presentation metadata with `peek()` inside the paint callback. Because the render effect did not subscribe to the catalog signal, a rotation or flip update could leave the current frame painted with stale presentation metadata until another frame arrived.

## Public API changes

None. The presentation helpers remain internal to `@moq/watch`.

## Test plan

- `nix develop --command just ci`
- Original contributor verification covered Android camera publishing to the web watch client, desktop Chrome/Safari/Firefox, and mobile Chrome/Safari.

The publish controls and demo test selector were split into draft PR #2507 so this receiver support can land independently.

(Written by GPT-5)


js/publish/src/element.ts
```diff
@@ -65,6 +66,14 @@ function parsePreview(value: string | null): Preview.Mode {
 	return "source";
 }
 
+function parseRotation(value: string | null): number | undefined {
```
kixelated: Do we need this, outside of testing? Would an application ever need to set rotation manually?

js/watch/src/video/renderer.ts
```diff
@@ -1,6 +1,7 @@
 import { Time } from "@moq/net";
 import { Effect, type Getter, getter, type Inputs, type Readonlys, readonlys, Signal } from "@moq/signals";
 import type { Decoder } from "./decoder";
+import { canvasPresentationTransform } from "./presentation";
```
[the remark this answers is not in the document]
kixelated: Fixed in `0741f3da5`. `#runRender` now tracks the catalog with `effect.get(...)` and passes the captured video metadata into the animation-frame callback. The new renderer regression test updates rotation after the initial paint and verifies that a second paint uses the rotated transform. Quarter-turn and flip geometry also have focused coverage.

(Written by GPT-5)

js/publish/src/element.ts
kixelated: Addressed by splitting the publisher controls and demo-only selector into draft PR #2507. PR #2458 now contains only receiver-side watch support.

(Written by GPT-5)

PR #2469, opened 2026-07-23
feat(moq-net): linger a broadcast across an ungraceful source loss

### LINKED ISSUE #2459: `moq export` terminates with `json: dropped` on relay restart / session drop instead of resuming (reconnect loop stays alive but the sink task is fatal)
## Environment

- `moq` / `moq-relay` **0.8.7**, commit `5eaf99bc` (branch `feat/mux-ts-dvb-service-layer`).
  Behaviour is in generic CLI wiring (`rs/moq-cli/src/main.rs`), so `main` is very
  likely affected too.
- macOS (Darwin 25.5.0), loopback, `--client-quic-gso=false --server-quic-gso=false`.
- Media-aware lane: `moq import ts` publisher, `moq export ts` subscriber, negotiated
  `moq-lite-05`.

## Expected behaviour

`moq export` already spawns the auto-reconnecting session handle
(`net.client(...).consume(origin)`), and the local `Origin` is persistent and
re-subscribes on each new session. So when a relay restarts (or any transient session
drop occurs), the exporter should **ride out the outage and resume writing** once the
session re-establishes — the same way `moq import` survives a restart (its publisher
reconnects and re-announces automatically). A relay restart should be a bounded gap,
not a fatal error.

## Observed behaviour

The instant the MoQ session drops, the `moq export ts` **process exits** with:

```text
Error: json: dropped

Caused by:
    dropped
```

The transport reconnect loop logs that it is *about to* reconnect, but the process is
already terminating:

```text
2026-…  WARN moq_native::reconnect: session closed, reconnecting url=https://localhost:4443/
2026-…  WARN moq_native::reconnect: reconnecting after backoff url=https://localhost:4443/ delay=1s
2026-…  WARN moq_net::model::track: track::Producer dropped without finish() or abort() track=catalog.json
2026-…   INFO moq_net::lite::session: session terminated
Error: json: dropped
```

Net effect for a broadcast subscriber: **a relay restart kills every `moq export ts`
consumer**, and the output stream does not resume on its own; an external supervisor
must relaunch the process. The publisher on the same drill reconnects and re-announces
correctly, so the failure is subscriber-side only.

## Reproduction steps

```bash
# 1. relay
moq-relay demo/relay/localhost.toml --server-quic-gso=false &

# 2. exporter to a file (disable-verify so the cert regenerated on restart is accepted)
moq --client-connect https://localhost:4443 --client-tls-disable-verify \
    --client-quic-gso=false --broadcast rec.hang export ts > out.ts &

# 3. publisher (looped, paced)
tsp -I file CNNiEMEA2.ts --infinite -P regulate --pcr-synchronous -O file - \
 | moq --client-connect https://localhost:4443 --client-tls-disable-verify \
       --client-quic-gso=false --broadcast rec.hang import ts &

# 4. let it flow a few seconds, then restart the relay:
pkill -f moq-relay
#   (optionally lower --client-quic-idle-timeout to ~3s on the clients so a hard-killed
#    relay is detected fast rather than after the 30s QUIC idle default; keep it above
#    --client-quic-keep-alive, default 5s)
moq-relay demo/relay/localhost.toml --server-quic-gso=false &
```

**Result:** the publisher re-announces to the restarted relay, but the exporter has
already exited with `json: dropped`; `out.ts` stops growing at the drop instant and
never resumes.

## Likely root cause

The reconnect handle and the sink run as **sibling tasks** in a `JoinSet`, and
`drive()` returns as soon as the **first** task finishes:

- `rs/moq-cli/src/main.rs` `run_export`: spawns the reconnect task
  (`reconnect.closed()`) *and* the stdout sink task (`run_stdout(...)`).
- `run_stdout` awaits `consumer.announced_broadcast(&name)` **once**, builds a
  `moq_mux::Source`, and runs the container export (`Subscribe::run` → `run_ts`).
- `drive()` (`main.rs`): `while let Some(res) = tasks.join_next().await { … Ok(Err(err)) => return Err(err) … }`
  — the first task to error ends the whole process.

When the session drops, the persistent origin drops the `catalog.json` track, so the
catalog/source consumer inside the export pipeline yields `Error::Dropped`
(`"json: dropped"`). That bubbles out of the sink task as `Err`, and `drive()` returns
it and exits — **even though the reconnect task is still alive and would re-establish
the session and re-announce the broadcast a second later.** The exporter simply never
re-awaits the broadcast/catalog after a drop; the first catalog disappearance is
treated as terminal.

Contrast with `moq import`, which is resilient because its stdin pipeline feeds a
persistent publish origin that re-announces on reconnect; the *consume* origin is
equally persistent, so the missing piece is only that the export sink does not loop
back to `announced_broadcast` / rebuild its `Source` across reconnects.

## Suggested fix

Make the export sink survive a session drop while the reconnect handle is live —
options, roughly in order of locality:

1. In `run_stdout` (and the other sink bridges), **loop**: on a transient
   `Error::Dropped` from the source/catalog, re-await `announced_broadcast(&name)`,
   rebuild `moq_mux::Source`, and resume, rather than returning the error. Exit only on
   Ctrl-C / genuine origin close / non-transient error.
2. Alternatively, have the container `Export`/`Source` treat a catalog-track drop as a
   re-acquire point (await the next catalog snapshot) instead of a fatal end, matching
   the persistent-origin semantics.
3. At minimum, `drive()` could distinguish "the reconnecting MoQ side dropped" (retry
   in place) from "a sink reached EOF/was cancelled" (terminate) so a transient session
   loss does not tear the process down.

kixelated:
Fixes #2459 (`moq export` dies with `json: dropped` on relay restart).

## Root cause

An origin front (`FrontState`) closed synchronously the instant its last source detached, regardless of *why* it detached. A session dying therefore aborted every downstream subscription with `Error::Dropped` even though the session's reconnect loop was about to re-establish it and re-attach a source seconds later. The graceful/abrupt distinction already existed at the source layer (`SourceGuard`: `finish()` on a clean unannounce in both the lite and IETF subscribers, `abort()` on session death, whose doc already anticipated "the broadcast may still linger for a reconnect") but was discarded at detach: `run_source` dropped the close reason and `detach_source` had a single teardown path.

## Change

**Model (`moq-net`):**

- `broadcast::Producer::finish()` now sets a `finished` bit that the origin reads when the source closes, so a **deliberate end (clean unannounce) still closes and unannounces immediately: it never lingers.**
- `origin::Info` gains `linger` (+ `Info::with_linger` / `origin::Producer::with_linger`, default zero): how long a front outlives the *ungraceful* loss of its last source. Within the window the path stays announced (so the linger composes across relay hops without any wire change) and a source re-attaching at the path splices into the same broadcast; tracks resume at the group boundary via the existing route-migration machinery, so consumers never observe the gap. `run_front` arms the countdown when the table empties, cancels it on re-attach, and closes the front if the window expires empty (re-checking under the write lock so a reconnect racing the deadline wins). An unrepresentable deadline (`Duration::MAX`) lingers forever.

**Wiring (what actually fixes the issue):**

- `moq-native`: `Client::consume()` derives the linger from the reconnect loop's own promise: `Backoff::linger()` = `backoff.timeout + 1s`, or forever when the timeout is 0 (unlimited retries). Broadcasts fed by a reconnecting session therefore survive a session drop exactly as long as the loop keeps retrying, and when it gives up, the loop's real error (`reconnect timed out after ...: <cause>`) surfaces from the CLI's JoinSet *before* the broadcasts abort, replacing the misleading `json: dropped`. The +1s epsilon exists to win that race. `moq export` (and every other `consume()` caller) picks this up with no CLI changes.
- `moq-relay`: new `[cluster] linger` / `--cluster-linger` (default 5s): an abruptly-disconnected publisher stays announced for the window so its reconnect resumes the same broadcast downstream; a clean unannounce still propagates immediately. `Cluster` now keeps its `origin::Info` so `with_cache` and the linger compose instead of the rebuild clobbering whichever was set first.

Why not one fixed 5s everywhere: with default exponential backoff (1s, 3s, 7s, ...) a fixed 5s window only survives outages of about 3s, so the window has to come from whoever knows the recovery promise. The relay's 5s covers publisher churn on a healthy link; the subscriber's `backoff.timeout` covers its own reconnects.

## Tests

- `test_linger_skipped_on_finish`: clean unannounce with a linger configured tears down immediately.
- `test_linger_reconnect_splices`: ungraceful loss is invisible during the window (no unannounce, no track error, mid-outage `request_broadcast` resolves), and the reconnect resumes at `group_start = Some(2)` after groups 0-1.
- `test_linger_expiry_closes`: expiry unannounces, aborts tracks, and a late re-create is a fresh broadcast.
- `test_linger_forever`: `Duration::MAX` never expires; a reconnect days later still splices.
- `test_backoff_linger`: the derived window outlives the give-up timeout; unlimited retries linger forever.
- `cli_does_not_clobber_toml_linger`: the new relay flag survives the clap+TOML merge.
- Existing `test_route_unannounce_immediate` / `test_route_detach_immediate` pin the zero-linger defaults.

`just rs check` passes (full workspace); rustdoc with `-D warnings` clean on the touched crates.

## Cross-package sync

- No wire change, so no draft update: the linger is expressed as "the announce has not gone away yet", which is already on the wire.
- `doc/bin/relay/config.md` documents the new `[cluster] linger` knob.
- No `moq-ffi` surface change; `js/net` has its own origin model and can mirror the linger when wired up there.

## Deferred
- Server-mode consumption (moq --server-bind export, serve_consume) gets no linger: inbound sessions carry no reconnect promise and the origin default is zero, so a publisher reconnecting to a moq-cli server still tears down its broadcast.
- An end-to-end relay-restart regression (two real sessions) reusing the sink-session harness; the mechanism is covered at the model layer here.

- An announce-end *reason* on the wire (a truncated broadcast currently reads as a clean end downstream of a relay once its linger expires); additive SETUP-gated extension, separate draft+impl PR.
- Splice-with-a-restarted-publisher (group sequences resetting) relies on the existing resume machinery; the same-publisher case is what this PR tests.

(written by Fable 5)

### DIFF
--- a/rs/moq-net/src/model/broadcast.rs
+++ b/rs/moq-net/src/model/broadcast.rs
@@ -170,6 +170,11 @@ struct BroadcastState {
 	// Set by an explicit `Producer::finish()` or `Producer::abort()` so `Drop` can
 	// tell a deliberate shutdown apart from a producer dropped by accident.
 	closing: bool,
+
+	// Set only by `Producer::finish()`: the broadcast ended deliberately, as
+	// opposed to aborting or losing its producer. The origin reads this to decide
+	// whether a detached source may linger for a replacement.
+	finished: bool,
 }
 
 /// The spliced (route-fed) half of a broadcast: logical tracks that outlive any
@@ -453,7 +458,11 @@ impl Producer {
 	/// declares the end, so it must not depend on the caller also surrendering the
 	/// handle.
 	pub fn finish(&mut self) {
-		self.state.lock().closing = true;
+		{
+			let mut state = self.state.lock();
+			state.closing = true;
+			state.finished = true;
+		}
 		// Ending the broadcast is what consumers wait on, so signal it here rather
 		// than leaving it to the last handle drop.
 		let _ = self.alive.close();
@@ -860,6 +869,13 @@ impl Consumer {
 	/// than a strike.
 	pub(crate) fn is_closing(&self) -> bool {
 		self.is_closed() || self.state.read().closing
+	}
+
+	/// Whether the broadcast ended via an explicit [`Producer::finish`], as opposed
+	/// to aborting or losing its producer. The origin uses this to close a front
+	/// immediately on a deliberate end instead of lingering for a replacement.
+	pub(crate) fn is_finished(&self) -> bool {
+		self.state.read().finished
 	}
 
 	/// Register a [`kio::Waiter`] that fires when the broadcast closes.
--- a/rs/moq-net/src/model/broadcast.rs
+++ b/rs/moq-net/src/model/broadcast.rs
@@ -680,7 +680,7 @@ impl Dynamic {
 
 	/// Poll until the broadcast closes; ready with the cause: the error passed to
 	/// [`Producer::abort`], or [`Error::Dropped`] for a [`Producer::finish`] or a
-	/// dropped producer (check [`is_finished`](Self::is_finished) to tell those apart).
+	/// dropped producer (check [`Consumer::is_finished`] to tell those apart).
 	pub fn poll_closed(&self, waiter: &kio::Waiter) -> Poll<Error> {
 		ready!(self.alive.poll_closed(waiter));
 		Poll::Ready(self.state.read().abort.clone().unwrap_or(Error::Dropped))
@@ -870,14 +870,15 @@ impl Consumer {
 		}
 	}
 
-	/// Block until the broadcast is closed, by [`Producer::finish`] or by every producer
-	/// dropping, and return the cause.
+	/// Block until the broadcast is closed, by [`Producer::finish`],
+	/// [`Producer::abort`], or every producer dropping, and return the cause.
 	///
-	/// Always returns [`Error::Dropped`]: a broadcast is just a collection of tracks, so it
-	/// only ends when every producer is gone. There is no way to abort it with a code.
+	/// Returns the error passed to [`Producer::abort`], or [`Error::Dropped`] for a
+	/// [`Producer::finish`] or a dropped producer (check [`Self::is_finished`] to
+	/// tell those apart).
 	pub async fn closed(&self) -> Error {
 		self.alive.closed().await;
-		Error::Dropped
+		self.state.read().abort.clone().unwrap_or(Error::Dropped)
 	}
 
 	/// Returns true if every [`Producer`] has been dropped.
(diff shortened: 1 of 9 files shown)


rs/moq-net/src/model/broadcast.rs
kixelated: Should this be pub and take an error? So it matches other Producers.

rs/moq-net/src/model/broadcast.rs
```diff
@@ -862,6 +871,13 @@ impl Consumer {
 		self.is_closed() || self.state.read().closing
 	}
 
+	/// Whether the broadcast ended via an explicit [`Producer::finish`], as opposed
+	/// to aborting or losing its producer. The origin uses this to close a front
+	/// immediately on a deliberate end instead of lingering for a replacement.
+	pub(crate) fn is_finished(&self) -> bool {
```
kixelated: Should this be pub? Could it match other Producers?

t0ms:
confirm tested and works.

PR #2475, opened 2026-07-24
feat(moq-mux,moq-boy): mark discontinuities, and never time a sample across one

kixelated:
## Summary

- A duration-less frame took its duration from the timestamp of whatever frame came next, **including one in a different group**. Groups are independently decodable, so the gap across a boundary is a discontinuity, not a duration; reading it as one makes a single sample last as long as the gap.
- Root cause of the report: `moq-boy` parks the emulator when nobody is watching, but its PTS is `start.elapsed()`, a clock that keeps running through the pause (the code says so: *"Re-anchor audio timestamps so the pause gap appears in PTS"*). It also ticks once at startup so the catalog gets codec config. A VOD recording that woke a box idle for 40 minutes therefore saw a warm-up frame at 63ms followed by the post-resume keyframe at 2405s, and emitted a video sample lasting **2405.05s**. Its `EXTINF` put the HLS video timeline 9620s ahead of audio, so hls.js picked a video position past the end of the audio playlist and stalled there forever, never buffering a byte. (moq-dev/moq.pro#814)
- The last frame before a boundary now falls back to `default_frame` from the catalog framerate / sample rate. On a continuous constant-rate source that is what the gap works out to anyway — the two answers only diverge when there *is* a gap. `base_dts` re-anchors on the real PTS every fragment, so a fallback that misses on a variable-rate source stays a local one-frame error rather than accumulating.
- This applies to the HLS fetch origin too, even though it walks a segment's groups in sequence and abandons the segment on the first missing one. **Consecutive sequence numbers say nothing about whether the publisher paused across the boundary**, so that path could reproduce the same sample. What it actually relied on was timing each one-packet audio group by the next packet; the Opus TOC fill-in that `emit_fragment` already did is now a shared `apply_codec_durations` step both callers run first, so those packets carry their real duration and never need a neighbour (otherwise every packet would drop onto the ~21.3ms `1024/sample_rate` fallback against real 20ms packets).
- `Producer::discontinuity()` makes a break visible: it publishes an **empty** group. A consumer can then see the break instead of inferring continuity from adjacent sequences, and — because a subscription starts at the latest group (`start_group.or_else(|| track.latest())`, and `max_sequence` advances on group *creation* regardless of frames) — a subscriber arriving mid-break lands on the marker and waits, instead of being served the pre-break group as though it were live. That is how the 40-minute-stale frame reached the recording in the first place.
- `discontinuity` takes no timestamp on purpose: a break is a gap between two groups, so any single timestamp is ambiguous about which side it belongs to. A caller that wants to bound the closing group's final frame calls `cut(end)` first. `moq-boy` marks the break **before parking**, not after resuming, so the marker is the live edge for the whole pause.

The marker carries no frames at all. Writing an empty frame to carry the group's end timestamp (the boundary `cut(end)` wants) is the eventual shape, once decoders are known to skip one; an empty *group* needs no such guarantee, since a consumer that ignores it just sees a group with nothing in it.

Publishers that know where a group's content ends should still say so with `Producer::cut(end)`; those durations arrive already set and are untouched.

## Public API changes

All additive, no renames/removals/signature changes to existing items:

- `moq_mux::container::Producer::discontinuity()`
- `moq_mux::codec::{h264,h265,opus}::Import::discontinuity()`
- `moq_video::encode::Producer::discontinuity()`
- `moq_audio::encode::Producer::discontinuity()`

Internal only (`pub(crate)`): `apply_codec_durations` added; `infer_missing_durations` / `pts_monotonic` behavior changed (signatures unchanged); `next_timestamp` renamed to `duration_bound`.

Removed from the public surface: `moq_mux::catalog::Rendition::{record_frame, record_reorder, record_group_end}` are now `pub(crate)`. They were `pub` but every caller is a codec importer inside the crate; no external consumer could meaningfully drive the catalog's private bitrate/jitter estimator, so this only sheds accidental surface.

Behavior change for existing callers: `Muxer::fragment` no longer times a sample by the next group's first frame. Its previous behavior is preserved for Opus via `apply_codec_durations`; other codecs get `default_frame` at a boundary instead of the cross-group delta.

## Test plan

- `cargo test -p moq-mux -p moq-hls -p moq-video -p moq-audio` — 427 / 49 / 48 / 18 pass.
- `cargo clippy` clean on those crates plus `moq-boy`, `--all-targets`.
- `cargo check --workspace --all-targets` clean.
- New tests, verified to fail without the fix — three of them reproduce the exact `Some(2405006756µs)` sample read out of the real recording's `trun`:
  - `infer_stops_at_a_group_boundary`, `infer_stops_at_an_interior_group_boundary` (audio buffers span groups; it has no keyframe roll), `infer_ignores_a_rewound_group_boundary`
  - `infer_crosses_a_mid_group_fragment_boundary` guards against over-applying it: a cap-flush successor in the same group is still used
  - `codec_durations_keep_one_packet_groups_exact`, `audio_fragment_takes_durations_from_the_codec`, `audio_fragment_does_not_absorb_a_pause`
  - `discontinuity_publishes_an_empty_group`, `discontinuity_moves_the_live_edge_off_stale_content`
- **Not run for real:** the `moq-boy` pause/resume path is compile- and unit-tested only; I have not driven an actual pause cycle to watch the empty group land on the wire.

(written by Opus 4.8)

### DIFF
--- a/rs/moq-boy/src/audio.rs
+++ b/rs/moq-boy/src/audio.rs
@@ -48,6 +48,14 @@ impl AudioEncoder {
 		self.producer.reset_epoch();
 	}
 
+	/// Publish an empty group marking the pause, so the gap the re-anchored epoch is about
+	/// to open reads as a break rather than one very long packet.
+	pub fn discontinuity(&mut self) {
+		if let Err(e) = self.producer.discontinuity() {
+			tracing::warn!(error = %e, "failed to mark the audio discontinuity");
+		}
+	}
+
 	/// Push interleaved signed-16-bit stereo PCM captured at `elapsed` (since
 	/// the emulator started, shared with the video clock).
 	pub fn push_samples(&mut self, samples: &[i16], elapsed: Duration) -> Result<()> {
--- a/rs/moq-boy/src/audio.rs
+++ b/rs/moq-boy/src/audio.rs
@@ -50,10 +50,9 @@ impl AudioEncoder {
 
 	/// Publish an empty group marking the pause, so the gap the re-anchored epoch is about
 	/// to open reads as a break rather than one very long packet.
-	pub fn discontinuity(&mut self) {
-		if let Err(e) = self.producer.discontinuity() {
-			tracing::warn!(error = %e, "failed to mark the audio discontinuity");
-		}
+	pub fn discontinuity(&mut self) -> Result<()> {
+		self.producer.discontinuity()?;
+		Ok(())
 	}
 
 	/// Push interleaved signed-16-bit stereo PCM captured at `elapsed` (since
--- a/rs/moq-audio/src/encode/producer.rs
+++ b/rs/moq-audio/src/encode/producer.rs
@@ -231,6 +231,18 @@ impl<E: CatalogExt> Producer<E> {
 		Ok(())
 	}
 
+	/// Mark a break in the published timeline: whatever is published next does not continue
+	/// what came before.
+	///
+	/// Call this when capture stops rather than merely gapping between packets -- going idle,
+	/// switching source, anything that resumes on a re-anchored epoch (see
+	/// [`reset_epoch`](Self::reset_epoch)). See
+	/// [`Producer::discontinuity`](moq_mux::container::Producer::discontinuity).
+	pub fn discontinuity(&mut self) -> Result<(), Error> {
+		self.track.discontinuity()?;
+		Ok(())
+	}
+
 	/// Flush any pending samples (zero-padded to a full frame) and finalize the
 	/// track.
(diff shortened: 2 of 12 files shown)


rs/moq-boy/src/audio.rs
```diff
@@ -48,6 +48,14 @@ impl AudioEncoder {
 		self.producer.reset_epoch();
 	}
 
+	/// Publish an empty group marking the pause, so the gap the re-anchored epoch is about
+	/// to open reads as a break rather than one very long packet.
+	pub fn discontinuity(&mut self) {
```
kixelated: Why not return an error?

PR #2484, opened 2026-07-24
fix(hang,watch): deliver non-sequential group ids incrementally

AaronKohlmeier:
Tracks numbered by DTS (e.g. CMSF) use large, non-+1 group-id gaps. The consumer assumed +1 numbering: it stalled delivery (constant stutter) and held whole groups until completion, then flushed them in a 1 Hz burst.

- consumer: advance #active to the next buffered group's actual sequence (not +1), add a next() promotion guard, and wake next() for the delivery head (#groups[0]) regardless of id scheme.
- decoder: bridge the decode buffer across non-adjacent group boundaries (group > prior.group) in both the legacy and CMAF paths.
- tests: non-sequential-gap incremental-delivery coverage.

### DIFF
--- a/js/hang/src/container/consumer.ts
+++ b/js/hang/src/container/consumer.ts
@@ -74,6 +74,30 @@ class Rewind {
 	discontinuity = 0;
 }
 
+// Two adjacent groups are treated as timeline-contiguous when the next group's first PTS is within
+// this slack of the current group's end. Per-sample durations and base-decode-times are each rounded
+// to microseconds independently, so a genuinely contiguous boundary can be off by ~1µs (seen on 48kHz
+// audio). A real missing group spans ~one group duration (orders of magnitude larger), so 1ms cleanly
+// separates rounding noise from an actual gap.
+const CONTIGUITY_TOLERANCE_US = 1000;
+
+/**
+ * True when `nextStart` continues the timeline that ends at `end`: it lands at or before `end`,
+ * within CONTIGUITY_TOLERANCE_US to absorb the µs rounding of independently-rounded per-sample
+ * durations and base-decode-times. Undefined on either side means continuity can't be proven.
+ *
+ * The bound is one-sided (upper only) by design: a next start at or before `end` continues the
+ * timeline, a start past `end` beyond the tolerance is a gap. A start well before `end` (a backward
+ * jump) is a rewind, handled by Rewind / #checkReset, not here.
+ */
+function ptsContiguous(end: Time.Micro | undefined, nextStart: Time.Micro | undefined): boolean {
+	return (
+		end !== undefined &&
+		nextStart !== undefined &&
+		(nextStart as number) <= (end as number) + CONTIGUITY_TOLERANCE_US
+	);
+}
+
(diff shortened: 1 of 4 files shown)


js/hang/src/container/consumer.ts
```diff
@@ -216,8 +224,15 @@ export class Consumer {
 			group.done = true;
 
 			if (group.consumer.sequence === this.#active) {
-				// Advance to the next group.
-				this.#active += 1;
+				// Advance to the next group's actual sequence. Some encoders number
```
kixelated: No I think this is too optimistic.

For example, lets say we're finishing group 4 and have group 6 in buffer. 

The current code will (or should) start at 5, and only skip ahead to 6 if it can prove that 5 would have expired anyway. We have to use the timestamps of group 6 to infer the final timestamp of group 5.

This code will always skip ahead to 6, even if 5 was being transmitted. It'll absolutely wreck audio.


I think we need *some* signal that group 5 is not coming before we can skip it. Either an empty group, or some drop message or something. Anything similar to a discontinuity in HLS that explicitly says do not download anything.

kixelated: And yeah the only reason we skip based on +1 is because it's definitive proof there's no gap. If you guys want to insert gaps explicitly (ex. group 0 -> 1000 -> 2000), then we also need some signaling that 1-999 are never coming and do not exist.

kixelated (changes requested):

js/hang/src/container/consumer.ts
AaronKohlmeier: i have updated the PR

js/watch/src/video/decoder.ts
```diff
@@ -391,9 +393,12 @@ class DecoderTrack {
...
 				}));
 
-				// Track decode buffer: frames sent to decoder but not yet rendered
+				// Track decode buffer: frames sent to decoder but not yet rendered.
+				// A finished prior group followed by frames from a later group is contiguous in
+				// playback even when group sequence numbers are not +1-adjacent (some encoders
+				// number groups non-sequentially with large gaps), so bridge on `prior.final`, not `+1`.
 				const prior = previous;
-				if (prior && (prior.group === group || (prior.final && prior.group + 1 === group))) {
+				if (prior && (prior.group === group || (prior.final && group > prior.group))) {
```
kixelated: This needs to use PTS instead of group count.

js/hang/src/container/consumer.ts
```diff
@@ -74,13 +74,40 @@ class Rewind {
 	discontinuity = 0;
 }
 
+// Two adjacent groups are treated as timeline-contiguous when the next group's first PTS is within
+// this slack of the current group's end. Per-sample durations and base-decode-times are each rounded
+// to microseconds independently, so a genuinely contiguous boundary can be off by ~1µs (seen on 48kHz
+// audio). A real missing group spans ~one group duration (orders of magnitude larger), so 1ms cleanly
+// separates rounding noise from an actual gap.
+const CONTIGUITY_TOLERANCE_US = 1000;
```
kixelated: cast as `Time.Micro`?

PR #2485, opened 2026-07-24
Pr/moq rust caller grouping

AaronKohlmeier:
Caller-driven group boundaries for audio + libmoq opt-in knobs

Lets a native publisher drive explicit, synchronized MoQ group boundaries instead of opening a new group (and QUIC stream) per audio frame — AAC flags every frame a keyframe, which was storming streams and desyncing co-multiplexed video. The spec only requires a group to start with a keyframe, not a new group at every one.

Features

moq-mux — caller-driven audio grouping. container::Producer gains keyframe_starts_group (default true, unchanged) + with_/set_keyframe_grouping; when off, keyframes accumulate and the caller bounds groups via seek/cut. AAC importer opts in; fmp4 import starts a group only at an explicit boundary, a video keyframe, or when none is open (audio rides the segment cadence, not one group per fragment).

libmoq — group-boundary C-ABI. moq_publish_media_group (forces a synchronized, explicitly-numbered boundary across a media importer's tracks) and moq_publish_media_keyframe_grouping (opt a track into caller-driven audio grouping).

libmoq — hardware video is now opt-in. moq-video moves behind a default-off hw-video feature, so a stock build needs no libva/CUDA; moq_consume_video_raw* stay in the ABI and return Unsupported without it.

libmoq — connection knobs. moq_tls_disable_verify (dev / self-signed relays) and moq_backoff_config (reconnect backoff)

### DIFF
--- a/rs/libmoq/Cargo.toml
+++ b/rs/libmoq/Cargo.toml
@@ -16,6 +16,15 @@ categories = ["multimedia", "network-programming", "web-programming"]
 name = "moq"
 crate-type = ["staticlib"]
 
+[features]
+# No hardware / system-dep video by default: a stock libmoq builds with no libva/CUDA.
+# Opt into native video decode (the moq_consume_video* C API) with `--features hw-video`,
+# which pulls in moq-video (NVENC/VAAPI/NVDEC); its vaapi backend links libva at build time.
+# Mirrors moq-cli's capture/transcode: hardware / system deps are opt-in, never assumed.
+# INVARIANT: libmoq MUST keep building with --no-default-features (no libva/CUDA). CI enforces this.
+default = []
+hw-video = ["dep:moq-video"]
+
 [dependencies]
 anyhow = { version = "1", features = ["backtrace"] }
 bytes = "1"
@@ -25,8 +34,9 @@ moq-json = { workspace = true }
 moq-mux = { workspace = true }
 moq-native = { workspace = true, default-features = true }
 moq-net = { workspace = true }
-# Enable the NVENC / VAAPI / NVDEC hardware codecs (default-off at the workspace level).
-moq-video = { workspace = true, features = ["nvenc", "vaapi", "nvdec"] }
+# Optional, behind the `hw-video` feature: native video decode for the moq_consume_video* C API.
+# Keeps its default NVENC/VAAPI/NVDEC backends; the vaapi backend links libva at build time.
+moq-video = { workspace = true, optional = true }
 serde_json = "1"
 thiserror = "2"
 tokio = { workspace = true, features = ["macros"] }
--- a/rs/libmoq/src/api.rs
+++ b/rs/libmoq/src/api.rs
@@ -406,6 +406,41 @@ pub extern "C" fn moq_error() -> *const c_char {
 	ffi::last_error_ptr()
 }
 
+/// Disable TLS certificate verification for all future sessions.
+///
+/// Global setting; call before [moq_session_connect]. Convenient for local development
+/// against a self-signed relay; use with caution in production.
+///
+/// Returns a zero on success.
+#[unsafe(no_mangle)]
+pub extern "C" fn moq_tls_disable_verify(disable: bool) -> i32 {
+	ffi::enter(move || {
+		State::lock().tls_disable_verify = disable;
+		Ok(())
+	})
+}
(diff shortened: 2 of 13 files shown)


rs/moq-mux/src/codec/aac/import.rs
```diff
@@ -57,71 +89,42 @@ impl<E: CatalogExt> Import<E> {
...
 		self.track.write(Frame {
 			timestamp,
-			payload: frame.into_bytes(),
+			payload: bytes::Bytes::copy_from_slice(frame),
 			keyframe: true,
 			duration: None,
 		})?;
-		self.track.cut(None)?;
-		self.rendition.record_frame(timestamp, bytes);
+		if !self.manual_groups {
```
kixelated: hmm, I'm not sure about this shape but it's close.

We could require the caller to `cut` manually. It's kind of lame that we automatically cut every frame into its own stream. In fact, fMP4 imports should not be automatically cutting like this; they should preserve fragment boundaries.

Or we could also have a `max_group_duration` or something that cuts automatically after some duration, with None being manual_groups.

AaronKohlmeier: Using this in an encoder i want full control of the content and groups. I plan to always us manual control. But I dont want to break anyone else using this so when someone isn't doing manual control my intention was to leave the functionality as is to prevent breaking the previous contract for anyone not using manual control.

But also I'm not just doing fmp4 control, i'm playing with hang too. And i have the same issue with hang. As production it makes sense to make it easier for players to make audio and video align by default when they join. That is you cant join an audio stream half way through the video streams and be misaligned by half a video group. Instead, following the patterns of hls as an example. Audio group edges should follow video group edges/durations.

kixelated: First off a disclaimer, I'm not entirely sure what to do about audio. It feels wrong.

> But also I'm not just doing fmp4 control, i'm playing with hang too. And i have the same issue with hang. As production it makes sense to make it easier for players to make audio and video align by default when they join. That is you cant join an audio stream half way through the video streams and be misaligned by half a video group. Instead, following the patterns of hls as an example. Audio group edges should follow video group edges/durations.

The problem is that aligning audio groups with video groups increases latency, as it causes HoLB between audio frames. It obviously makes sense for HLS because it's required, and in makes sense for VOD when we want guaranteed A/V at specific timestamps.

But even in your example of joining a live stream, we don't actually want to start playback at the latest keyframe. Real-time playback will download the keyframe and decode/drop as fast as possible until it caches up to the live playhead. We could do the same for audio but it's a waste of bandwidth, and the current behavior of a group per frame is closer to optimal. Not that it really matters though because the audio bitrate is so low.

One option is to have a IETF `sub-group` per audio frame, and a `group` per segment instead. But that causes the same late join behavior as above, as subscriptions start at the latest group, but we really want audio to start at the latest sub-group (whatever that means).

I'm thinking maybe layering the concept of `Segment` on top of the object model instead. Like the timeline track would produce an update per segment, not per group (super noisy anyway). ex. timeline segment 5 (timestamp 25) for video is group 5, while timeline segment 5 for audio is group 123-156. Audio is still lossy for real-time, but can be bundled into a nice aligned package for HLS.

And the HLS converter would use the timeline only for generating playlist.m3u8 files. I don't want it downloading media, only the segment generation should do that.

kixelated: > Using this in an encoder i want full control of the content and groups. I plan to always us manual control. But I dont want to break anyone else using this so when someone isn't doing manual control my intention was to leave the functionality as is to prevent breaking the previous contract for anyone not using manual control.

And yeah I'm okay with breaking the contract a little bit. Not everybody wants a group per audio frame, so it seems weird to force that behavior. I'm okay with a breaking change because manual control seems like a better API, but you're right we should rename the function or something so it fails to compile.

kixelated:
@AaronKohlmeier this was a lot harder than I expected.

I thiiiiink the right behavior is #2568 . Basically we create a new group on each keyframe, and for audio it's up to the caller to decide if each frame is a "keyframe" or not. It's very similar to your manual cutting but it preserves that keyframe: true for a frame is the same on the encode and decode side. 

Does this work for you? I'm totally fine going back to the drawing board.

PR #2490, opened 2026-07-24
feat(net, relay): GOAWAY with graceful drain and cluster migration

ksletmoe-aws:
## Summary

This implements **GOAWAY support** in `moq-net`, on both the moq-lite wire (lite-04+) and the IETF moq-transport wire (draft-14 through draft-19), in both directions. A session can **send** a GOAWAY to gracefully drain its peer (optionally with a replacement URI and a deadline), and can **observe** a GOAWAY it receives and react to it.

`moq-relay` consumes this in both directions:

- **Downstream (`--drain-timeout`)**: the first shutdown signal sends every accepted session (QUIC/WebTransport and the WebSocket fallback) an empty-URI GOAWAY and waits for clients to leave before force-closing; a second signal exits immediately.
- **Upstream (`--cluster-drain-timeout`)**: on a GOAWAY from a cluster peer, the relay dials the replacement while the old session keeps serving, then drains the old session in the background. Downstream sessions never see a GOAWAY of their own, and live subscriptions resume at a group boundary on the new connection.

Thanks for the GOAWAY groundwork already in place (the codecs and the receive-path stubs); this completes the lifecycle on top of it.

## Built on #2241: migration is the route machinery's job, not GOAWAY's

An earlier draft of this work (pre-#2241) hand-rolled a seamless-failover layer: failover markers on origin paths, silent announce swaps, successor-by-identity tracking, and a `Handoff` typestate to drive it. **All of that is gone.** #2241's `resume`/route model already provides exactly those semantics as first-class primitives:

- A replacement session sharing the same origin attaches its announcements as additional **routes** to the broadcasts the old session serves.
- When the old session closes, its routes detach and the origin hands live subscriptions over to the new route **at a group boundary** (`resume::Producer::switch`), with the first-hop-identity rule deciding route-change vs broadcast-replacement.

So GOAWAY here is deliberately thin: the protocol lifecycle in `moq-net`, and *policy* in the relay (when to dial, which redirect to trust, how long to drain). The cluster's migration handler is roughly 60 lines because the model does the heavy lifting. An empty-URI GOAWAY ("reconnect to me") needs no special-casing anymore: a restarted publisher resolves as a different first-hop identity and replaces the broadcast, exactly as the model specifies.

## moq-net API

- **Send / drain.** `Session::drain()` claims the one-GOAWAY-per-session slot and returns a `Drain` (or `None` on versions without GOAWAY, so a caller can't silently no-op into a hang; dropping an unstarted `Drain` releases the claim). `Drain::start(uri)` / `start_with_timeout(uri, deadline)` send the frame and return a `Draining`; `Draining::complete()` waits for the peer to leave, force-closing with the new `GoawayTimeout` code (33) when the deadline expires.
- **Receive / observe.** `Session::goaway()` resolves with `GoawayReceived { uri, timeout }`; `Session::is_going_away()` is a cheap synchronous check. A duplicate GOAWAY (a protocol violation) keeps the first payload and is logged, since an observer may already be acting on the first URI.
- **Request gating.** After a received GOAWAY, new SUBSCRIBE / FETCH / TRACK / announce-interest opens are rejected with `GoingAway` (32); PROBE is silently skipped; existing subscriptions keep flowing until the session closes. Under the resume model the rejection typically surfaces as a *stall* on the gated route (waiting for another route to serve it) rather than a torn-down subscription.
- **Wire channels per version.** lite-04+ uses the dedicated Goaway control stream; IETF draft-14-16 ride the shared control stream through the adapter; draft-17+ use the SETUP uni streams (ours to send, the peer's to receive). The receive reader stays alive after decoding: closing the peer's SETUP stream mid-session is a protocol violation (draft-19 §3.3) a strict peer would punish right in the middle of the drain.
- **Codec conformance.** Draft-18 (#1559) appends the required control-stream Request ID (0; tolerated-optional on decode from lenient peers); draft-19 (#1623) removes it. The 8,192-byte New Session URI cap is enforced on both wires. On moq-lite it is rejected from the length prefix alone, and the moq-lite draft now specifies the cap (changelog under lite-06).
- **Plumbing.** Everything follows the driver architecture: send paths are `TaskSet` children racing the drain trigger against transport close (nothing spawned, nothing blocking clean shutdown), and `Draining::complete` polls under a kio waiter.

## Two bugs found by the real-transport tests

- **The lite GOAWAY frame was silently lost on real QUIC.** `Writer`'s drop-reset raced the FIN and discarded the unacked frame; the in-memory mock physically cannot lose data, so only the new quinn tests caught it. Fixed by awaiting FIN acknowledgment before dropping, the same dance as `send_setup`.
- **The close reason vanished on quinn.** `Session::closed()` stringified the transport error, and quinn's `Display` drops the application close code and reason entirely ("connection error: closed"). `closed()` now surfaces `code=N: reason` when the transport carries them, so a peer can at least distinguish a `GoawayTimeout` force-close from a network failure. See the open question below about going further.

## Branch target and API compatibility

**Target: `main`.** Everything in `moq-net` is strictly additive: `Session::{drain, goaway, is_going_away}`, `Drain`/`Draining`, `GoawayReceived`, `Version::has_goaway`, and two new variants on the `#[non_exhaustive]` `Error` (wire codes 32/33; 31 was already taken by `Evicted`). The relay config fields land on `#[non_exhaustive]` structs.

**One deliberate exception, flagged in the open:** `moq_relay::Connection` gains a required `shutdown` field, and `Connection` is a plain pub-field struct that embedders construct with a literal, so this breaks external constructors (the in-repo fix was one line in the smoke test; `Shutdown::disabled()` is the drop-in value). By the letter of the branch rules that routes to `dev`. We propose `main` anyway because the blast radius is one line per embedder and the fix is mechanical. If you'd rather avoid the break entirely, the clean alternatives are an additive `Connection::run_with_shutdown(self, Shutdown)`, or biting the `#[non_exhaustive]` + constructor bullet now (this is the second time a new knob has hit this struct's literal-construction shape). Happy to re-shape either way.

## Known limitations

- **Resume is group-granular.** The resume model splices per-session tracks at group boundaries; there is no object-level resume (that requires carrying object ids through the relay plus `RESET_STREAM_AT`, unchanged from before and out of scope). Graceful drains are seamless because the in-flight group finishes on the old route; an abrupt mid-group death loses at most the group in flight.
- **Desynced replacements are untested.** Every failover test uses siblings sharing one origin. A replacement far behind stalls (by design) and one far ahead of the group cache falls forward; neither case has a test yet.
- **The force-close reason is best-effort on real QUIC.** The overstaying peer's own teardown close can race the server's CONNECTION_CLOSE and stomp the reason, so the structured-reason assertion lives in the deterministic mock test; the real-QUIC test asserts the enforcement (prompt close) only.

## Open questions

- **Structured close codes (deferred to a follow-up, raised here for direction).** `Session::closed()` currently surfaces close reasons as strings (`code=33: goaway timeout`). GOAWAY introduces the first codes where programmatic matching genuinely matters, and an earlier draft of this work had a full `Error::from_code` inverse (payload-free codes round-trip; codes whose variants carry non-wire payloads stay `Remote`; local-only codes never decode). We deliberately did **not** include it here: it changes how *every* close surfaces (not just GOAWAY), it touches the error architecture this repo just rewrote, and it is only as reliable as close-code delivery, which quinn does not currently guarantee (see the race above), suggesting the robust version also wants a `web-transport-trait`-level guarantee. If you want `closed()` to return matchable variants, we'll do it as a focused follow-up; the `from_code` sketch is ready. `Error` being `#[non_exhaustive]` and the string format being undocumented keeps the migration cost of deferring near zero.
- **GOAWAY redirect target trust.** The only validation on a redirect URI is a scheme security-tier check (encrypted/IPC ≥ plaintext ≥ unknown, no downgrades), plus carrying the cluster JWT. There is no allowlist or known-peer check on the redirect host, and `unix` in the top tier lets an authenticated upstream point the relay at a local socket. A host allowlist is marked as a deliberate follow-up in the code.

## Scope: Rust behavior + both wires; JS wire-level only

`@moq/net` decodes GOAWAY on both wires (with a byte-layout test locking draft-18 Request ID parity against the Rust encoder), so JS clients interoperate safely with a relay that sends GOAWAY. The JS session-level lifecycle (observing a GOAWAY, reconnecting) is deferred to a follow-up PR, consistent with #2241 deferring its own js/net mirror.

## Test plan

All green: 901 Rust tests across moq-net / moq-relay / moq-native, clippy clean, `cargo doc -D warnings` clean, `tsc` + biome clean on js/net, kramdown-rfc parses the draft. `bun test js/net`: all 95 GOAWAY/IETF codec tests pass; the full-suite run carries one pre-existing failure (`integration: lite draft-03`) that reproduces identically on a clean `main` checkout (3/3 runs, passes in isolation), so it is not introduced here.

- **moq-net** (544 lib + 8 GOAWAY integration on the in-memory mock; deterministic, no sleeps): send/receive across the full version matrix (lite-04/05, drafts 14/16/17/18/19, covering all three wire channels), wire deadline on draft-17+, `GoawayTimeout` force-close observed by the peer, drain-claim exclusivity and release-on-drop, `drain()` is `None` pre-lite-04, a duplicate-GOAWAY regression (injected as raw wire frames; the public API can't send two), and request gating with an existing subscription proven to keep flowing.
- **moq-native** (real QUIC / WebTransport): 6 GOAWAY tests, covering the send/receive lifecycle across lite-04/05 and drafts 14/17/19 plus force-close enforcement on draft-19. Real-transport coverage matters here: both bugs above were invisible on the mock.
- **Cross-language smoke** (`test/smoke/smoke.sh`, built from this checkout): the cells runnable on the dev host all pass — rust/js publishers x rust/js subscribers, which covers both implementations this PR touches. The python, c, js-native, and gst cells fail identically on a clean `origin/main` checkout on the same host (missing build deps: `just py build`, `cc`, native-lib `dlopen`, glib), so they are environmental; the full matrix should be confirmed via `nix develop --command just test smoke-full`. Note the smoke flows do not exercise GOAWAY itself (they verify media interop); cross-language GOAWAY behavior testing requires the deferred JS session-level support.
- **moq-relay** (148 lib incl. scheme-tier and redirect-resolution unit tests + 12 smoke + 3 integration): `cluster_migrates_on_upstream_goaway` (two siblings over real TCP), `cluster_reconnects_on_empty_uri_goaway`, and `cluster_diamond_goaway_seamless_failover`: a 5-node diamond (TOP -> MID-A/MID-B -> BOTTOM -> subscriber) publishing 24 groups of 3 frames each, 20 of them streamed at a steady cadence through the GOAWAY/reconnect/handover window, asserting every group arrives exactly once, in order, every frame intact and exactly counted; a positive gate that BOTTOM reconnects to MID-B; a post-drain phase with the old leg fully severed proving delivery flows via the new leg; no announce churn; and no GOAWAY cascade to the subscriber (sync flag plus a bounded async probe).

## Suggested follow-ups

1. **JS session-level GOAWAY** (`goaway()` / drain lifecycle in `@moq/net`).
2. **Structured close codes** + transport-level close-code delivery guarantee (open question above).
3. **Redirect host allowlist** for cluster GOAWAY targets (open question above).
4. **Desync hardening**: tests (and possibly bound-waits) for a replacement source far ahead of or behind the failed one.
5. **Object-granular resume** (object ids through the relay + `RESET_STREAM_AT`) for zero-loss mid-group failover. Pre-existing, unchanged by this PR.

(Written by Claude Fable 5)


doc/bin/relay/config.md
```diff
@@ -171,6 +182,13 @@ token = "cluster.jwt"
 # never notice. A clean unannounce always takes effect immediately. "0"
 # unannounces abrupt losses immediately too. Default: 5s.
 linger = "5s"
+
+# Optional. Fallback drain time in seconds when an upstream peer sends a GOAWAY
+# without its own deadline: after reconnecting to the replacement, the relay
+# keeps the old upstream alive this long so in-flight groups finish, then
+# force-closes it. A deadline on the received GOAWAY takes precedence.
+# Default: 10.
+drain_timeout = 10
```
kixelated: nit: `drain`

rs/moq-relay/src/cluster.rs
```diff
@@ -977,6 +1033,63 @@ fn connect_api_is_http(source: &str) -> bool {
...
+		let token = fallback
+			.query_pairs()
+			.find(|(k, v)| k == "jwt" && !v.is_empty())
+			.map(|(_, v)| v.into_owned());
+		if let Some(token) = token {
+			parsed.query_pairs_mut().append_pair("jwt", &token);
+		}
+	}
+	parsed
+}
```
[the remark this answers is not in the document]
ksletmoe-aws: Good catch, and I agree the JWT carry-over is the part that turns this from a wasted dial into credential exposure. I don't want to fix it blindly, though, because the obvious fix conflicts with what the feature is for.

The tension: the suggested "only forward the JWT when the redirect host matches the configured peer" would close the hole, but it also breaks the main use case. A seamless failover redirect points at a *different* sibling host, and in a JWT-authed cluster that sibling needs the token to authenticate. Same-host-only forwarding defeats cross-host failover, which is the whole point of the redirect.

So this is really a trust-model decision. @kixelated, how would you like to handle it? Options I see:

1. **Leave as a documented follow-up** (current state). The PR already flags "constrain redirect targets to known peers" as an open question; the scheme-tier no-downgrade check stays as the only guard for now.
2. **Redirect-host allowlist** — dial (and forward the JWT to) a redirect only if its host is a known cluster peer (e.g. from `--cluster-connect` / gossip / `connect-api`). This is the real fix and preserves cross-host failover, but it's a bit more scope: it needs the reconnect path to consult the peer set.
3. **Opt-in cross-host forwarding** — keep same-host JWT forwarding by default (safe), and gate cross-host redirect JWT behind an explicit config flag for operators who trust their redirect sources.

My lean is (2) as the proper fix, done as a follow-up PR rather than growing this one, with (1) holding until then. Happy to implement whichever you prefer here or in a follow-up.

(Written by Claude Fable 5)

rs/moq-net/src/lite/subscriber.rs
```diff
@@ -1247,6 +1270,7 @@ impl<S: web_transport_trait::Session> TrackServe<S> {
 
 		tracing::info!(id, broadcast = %self.subscriber.log_path(&self.path), track = %self.name, "subscribe started");
 
+		self.subscriber.check_going_away()?;
```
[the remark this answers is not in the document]
ksletmoe-aws: Fixed. Dropped the dead inline SUBSCRIBE open/encode; `open_subscribe(&msg)` is now the sole path (it already carries the `check_going_away` gate), so no stray subscribe stream is opened and immediately reset. Verified the lite subscribe round-trips still pass.

_(AI-assisted comment. Written by Claude Fable 5.)_

rs/moq-net/src/session.rs
```diff
@@ -110,6 +111,155 @@ impl Session {
...
+	/// Send the GOAWAY frame with a deadline for the peer to disconnect.
+	///
+	/// [`Draining::complete`] force-closes the session with
+	/// [`Error::GoawayTimeout`] when `timeout` elapses. The deadline also rides
+	/// the wire on versions with a timeout field (IETF draft-17+) so the peer can
+	/// observe it; the wire encodes 0 as "no deadline", so pass a non-zero
+	/// duration for a deadline the peer can see.
+	pub fn start_with_timeout(self, uri: impl Into<Arc<str>>, timeout: Duration) -> Draining {
+		self.start_inner(uri, Some(timeout))
+	}
```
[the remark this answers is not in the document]
ksletmoe-aws: Fixed. `start_with_timeout` now normalizes `Duration::ZERO` to no local deadline via `(!timeout.is_zero()).then_some(timeout)`, so a zero timeout matches the wire's "0 = no deadline" on both sides instead of force-closing right after the GOAWAY. Doc updated to match.

_(AI-assisted comment. Written by Claude Fable 5.)_

rs/moq-net/tests/goaway.rs
```diff
@@ -0,0 +1,340 @@
...
+		// (waiting for another route), not an error, so the observable behavior
+		// is that a track with content ready produces nothing. The mock delivers
+		// synchronously, so the bounded wait is generous.
+		match bc.track("audio").unwrap().subscribe(None).await {
+			Err(_) => {} // Rejected outright is also acceptable gating.
+			Ok(mut gated) => {
+				let delivered = tokio::time::timeout(Duration::from_millis(500), gated.recv_group()).await;
+				assert!(delivered.is_err(), "new subscribe after GOAWAY must not deliver");
+			}
+		}
```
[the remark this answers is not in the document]
ksletmoe-aws: Tightened. The error arm now matches `Error::GoingAway` specifically and panics on any other error, so a regression that fails for a different reason still fails the test. Kept the bounded wait for the resume-model stall case with a comment explaining it. A wire-level "no upstream subscribe reached the wire" assertion would be stronger but needs a hook the mock doesn't expose today.

_(AI-assisted comment. Written by Claude Fable 5.)_

rs/moq-net/tests/support/mock.rs
```diff
@@ -0,0 +1,428 @@
...
+	}
+
+	fn protocol(&self) -> Option<&str> {
+		self.side.protocol
+	}
+
+	fn close(&self, code: u32, reason: &str) {
+		*self.side.conn.close_state.lock().unwrap() = Some((code, reason.to_string()));
+		self.side.conn.close_notify.notify_waiters();
+	}
```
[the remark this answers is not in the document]
ksletmoe-aws: Fixed. Mock `close()` is now set-once (first reason wins), matching a real QUIC CONNECTION_CLOSE and the documented contract. This is exactly what keeps the GOAWAY-timeout assertion deterministic against a later local teardown close from the peer's own driver, so thanks for catching the contradiction.

_(AI-assisted comment. Written by Claude Fable 5.)_

rs/moq-relay/src/cluster.rs
```diff
@@ -929,12 +913,84 @@ impl Cluster {
...
+								tracing::warn!(?timeout, "old upstream did not drain in time; force-closing");
+								old.abort(moq_net::Error::GoawayTimeout);
+							}
+						}
+					});
+
+					tracing::info!("upstream reconnect successful; new session active");
+				}
+			}
+		}
```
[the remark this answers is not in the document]
ksletmoe-aws: Fixed. Added a sliding-window churn guard: more than 5 migrations within 10s makes `run_remote_once` return an error so `run_remote`'s exponential backoff engages, instead of migrating at handshake speed and spawning a detached drain task each cycle. A redirect loop (A names B, B names A) or an immediate-GOAWAY peer now backs off.

_(AI-assisted comment. Written by Claude Fable 5.)_

rs/moq-relay/src/shutdown.rs
```diff
@@ -0,0 +1,75 @@
...
+	///
+	/// Sessions on versions without GOAWAY (moq-lite-03 and earlier) are closed
+	/// immediately with [`moq_net::Error::GoingAway`]; there is no wire message
+	/// to warn them with.
+	pub async fn drain_session(&self, session: &moq_net::Session) {
+		match session.drain() {
+			Some(drain) => drain.start_with_timeout("", self.drain_timeout).complete().await,
+			None => session.abort(moq_net::Error::GoingAway),
+		}
+	}
```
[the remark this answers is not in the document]
ksletmoe-aws: Fixed. `drain_session` now force-closes only GOAWAY-less versions (`None if !session.version().has_goaway()`). When `drain()` returns `None` because a drain is already in flight, the session is left to that drain rather than being force-closed a second time.

_(AI-assisted comment. Written by Claude Fable 5.)_

rs/moq-relay/tests/goaway_cluster.rs
```diff
@@ -0,0 +1,575 @@
...
+	tokio::task::JoinHandle<()>,
+) {
+	// Pick a free TCP port, then drop the probe so the listener can bind it.
+	let probe = TcpListener::bind("127.0.0.1:0").expect("bind probe");
+	let port = probe.local_addr().expect("local addr").port();
+	drop(probe);
+
+	let mut config = moq_native::ServerConfig::default();
+	config.tcp.bind = Some(format!("127.0.0.1:{port}").parse().expect("parse addr"));
+	let mut server = config.init().expect("server init");
```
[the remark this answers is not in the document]
ksletmoe-aws: Fixed. Extracted a `bind_free_tcp_server()` helper that retries the probe -> bind sequence (up to 20 fresh ports) on `init()` failure, and both spawn helpers use it. Closes the TOCTOU window between dropping the probe and the real bind that could spuriously panic under parallel tests.

_(AI-assisted comment. Written by Claude Fable 5.)_

rs/moq-relay/tests/goaway_cluster.rs
```diff
@@ -0,0 +1,575 @@
...
+	.expect("broadcast announced");
+	let mut sub = within(
+		"subscribe to the video track",
+		bc.track("video").expect("track handle").subscribe(None),
+	)
+	.await
+	.expect("subscribe");
+
+	// ── group 0 flows through the MID-A leg (2 frames, fully verified) ──
+	const FRAMES_PER_GROUP: u64 = 3;
```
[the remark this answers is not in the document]
ksletmoe-aws: Fixed. Corrected the stale comment to match `FRAMES_PER_GROUP = 3`.

_(AI-assisted comment. Written by Claude Fable 5.)_

doc/concept/layer/moq-lite.md
```diff
@@ -135,6 +135,29 @@ When combined with a local jitter buffer, this should result in different user e
...
 
+### GOAWAY (Graceful Shutdown)
+
+Either endpoint can gracefully drain a session by sending a `GOAWAY` message.
+It tells the peer to reconnect, either to a different endpoint (a URI in the message) or to the same endpoint (empty URI), instead of being cut off when the sender shuts down.
+
+The lifecycle on the sending side (Rust `moq-net`):
+
+1. `session.drain()` claims the session for draining (one GOAWAY per session). Returns a `Drain` handle, or `None` on versions without GOAWAY (moq-lite-03 and earlier).
+2. `drain.start(uri)` or `drain.start_with_timeout(uri, duration)` sends the GOAWAY frame.
```
kixelated: not a fan of xxx_with_yyy methods. If we ever wanted to make another option, then we're out of luck.

Make a separate options struct with a builder instead.

rs/moq-net/src/ietf/adapter.rs
```diff
@@ -509,14 +509,17 @@ impl<S: web_transport_trait::Session> ControlStreamAdapter<S> {
...
 	/// This reads from the control stream and routes messages to virtual streams,
-	/// and also drains the write queue to the control stream writer.
+	/// and also drains the write queue to the control stream writer. A decoded
+	/// GOAWAY is surfaced through `goaway` so [`crate::Session::goaway`] resolves
+	/// instead of the session tearing down.
 	pub async fn run(
 		&self,
 		reader: Reader<S::RecvStream, Version>,
 		writer: Writer<S::SendStream, Version>,
+		goaway: crate::goaway::Protocol,
```
kixelated: Should this be in the adapter? I guess, but it should be session.rs or something.

rs/moq-net/src/ietf/session.rs
```diff
@@ -270,9 +321,48 @@ async fn run_setup<S: web_transport_trait::Session>(
...
-	// Hold the writer alive until the session closes.
-	session.closed().await;
-	writer.finish().ok();
+	// Hold the writer alive until the session closes, sending a GOAWAY if the
+	// drain trigger fires meanwhile. The trigger resolves `None` when the session
+	// drops without draining; keep holding either way (closing this stream
+	// mid-session is a protocol violation on strict peers).
+	let payload = {
+		let mut closed = std::pin::pin!(session.closed());
+		let mut triggered = std::pin::pin!(goaway.triggered());
```
kixelated: ideally we can poll_closed and poll_triggered instead.

rs/moq-net/src/goaway.rs
```diff
@@ -0,0 +1,137 @@
...
+			*state = Some(received);
+		}
+		true
+	}
+
+	/// Wait for the send trigger to fire, returning the payload to encode.
+	///
+	/// Returns `None` if the trigger was dropped without firing (the session is
+	/// closing without a drain), so no GOAWAY should be sent.
+	pub async fn triggered(&self) -> Option<Payload> {
```
kixelated: poll_triggered

rs/moq-net/src/session.rs
```diff
@@ -110,6 +111,159 @@ impl Session {
 	pub async fn closed(&self) -> Error {
 		Error::Transport(self.shared.inner.closed().await)
 	}
+
+	/// Initiate a graceful GOAWAY drain of this session.
```
kixelated, line 115: Why do we need to call `Drain::start` if this initiates a drain?

rs/moq-net/src/session.rs
```diff
@@ -110,6 +111,159 @@ impl Session {
...
+	///
+	/// Returns a [`Drain`] handle; call [`Drain::start`] (or
+	/// [`start_with_timeout`](Drain::start_with_timeout)) to send the GOAWAY frame
+	/// and transition into the [`Draining`] state, then await
+	/// [`Draining::complete`] for the peer to disconnect.
+	///
+	/// Returns `None` when the negotiated version has no GOAWAY message
```
kixelated, line 121: IDK I think it's fine to just wait until the timeout then abort for older versions. It's not worth applications handling None.

rs/moq-net/src/session.rs
```diff
@@ -110,6 +111,159 @@ impl Session {
...
+	/// (moq-lite-03 and earlier), or when a drain is already in progress (only
+	/// one GOAWAY per session). The claim is released if the returned [`Drain`]
+	/// is dropped before starting, so a caller that bails out can retry later.
+	pub fn drain(&self) -> Option<Drain> {
+		// Pre-GOAWAY versions have no send path listening on the trigger, so a
+		// Drain would silently no-op and Draining::complete could hang.
+		if !self.version.has_goaway() {
+			return None;
+		}
+		// Atomically claim the drain so two concurrent callers can't both get a handle.
```
kixelated, line 131: Kind of weird that this returns None if unsupported OR it was called twice. Should return an Err on the second drain.

rs/moq-net/src/session.rs
```diff
@@ -110,6 +111,159 @@ impl Session {
...
+			session: Some(self.clone()),
+		})
+	}
+
+	/// Wait until a GOAWAY is received from the peer.
+	///
+	/// Resolves with the payload (redirect URI and optional deadline) once the
+	/// remote endpoint signals that this session should reconnect elsewhere.
+	/// Returns `None` if the session closes before a GOAWAY arrives.
+	pub async fn goaway(&self) -> Option<crate::GoawayReceived> {
```
kixelated, line 150: Not a fan of the name. `goaway` sounds like we're telling the peer to go away.

rs/moq-net/src/session.rs
```diff
@@ -110,6 +111,159 @@ impl Session {
...
+		kio::wait(|waiter| {
+			match self.goaway.received.poll(waiter, |state| match &**state {
```
kixelated, line 152: IDK this should be simplified. `ready!` at least. goaway::Handle should do this anyway, instead of reaching into internal state.

rs/moq-net/src/session.rs
```diff
@@ -110,6 +111,159 @@ impl Session {
...
+	// `Some` until start consumes it; `Drop` uses the remainder to release the claim.
+	session: Option<Session>,
+}
+
+impl Drain {
+	/// Send the GOAWAY frame with no deadline.
+	///
+	/// `uri` is the new session URI the peer should reconnect to; empty tells the
+	/// peer to reconnect to the same endpoint.
+	pub fn start(self, uri: impl Into<Arc<str>>) -> Draining {
```
kixelated: Clients are not allowed to send a URL, but they can still GOAWAY.

rs/moq-net/src/session.rs
```diff
@@ -110,6 +111,159 @@ impl Session {
...
+		// The drain claim stays set: only one GOAWAY per session.
+		Draining { session, timeout }
+	}
+}
+
+impl Drop for Drain {
+	fn drop(&mut self) {
+		// Not started: release the claim so a later drain() can retry.
+		if let Some(session) = &self.session {
+			session.goaway.draining.store(false, Ordering::Release);
```
kixelated, line 244: This is dumb, combine the start + drain method to avoid this.

rs/moq-net/src/session.rs
```diff
@@ -110,6 +111,159 @@ impl Session {
...
+	timeout: Option<Duration>,
+}
+
+impl Draining {
+	/// Wait for the peer to close the session after receiving the GOAWAY.
+	///
+	/// With a deadline (from [`Drain::start_with_timeout`]), the session is
+	/// force-closed with [`Error::GoawayTimeout`] when it expires; the timer is
+	/// cancelled if the peer closes first.
+	pub async fn complete(self) {
```
kixelated: Pretty weird API. Maybe impl Future instead, or return a Result.

rs/moq-relay/src/cluster.rs
```diff
@@ -348,6 +347,18 @@ pub struct ClusterConfig {
 	)]
 	#[serde(default, with = "humantime_serde")]
 	pub linger: Option<Duration>,
+
+	/// Fallback drain time in seconds for an upstream peer that sends a GOAWAY
+	/// without its own deadline. After a successful reconnect the relay keeps
+	/// the old upstream alive this long so its in-flight groups finish, then
+	/// force-closes it. A deadline carried on the received GOAWAY takes
+	/// precedence over this value. Defaults to 10 seconds.
```
kixelated: The deadline on the GOAWAY should not take precedence; we should take the min.

rs/moq-relay/src/cluster.rs
```diff
@@ -348,6 +347,18 @@ pub struct ClusterConfig {
 	)]
 	#[serde(default, with = "humantime_serde")]
 	pub linger: Option<Duration>,
+
+	/// Fallback drain time in seconds for an upstream peer that sends a GOAWAY
+	/// without its own deadline. After a successful reconnect the relay keeps
+	/// the old upstream alive this long so its in-flight groups finish, then
```
kixelated, line 353: I'm kinda unsure why this exists? The idea is that the client should establish the new connection in parallel. I'm not sure who closes the connection when it's fully drained, but I wouldn't keep a dead connection "alive" for this arbitrary amount of time.

rs/moq-relay/src/cluster.rs
```diff
@@ -929,12 +940,109 @@ impl Cluster {
...
 		}
-		let cs = client
+		let mut session = client
 			.connect(url.clone())
 			.await
 			.context("failed to connect to cluster peer")?;
 
-		Err(cs.closed().await.into())
+		// Fallback drain window for an upstream GOAWAY that carries no timeout.
+		let drain_timeout =
```
kixelated: IDK, shouldn't this whole thing be in `Reconnect` instead? The code looks pretty terrible.

rs/moq-relay/src/cluster.rs
```diff
@@ -977,6 +1085,63 @@ fn connect_api_is_http(source: &str) -> bool {
...
+	if scheme_security_tier(parsed.scheme()) < scheme_security_tier(fallback.scheme()) {
+		tracing::warn!(
+			redirect_scheme = parsed.scheme(),
+			current_scheme = fallback.scheme(),
+			uri,
+			"GOAWAY redirect is a security downgrade; falling back to original URL"
+		);
+		return fallback.clone();
+	}
+	// Carry the cluster JWT over to the redirect target unless it brings its own.
```
kixelated: Yeah don't do this? The server SHOULD echo back the JWT in the GOAWAY URL. It seems pretty useful for forcing the client to switch auth tokens or something.

kixelated (changes requested):
It's a good first attempt and most of the stuff is salvageable.

I would want to see client support via Reconnect in moq-native and moq/net. It should be transparent to the client and there should be (at least) two concurrent connections. Starting with the cluster connections is okay but less testable, and really cluster connections should be leveraging moq-native instead.

We also need some way of marking the old connection/route as lower priority so we migrate the subscription. I know it's a follow up, but it seems required for any sort of draining. We need to somehow cost a draining connection as infinite so its never used as a primary.

ksletmoe-aws:
@kixelated — pushed an update working through the review. What's in this revision:

**API-shape feedback (all addressed):**
- `Drain::start` / `start_with_timeout` → a single `start(DrainConfig)` with a builder
- `triggered()` → `poll_triggered`; the lite + IETF session loops now drive off the poll methods
- `Draining::complete` returns `Result` so a graceful close is distinguishable from a timeout-forced abort
- moved the GOAWAY plumbing out of the IETF adapter into the session driver
- `drain_timeout` → humantime `drain` / `cluster.drain` (old keys kept as hidden deprecated aliases)
- drain deadline is now `min(local, goaway)` rather than the GOAWAY deadline winning outright
- dropped the JWT carry-over on redirect; the sending server echoes the JWT in the GOAWAY URI instead

**The part I'd most like your eyes on — the reconnect API:**
I built the transparent client-side migration in `moq-native`: `Reconnect` gains a `Drain` config and a `Status::Migrating`, and on GOAWAY it dials the redirect in parallel, drains the old session for `min(drain, goaway)`, and falls back to backoff if the redirect fails. The "cost a draining connection as infinite" piece lives down in `moq-net`: on GOAWAY receipt the subscriber bumps its own routes to `broadcast::DRAIN_COST`, so `best_route` migrates live subscribers off the dying connection — no new public API, works for both lite and IETF. Since you flagged the interface as the uncertain part, I'm posting this shape specifically for you to react to before I build on it.

**Held back deliberately:** I haven't touched `cluster.rs` or the `@moq/net` `Reload` mirror yet. Both consume this reconnect API, and there's no point mirroring something we might reshape — I'll do the cluster rewrite and the JS side once we've settled the reconnect API.

**Target branch:** this includes breaking `moq-net` pub API changes (`Drain::start` signature, `Draining::complete` return type), so per the branch-targeting rules I think this should target `dev` rather than `main`. Let me know if you agree or want it split differently.

(Written by claude-opus-4.8)

ksletmoe-aws:
@kixelated pushed a few more commits, mostly housekeeping to get this mergeable again:

- merged current `main` in to clear the conflicts. One real conflict in `moq-net/src/client.rs`: your `poll_ready(waiter)` refactor from #2515 against the added `start.goaway` arg on `Session::new`, kept both.
- dropped the relay's `MOQ_DRAIN_TIMEOUT` handling. It collided with the `--drain-timeout` env on the moq-native reconnect drain, so setting the client drain was silently reconfiguring the relay's shutdown drain and warning at a correct use.
- stripped the deprecation aliases on the drain config keys. I had added hidden aliases for `drain_timeout` -> `drain`, then realized `drain_timeout` only ever existed on this branch, so I was carrying back-compat machinery for a name we invented and never shipped. One name per setting now.

I also saw #2542. That looks like a cleaner shape than what I had, and I am happy to just follow up there instead of here if you would rather. Say the word and I will close this one out.

One thing that looks missing from #2542 though, and it is the bit you originally flagged: I do not see the draining route getting deprioritized. #2542 keeps the old routes attached at their normal cost and lets tracks hand over at a group boundary, which works when the replacement is genuinely cheaper, but on equal cost `route_order` falls through to hop count and then hash, so the draining connection can stay primary. That is the "cost a draining connection as infinite so it is never used as primary" case.

What I have on this branch for that: on GOAWAY receipt the subscriber bumps its own routes to `broadcast::DRAIN_COST` (`u64::MAX`), so `best_route` moves live subscribers off the dying connection. It sits at the session layer rather than in `Reconnect`, since that is where the broadcast producers actually live, so it needs no new public API and both lite and IETF get it. Happy to port that onto `goaway-api` as a separate commit if you want it, or drop it if you would rather solve the handover a different way.

(Written by claude-opus-4.8)

ksletmoe-aws:
Closed in favor of https://github.com/moq-dev/moq/pull/2542

Issue #2478, opened 2026-07-24
moq-audio: playback engine (device output, mixing) + echo cancellation, ported from iroh-live


kixelated:
Part of the iroh-live merge (#2481). This is the largest single audio asset: iroh-live's full-duplex device engine (`moq-media/src/audio_backend.rs` + `audio_backend/aec.rs`, ~2.8k LOC), which fills the two gaps that block native conferencing per #2282: moq-audio has no playback path at all, and AEC needs the far-end reference from a playback path we don't have.

It also answers #2282's central open question. We believed no good pure-Rust AEC existed; iroh-live ships one via [`sonora`](https://crates.io/crates/sonora) (a pure-Rust port of WebRTC audio-processing: AEC3, noise suppression, AGC), on crates.io, no CMake, no C++ dep. The pure-Rust stance holds.

## What to port

The engine, not its API. Internals carry over close to as-is (this is the part that is hard to get right after the fact):

- A dedicated driver thread owning the cpal streams, driven by a bounded command channel.
- Output: mixing of N registered streams with per-stream volume, a declicker fade state machine on pause/resume (~3 ms fades), clamping, smoothed peak metering.
- Input: channel mapping (mono/N-channel devices to stereo), in-place AEC, fan-out to N registered consumers (one mic feeds any number of encoders).
- Realtime discipline: preallocated lock-free resampling ring buffers at every stream endpoint (`fixed-resample`), bounded control channels, allocation-free callbacks.
- Recovery: live device switching without recreating caller-side streams, restart with exponential backoff on stream errors/underruns.
- AEC: the output callback writes the post-mix render reference into a ~100 ms lock-free SPSC ring; the input callback drains it and runs render+capture 10 ms frames serialized on one thread (no cross-callback locking). Runtime toggle, no stream restart.

## The moq shape (where we deviate from a straight port)

- Two non-default features mirroring `capture`: `playback` (cpal, fixed-resample, ringbuf) and `aec = ["playback", "capture", dep:sonora]`. Default builds compile none of it.
- Role-module surface per #2321's alignment direction: a `playback` module with a `Config` options bag and a small owned sink/stream handle. Drop closes the stream; no user-supplied hooks or trait glue (their `AudioSink`/`AudioSource`/`AudioStreamFactory` traits are dropped; we dispatch on concrete types). Structs `#[non_exhaustive]`.
- Device enumeration async, returning `id` + `name` + `default`, consistent with the video enumerators (#2321).
[...]

Issue #2282, opened 2026-07-15
moq-audio: echo cancellation / AGC / noise suppression for native capture


kixelated:
You cannot build a native MoQ conferencing client today, because two participants on speakers will howl. The browser gets echo cancellation, noise suppression, and auto gain control free from getUserMedia; native gets nothing. That asymmetry blocks every native two-way use case.

## What exists today

**`rs/moq-audio` does codec + format + resample, and no processing:**
- `codec.rs` — Opus `Encoder`/`Decoder` via `unsafe-libopus` 0.2 (pure-Rust transpile of libopus 1.3.1)
- `format.rs` — `AudioFormat` (mirrors WebCodecs `AudioData.format`), layout conversion to interleaved f32
- `resample.rs` — `Resampler`, rubato 3.0, sinc only. **Sample-rate only**; channel up/downmix is rejected upstream.
- `producer.rs` / `consumer.rs`, `frame.rs`, `error.rs`
- `capture.rs` + `capture/permission.rs`

No echo cancellation, no AGC, no noise suppression, no VAD, no mixer. The only signal-conditioning-adjacent knobs are Opus-internal (DTX, `signal=voice`), and those are **JS-side only** (`js/publish/src/audio/types.ts`, `Kind = "voice"|"music"|"auto"`) — the native encoder has no equivalent plumbing.

**Native mic capture already exists**, which is what makes this gap actionable rather than theoretical: `rs/moq-audio/src/capture.rs` (behind the `capture` feature) uses **cpal 0.18** (CoreAudio/WASAPI/ALSA) with `capture::Config { device, sample_rate, channels }` and `capture::Microphone` (realtime callback → `mpsc` → interleaved-f32 `Frame`s). `capture/permission.rs` handles macOS TCC up front (objc2 + objc2-av-foundation), and `FIRST_BUFFER_TIMEOUT` = 5s surfaces a denial as an error instead of a silent hang. Wired to the CLI via `ImportSource::Capture` (`--microphone`, `--audio-bitrate`).

So we can capture a mic natively. We just can't make it usable in a two-way call.

[...]

Issue #2272, opened 2026-07-15
moq-cli: remaining work for capture and playback (window/app capture, device enumeration, native player)


kixelated:
Tracking issue for the gaps in `moq import capture` and the (nonexistent) native playback path, in the same spirit as #1837 (which tracks the `moq-video` backend work this sits on top of). Audit done against `rs/moq-cli`, `rs/moq-video`, and `rs/moq-audio` on the `dev` branch.

Goal: `moq` should be able to capture **a window / an app / a screen / a microphone or device**, and play a broadcast back **like `ffplay`**, without an external ffmpeg.

Current state:
- **Capture** (`moq import capture`, feature-gated): camera (AVFoundation / Media Foundation / V4L2) and whole-display (ScreenCaptureKit / DXGI Desktop Duplication / xdg-desktop-portal + PipeWire), encoded to H.264/H.265; microphone via cpal, encoded to Opus. Flags today are `--camera`, `--screen`, `--width/--height/--fps/--bitrate/--codec`, `--hardware/--software`, `--microphone`, `--audio-bitrate`, `--no-video/--no-audio`.
- **Playback**: none. `moq export <container>` writes fmp4/mkv/ts/flv/h264/h265 to stdout; playing it means piping to ffplay (`doc/bin/cli.md:205`). `moq-video::decode` exists on `dev` but `moq-cli`'s export path never calls it.

---

### Shipping: no released binary can capture at all

- [ ] **`capture` is off by default and no shipped artifact turns it on.** `rs/moq-cli/Cargo.toml` omits `capture` from `default`, and every distribution builds default features: Nix (`nix/overlay.nix:59`, `cargoExtraArgs = "-p moq-cli"`), Docker, winget, and `cargo install moq-cli`. So `moq import capture` is unreachable for every user who doesn't build from source with `--features capture`. The feature is off for good reasons (Linux `capture` pulls libclang + V4L2 headers via bindgen, and ALSA via cpal), so pick one: default-on with the existing trimmable sub-features, or a separate `moq-cli-capture` artifact. (`doc/bin/cli.md` does document the `--features capture` build, so this is a packaging gap, not a documentation one.)

### Capture sources

[...]

PR #2529, opened 2026-07-27
feat(moq-audio): play decoded PCM out a speaker

kixelated:
## Summary

moq-audio could capture, encode, and decode but never play anything. That is the first of the two blockers #2282 names for native conferencing, and what a native player (#2272) needs from us. This ports the **playback half** of iroh-live's device engine (#2478, part of #2481) as a `playback` role module behind a non-default feature, alongside `capture`.

`Engine` owns the output device and `Sink` is one stream playing through it. The split is deliberate: a call with several participants mixes into one device stream rather than opening one per speaker, and echo cancellation will need a single post-mix render reference from exactly there.

The engine, not its API, is what carried over. Where their shape and ours differ:

- No `AudioSink`/`AudioSource`/`AudioStreamFactory` traits or boxed handles; concrete types and a `Config` options bag.
- **No 4-state declicker plus a separate pause flag.** Volume changes ramp over ~3 ms, so pause is `set_volume(0.0)` and one mechanism declicks every gain change. A muted sink keeps draining its input rather than queueing up and jumping when it unmutes, since this is live audio.
- **Peak metering reports the raw peak since the last read, per sink.** Smoothing and dB normalization are presentation policy and belong in the caller. Theirs also stored the shared mix peak on every stream, so per-stream meters all read alike.
- `Engine::open` fails fast when there is no device, instead of logging and handing back a backend that plays into nothing.
- Sink registration goes through shared state rather than a request/reply to the driver, so `Engine::sink` is synchronous and never waits behind a device open. Only device-plane work (switch, restart) is a driver message.
- **The driver waits on one channel with a deadline, and cpal's error callback posts to it.** Theirs drained errors only when an unrelated message arrived, so a device that failed while the app was quiet was never restarted.
- Rate conversion is per sink, caller rate straight to device rate. There is no fixed internal rate, only a stereo mix bus.

### Two bugs found while validating on hardware

1. **`plughw:` devices could not be opened.** An ALSA `plughw:` node advertises every format ALSA can convert to, **starting at 8-bit**, so choosing a stream config by sample rate alone handed back a format we cannot write (`unsupported audio configuration: output sample format I8`). Format selection is now filtered to writable formats before rate preference applies, and every `plughw:` node on the test box opens.

2. **The Nix dev shell could not reach ALSA's default device** (second commit). On any desktop running a sound server the default device is a plugin (`pipewire`, or `pulse` on an older setup), and the shell's alsa-lib only searches its own store path, which holds neither. Opening the default failed with ENXIO, so audio work was limited to raw `hw:` devices. The host's plugins are not a substitute: they link the host's libasound and libpipewire, so loading one into a Nix-built binary hits a glibc version mismatch. `pipewire` was already in the shell for moq-video's screen capture and ships the plugin, so this is just `ALSA_PLUGIN_DIR` pointing at it (joined with `alsa-plugins` for the PulseAudio case). This also affects the existing `capture` device tests.

### Dependency

One new crate: `fixed-resample` (MIT), a preallocated ring buffer that resamples and corrects drift without allocating or locking in the audio callback. It wraps the same `rubato` we already depend on at the same version, so it adds no second resampler and no duplicate in the tree. `cpal` comes from crates.io 0.18, which already has the `DeviceId` / `device_by_id` APIs iroh-live pinned a git rev for, so the "no git dependency" prerequisite in #2478 is satisfied.

## Public API changes

All additive, all new, all behind the non-default `playback` feature except the error variant. No renames, removals, or signature changes, so this targets `main`.

`moq_audio::playback` (new module):
- `Engine` + `Engine::{open, sink, switch}`
- `Sink` + `Sink::{write, buffered, input, control, set_volume, volume, peak}`
- `Control` (clonable) + `Control::{set_volume, volume, peak}`
- `Config` (`#[non_exhaustive]`, `Default`), `Input`, `Device`, `devices()`

`moq_audio::Error`:
- New `Playback(String)` variant. `Error` is already `#[non_exhaustive]`, so this is additive. Present in every build, not just `playback` ones (it carries no dependency).

Everything else in the module (`Mixer`, `Gain`, `Registration`, the driver) is `pub(crate)`/`pub(super)`.

## Test plan

- `just check` (default features): clean.
- CI-equivalent feature passes, since `just check` does not cover non-default features: `cargo check --workspace --no-default-features`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --all-features --no-deps`. All clean.
- `cargo nextest run --workspace --all-targets --all-features`: 2341 passed, 7 skipped (the `#[ignore]`d hardware tests).
- Hardware-free unit tests (9, run in CI): summing and clipping, gain-ramp monotonicity over the full ramp, per-sink peak and reset, sink removal, silence with no sink, mono downmix, surround silence past the front pair, underflow reading as silence, and that writes are discarded until the device reads.
- Device round trip (4, `#[ignore]`d, `rs/moq-audio/tests/playback.rs`): a tone plays and meters, two sinks mix with volume honored, switching across every openable device keeps the sink draining, and a sink outlives the `Engine` it came from. **Run and passing twice: against a raw ALC1220 analog output, and (after the flake fix) against the system default through PipeWire, with no manual environment.**

Not verified on macOS or Windows: no hardware to hand. The cpal path is the same one `capture` already uses on those platforms.

## Cross-Package Sync

No row in the table applies. Two notes on what is deliberately skipped:

- **`doc/`**: moq-audio has no `doc/` page at all today (nor does moq-video), so rustdoc module docs are the documentation surface. A page covering capture/encode/decode/playback together is worth its own PR rather than a playback-only stub. #2481's release gate asks for one before the wave ships.
- **`moq-ffi` and the language wrappers**: explicitly out of scope per #2478.

## Follow-up

AEC is the other half of #2478. It needs the mixed render reference this PR lands, plus the license and maturity check on `sonora` that #2481 asks for.

Refs #2478, #2481, #2282, #2272

https://claude.ai/code/session_019zwRXGuB6oaAWkVJZUJR2K

(Written by Claude Opus 5)


kixelated:
All three confirmed against the code and fixed in 953b37d. Each has a regression test that I verified fails with its fix reverted.

**P1, lost mixer commands.** Correct, and the removal half is the worse of the two: a refused `Remove` was dropped outright, so the mixer kept reading a sink nobody owns for the life of the stream. Refused updates are now recorded (`State::detaching` for removals; the pending consumer already existed for registrations) and replayed by `Shared::sync`.

The missing piece was that nothing called a retry, so `Engine::sink` and `Sink::drop` now wake the driver, which runs `sync` and retries for a few device periods. There is a test for the replay and a separate one for the wake, since the replay is useless if nobody triggers it.

**P1, heap work in the callback.** Correct, and it contradicted the invariant in the module's own header. Both halves are gone:

- The entry list is allocated once at `MAX_SINKS` (64) and the driver refuses to register past it, so `Engine::sink` returns an error rather than handing back a sink that plays nothing. That is a new public constraint, called out in the `sink` docs.
- Removed entries go back to the driver over a channel and are dropped there. The channel holds `MAX_SINKS`, and filling it takes a driver that has stopped running entirely; the fallback drops in place, which costs one free rather than leaking the entry for the life of the stream.

Asserting on the absence of a `free` is not directly testable, so the tests assert the mechanism instead: the entry list's capacity is unchanged after `MAX_SINKS` registrations, and a removed sink arrives on the retirement channel.

**P2, stale failures.** Correct. Streams now carry a generation and `Command::Failed` is tagged with it; `Driver::should_restart` ignores anything from a replaced stream.

One note on the queue depth, since it is adjacent to your first finding: it is now `2 * MAX_SINKS` rather than a flat 64. That is not the fix, only a reason the retry should stay rare, given add/remove churn is not bounded by the live sink count.

`just check`, the all-features clippy/doc/nextest passes, and the four device tests against the default output are all green.

(Written by Claude Opus 5)

kixelated:
Reviewed my own diff and worked through the bot findings. Fixed in c98ebf1.

**One real bug, and it was the best catch of the round.** `f32::clamp` propagates NaN rather than clamping it, so `set_volume(f32::NAN)` stored NaN, the gain ramp carried it into the applied gain, and it never washed out: every sample that sink contributed was NaN for the life of the stream. `Control::set_volume` is public and takes an arbitrary `f32`, so a UI slider dividing by zero reaches it. Non-finite volumes are now ignored, with two tests I verified fail without the guard (the setter keeps the old volume; the device buffer stays finite).

**A silent-death path.** An error kind the driver had no rule for logged and returned, leaving nothing to wake it. If such an error was terminal, playback stopped for good with one warning. Unclassified errors now escalate to a rebuild after a few in the same window that counts underruns, and `DeviceChanged` / `RealtimeDenied` are split out as non-fatal since cpal documents them as survivable.

**API, taken while it is still free.** `Engine::switch` now takes a `Config` instead of a bare device id, matching `open`, so a future playback option applies to a switch without a breaking change.

Also: `Driver::restart` replaces three different spellings of "rebuild and schedule a retry if that failed"; a sustained overflow logs once per spell rather than ~100 lines/sec; the crate map lists the playback types it had missed; the hardware-test doc points at `nix develop` (bare `cargo` misses the ALSA plugin path); and that doc comment no longer claims every enumerated device opens, since the loop skips the ones that don't.

Two I'm declining:

- **`#[non_exhaustive]` on `playback::Input`.** It mirrors `encode::Input` exactly and describes a closed concept, the PCM layout of the buffers you hand in: format, rate, channels. The attribute also forbids struct literals outright, including `..Default::default()`, which is the primary way to build one and what the doc example uses. `Config` earns it because it is an options bag with one rarely-set field. A future per-sink knob (a latency target) belongs in a sink config, not in the layout description.
- **Rejected `mixer::Command`s are never re-delivered.** Already fixed in 953b37d, before this review ran: refused commands are recorded in `Shared` and replayed by the driver.

`just check`, the all-features clippy/doc/nextest passes, `--no-default-features`, and the four device tests against the default output are all green. Enabling auto-merge.

(Written by Claude Opus 5)

Issue #2330, opened 2026-07-17
Add an epoch to announcements so a same-origin republish is detectable


kixelated:
## Context

#2241 makes subscriptions migrate transparently across connections. A restart (ANNOUNCE_RESTART on lite-06, duplicate ANNOUNCE on lite-05) is applied as an in-place route update only when the first hop of the reconstructed chain is unchanged: the first hop identifies the original publisher, so a different first hop means the broadcast was replaced and the subscriber reloads (Ended + Active, cached TRACK_INFO discarded). The draft now specifies this under ANNOUNCE_RESTART.

## Limitation

The first hop identifies the *publisher*, not a particular *broadcast instance*. A publisher that deliberately ends a broadcast and republishes new content at the same path advertises the same first hop, so downstream cannot tell the republish from a route change: it resumes subscriptions and keeps cached TRACK_INFO across what is actually new content (unrelated group sequences, possibly a different timescale). The same ambiguity underlies the same-path standby assumption: two routes announcing one path are assumed to carry the same content.

## Proposal

Add an explicit `epoch` to announcements (ANNOUNCE_START / ANNOUNCE_RESTART), assigned by the origin per broadcast instance and forwarded unchanged by relays:

- restart with same first hop + same epoch: alternate route, resume subscriptions, TRACK_INFO stays valid
- same first hop + different epoch: the same origin explicitly made a new broadcast; reload (END+START semantics)
- routes for the same path only join as standbys when first hop and epoch both match, making the backup-route assumption exact

Natural fit with the lite-06 announce-id work (and the cost-on-the-wire field deferred from #2241).

fperex:
## A measured case where undetectable republish costs a subscriber minutes of data

We ran into this from the consumer side and have numbers that might be useful motivation, so
recording them here rather than opening anything new.

Setup: stock `moq-relay` binaries, a publisher session writing two tracks on one connection
(50 groups/s and 1 group/s), a subscriber that subscribes once and also does the usual
`reload: true` thing — watch the announcement stream, re-attach when the broadcast is announced
active again. We `SIGKILL` the publisher and start a second one on the same path a second later.

On **v0.14.1** the subscriber sees `active=false`, its reload fires ~250 ms later, and it loses
nothing (`1158/1158` groups delivered).

On **v0.14.2+** the announcement correctly stays `active` for the linger window — that is
#2469's documented intent, *"consumers never observe the gap"* — but the replacement publisher is a
[...]

PR #2534, opened 2026-07-27
fix(moq-net): deliver groups from a takeover source that restarted its numbering

wrangelvid:
## Summary

- Root cause: the linger takeover (#2469) splices a replacement source at `latest + 1`, assuming it resumes the old sequence space. A reconnecting publisher is a fresh session with a fresh identity and restarted numbering, so the boundary starves it twice: the demand floor asks it for groups it will never number, and late-join floors resolved against the pre-restart `latest()` filter anything that does arrive. One-group-per-event feeds lose the next N events after every publisher reconnect.
- The takeover now gates on the announce's first hop, the producer-instance identity. A matching first hop splices at the boundary exactly as before (eager demand floor, unchanged route behavior). A differing first hop is a new broadcast: the logical track resets onto it unbounded and a numbering era bumps, clearing subscriber cursors and late-join floors anchored in the old numbering.
- A same-identity source producing below the splice boundary logs a one-shot warning (an identity-contract violation that would starve those groups); the `Origin` doc now states that ids are minted per producer instance.
- The route tests' synthetic hops are adjusted so alternate routes to the same producer share a first hop; the restart tests model a reconnect as a fresh session with a fresh identity.

## Public API changes

None. New items are `pub(crate)` (`resume::Producer::reset_onto`, `resume::Producer::splice_boundary`, `track::Consumer::poll_latest_changed`). One doc addition on `Origin`.

## Test plan

- `test_linger_reconnect_restarted_sequences_still_deliver` fails on main, passes here.
- `test_linger_restart_resets_late_join_floors` fails without the era cursor reset, passes here.
- `test_linger_reconnect_splices` and the route handover/failover/cost tests pass with their original assertions; 559 moq-net tests green.
- An out-of-tree integration suite driving the relay binary reproduces the starvation on stock 0.14.2/0.14.3 (an event-feed test failing ~90% of runs after a publisher reconnect, a publisher-rotation test starving 3/5); both pass consistently with this branch's relay.

(Written by Claude Fable 5)

### DIFF
--- a/rs/moq-net/src/model/origin.rs
+++ b/rs/moq-net/src/model/origin.rs
@@ -3676,6 +3676,65 @@ mod tests {
 		sub.assert_not_closed();
 	}
 
+	/// A replacement source that RESTARTS its group numbering (a fresh session's
+	/// fresh producer beginning at sequence 0, which is what a real reconnecting
+	/// publisher does) must still reach the spliced consumer. The resume hint
+	/// (`group_start`) is advisory for a source that resumes; it must not filter
+	/// a restarted source's groups into a permanent stall.
+	#[tokio::test(start_paused = true)]
+	async fn test_linger_reconnect_restarted_sequences_still_deliver() {
+		let origin = Info::new(Origin::random())
+			.with_linger(Duration::from_secs(5))
+			.produce();
+		let consumer = origin.consume();
+		let mut announced = consumer.announced();
+
+		let hops = OriginList::try_from(vec![Origin::new(1).unwrap()]).unwrap();
+		let source = origin
+			.create_broadcast("test", announce().with_hops(hops.clone()))
+			.unwrap();
+		let mut dynamic = source.dynamic();
+		settle().await;
+		settle().await;
+		let broadcast = consumer.request_broadcast("test").await.unwrap();
+		announced.assert_next_some("test");
+
+		let subscribing = broadcast.track("video").unwrap().subscribe(None);
+		let mut producer = accept_track(&mut dynamic, "video").await;
+		settle().await;
+		let mut sub = subscribing.await.unwrap();
+
+		producer.append_group().unwrap();
+		producer.append_group().unwrap();
+		assert_eq!(sub.assert_group().sequence, 0);
+		assert_eq!(sub.assert_group().sequence, 1);
+
+		// The session dies without unannouncing: the broadcast lingers.
+		drop(producer);
+		source.abort(Error::Dropped).unwrap();
+		drop(dynamic);
+		settle().await;
+		sub.assert_no_group();
+		sub.assert_not_closed();
+
+		// The publisher reconnects as a NEW session: fresh broadcast, fresh
+		// producer, group numbering restarted from zero.
+		let source = origin.create_broadcast("test", announce().with_hops(hops)).unwrap();
+		let mut dynamic = source.dynamic();
+		settle().await;
+		settle().await;
+
+		let mut producer = accept_track(&mut dynamic, "video").await;
+		settle().await;
+
+		// The restarted source publishes an event the consumer must see.
+		producer.append_group().unwrap();
+		settle().await;
+		let group = sub.assert_group();
+		assert_eq!(group.sequence, 0, "the restarted source's first group is delivered");
+		sub.assert_not_closed();
+	}
+
 	/// The linger window expiring without a replacement closes the broadcast: the
 	/// path unannounces, tracks abort, and a later re-create is a fresh broadcast.
 	#[tokio::test(start_paused = true)]
--- a/rs/moq-net/src/model/origin.rs
+++ b/rs/moq-net/src/model/origin.rs
@@ -1526,11 +1526,19 @@ async fn serve_track(state: kio::Producer<FrontState>, name: Arc<str>, mut resum
 		Idle,
 		/// A reader arrived or the last one left: recompute the demand gate.
 		Demand,
+		/// The freshly spliced copy produced a group: decide whether its source
+		/// resumed the old numbering or restarted its own.
+		Produced(Option<u64>),
 	}
 
 	let mut fails = 0u32;
 	// The source whose copy is currently spliced in, and that copy.
 	let mut serving: Option<(u64, track::Consumer)> = None;
+	// Restart watch over the newest splice: the last production edge observed
+	// on the serving copy, and whether the resumed-vs-restarted question is
+	// already settled (see resume::Producer::reanchor_restarted).
+	let mut splice_latest: Option<u64> = None;
+	let mut splice_settled = true;
 	// A source whose splice failed because it had already closed. Its watcher is
 	// about to detach it, so wait for the table to move on rather than burning
 	// the strike budget on a corpse (ids are never reused, so this cannot wedge).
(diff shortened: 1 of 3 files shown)


kixelated:
I'll check in a second, but we shouldn't be trying to resume broadcasts with different origin IDs. Like the relay needs soooome indication that a broadcast is new versus continuous, and right now we should be using the first hop as that flag.

wrangelvid:
> I'll check in a second, but we shouldn't be trying to resume broadcasts with different origin IDs. Like the relay needs soooome indication that a broadcast is new versus continuous, and right now we should be using the first hop as that flag.

Oh I didn't see that. I reworked this by keeping the boundary splice on the same first hop, and a different one resets the logical track onto the new source with an era bump, which should also clear later join floors resolved against the old numbering.

fperex:
## Independent reproduction of the restarted-numbering case, with a version bisect

We hit this in production (a meeting app: peers publishing Opus + H.264 through a relay, one peer's
track going silent for minutes while the session stayed healthy) and then reproduced it offline
against stock release binaries. Posting the data since this PR is the active thread on it.

Worth noting up front: #2469's own **Deferred** list already called this out —

> Splice-with-a-restarted-publisher (group sequences resetting) relies on the existing resume
> machinery; the same-publisher case is what this PR tests.

This is that deferred case, measured.

### Setup

Two ~60-line scripts using `@moq/net` directly. No browser, no media, no application code.

- **Publisher** — one session, one broadcast, **two tracks on the same connection**: `fast` at
  50 groups/s and `slow` at 1 group/s (one group per event, matching how `js/publish` writes Opus —
  each audio frame is its own group — versus video's one group per GOP).
- **Subscriber** — subscribes once, and also does what real clients do
  (`Watch.Broadcast({reload: true})`): watches the announcement stream and re-attaches when the
  broadcast is announced active again.
- **Runner** — attaches the subscriber, `SIGKILL`s the publisher (ungraceful loss), starts a second
  publisher on the same path 1 s later, reports per-track delivery.

Stock binaries, `demo/relay/localhost.toml` (anonymous auth, self-signed), client over the WebSocket
fallback, ALPN `moq-lite-05`. Each run ~45 s.

Two harness properties we needed before the numbers meant anything, in case anyone re-runs this:

1. **A unique broadcast path per run.** Reusing one path contaminates the next — the previous
   generation's high-water mark is still held for that path, so the next run's *first* publisher is
   itself treated as a restarted takeover source and starves from the start.
2. **A validity gate.** Only count runs where the subscriber was demonstrably streaming *before* the
   cut; a failed initial subscribe is indistinguishable from "starved forever" in the output. Four
   of our first five runs were invalid this way.

### Result

Identical scenario, ~16 s of uptime before the loss:

| | last seq delivered before the cut | new source produced | first seq delivered after | gap before delivery resumed | delivered / produced by the new source |
| --- | --- | --- | --- | --- | --- |
| **v0.14.1** `fast` | 764 | 0..1157 | **0** | 0.2 s | **1158 / 1158** |
| **v0.14.1** `slow` | 16 | 0..24 | **0** | 0.2 s | **25 / 25** |
| **v0.14.4** `fast` | 767 | 0..1165 | **768** | 16.4 s | 398 / 1166 |
| **v0.14.4** `slow` | 16 | 0..24 | **17** | 17.0 s | 8 / 25 |

On v0.14.2+ the first group delivered after the reconnect is exactly `last_delivered + 1` — the
splice boundary — and everything the replacement produced below it never arrives. On v0.14.1 the
replacement's own numbering is accepted from 0 and nothing is lost. (The v0.14.1 gap is just our
client's 250 ms reload backoff; those groups then arrive from the backlog, hence `1158/1158`.)

### Version bisect

Same scenario at a 10 s uptime, 3 runs per version, unique path per run, invalid runs discarded
(20/20 valid):

| moq-relay | announcement withdrawn | `fast` starved for | delivered / produced by the new source |
| --- | --- | --- | --- |
| **v0.14.1** | yes | 0.2s, 0.2s, 0.2s | **1391 / 1391** |
| **v0.14.2** | no | 11.4s, 11.6s, 11.6s | 860 / 1394 |
| **v0.14.3** | no | 11.4s, 11.5s, 11.5s | 854 / 1385 |
| **v0.14.4** | no | 11.5s, 11.5s, 11.5s | 850 / 1380 |

v0.14.2 is the first release containing #2469.

### It scales with the broadcast's age at the moment of loss

v0.14.4, 2 runs each. Our publisher runs 1.5 s before the subscriber attaches, so broadcast age at
loss = listed uptime + 1.5 s — which the starvation column reproduces to within 0.2 s:

| uptime before the loss | `fast` (50 groups/s) | `slow` (1 group/s) |
| --- | --- | --- |
| 5s | 6.5s, 6.5s | 7.0s, 7.0s |
| 10s | 11.5s, 11.4s | 12.0s, 12.0s |
| 20s | 21.5s, 21.5s | 22.0s, 22.0s |
| 30s | 31.7s, 31.2s | 32.0s, 32.0s |

On a long-lived session that is effectively permanent from a user's point of view.

Note the duration is the same for **every** track regardless of group rate, because each track's
boundary scales with its own cadence. We had expected high-rate tracks to be hit disproportionately
in wall-clock terms; measuring showed otherwise.

### Interaction with the linger's "invisible gap" promise

This part isn't a defect report — the behaviour is exactly what #2469 documents — but the
combination is what made it expensive for us to diagnose, so it may be worth recording.

#2469's design is that within the window *"the path stays announced … so consumers never observe the
gap"*, pinned by `test_linger_reconnect_splices` (*"ungraceful loss is invisible during the window
(no unannounce, no track error)"*). With a same-publisher reconnect that promise holds.

With a **restarted** publisher the invisibility inverts. We confirmed the announcement behaves as
designed — it goes `active=true` … `active=true` with no `active=false` — but there is now a real
gap behind it:

```
v0.14.1     +1.5s  ANN active=true
           +16.5s  END fast / END slow
           +16.5s  ANN active=false      <-- loss surfaced
           +16.7s  SUB fast / SUB slow   <-- client reload fires
           +17.5s  ANN active=true
           +17.7s  SUB fast / SUB slow   -> continuous delivery, nothing lost

v0.14.4     +1.5s  ANN active=true
           +17.5s  ANN active=true       <-- correctly no unannounce; but no END, no resubscribe
                                             and no delivery for 16.4s
```

So pre-linger the failure was loud and client-recoverable; in the restarted-publisher case it is
silent and the client's own reload never engages, because there is nothing for it to engage on. That
seems closely related to #2330 (*"Add an epoch to announcements so a same-origin republish is
detectable"*) — our numbers are a concrete instance of why detectability matters here, and we've
added them there too.

### Where it starts, for anyone confirming

`resume.rs` `Producer::takeover()` splices the replacement at `latest + 1` and `Segment::produced()`
discards groups below the segment start, which matches the observed `last_delivered + 1` boundary
exactly. `resume.rs` carries no tracing, so none of this appears in the relay log even at
`level = "debug"` — we only found it by counting group sequences at the subscriber.

### What we have not established

- We could not determine which build our production relay runs (`/health` and `/metrics` aren't
  reachable from outside, no version header), so we can't claim our incident is this code path —
  only that the symptom matches and that it reproduces on v0.14.2+.
- Our production symptom had one track silent while another kept flowing on the same session. The
  reproduction above starves all tracks equally, so it does **not** explain that asymmetry, and we
  didn't want to imply otherwise. We're still chasing that end separately.

### Environment

macOS 26.5.1 arm64 · stock `moq-relay` release binaries v0.14.1 / v0.14.2 / v0.14.3 / v0.14.4 ·
`@moq/net` client over the WebSocket fallback · `demo/relay/localhost.toml` with `[auth] public = ""`.

Happy to share the harness, run any variation that would help, or open a standalone issue for the
defect if you'd rather track it separately from this PR — we noticed there isn't one linked.

kixelated:
I think this is fixed by #2556 

I'm doing more takeover stuff on the dev branch.

wrangelvid:
> I think this is fixed by #2556 
> 
> 
> 
> I'm doing more takeover stuff on the dev branch.

Sounds good, thanks!

PR #2538, opened 2026-07-27
feat(moq-audio): cancel the speaker's echo out of the microphone

### LINKED ISSUE #2478: moq-audio: playback engine (device output, mixing) + echo cancellation, ported from iroh-live

kixelated:
## Summary

moq-audio can now both capture and play, which means it can also hear itself: on any laptop without a headset the microphone picks up the speaker and sends the whole call back. That is the second of the two blockers #2282 names, and the half of #2478 that #2529 left open.

`playback::Engine::canceller` taps the post-mix signal and `capture::Config::aec` subtracts it from the microphone, so buffers leaving capture are already clean and `encode::publish_capture` is untouched. Two things you get from an open output device now: `sink()` to play, `canceller()` to stop hearing it.

The engine, not iroh-live's API, is what carried over. Where the shape differs:

- **A canceller is a handle, not a graph node.** No `AudioSink`/`AudioSource` traits or firewheel nodes; `Config` in, an owned `Canceller` out, `Drop` removes the tap.
- **The tap is a `fixed-resample` channel, not a raw ring.** It resamples the device's rate to a fixed 48 kHz, so switching to a 44.1 kHz output doesn't invalidate an adaptive filter that spent seconds converging, and it costs nothing on the 48 kHz devices that are the norm. Theirs pinned the whole engine to one internal rate to get the same property.
- **The tap is drained every pass, not paced against capture.** Anything left queued is delay on top of the hardware's, and on a device that buffers less than the queue holds it inverts the ordering the echo model needs: the echo arrives before the reference describing it, and cancellation stops working outright rather than degrading. Draining means the reference is never older than the last output callback, so it keeps the full head start the output device's buffering gives it. The channel's underflow autocorrect is off for the same reason: a near-empty tap is the normal state here, not a shortfall to pad with silence the speaker never played.
- **The tap is drained to empty while cancellation is off, and on every microphone open.** Nothing reads it while the microphone is closed, and capture is demand-gated, so a resumed call would otherwise start on a reference describing whatever was playing minutes ago.
- **The adaptive filter survives demand gating.** Reopening the same format keeps the processor and only drops the samples in flight, since it is still the same room. Rebuilding it on every subscriber transition would cost a re-convergence each time.
- **Neither audio callback allocates or frees.** Planar scratch is channel-major in one buffer, split with `split_at_mut` per channel count rather than collecting a `Vec<&[f32]>` per 10 ms frame. On the output side the tap owns a heap ring, so a replaced or cleared one leaves by the same retire channel a removed sink uses; that channel now carries a `Retired` enum instead of a bare sink entry.

### Two lifecycle bugs found in review

**A deadlock.** The tap registration is the mirror of a sink's, and the driver rebuilds it inside its own lock. Holding a strong handle on the canceller there means the last one can drop *there*, and its `Drop` deregisters, which re-enters that same lock. `State` sits behind its own `Arc` so the driver only ever holds a `Weak` on the canceller itself and never upgrades it.

**A retry that never fired.** Registering or removing a tap recorded a too-full mixer queue for retry but never woke the driver, so the retry never happened and the canceller sat silently unattached. `Shared` now carries the sync channel, since a canceller deliberately holds no `Handle` (holding one would keep the output device open).

## Dependency

One new crate, optional: [`sonora`](https://crates.io/crates/sonora), a pure-Rust port of WebRTC's audio processing (AEC3, noise suppression, AGC2). BSD-3-Clause like the upstream it was ported from, already on `deny.toml`'s allow list, and it adds no C++ toolchain, which was the whole reason #2478 asked whether a pure-Rust AEC existed. No duplicate versions enter the tree; no git dependency, per that issue's prerequisite.

**Requires 0.2.0 or newer.** 0.1.0 panicked when the adaptive filter shrank, which is what happens whenever the echo path gets shorter: a device switch, an output stall, somebody picking the laptop up. Since this workspace builds `panic = "abort"`, that was a process kill mid-call, and it held this PR up until upstream cut a release carrying the fix ([sonora#14](https://github.com/dignifiedquire/sonora/issues/14), [#15](https://github.com/dignifiedquire/sonora/pull/15)). `survives_a_moving_echo_delay` covers the path and was mutation-checked against 0.1.0, where it reproduces the reported panic verbatim (`slice index starts at 13 but ends at 12`). The bump needed no code changes.

On the port's provenance, since it matters for a dependency this load-bearing: it is AI-assisted and says so, but it is validated by running the upstream WebRTC C++ test suite (2400+ tests) against the Rust backend through an FFI bridge, in CI on Linux and macOS, per push and weekly. That is a stronger correctness story than a hand-written port with hand-written tests.

## Public API changes

All additive, all behind the new non-default `aec` feature. Nothing renamed, removed, or signature-changed, so this targets `main`.

- `moq_audio::aec` (new module)
  - `aec::Config` — `#[non_exhaustive]`, `Default`, with `noise_suppression: bool` and `auto_gain: bool` (both on, matching what a browser gives a call).
  - `aec::Canceller` — `Clone + Debug`, with `set_enabled(bool)` / `enabled() -> bool`. Everything else on it is `pub(crate)`.
- `moq_audio::playback::Engine::canceller(aec::Config) -> aec::Canceller`.
- `moq_audio::capture::Config::aec: Option<aec::Canceller>` — a new field on a `#[non_exhaustive]` struct, so callers already build it via `default()`.

`playback::{Shared, BUS_CHANNELS}` became `pub(crate)` re-exports for the `aec` module; neither is public surface.

## Cross-Package Sync

`moq-audio` has no row in the table and no `doc/` page. `moq-ffi` and wrapper exposure is explicitly out of scope in #2478; nothing in the FFI surface changed. `rs/CLAUDE.md`'s crate map is updated.

## Test plan

`cargo nextest run -p moq-audio --all-features` — 76 passed. Clippy `-D warnings` across `--all-features`, `--no-default-features`, `--features capture`, and `--features playback`, plus the workspace `--all-features` / `--no-default-features` gates; `cargo doc -D warnings`, `fmt`, `sort`, `shear`, and `deny check` all clean.

New tests, all hardware-free:

- **It actually cancels.** A synthetic room: noise into the reference, the same noise attenuated and delayed into the microphone, nobody talking. The last half second comes back **51 dB** below what went in (asserted at 20 dB). This is the test that fails if the reference path is wired up wrong rather than merely compiling.
- Frame accumulation: whole 10 ms callbacks pay no latency; a callback that straddles a boundary is short by the leftover exactly once and never loses samples across 20 rounds of 15 ms callbacks.
- Disabled is a bit-exact passthrough, which is also the assertion that sonora is not being called on that path.
- Turning it off drops the partial frame in flight, rather than replaying it seconds later.
- A missing reference, a missing microphone, and a format sonora would reject are each handled without dropping audio.
- A second canceller takes the engine's one tap, and the first does not take it back on its way out.
- A pass drains the whole tap rather than one frame per microphone frame, so no standing delay accumulates. Mutation-checked: it fails against the old pacing behavior.
- An echo path that keeps changing length runs through without killing the process, and the filter re-converges afterwards. Mutation-checked against sonora 0.1.0, where it reproduces the upstream panic.
- On the mixer side: the tap receives the summed and clipped mix rather than any one sink, and a replaced or cleared tap is handed back to the driver rather than dropped in the callback.

One `#[ignore]`d device round trip (`taps_a_real_output_device`) confirms a real output device reaches the tap; run it with `cargo test -p moq-audio --features aec -- --ignored`. Whether the microphone then stops hearing the speaker is an acoustic question about the room, so it is a manual check, and CI compiles this feature without running it.

Closes #2478.

(Written by Claude Opus 5)


dignifiedquire:
published sonora@0.2 please ping me if you run into any more troubles with it

Issue #2461, opened 2026-07-23
Active/active source failover: a standby publisher's route is never presented to the relay serving the active source (mesh), so a dead source is not failed over — even with `moq-lite-06` cost routing


t0ms:
## Why this matters (the requirement)

We are evaluating MoQ as a primary-distribution transport for linear broadcast, where
the baseline expectation is **no visible failure during contracted content**. The
standard broadcast pattern is **1+1 active/active**: two independent source→publisher
chains run hot, and if the active one dies the receiver keeps playing with no (or a
bounded, hitless) artefact. Today broadcasters achieve this with ST 2022-7 at the IRD.

We want to understand the intended MoQ answer, because MoQ already ships most of the
machinery, and we would like to help land whatever is missing rather than route around
it. This is a **requirements + reproduction + "which direction do you want"** issue,
not a bug report.

## What already works (confirmed by source + drills)

- **The origin's multi-source route table exists and does the right thing.** An origin
  holds every route it knows for a broadcast path, ranks them
  (`route_order` = announced-before-offline → cumulative cost → hop length → stable
  hash), parks losers as silent standbys, and on active-source death `reselect`
  promotes the winner with tracks re-splicing at the next group boundary — covered by
  `test_route_failover` (`rs/moq-net/src/model/origin.rs`). **The splice/failover
  primitive is not the gap.**
- **Cost-based ranking landed** (#2424: cumulative route cost ranks above hop count on
  lite-06 announcements) and **announce ids** (#2160), both in `moq-lite-06-wip`.
- **Transport reconnection and fan-out work.** A publisher survives a relay restart and
  re-announces; N subscribers get byte-identical output.

## The gap (reproduced)

The origin's failover primitive is **never fed a second live route** for the same
broadcast at the node that needs it.

1. **Two publishers, one relay (same broadcast).** The second `active` announce makes
   the path `unroutable` and **both** publishers are torn down — there is no
   "standby publisher for this path" concept on lite-05.
2. **Two publishers, two-relay mesh** (`pubA→relayA`, `pubB→relayB`, meshed). The pair
   coexists, but when `pubA` is killed **the output freezes permanently — no failover.**
   Relay A's session log shows it only ever held its **own** local route for the
   broadcast (its connections were: 2 subscribers + 1 publisher + the cluster peer);
   **relay B never advertised its standby `pubB` across the cluster link.** So relay A
   had nothing to reselect.

### `moq-lite-06` cost routing was tested and does **not** by itself close this

[...]

PR #2545, opened 2026-07-27
test(failover): two-relay 1+1 source-failover drill

t0ms:
The end-to-end drill from #2461, cleaned up for the tree as `just test failover`.
It is the live counterpart to the routing unit tests in `model/origin.rs`: two
real relays, real publishers whose tracks are created lazily by the demuxer, and
the QUIC idle timeout in the loop.

```
pubA ──▶ relayA(:4443) ◀──cluster──▶ relayB(:5443) ◀── pubB
          ▲   ▲                          ▲
        sub1 sub2                       sub3
```

`sub3` is load-bearing rather than decorative: it forces relayB to carry the
broadcast via relayA, which is what makes relayA a hop in relayB's own route and
therefore the peer that per-peer announce selection has to treat specially.
Without it the interesting case never arises.

**CHECK 1 — failover.** Kill `pubA`; `sub1` on relayA must resume, which requires
relayB to have already advertised the local `pubB` standby to relayA.

**CHECK 2 — standby join.** `sub3` on relayB must survive `pubB` joining. This is
the case a model-level standby cannot reproduce: it accepts every track request
immediately, whereas `moq import` announces on connect and creates each track
only once its demuxer reaches it, so a freshly attached standby legitimately
cannot serve some tracks yet.

## Results on #2473 (head `cc11cbaf`, release, macOS)

Both checks pass, repeatably:

```
CHECK 1 failover: sub1 at kill(t=32)=5293892 -> t=81=8647624
PASS: sub1 resumed at t=65, 33s after the kill (+3353732 bytes)
CHECK 2 standby join: sub3 at join(t=10)=1624696 -> t=31=3647388
PASS: sub3 survived pubB's join (+2022692 bytes)
WARN: sub3 stalled 9s at the standby join before recovering
```

`sub1` resumes 30–33 s after the kill across runs, which is one QUIC idle
timeout — so detection dominates and the reselect itself is essentially free.

The standby join is **not** yet transparent: `sub3` stops for 8–9 s before
recovering at full rate. That is a large improvement on being torn down, but it
is still a visible outage for a viewer of a relay that is only carrying the
broadcast, so the drill measures it and prints a `WARN` rather than hiding it
behind a pass. Details in #2473.

## Two things about the harness that are easy to get wrong

Both cost me a wrong conclusion, so both are commented in the script.

**The timeline is derived, not hard-coded.** Killing a publisher sends no
`CONNECTION_CLOSE`, so the relay keeps serving the dead source until the QUIC
idle timeout expires and logs nothing at all in between. A grading window shorter
than that budget cannot pass on any build — my original drill killed at t=22 and
graded at t=43, i.e. 21 s into a 30 s timeout, which is how I first reported a
failover that did not exist. `--idle` lowers the timeout and the window together.

**The publisher pipeline is SIGKILLed in one pass.** Killing `tsp` first leaves
`moq import` reading a truncated stream plus EOF, so it shuts its broadcast down
*cleanly*, the relay unannounces immediately, and you grade a graceful detach
instead of a source failure. Worth knowing that the two paths differ sharply: on
the clean-detach path `sub1`'s `export ts` dies with `TS track layout changed
after PAT/PMT was emitted: '0.avc3' removed` instead of failing over, even with
the standby announced. I have not chased that one down; noting it in case it is
interesting.

## Dependency

Requires `moq --origin`, which arrives with #2473, so this should land with or
after it. On a build without the flag the drill exits with a diagnostic rather
than reporting a misleading zero-byte failure. Verified: on `main` it stops at
the preflight; on #2473's head both checks pass.

Not wired into the default `just test` aggregate, same as `just test smoke`.
Needs `ffmpeg` and TSDuck's `tsp`; the source clip is generated, with `--source`
to publish a real capture instead.

### DIFF
--- a/test/failover/README.md
+++ b/test/failover/README.md
@@ -1,0 +1,111 @@
+# Two-relay 1+1 source-failover drill
+
+`just test failover` stands up two meshed relays, two publishers of the same
+broadcast, and three subscribers, then kills the active publisher and checks that
+the surviving standby takes over.
+
+This is the end-to-end counterpart to the routing unit tests in
+`rs/moq-net/src/model/origin.rs`. Those cover the model. This covers the wiring:
+two real `moq-relay` processes, real `moq import ts` publishers whose tracks are
+created lazily by the demuxer, and the QUIC idle timeout in the loop.
+
+## Topology
+
+```text
+pubA ──▶ relayA(:4443) ◀──cluster──▶ relayB(:5443) ◀── pubB
+          ▲   ▲                          ▲
+        sub1 sub2                       sub3
+```
+
+`pubA` and `pubB` publish the same broadcast with the **same `--origin` id**,
+which is what declares them interchangeable sources rather than different
+content. (`moq --origin` arrives with
+[#2473](https://github.com/moq-dev/moq/pull/2473); on a build without it the
+drill exits with a diagnostic rather than reporting a bogus zero-byte failure.)
+
+`sub3` matters more than it looks. It subscribes on relayB, which has no local
+publisher at first, so relayB has to carry the broadcast across the mesh from
+relayA. That makes relayA a hop in relayB's own route, and therefore the peer
+that per-peer announce selection has to treat specially: relayB must not
+advertise the route that runs *through* relayA back to relayA, but it must
+advertise the local `pubB` standby once that exists. Without `sub3` the
+interesting case never arises.
+
+## What is checked
+
+| Check | Proves |
+|---|---|
+| **1, failover** | After `pubA` is killed, `sub1` on relayA resumes. relayA had already been told about the `pubB` standby and reselected onto it. |
+| **2, standby join** | `sub3` on relayB keeps its subscription when `pubB` attaches locally. A standby wins dispatch the moment it attaches, which is *before* a real publisher has created every track, so a per-track refusal must not abort the whole subscription. |
+
+Check 2 is a regression test for a bug that only appears with real publishers.
+`moq import` announces its broadcast on connect and creates each track only once
+its demuxer reaches it, so a freshly attached standby legitimately cannot serve
+some tracks yet. A model-level standby accepts every track request immediately
+and never reproduces it.
+
+## Three things about the harness that are load-bearing
+
+Changing any of them turns a real result into a meaningless one, quietly. Each
+cost a wrong conclusion at least once, so each is commented where it is set.
+
+**The observation window is derived, not chosen.** Killing a publisher never
+sends `CONNECTION_CLOSE`, so the relay keeps serving the dead source until the
+**QUIC idle timeout** expires, 30 s by default. The relay logs nothing at all
+between the kill and `connection closed err=timed out`, and only then can it
+reselect. **A grading window shorter than that budget cannot pass on any build.**
+So the drill kills the active publisher at t=32 and grades at least 20 s *after*
+the budget elapses. Recovery time is dominated by detection, not by the reselect:
+expect `sub1` to resume roughly one idle timeout after the kill (30 to 33 s
+measured). `--idle` pins a lower timeout on the relays and shortens the run with
+it; it must stay above the keep-alive interval or healthy sessions will flap.
+
+**The kill is atomic.** The publisher pipeline is SIGKILLed in one pass, because
+killing `tsp` first would leave `moq import` reading a truncated stream plus EOF.
+It would then shut its broadcast down *cleanly*, the relay would unannounce
+immediately, and the drill would be grading a graceful detach rather than a
+source failure. The two paths behave nothing alike: one waits out an idle timeout
+and fails over, the other unannounces at once, and on that path the subscriber's
+`export ts` currently dies with `TS track layout changed after PAT/PMT was
+emitted` instead of switching to the standby.
+
+**The standby joins early, and that is deliberate.** The two publishers replay
+independent copies of the same clip *from its start*, so the standby's media
+timeline lags the active one by roughly its join delay. On the splice the
+subscriber's muxer has to wait for the new source's timestamps to overtake the
+last ones it wrote, and that wait scales one-for-one with the join delay (join at
+t=4 costs under 2 s, t=10 costs 9 s, t=20 costs 18 s). It is a property of this
+harness, not of the relay, which reselects in the same millisecond the standby
+connects. A real 1+1 pair shares a timestamp-aligned feed and does not pay it.
+Hence the default join at t=4, and hence check 2 only warns when a stall exceeds
+what that offset explains.
+
+## Usage
+
+```bash
+just test failover                  # generated clip, 30s detection budget (~90s)
+just test failover --idle 10s       # faster run (~70s)
+just test failover --source cap.ts  # publish a real capture instead
+just test failover --keep-logs      # keep the work dir (relay logs, sizes.csv)
+```
+
+Requires `cargo`, `ffmpeg`, `curl`, `pgrep`, `pkill`, and TSDuck's `tsp` (for PCR
+pacing).
+
+Environment overrides: `FAILOVER_PORT_A`, `FAILOVER_PORT_B`, `FAILOVER_PROFILE`
+(`debug`/`release`), `FAILOVER_ORIGIN`, `FAILOVER_PRE` (standby join time),
+`FAILOVER_KILLA` (when the active source is killed).
+
+## Debugging a failure
+
+`--keep-logs` prints the work dir, which holds both relay logs, every client log,
+and `sizes.csv` (per-second byte counts for `sub1` and `sub3`).
+
+`RUST_LOG=moq_net=debug` makes the announce decisions visible, which is usually
+what you want:
+
+- `no advertisable route for this peer exclude_hop=…`: relayB correctly
+  advertising nothing while it is merely carrying via relayA.
+- `announce broadcast=… hops=2`: relayB advertising the local standby to relayA
+  once `pubB` joins. This is the line that must appear at the join; without it
+  there is nothing for relayA to reselect onto.
--- a/test/failover/run.sh
+++ b/test/failover/run.sh
@@ -1,0 +1,373 @@
+#!/usr/bin/env bash
+# 1+1 active/active source-failover drill across a two-relay cluster.
+#
+# Two relays are meshed, two publishers publish the SAME broadcast with a shared
+# origin id (so they are interchangeable sources), and the active publisher is
+# killed. The point is to prove, with real processes rather than model tests,
+# that the surviving standby is already advertised across the mesh and that the
+# relay serving the dead source reselects onto it.
+#
+# ```text
+#     pubA --> relayA(:4443) <--cluster--> relayB(:5443) <-- pubB
+#               ^   ^                          ^
+#             sub1 sub2                       sub3
+# ```
+#
+# sub3 subscribes on relayB, which has no local publisher at first, so relayB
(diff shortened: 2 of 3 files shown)


t0ms:
Rebased on merged `main` and updated. Force-pushed as a single commit, so here is what changed
since the review.

**One of the two things this drill reported was the drill's own fault, and it is gone.** The
8 to 9 s stall at the standby join was not routing: the two publishers replay independent copies
of the same clip *from its start*, so the standby's media timeline lags the active one by its join
delay, and on the splice the subscriber's muxer waits for the new source's timestamps to overtake
the last ones it wrote. It scales one-for-one with the join delay (t=4 costs under 2 s, t=10 costs
9 s, t=20 costs 18 s) while the relay reselects in the same millisecond the standby connects.
Retracted on #2473. The drill now joins the standby at t=4 and only warns when a stall exceeds
what that offset explains, and the offset is documented next to the other two traps. Both checks
pass clean on merged `main`, no warning, `sub1` resuming 33 s after the kill.

**CodeRabbit's review, addressed:**

- `--idle` now rejects anything this script cannot do arithmetic on. It was stripping a trailing
  `s`, so `1m` or `500ms` would have silently mis-derived the observation window, which is the one
  thing this harness cannot afford to get wrong.
- Em dashes removed. I had missed the house rule in `AGENTS.md`, sorry.
- Diagram fence tagged `text`; `pkill` listed in the prerequisites.
- Check 2 is now described by what it proves rather than by the bug it came from.
- Runtime estimates corrected to match the timeline (~90 s default, ~70 s with `--idle 10s`).

Also exposed the join and kill times as `FAILOVER_PRE` / `FAILOVER_KILLA`, which is what made the
offset above diagnosable in the first place.

`--origin` is on `main` now that #2473 has merged, so the dependency note is historical; the
preflight check stays for anyone running this against an older checkout.

kixelated:
With the timeline stuff, if you're publishing two separate broadcasts (ie. different timestamps, different codecs, different segments, etc) with the same name then failover is impossible. We need explicit cooperation from publishers to produce the exact same broadcast.

If they're not the same, then they could be published with different names. foo/primary and foo/backup for example. Either way the application needs to reinitialize everything on a switch.

t0ms:
Following up with data. We rebuilt the drill around one source feeding both publishers, since a 1+1 pair should be two views of one feed rather than two independent encodes. On current main (moq-relay 0.14.3):

1. **Byte-identical** (one file duplicated into two co-started importers): the two subscribers' output is byte-for-byte identical up to the kill, confirming the publishers produce the same broadcast. After a hard kill, the subscriber reselects onto the standby ~31 s later (the 30 s idle timeout), continuity-clean TS, a single current group evicted; skipping to next buffered group.

2. **Mid-stream standby** (shared feed; standby joins ~11 s after the primary, so its group numbering is offset): failover still succeeds and the TS is still continuity-clean; the exporter subscribes once, never re-subscribes, and skips to the standby's live edge.

So in our testing the requirement was a common source (same PMT/track layout, consistent PTS), not byte-identical segmentation, the consumer resumes at the new live edge rather than demanding group-number continuity. The switch is not hitless: the ~30 s outage is a PCR/PTS discontinuity (continuity-counter stays clean, but there's a content hole). Lowering --server-quic-idle-timeout to 10 s gives an ~11 s switch; 5 s was too aggressive and the switch didn't complete. 

Is a lower idle default on the table? 
Is receiver-side dual-subscribe the intended path for anyone who needs sub-second/hitless?

kixelated:
> So in our testing the requirement was a common source (same PMT/track layout, consistent PTS), not byte-identical segmentation, the consumer resumes at the new live edge rather than demanding group-number continuity.

So I think this requirement is quite difficult and out of scope.

The problem is that we lack information about what is continuous. The relay sees two broadcasts with the same name but different IDs and group numbering. A seamless resume is impossible from the relay's point of view because it's content agnostic; the bytes are either the same or they are different. We purposely don't want the relay to try to interpret the data to decide that these bitstreams are close enough so we could merge them by rewriting timestamps or something.

There are a few options to get the desired behavior you want:

1. Produce explicitly compatible content. This is something elemental is focusing on, basically setting the group ID based on the wall clock timestamp of the encoder. Requires some annoying clock sync stuff and a lot of guarantees that encoder A produces compatible (doesn't have to be bit identical) content as encoder B.

2. Have the clients resume. This is how `@moq/watch` works; we reinitialize the decoder/timestamps rather than assuming the new broadcast is a continuation of the previous one. It might be possible with `export ts` to optimistically try to stitch in the new broadcast by rewriting timestamps and validating the PMT/PAT is the same. But yeah it's gross and we'd need some way of aligning timestamps.

3. Just putting this out there in case it's possible, but you could use the same broadcast with two routes. ie. one encoder, but you would publish the broadcast twice over separate routes (ex. satellite and ethernet). This should work today, as if original publisher uses the same origin ID then we'll seamlessly stitch tracks. The issue you're running into is that both publishers are explicitly saying they're NOT the same ID, so we have no idea if they're collaborating or competing. You might want collaboration (same source feed), but somebody else might want it to be a brand new broadcast using the same name (ex. kixelated starts streaming from a different PC). 

2 is the type of thing you would do within a company with a lot of safeguards in place. It's difficult to do it generically, because even something like the `moq` binary version matters a lot and of course so does the encoder.

kixelated:
and yeah just to clarify, using the same `--origin <id>` means you are explicitly stating you produce the same content (ex. like elemental's transcoders). If you actually produce different content, then that will blow stuff up.

t0ms:
Clear: the relay is content-agnostic, so a *seamless* merge of two independently-numbered broadcasts isn't something it can (or should) infer.

It's probably worth clarifying our actual architecture, it reframes your three options for our case.

**How we run primary distribution.** Everything is 1+1 active/active and doubled end-to-end, with the
two legs kept fully separate: two sources (same content) → two publishers → two relays → two subscribers → two
pacers, and two receivers (IRDs) do the hitless switch with ST 2022-7. The routes are permanently
provisioned, so we control the names/IDs. The load-bearing property for us is **resilience of each
leg** and a **make-before-break switch at the receiver**. This is designed to minimise any stall if avoidable.
Given that, the failovers *inside* MoQ are nice-to-haves rather than requirements:

- Subscriber → alternate relay is the one that matters most (and we get it by doubling the
  subscriber, i.e. a receiver watching both legs).
- Relay switching from a dead publisher to a standby is welcome but not critical.
- Publisher switching to a different upstream source is likewise a plus, not a must.

Most primary IP distribution today has little or no failover at all, so even the bounded behaviour is
an improvement, but the thing we actually rely on for "no stall" is the doubled chain + receiver-side
selection, not a relay-side reselect.

**Your three options mapped to that:**

1. *Compatible-content encoders (Elemental / wall-clock group IDs)* removes the numbering discontinuity at the source. But it still doesn't remove the **detection outage**: a hard-killed source is ~one idle timeout of silence regardless of numbering. So it makes relay failover *cleaner*, not *hitless*. Useful long-term; not on our critical path.
2. *Client resumes / reinitialises (`@moq/watch`; optimistic `export ts` stitching)* this is a
   re-init, i.e. still a gap, and it's a single-consumer story. Our receivers are doubled and merged
   downstream, so we'd rather run two hot legs than have one consumer reinitialise through a hole.
3. *Same broadcast, same `--origin`, two routes* closest to our world and works today; it's
   effectively what our single-source drill exercised. The one caveat for us is that it couples the
   two legs *at the relay*, whereas we deliberately keep them independent and merge at the IRD (no
   shared-relay dependency). Still carries the detection gap on a hard loss.

**The point we'd underline:** none of the three gives no-stall on a hard failure — they each carry a
detection or re-init gap. Sub-second/hitless is a *receiver-side make-before-break* property, and crucially the
common source is precisely what makes that feasible (two legs that are the same broadcast → an IRD can
merge them). So we're aligned with you that relay reselect is a bounded enhancement, not the seamless
path.

**On this PR specifically:** the drill validates the *shipped* #2473 behaviour honestly — a bounded
route reselect on an *ungraceful* source loss, with a shared source and shared `--origin`,
continuity-clean but not hitless (a PCR/PTS discontinuity across ~one idle timeout; ~11 s at
`--server-quic-idle-timeout 10s`, and 5 s was too aggressive to complete). It's a test of that
behaviour, not a feature. Happy either way on disposition — **land it as a regression test with its
scope stated that way, or close it now that the model is settled.** Your call; we don't want it
lingering if it's not useful to you.

One small question that would help the nice-to-have case for anyone using relay reselect without an
IRD: is a lower `--server-quic-idle-timeout` default on the table? ~10 s was reliable in our runs.

kixelated:
> One small question that would help the nice-to-have case for anyone using relay reselect without an
IRD: is a lower --server-quic-idle-timeout default on the table? ~10 s was reliable in our runs.

It used to be lower but I don't remember why I raised it.

And yeah, I think what you're saying makes sense. Something at the media layer should support splicing semi-related broadcasts, resuming tracks by rewriting timestamps. It's just difficult to make it generic, maybe if there's an explicit opt-in somewhere in the catalog that `Broadcast A Track B` is the same content as `Broadcast C Track D` even if the timestamps and codecs don't match.

kixelated:
We've got unit tests for all of this behavior now, which is easier to maintain than manual executions. I also had Codex run this test:

```
just test failover --idle 10s: passed
Failover resumed 14 seconds after killing publisher A
Subscriber 3 survived publisher B joining
```

t0ms:
Closing this makes sense, and agreed on the principle: unit tests run on every PR, a hand-run drill never does.

Three things worth leaving on the record.

**The numbers line up.** `just test failover --idle 10s` resuming at 14 s matches what we measured out of band, about 11 s at a 10 s idle timeout, with 5 s too aggressive for the switch to complete at all. So detection dominates and the reselect itself is essentially free, which is the model we have been working to.

**On the idle-timeout default.** Since it used to be lower, the thing worth recovering is *why* it went up, before changing it back. The usual reason is a NAT or radio path where a peer with nothing to send gets dropped by an intermediary before the other end notices, so a low timeout buys spurious disconnects on mobile. That trades directly against failover detection, and the two callers want opposite defaults. If it is useful, we can measure the failover side of that trade across a few values on our rig and post the numbers, so any change rests on something more than our single data point. Equally happy to leave it alone.

**On catalog opt-in splicing.** It is worth being explicit about which consumer that feature would serve, because it is probably not us. In primary distribution the two legs are the *same* content, and the receiver merges them at the packet level (ST 2022-7) precisely because they are identical. Rewriting timestamps to splice two semi-related broadcasts is a different operation, and its output would not be acceptable on a primary feed: a codec or parameter change across the join is a discontinuity the downstream equipment has to reacquire through, which is the visible event we are trying to avoid in the first place. So we would only ever use the strict form of that opt-in, "these are the same content", which is what `--origin` already asserts. The loose form is a real feature, just for a different consumer, and probably worth scoping as one rather than as the generic answer to redundancy.

**Following your unit-test point**, we opened #2713 as a draft with two model tests drawn from this work. Every existing takeover and reselect test subscribes to a single track, so nothing currently pins that a reselect is decided and served per track. That matters for TS, where one broadcast carries video, audio and data together, and a takeover that restores video while audio quietly stops is worse than a clean failure because it survives monitoring. As a check that the tests are not just restating coverage you already have, we injected a fault that silently drops all but one queued track request: 705 of the 707 existing `moq-net` tests still pass, and the only two that fail are the new ones. Happy to reshape it however you prefer, or to drop it if you would rather keep the model tests lean.

(Written by Opus 5)

kixelated:
Yeah, if you really want redundancy, you would do active-active. Don't wait for the QUIC idle timeout; always pull both broadcasts and splice them.

PR #2546, opened 2026-07-27
docs(drafts): address relay-hops review feedback

kixelated:
Addresses the review feedback in moq-dev/drafts#50 (from afrind), applied to `drafts/draft-lcurley-moq-relay-hops.md` here since the drafts are moving into the monorepo, plus a follow-up simplification pass.

## Review feedback (commit 1)

- **"MUST include HOP_PATH... or else what?"** The consequence is now specified: a receiver that negotiated the extension and gets a PUBLISH_NAMESPACE or NAMESPACE without HOP_PATH MUST close the session with a PROTOCOL_VIOLATION.
- **Parameter handling vs. MOQT draft 16.** Dropped the even/odd rationale everywhere (HOP_PATH definition, EXCLUDE_HOP definition, IANA section). The draft now follows the current MOQT convention: each message parameter has a bespoke serialization defined by its spec, and there is no generic skip rule. Consequently the "unknown parameters are ignored" graceful-degradation claim was wrong and is replaced: endpoints MUST NOT send HOP_PATH or EXCLUDE_HOP on a session that did not negotiate the extension, and a relay strips them (and the appended NAMESPACE Parameters field) when forwarding into a non-supporting session. The wire formats themselves are unchanged.
- **Hop ID uniqueness.** Uniqueness is now a MUST, with the failure modes spelled out (false loop drops, conflated origins). Configured unique identifiers are the SHOULD; a full-width random value is the fallback for endpoints with no configured ID, no longer the recommendation. Security Considerations adjusted since Hop IDs are no longer necessarily random.
- **Upstream/downstream.** Defined in Conventions and Definitions relative to the flow of each advertisement rather than the endpoints, with a note contrasting MOQT's publisher-relative usage in a mesh.
- **Number of Parameters redundancy.** Kept as-is, since it mirrors PUBLISH_NAMESPACE in MOQT (reviewer agreed it's fine).
- **Nit.** "origin publisher" is now "original publisher" throughout.
- **Editorial.** Trimmed some of the parsing pedantry.

## Simplification pass (commit 2)

- **EXCLUDE_HOP scoping fix.** The suppression rule named PUBLISH_NAMESPACE while also claiming subscription scope, but matching advertisements for a SUBSCRIBE_NAMESPACE are delivered as NAMESPACE messages. It now suppresses any namespace advertisement sent for that subscription, and states that other subscriptions and proactive advertisements are unaffected.
- **Stable stand-in Hop ID.** The Hop ID a bridging relay assigns for a non-supporting upstream must now be stable per upstream session; a fresh value per advertisement would give the same broadcast a different origin each time and break origin-based deduplication.
- **Consolidated negotiation gating.** The "don't send this without negotiation" rule is stated once in Setup Negotiation instead of three times.
- **Folded "Why per-hop, not end-to-end" into the Introduction** (it repeated the intro and Relay Behavior) and trimmed advisory/flavor text (single-best-path advertising paragraph, BGP confederation analogy, greasing aside).

EXCLUDE_HOP itself was deliberately kept despite being loop-detection-redundant: moq-lite implements the same mechanism with the same single-value shape (`AnnounceRequest.exclude_hop` in `rs/moq-net/src/lite/announce.rs`), so the moqt extension keeps parity with the shipped design.

No code or other-draft changes needed: this extension has no moqt-side implementation in the repo, and the lite wire protocol (`draft-lcurley-moq-lite`) is untouched.

Validated with `kramdown-rfc --v3` (the `just drafts check` recipe).

Fixes https://github.com/moq-dev/drafts/issues/50

(Written by Claude Fable 5)

### DIFF
--- a/drafts/draft-lcurley-moq-relay-hops.md
+++ b/drafts/draft-lcurley-moq-relay-hops.md
@@ -24,7 +24,7 @@ informative:
 --- abstract
 
 This document defines a Relay Hops extension for MoQ Transport {{moqt}}.
-Each namespace advertisement carries an ordered list of Hop IDs identifying the relays it has traversed, starting with the origin publisher.
+Each namespace advertisement carries an ordered list of Hop IDs identifying the relays it has traversed, starting with the original publisher.
 This lets a subscriber prefer the shortest of several paths to the same namespace, identify which advertisements refer to the same broadcast (same origin), and lets a relay cluster detect and avoid routing loops.
 A namespace subscription MAY carry a single Hop ID to exclude, which a relay uses to suppress advertisements that have already passed through that hop.
 
@@ -32,6 +32,9 @@ A namespace subscription MAY carry a single Hop ID to exclude, which a relay use
 
 # Conventions and Definitions
 {::boilerplate bcp14-tagged}
+
+This document uses **upstream** and **downstream** relative to the flow of a namespace advertisement: an advertisement travels from its original publisher downstream toward subscribers, so on a given session the peer that sends an advertisement is the upstream peer and the peer that receives it is the downstream peer.
+{{moqt}} uses these terms relative to the original publisher of the content; in a relay mesh the same pair of relays can carry advertisements in both directions, so here the direction is a property of each advertisement, not of the endpoints.
 
 
 # Introduction
@@ -47,13 +50,13 @@ This redundancy is desirable for failover, but it leaves a receiver with no info
 - **Broadcast identity**: two advertisements for the same namespace may refer to the same broadcast or to two distinct origins reusing a namespace. With no origin identity a receiver cannot tell them apart, nor deduplicate redundant paths to one broadcast.
 - **Routing loops**: relay A advertises a namespace to relay B, which advertises it back to A (directly or through a cycle). Without a way to recognize an advertisement it has already seen, a relay will re-advertise it indefinitely.
 
-This extension solves all three with a single mechanism: an ordered list of **Hop IDs** that records the path an advertisement has taken, starting with the origin publisher and with one entry appended per relay.
+This extension solves all three with a single mechanism: an ordered list of **Hop IDs** that records the path an advertisement has taken, starting with the original publisher and with one entry appended per relay.
 The first entry identifies the origin (broadcast identity); the list length gives the path length (path selection); a relay finding its own Hop ID already in the list detects a loop.
 
 ## Why per-hop, not end-to-end
 The Hop ID list is rewritten at every relay: a relay appends its own Hop ID before forwarding an advertisement downstream.
 A relay therefore detects a loop by finding its own Hop ID already present in an incoming advertisement, and a subscriber compares path lengths using the list length.
-Hop IDs are chosen randomly (see [Hop IDs](#hop-ids)) so they are unique with overwhelming probability without any central coordination, even across independently operated relays.
+Hop IDs are unique (see [Hop IDs](#hop-ids)), even across independently operated relays.
 
 
 # Setup Negotiation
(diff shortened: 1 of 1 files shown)


drafts/draft-lcurley-moq-relay-hops.md
```diff
@@ -47,13 +50,9 @@ This redundancy is desirable for failover, but it leaves a receiver with no info
...
 
-This extension solves all three with a single mechanism: an ordered list of **Hop IDs** that records the path an advertisement has taken, starting with the origin publisher and with one entry appended per relay.
+This extension solves all three with a single mechanism: an ordered list of **Hop IDs** that records the path an advertisement has taken, starting with the original publisher and with one entry appended per relay.
 The first entry identifies the origin (broadcast identity); the list length gives the path length (path selection); a relay finding its own Hop ID already in the list detects a loop.
-
-## Why per-hop, not end-to-end
-The Hop ID list is rewritten at every relay: a relay appends its own Hop ID before forwarding an advertisement downstream.
-A relay therefore detects a loop by finding its own Hop ID already present in an incoming advertisement, and a subscriber compares path lengths using the list length.
-Hop IDs are chosen randomly (see [Hop IDs](#hop-ids)) so they are unique with overwhelming probability without any central coordination, even across independently operated relays.
```
afrind, line 56: There might have been some valuable text here?  Was it duplicative?

drafts/draft-lcurley-moq-relay-hops.md
```diff
@@ -72,21 +71,23 @@ The extension applies to a single hop (one MOQT session) and is negotiated indep
...
 # Hop IDs
-A **Hop ID** is a variable-length integer that identifies a single relay (or the origin publisher) within the path of an advertisement.
+A **Hop ID** is a variable-length integer that identifies a single relay (or the original publisher) within the path of an advertisement.
 
-Each relay and each origin publisher chooses its Hop ID **randomly**.
-An endpoint SHOULD draw a full-width random value (up to the 64-bit varint maximum) so that the probability of two endpoints choosing the same Hop ID is negligible.
-Random assignment means there is no registry, no coordination, and no reserved values: a Hop ID is simply an opaque identifier that is, with overwhelming probability, unique.
+Hop IDs MUST be unique among the endpoints an advertisement can traverse.
+Loop detection and origin identification compare Hop IDs for equality, so two endpoints sharing a Hop ID are indistinguishable: advertisements get dropped as false loops, and distinct broadcasts get conflated into one origin.
+Deployments often already assign each node a unique identifier; an endpoint SHOULD use such a configured identifier as its Hop ID.
```
afrind, line 85: I would drop the SHOULD here.  Operators can do what they want.

drafts/draft-lcurley-moq-relay-hops.md
```diff
@@ -72,21 +71,23 @@ The extension applies to a single hop (one MOQT session) and is negotiated indep
...
+An endpoint with no configured identifier MAY instead draw a full-width random value (up to the 64-bit varint maximum), which is unique with overwhelming probability; a receiver cannot tell how a Hop ID was chosen.
```
afrind, line 86: Drop:  a receiver cannot tell how a Hop ID was chosen. ?

drafts/draft-lcurley-moq-relay-hops.md
```diff
@@ -72,21 +71,23 @@ The extension applies to a single hop (one MOQT session) and is negotiated indep
...
+There is no registry and no reserved values: a Hop ID is simply an opaque identifier.
 
 An endpoint SHOULD keep its Hop ID stable for the lifetime of a session (and MAY reuse it across sessions) so that loop detection and path comparison are consistent.
 
-When a relay bridges an advertisement from an upstream peer that did **not** negotiate this extension, the upstream carries no HOP_PATH. The relay MUST synthesize one (see [Relay Behavior](#relay-behavior)) by assigning a random Hop ID to stand in for the unknown upstream, so that loop detection and path length still work within the cooperating region of the mesh.
```
afrind, line 88: If the upstream relay that doesn't support the extension is connected to two relays that do, each relay selecting a unique ID seems bad?  It might be better to suggest a consistent algorithm, though I don't know how.  They may see different IP addresses.

drafts/draft-lcurley-moq-relay-hops.md
```diff
@@ -135,15 +133,19 @@ HOP_PATH Parameter {
...
 
 ## Relay Behavior
 When a relay forwards a namespace advertisement downstream on a session that negotiated this extension, it MUST append its own Hop ID to the HOP_PATH it received.
 The relay's own Hop ID is therefore always the last entry of the list it sends.
-If the advertisement arrived from an upstream that did not negotiate this extension (and so carried no HOP_PATH), the relay MUST first create a HOP_PATH whose single initial entry is a random Hop ID it assigns to stand in for that unknown upstream, then append its own Hop ID.
+If the advertisement arrived from an upstream that did not negotiate this extension (and so carried no HOP_PATH), the relay MUST first create a HOP_PATH whose single initial entry is a Hop ID the relay assigns to stand in for that upstream, then append its own Hop ID.
+The stand-in MUST be stable for the lifetime of the upstream session and unique per upstream (a random value chosen per session works); advertisements bridged from the same upstream then share an origin, so loop detection, path length, and origin-based deduplication keep working within the supporting region of the mesh.
```
afrind, line 148: See above

drafts/draft-lcurley-moq-relay-hops.md
```diff
@@ -157,16 +159,13 @@ This is advisory: the receiver MAY apply additional local policy (e.g. measured
 Two advertisements for the same namespace whose HOP_PATH begins with the same Hop ID share an origin and therefore refer to the same broadcast; a receiver MAY treat them as redundant paths and keep only the best one.
 If the first Hop IDs differ, the advertisements come from distinct origins that happen to reuse a namespace, and a receiver MUST NOT treat them as interchangeable.
 
-A publisher (or relay acting as one) SHOULD advertise only the single best path it currently knows for each namespace.
```
afrind, line 160: Why did you drop this?

afrind (commented):
@kixelated leaving you some feedback on this already merged PR.

PR #2556, opened 2026-07-29
fix(net): prefer the newest route so a reconnect takes over immediately

kixelated:
## Summary

A publisher reconnecting could not reclaim its path until the transport retired the session it replaced. For an ungraceful disconnect that is the QUIC idle timeout (30s by default), which is the failover-latency caveat called out at the bottom of #2473. Two separate mechanisms preferred the older route, depending on whether the reconnect keeps its origin id:

- **Same origin id** (a Rust publisher, or an upstream relay reconnecting): both sessions attach as routes to one front, but their hop chains and costs are *identical*, so `route_order` tied exactly and `min_by_key` returned the first inserted, i.e. the corpse. Recency (`Reverse(FrontRoute::id)`, the front's attach counter) is now the final tie-break, below the hop hash, so it only separates routes that are indistinguishable to the rest of the mesh. Local attach order never leaks into cluster convergence: what gets forwarded is the chain and cost, equal by construction wherever it applies.
- **Different origin id** (`js/net` mints one per connection, so *every* browser reconnect): the newcomer never entered the table at all. `attach_source` parked it on a first-hop mismatch and waited out the incumbent's idle timeout plus the origin linger. An **announced** newcomer now closes the incumbent front and falls through to the existing stale-entry replacement path, so consumers observe an unannounce followed by an announce. That preserves the guarantee the park was protecting (unrelated content is never *spliced* into a live subscription) while making the swap immediate.

Taking a path over requires announcing, which keeps the takeover consistent with the ordering it extends: an offline source (`Route::new()`, the shape a cache or on-demand handler uses) ranks below every announced route, so it still parks rather than unannouncing a live broadcast and cutting its subscribers. The second commit fixes that; the first commit alone would let an offline different-publisher source evict a live front, which `test_offline_mismatch_never_evicts_a_live_front` reproduces.

### Behavioral change worth reviewing

Path ownership is now last-writer-wins among *announced* publishers: one announcing a path that a live different publisher holds evicts them rather than queueing behind them. Nothing can distinguish "dead but not yet timed out" from "alive" at attach time, so preferring the newest requires this. Authorization is what decides who may publish where; a takeover logs a warning. Redundant 1+1 publishers are unaffected, since they share an origin id and remain standbys.

The displaced source is orphaned rather than re-queued: it does not reclaim the path if the usurper later leaves. That is deliberate, since re-attaching on eviction would let two live publishers evict each other in a loop.

## Public API changes

None. No signature changes; `route_order`, `attach_source`, `Attach`, and `AttachContext` are private. `origin::Producer::create_broadcast` changes behavior only (an announced different-first-hop source replaces instead of parking), which reverses the corresponding bullet in #2473. Targets `main` per the branch-targeting rules.

## Wire / draft

No wire-format change. Both drafts normatively specified the old rule and are updated: `draft-lcurley-moq-lite.md` and `draft-lcurley-moq-relay-hops.md` now specify the recency tie-break and later-replaces-earlier instead of "keep serving the earlier one until it ends". Both are scoped to *advertisements*, which matches the announce requirement above. lite-06 is unreleased, so its changelog bullets are amended in place rather than appended.

## Cross-package sync

- `drafts/`: both drafts above.
- `doc/bin/cli.md`: the "Redundant Publishers (1+1)" section said a different-id publisher waits invisibly; it now describes the takeover and points at authorization for path ownership.
- `js/net`: no mirror needed. It is a leaf with no multi-route origin table, and nothing on the wire changed.

## Test plan

- `test_reconnect_wins_over_stale_route`: a same-id reconnect attaches alongside the stale route, and track requests dispatch to the new session while the stale source receives none.
- `test_carrying_reconnect_switches_immediately`: the same reconnect arriving while a subscription is already spliced onto the stale route, which re-splices onto the new session.
- `test_offline_mismatch_never_evicts_a_live_front`: an offline different-publisher source at a path with a live announced front and an active subscriber leaves both intact.
- `test_publisher_mismatch_replaces` (was `test_publisher_mismatch_parks`): the announced takeover is immediate, is a distinct face rather than a splice, and the displaced front is *closed* rather than merely unpublished.
- `test_carrying_switches_to_benign_routes` gains a tied-cost reconnect case, pinning that `reselect`'s simultaneous-activation gate stays strictly-cheaper.
- `test_offline_mismatch_never_evicts_a_live_front` also drives the park's *exit*: the incumbent ends and the parked source takes over. A review pass caught that nothing covered this (a panic planted in the wake path left all 602 tests green), which matters because a lost wakeup there strands a publisher invisibly rather than failing anything.
- Every guard above is mutation-checked: dropping `Reverse(route.id)`, dropping `s.closed = true`, dropping the `!route.announce` park, stubbing the park's incumbent poll to `Pending`, and loosening the gate to `<=` each fail exactly one test. (The eviction was invisible until the `is_closed` assertion was added, since the replacement path fires either way.)
- `just check` clean; `cargo test -p moq-net` 602 passing; `just rs loom` passes (this touches `moq-net/src/model/`).

Tooling note: finding the park gap took two runs that hung and pegged a core, because `cargo test` has no timeout. That fix (nextest by default, plus an `opt-level` override that takes the RSA keygen tests from 16s to 0.8s) is split out into #2558 rather than bundled here.

## Review notes

CodeRabbit's two maintainability findings are addressed in the second commit (`attach_source` down to 4 parameters via `AttachContext`; the `FrontState::closed` doc now lists all three triggers), as is its suggested carrying-reconnect test. Its premise for that test was half right: for a single-hop publisher chain the gate is unreachable on `hops.len() >= 2` regardless of the cost comparison, so the `<` boundary is pinned by the unit-level case in `test_carrying_switches_to_benign_routes` instead.

(Written by Opus 5)

### DIFF
--- a/doc/bin/cli.md
+++ b/doc/bin/cli.md
@@ -128,7 +128,7 @@ whenever routing prefers another (not only on failure), splicing at the next
 group boundary, so encoders that drift (for example segment-numbered tracks
 from processes started at different times) tear down subscribers on the
 switch. Independent publishers with different tracks or timelines MUST use
-different origin ids; the newcomer then waits as a replacement instead of
+different origin ids; the newcomer then takes the broadcast over instead of
 joining. Run the same command from two aligned encoders, pinning the same id
 on both:
 
@@ -139,9 +139,12 @@ moq --origin 42 --client-connect https://relay-b.example.com/anon --broadcast ev
 
 Leave `--origin` unset everywhere else. The default fresh id per run is what
 makes a restarted encoder look like new content (ending subscriptions and
-invalidating caches) instead of silently splicing mid-stream, and a second
-publisher with a *different* id waits invisibly until the first ends rather
-than joining it.
+invalidating caches) instead of silently splicing mid-stream. A publisher with
+a *different* id takes the broadcast over the moment it announces, ending the
+incumbent rather than waiting behind it, so a reconnect is live again without
+waiting for the relay's transport to time out the connection it replaced. The
+last publisher to announce a path owns it, so use authorization to decide who
+may publish where.
 
 ### Capture a Webcam
 
--- a/drafts/draft-lcurley-moq-lite.md
+++ b/drafts/draft-lcurley-moq-lite.md
@@ -299,10 +299,12 @@ The first entry of the reconstructed path identifies the original publisher (see
 A publisher MUST NOT keep multiple current advertisements for the same broadcast on the same stream: each broadcast has at most one current advertisement at a time, and a second ANNOUNCE_START for an already-available path is a protocol violation (use ANNOUNCE_RESTART).
 A subscriber that sees the same broadcast advertised across multiple streams SHOULD route subscriptions to the advertisement with the lowest Route Cost after adding each arriving link's cost, breaking ties by the shortest total path length (see [ANNOUNCE_START](#announce-start)).
 Advertisements from peers that predate the Route Cost carry an effective cost of 0, so a mixed mesh degrades to shortest-path routing.
+Advertisements that tie on every advertised property SHOULD be broken toward the most recently received: they are indistinguishable to the rest of the mesh, and a publisher reconnecting over a new session is otherwise outranked by the session it replaced until the transport declares that one gone.
 
 Advertisements for the same broadcast path whose reconstructed paths share the same first entry carry interchangeable content: a relay MAY hold them as redundant routes for one broadcast and splice a live subscription across them at a Group boundary, e.g. when the serving route ends.
 Cooperating redundant publishers MAY share a Hop ID to opt into this.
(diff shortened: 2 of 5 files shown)


kixelated:
Confirmed that this also resolves the restarted-numbering/stitching issue investigated in #2534. When the replacement's first hop changes, it is now surfaced as a new broadcast instead of being stitched at the incumbent's `latest + 1` boundary.

I tested head `83c03fe7` with a live incumbent subscriber. The takeover closed the old broadcast face, produced the expected unannounce/announce transition, and a new subscriber received the replacement's restarted group 0. The complementary same-first-hop case still re-spliced the existing subscriber at the next boundary and delivered group 1.

(Written by GPT-5)

PR #2559, opened 2026-07-29
fix(publish): recover a capture whose track dies

kixelated:
Follow-up to #2557, found by an adversarial review of that PR. Two further review rounds on this branch found more bugs, all of them in terminal and failure states; the design below is what answers them.

## Summary

**Root cause.** Nothing in `js/publish` watched a `MediaStreamTrack` after handing it to `out.source`. `grep -rn '"ended"\|onended\|readyState' js/publish/src` returned nothing. A track that ends on its own left the broadcast silently dead: `Capture` breaks its read loop and stops, and no source reopened. The worst case is screen share, because it is a routine user action rather than a hardware event: clicking the browser's own "Stop sharing" ended the track while `out.source` kept pointing at the corpse.

**The invariant.** `out.source` is written only by the effect run, never from a listener. Every outcome, including running out of retries, reruns the effect, so the cleanup the previous run registered is what clears the source and stops the stream. No call site has to remember to do it, which is what makes the terminal states correct by construction instead of by inspection.

Concretely:

- `Camera` and `Microphone` reopen on `ended`. `getUserMedia` needs no user gesture once permission is granted, so recovery is invisible.
- A shared `Retry` holds the budget. Track deaths **and** failed reopens spend it, so a device that never works stops being asked. Spending the last of it still reruns the effect, which is what clears the source.
- **The budget belongs to the settings it was spent on.** Changing the device or the constraints is new intent rather than another go at the same thing, so it starts a fresh budget. Those are read before the budget check, so one change both reruns the effect and refunds it.
- A spent budget also watches the device list, so a replug recovers. That watch lives **only** in the out-of-budget branch.
- `Screen` deliberately never reopens: `getDisplayMedia` needs a user gesture, and "Stop sharing" is an explicit stop, not a failure. It reruns into a stopped branch, which releases every track. Switching the source off is the app's reset.
- A track that arrives already `ended` counts as a failure rather than being published, since it fired `ended` before anything could listen.

### Two triggers, deliberately different

**Settings** (device, constraints) are *intent*: they restart a healthy capture and revive a spent one. **The device list** is *environment*: it counts only while dead. Subscribing to it while healthy would restart a working capture whenever unrelated devices churn, since `devicechange` also fires for audio *output* changes. `unrelated device churn does not disturb a healthy capture` pins that.

`Screen` keeps reading its constraints after the stopped check, which looks like the same bug and is not: reviving there would re-prompt without a user gesture.

### Relationship to #2557

#2557 removed a code path that *incidentally* retriggered capture when the guessed default device changed, which had been masking part of this. With an unpinned `requested` now `undefined` in both states, unplugging the only camera and plugging it back in no longer recovered on its own. The device-list refund here closes that; the routing fix in #2557 stands.

## Public API changes

None. `Retry` lives in `source/retry.ts` and is deliberately **not** re-exported from `source/index.ts`, so it stays package-private. `Camera`, `Microphone`, and `Screen` keep their existing shapes; only behavior changes.

## Test plan

`js/publish/src/source/retry.test.ts`, 15 tests. Each behavior below was first confirmed to fail against the commit before it:

| behavior | previously |
|---|---|
| running out of retries clears the source | published a track with `readyState: "ended"` |
| a replug recovers a capture whose retries all failed | 0 further attempts, dead forever |
| screen releases every track when one ends | sibling track stayed `live`, `stopped: false` |
| picking a different device revives a spent capture | picker inert, capture stayed idle |
| fixing a constraint revives a spent capture | stayed idle until `enabled` was toggled |

Plus: both sources reopen on death, a born-dead track is not published, unrelated churn leaves a healthy capture alone, screen releases on either track ending, and screen re-prompts after an off/on toggle.

Also fixed the fake track in `device.test.ts`, which lacked `addEventListener` and was silently swallowing a throw.

Verified in a browser via `just dev` with `navigator.mediaDevices` stubbed (the browser pane blocks real capture):

- Screen, video-only end: `out.source` cleared, `audioStopped: true` (previously `live`), `getDisplayMedia` still 1 call.
- Camera + mic, budget exhausted: both `out.source` cleared, `getUserMedia` stops at the budget.
- Replug: camera revived on a video-list change, mic on an audio-list change. The refund is per device kind.
- Device picker: exhausted on `cam1` left the source cleared at 8 `getUserMedia` calls; selecting `cam2` produced exactly one more and republished a `live` track.

`just js check` and `just js fix` clean. Full `just js test` green across all 8 packages, `@moq/publish` 41 pass / 0 fail.

Not verified against real hardware, so unplug and "Stop sharing" are exercised through the stub rather than a physical device.

## Cross-package sync

No rows apply: no wire format, no `moq-ffi`, no CLI flags. `demo/web` consumes `<moq-publish>` but not these classes directly. No `/doc` page documents capture lifecycle.

(written by Opus 5)


kixelated:
@codex review

PR #2566, opened 2026-07-30
feat(moq-hls): serve DASH manifests from the timeline

kixelated:
Follow-up to #2547, verifying its premise by construction: the reworked timeline carries enough information to dynamically produce both HLS playlists and DASH manifests. HLS already rendered from it; this adds the DASH half to the same export origin.

## What

- `GET /{broadcast}/manifest.mpd` renders an MPD from the same per-rendition timeline windows the HLS playlists use: `dynamic` while live, `static` once the broadcast finishes. Representations group into AdaptationSets by codec (mirroring the HLS master's audio groups), with codecs/bandwidth/resolution/frameRate/audioSamplingRate/channels sourced from the catalog.
- Segments are `$Time$`-addressed via `SegmentTimeline`: `seg/t{pts}.m4s` is an alias for the same bytes as the HLS `seg/{number}.m4s`. Every `S` carries an explicit `t` (the record's `pts` in the timeline's own timescale, exact by construction), so gaps and mid-window joins can never mis-address a segment the way positional `$Number$` accumulation would.
- `availabilityStartTime` uses the catalog timeline's declared `wall` anchor when present; otherwise an anchor is estimated once, when the first record arrives (a record publishes when its segment completes, so pts 0 maps to `now - (pts + duration)`), and reset with the window if the timeline restarts.
- The `?query` credential propagation (e.g. `?jwt=`) rides every child URL exactly like the HLS master, XML-escaped.
- Docs the change touches (crate/module headers, route list, moq-cli `hls export` wording) updated in the same diff.

## Verification

- 7 new tests: MPD renderer units (dynamic/static, presentationTimeOffset, query escaping, empty-timeline representations omitted) plus in-process integration through a real origin (aligned video+audio S entries, `$Time$` bytes == number-addressed bytes, live→static transition). 56 total moq-hls tests green; clippy and fmt clean.
- End-to-end with a stock player: imported a 12s H.264+AAC MPEG-TS (2s GOPs) through `ts::Import` into an origin served by `Server::router`, then drove it with ffmpeg's DASH demuxer. The dynamic manifest probes correctly (both streams identified), the static manifest fully decodes (360 video frames, 12s audio), and `curl`'d `seg/t{pts}.m4s` bytes are identical to `seg/{n}.m4s`. The HLS side decodes the same content through `master.m3u8` as before.

One observed-and-kept behavior: a clean finish promotes the live-edge group into a final segment whose duration is 0 when its last frame has no successor to bound it; the MPD lists it with `d="0"`, mirroring HLS's trailing `EXTINF:0`.

(written by Fable 5)

### DIFF
--- a/rs/moq-cli/src/args.rs
+++ b/rs/moq-cli/src/args.rs
@@ -218,7 +218,7 @@ pub enum ExportSink {
 	H264(Container),
 	/// H.265 Annex-B elementary stream to stdout.
 	H265(Container),
-	/// Serve HLS / LL-HLS over HTTP.
+	/// Serve HLS / LL-HLS and DASH over HTTP.
 	Hls(crate::hls::ExportArgs),
 	/// RTMP: push to a remote (`--connect`) or serve plays (`--listen`).
 	Rtmp(crate::rtmp::ExportArgs),
--- a/rs/moq-cli/src/hls.rs
+++ b/rs/moq-cli/src/hls.rs
@@ -1,5 +1,5 @@
-//! HLS endpoints: pull a remote playlist into MoQ (import), or serve HLS over
-//! HTTP from MoQ broadcasts (export), fetching media groups on demand.
+//! HLS endpoints: pull a remote playlist into MoQ (import), or serve HLS and
+//! DASH over HTTP from MoQ broadcasts (export), fetching media groups on demand.
 
 use std::net::SocketAddr;
 use std::time::Duration;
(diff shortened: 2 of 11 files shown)


kixelated:
Addressed three findings from an adversarial (Codex) review in d2399a2:

1. **Stale media across publisher restarts**: segment/init URLs aren't globally unique (a new timeline can repeat pts and numbers with different media), so the one-year `immutable` cache policy could pin a previous generation's bytes. Media responses are now capped at `max-age=<window>` — every concurrent viewer of the live window still hits the cache, and a stale generation ages out as fast as the window that stopped listing it. This covers the pre-existing HLS number addressing too, which had the same exposure. (The fully-correct fix — a generation/epoch discriminator in the URLs — needs a publisher-supplied identity, e.g. a timeline epoch in the catalog section, to avoid fragmenting the CDN cache per origin instance; left as a possible upstream follow-up.)

2. **presentationTimeOffset shifted on live→static**: the static render used to offset the presentation to its first listed segment, changing representation timing in place on the finishing reload, which DASH forbids. Presentation time now stays anchored at pts 0 in both states (no PTO at all); verified with ffmpeg that a static manifest with a lead-in starts playback at the first listed `S@t` and fully decodes.

3. **Zero-duration tail segments**: the terminal record is `d=0` when the publisher can't bound its final sample, and such a SegmentTimeline entry sits exactly at the presentation end where a conforming player never requests it. These are now omitted from the MPD (HLS keeps its `EXTINF:0`, which players fetch by URL rather than by timing). Bounding the final sample for real belongs upstream in the timeline finalizer.

All 56 moq-hls tests green; re-verified end to end with ffmpeg against the 12s H.264+AAC TS import in both dynamic and static states.

(written by Fable 5)

PR #2584, opened 2026-08-01
fix(moq-video): keep Media Foundation decoded frames on the GPU, and stop losing frames at group boundaries

kixelated:
## Summary

The Windows decode-surface retention from #2481, plus two more Media Foundation frame-lifetime bugs the hardware surfaced along the way. All of it validated on a real GPU (RTX 3070 Ti); none of it is reachable by CI, which never compiles these paths.

**Decoded frames stay on the GPU.** A Media Foundation decode now yields `Surface::Texture` instead of downloading every picture to I420, so decode -> render and decode -> re-encode never touch the CPU.

The plan's literal shape (retain the DXVA surface, the mirror of the VideoToolbox change in #2467) does not work on Windows, and the hardware says so three ways. The decoder's output is an 8-slice texture array bound `D3D11_BIND_DECODER` and nothing else:

- No shader can sample a slice and no encoder can read it, so retention would hand out a frame nobody can use.
- Slices are recycled the moment their sample is released. Holding a pool's worth wedges the decoder with no error to report: the MFT blocks *inside* `ProcessInput` waiting for a picture buffer a consumer is holding.
- Releasing a retained `IMFSample` after the decoder's `MFShutdown` is an access violation, which any frame outliving its decoder would hit.

So each picture takes one GPU-to-GPU blit into a texture we own, with bind flags read from `CheckFormatSupport` rather than assumed (this GPU grants `SHADER_RESOURCE | RENDER_TARGET | VIDEO_ENCODER`).

**Root cause 2: decoded frames were the wrong size.** The blit crops coded to display size, which exposed it. `MF_MT_FRAME_SIZE` is the coded size, so every decode of a stream whose height is not a whole number of macroblocks came back padded, 1080p as 1088 rows. The size now comes from the stream's minimum display aperture where it declares one.

**Root cause 3: the hardware encoder lost a frame per group.** `finish` swept whatever events were already queued, but a hardware MFT holds a frame or two and posts those events *after* the drain command. On a fetched group that dropped the tail. On a live track it was worse: nothing emptied the codec at a group boundary, so the tail of one group surfaced in the next ahead of its keyframe, where a subscriber joining there cannot decode it. `Backend` grows a `flush` that empties the pipeline and leaves the encoder running, `finish` waits the drain out, and the transcode live loop flushes at each boundary. This is what fixes the two `moq-transcode` hardware tests, which failed on every Windows machine with a GPU before this PR.

**Root cause 4: capture had the same pool-recycling hazard.** The source reader hands out textures it recycles as soon as the sample is released, while a captured frame waits in a 4-deep channel. It now takes the same blit out of the pool, which leaves `copy_from_sample` as the only way to build a `Texture` at all.

## Public API changes

Additive only, so this targets `main`.

- `moq_video::frame::d3d11::Texture` (Windows): new `texture()`, `device()`, `width()`, `height()`. Without them `Surface::Texture` is opaque to anyone outside the crate, and the retention only pays off in-tree.
- `moq_video::windows` (Windows): the `windows` crate re-exported, matching the existing objc2 re-exports, so a consumer naming a texture or a device is not guessing at a compatible version. A major bump there is a breaking change for this crate.
- `moq_video::encode::Encoder::flush()`: new.
- `pub(crate)` only, not public surface: `Backend::flush` (no default, so a future pipelined backend has to think about it), `Texture::copy_from_sample` replacing `Texture::new`.

## Test plan

Six Media Foundation tests on hardware, all skipping cleanly on a machine without a GPU, plus one portable test that runs everywhere:

- `mediafoundation_decode_stays_gpu_resident`, `mediafoundation_held_frames_keep_their_pixels`, `mediafoundation_decode_crops_coded_padding`, `mediafoundation_decoded_texture_reencodes_in_place`.
- `a_flush_empties_a_pipelined_backend_and_leaves_it_running`, over the existing `Delayed` test backend, so the group-boundary invariant is covered on Linux and macOS CI too.
- Each regression test was verified to fail against the bug it encodes. Reverting to the zero-copy hand-out gives `frame 0 decoded to luma 153, expected about 43`, a later picture written over a held frame. The re-encode test decodes its own output back to pixels, so a blit that never landed fails rather than merely producing packets.
- `moq-transcode`'s `end_to_end_hardware` and `live_multi_rung_hardware` now pass; both failed (4 of 5 frames) on `main`.
- `cargo test -p moq-video -p moq-transcode`, and `cargo check --workspace --exclude moq-gst --all-targets` (the `just rs windows` gate). Clippy and `cargo doc -D warnings` clean; the two pre-existing `collapsible_if` warnings in the MF backends are fixed in passing since Windows clippy could not be clean otherwise.

## Not in this PR

- **GPU resize for D3D11 textures**, the counterpart to #2512 on macOS. Decoded frames are now textures and `Surface::resize` has no GPU scaler, so a transcode ladder downloads once per rung instead of once per frame. I implemented it with `ID3D11VideoProcessor` and backed it out: `VideoProcessorBlt` deadlocks on a device that has a live DXVA decoder session. Two experiments isolate it. A cold processor blitting a copy destination on a *fresh* device works; the same blit on the *decoder's* device hangs, and hangs inside `VideoProcessorBlt` itself. A warm processor (one that already blitted a texture untouched by the decoder) then works on the decoder's device. Shipping a resize that can wedge the decode thread is worse than a download per rung, so this needs the `IMFDXGIDeviceManager::LockDevice` protocol investigated first.
- **Cross-package sync**: no rows apply. No wire format, no `moq-ffi` surface, and `moq-video` has no `doc/` page yet (#2481 tracks writing one).

## Review round

All six findings from Codex and CodeRabbit are addressed in 06066d58. Two were worth more than their labels suggested:

- The display aperture comes off the network and was used unclamped. `CopySubresourceRegion` reports no status, so a stream declaring an aperture larger than its coded frame would have produced a frame claiming dimensions over a texture nothing ever wrote. Now clamped to the coded size.
- `Texture::subresource()` is gone rather than re-documented. Every texture is now one this crate allocated with a single slice, so the accessor could only ever return 0.

The one suggestion not taken is CodeRabbit's request for a test that makes `MFT_MESSAGE_NOTIFY_END_OF_STREAM` fail: the backend drives a real MFT enumerated from the system, so there is nothing to inject the failure through short of a mock transform layer this crate does not have.

## One thing I could not explain

`live_multi_rung_hardware` failed once (4 of 5 frames) in a run where the `moq-video` and `moq-transcode` test binaries were executing concurrently, then passed 13 consecutive runs including 8 more of the same concurrent shape. I could not reproduce it or find a mechanism. Contention for NVENC sessions across two processes is the obvious suspect, but that is a guess, not a diagnosis.

(Written by Claude Opus 5)

### DIFF
--- a/rs/moq-transcode/src/rung.rs
+++ b/rs/moq-transcode/src/rung.rs
@@ -144,6 +144,13 @@ async fn live(rung: &Rung, producer: &mut moq_net::track::Producer) -> Result<()
 
 			match item {
 				Some(Item::Group(sequence)) => {
+					// Empty the codec before opening the next group even though this one
+					// is being abandoned: a pipelined encoder still holding the previous
+					// group's tail would otherwise emit it into the new group, ahead of
+					// the keyframe requested just below.
+					if let Some(encoder) = &mut encoder {
+						encoder.flush()?;
+					}
 					if let Some(output) = current.take() {
 						// A group boundary without an end: treat as incomplete.
 						output.abort(moq_net::Error::Cancel)?;
@@ -205,6 +212,11 @@ async fn live(rung: &Rung, producer: &mut moq_net::track::Producer) -> Result<()
 				}
 				Some(Item::End) => {
 					if let Some(mut output) = current.take() {
+						// The source group is complete, so this one has to be too: a
+						// hardware encoder is still holding its last frames.
+						if let Some(encoder) = &mut encoder {
+							write(&mut output, encoder.flush()?)?;
+						}
 						output.finish()?;
 					}
 				}
--- a/rs/moq-video/src/capture/mediafoundation.rs
+++ b/rs/moq-video/src/capture/mediafoundation.rs
@@ -3,10 +3,11 @@
 //! Drives an [`IMFSourceReader`] over the selected capture device. When a
 //! Direct3D11 device is available the reader runs with that device's DXGI
 //! manager and the advanced video processor, so each sample arrives as a
-//! GPU-resident NV12 texture ([`Surface::Texture`]) that the hardware encoder MFT
-//! consumes zero-copy. Without a GPU (e.g. a headless VM) it falls back to the
-//! source reader's software video processor, copying each sample to a packed CPU
-//! [`I420`] ([`Surface::I420`]) the encoder uploads.
+//! GPU-resident NV12 texture: one blit out of the reader's pool (see
+//! [`Texture::copy_from_sample`]) and it is a [`Surface::Texture`] the hardware
+//! encoder MFT consumes with no further copy. Without a GPU (e.g. a headless VM)
+//! it falls back to the source reader's software video processor, copying each
+//! sample to a packed CPU [`I420`] ([`Surface::I420`]) the encoder uploads.
 
 use std::ffi::c_void;
 use std::ptr;
@@ -208,9 +209,15 @@ impl Camera {
 		})
 	}
 
-	/// Wrap a GPU sample's DXGI texture as a zero-copy [`Surface::Texture`].
+	/// Take a GPU sample's picture into a [`Surface::Texture`] of our own, on the
+	/// GPU.
+	///
+	/// The reader hands out textures from a pool it recycles as soon as the sample
+	/// is released, and a captured frame outlives that: it waits in the frame
+	/// channel while the pump reads on. Blitting it out is what stops the next
+	/// picture landing on a frame the encoder has not reached yet.
 	fn sample_to_texture(&self, device: &ID3D11Device, sample: &IMFSample) -> Result<Surface, Error> {
-		let texture = Texture::from_sample(device, sample, self.width, self.height)?;
+		let texture = Texture::copy_from_sample(device, sample, self.width, self.height)?;
 		Ok(Surface::Texture(texture))
 	}
 
--- a/rs/moq-video/src/capture/mediafoundation.rs
+++ b/rs/moq-video/src/capture/mediafoundation.rs
@@ -12,10 +12,10 @@ use std::ffi::c_void;
 use std::ptr;
 use std::slice;
 
-use windows::Win32::Graphics::Direct3D11::{ID3D11Device, ID3D11Texture2D};
+use windows::Win32::Graphics::Direct3D11::ID3D11Device;
 use windows::Win32::Media::MediaFoundation::{
-	IMF2DBuffer, IMFActivate, IMFAttributes, IMFDXGIBuffer, IMFDXGIDeviceManager, IMFMediaSource, IMFSample,
-	IMFSourceReader, MF_DEVSOURCE_ATTRIBUTE_FRIENDLY_NAME, MF_DEVSOURCE_ATTRIBUTE_SOURCE_TYPE,
+	IMF2DBuffer, IMFActivate, IMFAttributes, IMFDXGIDeviceManager, IMFMediaSource, IMFSample, IMFSourceReader,
+	MF_DEVSOURCE_ATTRIBUTE_FRIENDLY_NAME, MF_DEVSOURCE_ATTRIBUTE_SOURCE_TYPE,
 	MF_DEVSOURCE_ATTRIBUTE_SOURCE_TYPE_VIDCAP_GUID, MF_DEVSOURCE_ATTRIBUTE_SOURCE_TYPE_VIDCAP_SYMBOLIC_LINK,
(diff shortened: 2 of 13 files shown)


kixelated:
All six findings addressed in 06066d58. Two deserve more than a checkmark:

**The aperture clamp is the best catch here.** `MF_MT_MINIMUM_DISPLAY_APERTURE` comes out of the bitstream, so it comes off the network, and `CopySubresourceRegion` returns no status. An aperture larger than the coded frame would have produced a frame reporting oversized dimensions over a texture nothing ever wrote, silently. Clamped to the coded size.

**`Texture::subresource()` is removed rather than re-documented.** The suggestion was to document it as "always 0", which is accurate but leaves an accessor that can only ever return one value. Since `copy_from_sample` is now the only constructor and it allocates a single-slice texture, the field itself was dead: it and the accessor are gone, and the two internal uses pass 0 directly.

The rest went in as proposed: the end-of-stream failure now fails the drain rather than going on to block for a completion the MFT never promised; `windows` is re-exported alongside the objc2 bindings; the module doc no longer claims a zero-copy render path that does not exist on Windows yet; and the re-encode test asserts its frame count, which matters because a per-frame check alone cannot see the pipelined tail loss the flush exists to prevent.

One suggestion not taken: the regression test for a failing `MFT_MESSAGE_NOTIFY_END_OF_STREAM`. The backend drives a real MFT enumerated from the system, so there is no seam to inject that failure through without a mock transform layer this crate does not have. The `Delayed` test backend covers the flush contract at the `Backend` level instead, and that one runs on every platform.

(Written by Claude Opus 5)

kixelated:
Good catch, fixed in fe0827e7.

`process_output` returns `true` for a handled `MF_E_TRANSFORM_STREAM_CHANGE` without pushing a frame, so the guard's frame-count check really would have turned a mid-stream resolution change into a fatal codec error, and precisely in the case where retrying the same access unit is correct. `drain_output` now returns whether the decoder got anywhere at all, which covers a picture and a renegotiation alike, and the guard fires only when neither happened.

Worth noting for anyone reading the guard later: it exists because `ProcessInput` returning `MF_E_NOTACCEPTING` with nothing drainable is an unbounded retry loop, which is how this backend used to wedge when a consumer held the decoder's picture buffers. That failure mode is gone now that frames are copied out of the pool, but the loop still needs a floor.

(Written by Claude Opus 5)

kixelated:
Taken, in e34a5e38, with one correction to the framing: this is not currently reachable.

The mechanism is real. A drain restarts the MFT's stream, so the picture after it is not temporally continuous with the one before, and nothing said so. The first sample after a flush now carries `MFSampleExtension_Discontinuity`, and the flag clears only once `ProcessInput` accepts the sample, so a rejected one keeps the mark for the retry.

Where I would not describe it as a data-integrity bug today: `moq-transcode` flushes at a group boundary and then calls `keyframe()`, because a subscriber has to be able to start at that group. The next picture is an IDR either way, and an IDR is a clean restart whether or not it is labelled one. What the fix actually protects is `Encoder::flush`, which this PR makes public, for a caller that flushes without asking for a keyframe.

No regression test, for the same reason as the end-of-stream one: the attribute goes onto an `IMFSample` handed to a real MFT enumerated from the system, and nothing on the way out reflects it back, so there is nothing to assert against short of a mock transform layer this crate does not have. Verified only that the existing hardware tests still pass, which is weak evidence and I would rather say so than dress it up.

(Written by Claude Opus 5)

kixelated:
Skipping this one, and not because it is unimportant: the fix as described cannot do what its rationale asks for.

Every operation the comment lists happens in `start()`, not `open()`: the D3D manager, `configure_codec_api`, `set_output_type`, `set_input_type`, `read_output_info`. And `start()` is deliberately deferred to the **first frame**, because the Direct3D11 device to bind comes from the frame itself. By the time any of it can fail, `open()` has long since returned `Ok`, the backend-selection layer in `encode/backend/mod.rs` has committed to this backend, and `enumerate_encoder` is out of scope entirely. Looping inside `enumerate_encoder` would not retry a single one of them.

There is a narrower real point underneath. `open()` can fail *after* activation, in the async unlock or the `IMFMediaEventGenerator` / `ICodecAPI` casts, and a machine with two hardware encoder MFTs would give up on the whole backend rather than trying the second. That is worth having. But it is a change to backend selection that predates this PR (my only edit to that function was a clippy-driven let-chain), it wants the same treatment in `enumerate_decoder`, and doing it for the configuration failures the comment actually cares about means unpicking the lazy-start design so a failed `start()` can fall back to another MFT. That is a different PR with a different test plan, not a quick win on this one.

Filing it as a follow-up rather than smuggling it in here. This PR is already four root causes deep and green; I would rather it stay reviewable.

(Written by Claude Opus 5)

PR #2593, opened 2026-08-02
feat(cli): add a `token` subcommand to moq-cli

kixelated:
## Summary

- `moq token generate|sign|verify` now does everything the standalone `moq-token` binary does, so a relay operator with `moq` installed needs no second CLI.
- The command tree moved out of `moq-token-cli`'s `main.rs` into that crate's new `[lib]` target, as `moq_token_cli::Args`. The binary flattens it, `moq-cli` nests it under `token`. One implementation, two entry points, rather than a copy in each binary.
- `token` joins `devices` as a *local* verb: handled before any transport is bound, and it rejects `--client-connect` / `--server-bind` instead of silently ignoring them. `MoqSide::reject` lost its `#[cfg(feature = "capture")]` gate, since `token` is always compiled in.
- `moq-token` keeps working unchanged, same flags and same binary name. Nothing is deprecated or removed here, so the release/packaging surface (nfpm, brew, winget, nix, Docker) is untouched. Happy to follow up with a removal if you want the crate gone.

The shared module first landed in the `moq-token` *library* behind a `cli` feature. That was wrong: it put a clap `Args` type and an `anyhow::Result` into a published library's API, neither of which the crate wraps or re-exports. Swapping the error type alone would have missed the clap half, so the module moved to the CLI crate instead, where clap and anyhow are the house style. Details in [this comment](https://github.com/moq-dev/moq/pull/2593#issuecomment-5159490130).

Flattening the shared `Args` into `moq-token-cli`'s own `Cli` initially aborted at startup with `Argument group name must be unique: 'Cli' is already in use` - clap derives an arg group per struct name, and both were called `Cli`. Hence `Args`, plus a `debug_assert` test in each binary, since the collision only manifests in whichever crate flattens it.

## Public API changes

- **Added** a `[lib]` target to `moq-token-cli` exposing `pub struct Args` (clap `Args`) and `pub fn Args::run(self) -> anyhow::Result<()>`. The subcommand enum and its fields stay private, so the surface is one opaque type plus one terminal method that consumes `self`.
- **No change** to `moq-token`'s API. It gains no feature, no clap, and no anyhow (`cargo tree --edges normal` confirms; anyhow remains only as its pre-existing dev-dependency).
- Additive only, so this targets `main`.

## Bug fix

`parse_unix_timestamp` built its result with `SystemTime + Duration`, which panics on overflow, so a large `--expires` aborted the process instead of returning the intended parse error. It's now `checked_add`.

Reachability is platform-dependent, and worth stating precisely since it decides what the test can assert: the input is parsed as `i64` before converting to `u64`, so the largest reachable value is `i64::MAX` seconds. That is exactly representable in a timespec, so Linux and macOS return `Ok` and never panic. Windows counts 100ns ticks and overflows well before that. The regression test therefore asserts the maximum never panics, and separate cases cover the pre-epoch and non-numeric inputs, which fail deterministically on every platform.

## Cross-Package Sync

- `rs/moq-cli` -> `doc/bin/cli.md`: documented the verb and the local-verb exception to the grammar.
- `rs/moq-token-cli` -> `doc/bin/relay/auth.md`, `doc/lib/rs/crate/moq-token.md`, `doc/lib/rs/index.md`: noted that the same commands are available as `moq token ...`. Existing `moq-token` examples still parse, so they were left alone.
- `rs/moq-token` -> `js/token`: skipped. No token semantics, claims, or key handling changed.
- No wire format change, so no draft updates.

## Test plan

- `cargo test -p moq-token -p moq-token-cli -p moq-cli`: 147 passing. New tests:
  - `moq_token_cli`: the `--out`/`--public` pair round-trips a real ES256 signature, `sign` runs through the CLI grammar so both timestamp parsers are exercised, `--out-dir` names the file after the kid, both the two-keys-to-stdout and two-inputs-from-stdin collisions are rejected, and the timestamp cases above. This logic was previously unreachable from tests, since it lived in a binary's `main()`.
  - `moq-cli`: clap `debug_assert`, plus a `moq token` test asserting it needs no MoQ side *and* rejects one when given.
- `cargo clippy --all-targets`, `cargo fmt --all --check`, `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps` on all three crates.
- The `doc/bin/cli.md` walkthrough run verbatim end to end, then the resulting token re-verified with the standalone `moq-token` binary.
- Not run locally: `just check` (no nix/just in this environment), `cargo shear` / `cargo sort` (not installed). Manifest additions were kept alphabetical by hand.

## Review follow-ups

Codex and CodeRabbit both reviewed; findings and the reasoning for the two I skipped are in the [first](https://github.com/moq-dev/moq/pull/2593#issuecomment-5159380380) and [second](https://github.com/moq-dev/moq/pull/2593#issuecomment-5159490130) comments. One is worth surfacing here: **`moq_token::Key::to_file` writes private key material with default permissions**, so generated `.jwk` files are not owner-only on Unix. That affects every caller of the library method, not just this CLI, so it wants its own change against `moq-token` rather than a patch at this call site.

(Written by claude-opus-5)


kixelated:
Both Codex findings on the `doc/bin/cli.md` walkthrough were real; fixed in fbca186.

- **Missing `alice.jwt`**: `sign` now redirects to the file the `verify` step reads.
- **Expired `--expires 1735689600`**: dropped the literal rather than replacing it with another future timestamp, which would just rot again. The comment above the command names `--expires <unix-timestamp>` so the flag is still discoverable, and the block now copy-pastes correctly on any platform (no `date -d` portability trap).

Two pre-existing copies of the same stale timestamp live outside this change, at `doc/lib/rs/crate/moq-token.md:89` and `doc/lib/rs/index.md:200`. They mint already-expired tokens too, but they are standalone snippets rather than a chained walkthrough, so I left them for a separate pass rather than widening this diff. Happy to fold them in if you'd rather.

Same commit adds test coverage for the routing logic, which only became reachable from tests once the command surface moved into the library: the `--out`/`--public` pair round-trips a real signature, `--out-dir` names the file after the kid, and both stdin/stdout collisions are rejected. 130 tests pass in `moq-token` under `--all-features`, which is what CI runs.

Sourcery hit its weekly rate limit instead of reviewing, so per CONTRIBUTING.md I also ran the review skill locally against the diff. Its only findings were the two above (the `alice.jwt` one independently) plus the missing test coverage; the moved code itself diffs clean against the original modulo the structural `main()` -> `Args::run` change.

(Written by claude-opus-5)

---

kixelated:
Addressed in 6e27768.

**Codex on the `anyhow` leak: taken, via its second suggestion.** Returning a crate-owned error would have fixed half the problem, since `pub struct Args` is `#[derive(clap::Args)]` and so puts clap in the signature too, and it would have meant inventing variants nobody can match on when the only caller is `fn main()`. The root cause was CLI code living in a library crate, so the module moved to `moq-token-cli`, which grows a `[lib]` target that `moq-cli` depends on. `moq-token`'s public API is back to where it started, with no clap and no anyhow (`cargo tree --edges normal` confirms both are gone; anyhow remains only as the pre-existing dev-dependency). Still one implementation behind both binaries.

**CodeRabbit's `parse_unix_timestamp` overflow: taken, but not as written.** The `checked_add` hardening is right, and it's in. The suggested regression test is not: `SystemTime` on Linux and macOS is a timespec with `i64` seconds, and the input is parsed as `i64` before converting to `u64`, so the largest reachable value is exactly representable and returns `Ok`. I verified this directly rather than assuming:

```
secs=9223372036854775807: checked_add -> Some
i64::MAX add -> Ok(SystemTime { tv_sec: 9223372036854775807, tv_nsec: 0 })
```

So `assert!(parse_unix_timestamp(&i64::MAX.to_string()).is_err())` would have gone red on CI. The panic is reachable on Windows, where `SystemTime` counts 100ns ticks and runs out far sooner, which is why the fix is still worth landing. The test asserts it never panics, and a separate test covers the pre-epoch and non-numeric cases, which are deterministic everywhere.

**Also taken:** the `moq token` rejection path is now actually asserted (the old test only covered the no-side case, so the new behavior was untested); `sign` runs through the CLI grammar so both timestamp parsers are exercised; test scratch dirs use `tempfile` instead of a fixed shared path; `--root` documents that it applies only alongside `--publish`/`--subscribe`; and `moq devices` notes the `capture` gate, matching how the capture section of the same page already words it.

**Skipped, with reasons:**

- **Owner-only permissions on private key files.** Real issue, wrong layer and wrong PR. `write_key` just calls `moq_token::Key::to_file`, which uses `std::fs::write`; so does `set::to_file`. Every caller of that library method has the same exposure, not just the CLI, so fixing it at this call site would leave the hole open for the next caller. It's pre-existing and deserves its own change against `moq-token`. Worth doing, and I'm happy to take it separately.
- **Move `anyhow`/`clap` into `[workspace.dependencies]`.** Moot for `moq-token` now that both are gone from it. For the CLI crates, literal versions are the existing convention (`moq-cli`, `moq-relay`, and `moq-token-cli` all declare clap that way today), so changing it here would be an unrelated sweep.

Verified: 147 tests pass across the three crates, `cargo clippy --all-targets`, `cargo fmt --check`, `RUSTDOCFLAGS="-D warnings" cargo doc`, and the documented walkthrough end to end through both binaries.

(Written by claude-opus-5)

---

kixelated:
Good catch on the crate map, fixed in b361165. I updated `rs/CLAUDE.md` for the original design and then didn't revise it when the module moved, so it described a `moq-token` `cli` feature that no longer exists. `moq-token-cli` now gets its own entry saying it's lib+bin and owns `moq_token_cli::Args`.

The earlier thread on `Args::run` returning `anyhow::Result` is resolved by that same move: the method is no longer part of the `moq-token` library's API, and `moq-token-cli` is a CLI crate where anyhow is the house style per `rs/CLAUDE.md`.

(Written by claude-opus-5)

---

kixelated:
Good find on `--broadcast`, fixed in 288db0c. I'd spotted the same hole reviewing my own diff and talked myself out of it as pre-existing on `devices`, which was the wrong call: this PR adds documentation promising local verbs reject a MoQ side, so the gap became a mismatch I introduced. `moq --broadcast rooms/123 token generate` now errors, and the message names the offending flag rather than listing all of them.

I did **not** take "reject every non-default MoQ-side option", because it would break the common case. `--origin` is declared `env = "MOQ_ORIGIN"`, and `--client-connect` / `--server-bind` read `MOQ_CLIENT_CONNECT` / `MOQ_SERVER_BIND`, so clap populates those from the environment with no flag on the command line. Rejecting them wholesale means `moq token generate` fails in any shell configured for publishing, which is precisely where an operator mints a token. Verified both sides:

```
$ moq --broadcast rooms/123 token generate
Error: `token` runs locally and takes no MoQ side; drop --broadcast

$ MOQ_ORIGIN=5 moq token generate
  ok: still works with MOQ_ORIGIN exported
```

So the check covers `--client-connect`, `--server-bind`, and `--broadcast`, and `--origin` is exempt with the reasoning in a comment on `reject`. The transport-tuning flags stay unchecked for the same env reason, and passing `--client-tls-cert` without a connection is already inert for every verb rather than misleading.

Worth noting the distinction this rests on: a flag typed on the command line states intent about *this* invocation, while an env var is ambient shell config. Only the first should fail a local verb. clap can tell them apart via `ArgMatches::value_source`, but not through the derive API we parse with, so the env-provided values still pass silently. Happy to wire that up if you want the stricter behavior.

Local gate before pushing: `cargo fmt --check`, `cargo sort --check`, `cargo shear`, `cargo clippy --all-targets`, and 148 tests across the three crates.

(Written by claude-opus-5)

---

PR #2604, opened 2026-08-03
feat(kio): add Fan, a waker for a whole WaiterList

kixelated:
## Summary

- Add `Fan<T>`, a cloneable owner or projection of a `WaiterList` that exposes one stable `Waker` for polling a foreign future on behalf of every parked caller.
- Add `Hold<T>`, which records wakes while a caller is polling that foreign future under a lock, then delivers the deferred wake after the last hold drops.
- Keep projected fans weak so state can retain the fan or its waker without creating a reference cycle.
- Fix two opposing lock-order races found during review. `notify` first checks for a hold before acquiring projected state, which lets an inline wake defer instead of self-deadlocking on the caller's lock. It then re-checks under the target lock and keeps the deferral guard through the waiter drain, which covers a hold created while the wake was blocked on that target. User wakers are still invoked after both locks are released.
- Document the new primitive and remove prohibited em dashes from the changed prose.

## Public API changes

- Add `Fan<T = WaiterList>` with `new`, `project`, `register`, `wake`, `waker`, and `hold`.
- Add the `Hold<T = WaiterList>` guard returned by `Fan::hold`.
- Export the existing `Lock`, `LockGuard`, and `WeakLock` types required by `Fan::project`.

All changes are additive and target `main`.

## Test plan

- `nix develop --command just rs loom`: 15 kio models and 5 moq-net models passed.
- Mutation checks: the blocked-target model fails on the original ordering with `wake escaped a live hold`; the projected inline-wake model deadlocks when the pre-lock check is removed. Both pass with the two-phase fix.
- `nix develop --command just ci`: dependency policy, clippy, docs with warnings denied, and all 2,545 tests passed. The final flake evaluation then hit an unrelated host-toolchain limitation because Nixpkgs 26.11 dropped `x86_64-darwin`, which the current flake still evaluates. Fresh GitHub CI is required before merge.

No cross-package update is needed because this adds a runtime-independent kio primitive without changing a wire format or binding surface.

(Written by Opus 5 and GPT-5)

### DIFF
--- a/rs/kio/README.md
+++ b/rs/kio/README.md
@@ -17,7 +17,10 @@ The channel auto-closes when all producers are dropped.
 `Shared` is the role-less sibling for state that both sides mutate, and `Queue` is a
 poll-native FIFO queue (bounded or unbounded) built in the same style. `Park` holds a
 waiter for as long as a poll stays pending, bridging a `std::task::Context` to kio's
-waiter-based polls when implementing `Future` or `poll_*` on top of kio channels.
+waiter-based polls when implementing `Future` or `poll_*` on top of kio channels, and
+`Fan` hands out a waker that wakes a whole waiter list, for driving a foreign future on
+behalf of everyone parked on it — either over a list it owns, or over one already inside
+a `Lock`.
 
 It's used internally by [moq-net](https://github.com/moq-dev/moq/tree/main/rs/moq-net) and friends, but is generic enough to be useful on its own.
 
--- a/rs/kio/README.md
+++ b/rs/kio/README.md
@@ -19,8 +19,8 @@ poll-native FIFO queue (bounded or unbounded) built in the same style. `Park` ho
 waiter for as long as a poll stays pending, bridging a `std::task::Context` to kio's
 waiter-based polls when implementing `Future` or `poll_*` on top of kio channels, and
 `Fan` hands out a waker that wakes a whole waiter list, for driving a foreign future on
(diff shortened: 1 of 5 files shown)


kixelated:
Independently verified the deferral fix in 5a4d5b2 + bbe2291, and it corrects my earlier reading. I had argued the window could not be closed, because I only considered closing it by keeping the deferral lock across delivery. That shape does invert against a caller taking `hold` from inside its own state lock (state then defer), and it deadlocks. Checking `held` *behind* the target lock is the ordering I missed: it makes `notify` state then defer as well, matching the caller, with the cheap pre-check ahead of it so an inline wake never reaches for a lock its own thread already holds.

Both halves are load-bearing, and each is covered by one of the new Loom models. Confirmed by mutation:

- Removing the recheck under the target lock: `a_hold_covers_a_wake_waiting_for_the_list` fails.
- Removing the pre-check: Loom reports a deadlock, single thread blocked, which is the inline self-deadlock.

Also ran the full `cargo test -p kio` (47 unit plus doctest) and the whole Loom suite against the branch head. All green.

I had a follow-up commit prepared that documented the window as permanent and added a termination-only Loom model for it. Both are obsolete now, so I dropped it rather than pushing: your models assert deferral, which is the stronger property, and the doc paragraph would have been wrong.

One thing I got wrong that is worth recording: the em dashes you had to strip were mine. The rule is in `CLAUDE.md` line 81, and I wrote that code without reading it. Noted for anything further I send here.

(written by Opus 5)

PR #2647, opened 2026-08-05
feat: fail-fast retries: jittered backoff bounded by time, not error type

kixelated:
Addresses the `moq-dev/moq` half of moq-dev/moq.pro#921 with a fail-fast policy: **retries are bounded by time, with one narrow exception for HTTP statuses.** The moq.pro half is untouched.

## The policy

Every retry loop is bounded by *time*, not by guessing at error types. An ephemeral failure is, by definition, one that clears within a few seconds; so each loop retries everything indistinguishably with capped exponential backoff and jitter, and gives up after a short budget (~10s), surfacing the **last real error** rather than a synthetic timeout.

The one exception is an answer a peer actually gave, where the protocol defines the meaning: `Error::status()` on moq-native and moq-hls reports the HTTP status a server sent, and a status that isn't an invitation to retry (`404`, `405`, ...) fails immediately instead of burning the budget, while `408`, `429`, and the gateway/overload statuses ride the backoff. That is reading a response, not inferring intent from a failure; nothing without a status is classified.

Concretely:

- **Reconnect (`moq-native`, `js/net` Reload)**: default give-up budget drops from 5 minutes to 10 seconds, max delay from 30s to 5s. A dead relay or a typo'd URL now errors in ~10s instead of silently reconnecting for 5 minutes. Both remain configurable (`Backoff` / `ReloadDelay`), and `timeout = 0` still means retry forever for supervised publishers (the gst sink keeps using it).
- **HLS import**: retry budget drops from 300s to 10s. A broken origin surfaces as an error within seconds instead of the import papering over it while publishing nothing; a `404` playlist ends it immediately.
- **Supervisor loops with nobody to return an error to** (cluster peer dials, audio device reopen, the RTMP accept loop) necessarily retry forever: erroring out would mean the mesh never redials a peer that comes back an hour later, which is the truly silent failure. Instead their delay ceiling is seconds (cluster: 300s -> 10s), so a returning peer is picked up promptly and the warn log fires often enough that a dead one is loudly broken.

## Bug fixes

- **RTMP accept stalled ready handshakes.** The backoff was slept on inline, so a handshake that completed during it waited out the full delay. It's a deadline now, and the loop keeps polling `pending` through the pause.
- **Cluster backoff never reset, then reset on the wrong clock.** `run_remote_session` reports even a clean close as an error, so the reset keyed on `Ok(())` could never fire. Keying it on elapsed time then counted a slow dial as a healthy session. It now keys on the session's own lifetime.
- **HLS import read a quiet pass as a dead one.** `step(OnError::Warn)` returns `Ok` even when every rendition failed. Promoting that to a failure by segment count would break multi-rendition imports (a healthy live playlist writes no segments between passes), so it keys on renditions that ingested without error.
- **Reconnect's budget didn't cover the attempts**, only the sleeps, so a loop whose dials hung could outlive it and land past the linger window `Client::consume` sizes from it.

All six fixed-delay retry loops now escalate with jitter (native reconnect, cluster dials, HLS import/export, RTMP accept, audio device reopen, JS reconnect and capture reopen). `nix/modules/moq-relay.nix` and `packaging/moq-relay/moq-relay.service` gain `RestartSteps`/`RestartMaxDelaySec` (systemd 254+; older versions warn and keep the flat `RestartSec`).

## Public API changes

**One method**, so this targets `main`: `Error::status() -> Option<u16>` on `moq-native` and `moq-hls`. Nothing removed or signature-changed. Default *values* changed on `moq_native::Backoff` (max 30s -> 5s, timeout 5m -> 10s) and `js/net` `ReloadDelay` (max 30000 -> 5000, timeout 300000 -> 10000). `moq-relay`, `moq-audio`, and `moq-hls` gain a `rand` dependency (already in the lockfile) for jitter.

No wire-format change, so no `drafts/` update. No `moq-ffi` change, so no binding/wrapper sync. CLAUDE.md gains a short Retries section stating the policy.

## Test plan

- `just fix` + `just check`: clean (after rebasing onto main; the branch had fallen behind #2648's macOS lint fixes).
- `just rs test` for moq-native, moq-hls, moq-relay, moq-rtmp, moq-audio: 672 tests green. `bun test` for `@moq/net` + `@moq/publish`: green.
- New integration test `rs/moq-native/tests/reconnect.rs` asserts the budget ends the loop and the give-up error names the underlying cause; `rs/moq-native/src/quinn.rs` tests assert a rejected CONNECT carries its status through.

## Not verified locally

The `quiche` backend never compiles on this host (`boring-sys` needs cmake/nasm), nor does `moq-gst` (no GStreamer). CI covers both.

(Written by Claude Fable 5)


kixelated:
Pushed `9ffc69fc` for the review. Taken as flagged: the RTMP accept stall, the cluster backoff never resetting after a stable session, the jitter `u64` narrowing panic, the `Reload.closed` unhandled rejection, the JS multiplier clamp, the exhaustive `moq_mux` match, the WebSocket status helper, the stale `next()` in both guides, the flaky JS budget test, and the capture test's constants and constraint framing. The quinn `Client` arm was already covered by `c861b909`.

Two I'm declining, with reasons:

**`retry::Config` struct literal instead of `default()` + field set.** `Config` is `#[non_exhaustive]`, so a struct literal is illegal outside `moq-net` even with `..Self::default()`; the suggested diff wouldn't compile in `moq-native`. The root `CLAUDE.md` calls for exactly this shape ("callers build via `default()`/`new()` + field set, not struct literals"), and clippy's `field_reassign_with_default` doesn't fire here for the same reason.

**A give-up budget on the HLS export catalog wait.** This one is deliberate and the comment now says so. The loop isn't repeating a failed request; it's waiting for a publisher to create its catalog track, and the broadcast closing is what ends the wait. A relay-side broadcast outlives its publisher's session, so any deadline here becomes a window in which a publisher outage leaves the broadcaster permanently empty with nothing to recover it. That's the same failure mode as the earlier `NotFound` finding, just on a longer fuse.

Two fixes ship without a regression test, both because the loop has no seam to drive from a test: forcing `TcpListener::accept` to fail portably, and running a cluster peer session long enough to be stable. The behavior each encodes is stated in a comment at the site instead.

(Written by Claude Opus 5)

kixelated:
Pushed `c36f991a`. All four findings from this round were real, and two of them were holes the classification removal opened up:

- **Cluster reset keyed on the wrong clock** (CodeRabbit). It timed the whole dial rather than the session, so a peer that blackholed until the 30s connect timeout looked "stable" and reset the backoff on every attempt — the escalation never happened. `run_remote_session` now reports when the session actually came up, via a small `Attempt { connected, result }`, and the reset keys on the session's own lifetime. This was my regression from the earlier "reset after a stable session" fix.
- **HLS import reset on a failed pass** (Codex). `step(OnError::Warn)` returns `Ok` even when every rendition failed, so a source returning 404 forever would reset the backoff each pass and spin at the refresh cadence publishing nothing, never reaching the status arm or the budget. `StepOutcome` now carries the failure and a pass that imported nothing while a rendition was failing is treated as one.
- **Delay could outlive the budget** (CodeRabbit). Both twins now cap the jittered delay to what's left, so the sequence lands on the deadline instead of overshooting by up to a full window. `initial: 100, timeout: 5` no longer sleeps 100ms against a 5ms budget.
- **`status()` dropped on a lost transport race** (Codex). It reports one now, but only when both halves were answered and neither answer invites a retry — one transport being refused says nothing about the other.

No regression tests for the cluster or RTMP accept fixes: neither loop has a seam to drive a slow-dial failure or a failed `accept` from a test without inventing one. The invariant each encodes is stated at the site instead. The other two are covered (`a_delay_never_outlives_the_budget` in both languages, plus the budget test now pinning that the sequence doesn't overshoot by a window).

Context for anyone reading the diff cold: the retryable/non-retryable classification that earlier reviews were built around is gone as of `8e8adb54` — the budget is what stops a loop now. The one thing still read off a failure is an HTTP status a peer actually sent. `CLAUDE.md` has the reasoning.

(Written by Claude Opus 5)

kixelated:
Pushed `cac2ce17`. `moq_net::retry` is gone — `Backoff` now lives in `kio::time` next to `Deadline`, which is where a scheduling utility belongs rather than in the wire layer's published surface. `status_retryable` moved to private copies in moq-native and moq-hls, so neither crate exports anything about retry policy. `moq-net`'s only change from this PR is now zero: no new public API there at all.

Three findings from this round, all valid:

- **The budget didn't cover the attempts** (Codex). It ran from the first delay, so a reconnect whose every dial hung until the 30s connect timeout could outlive its budget by that much — and land past the linger window `Client::consume` sizes from `Backoff::timeout`. The clock now runs from construction or the last `reset`.
- **`doc/bin/relay/cluster.md` promised a give-up that no longer happens** (Codex). Good catch: that paragraph documented the terminal path removed in `8e8adb54`. It now describes the redial-forever behavior and what to watch for instead.
- **HLS multi-rendition failure ordering** (CodeRabbit) — fixed, but **inverted from the suggestion**. The proposal was to retain a non-retryable error once observed. That makes one permanently-dead variant end an import the other renditions could still serve, which is the opposite of what `OnError::Warn` exists for. The actual defect was order-dependence: whichever rendition failed last won, so `[transient, 404]` terminated while `[404, transient]` retried. The step now keeps the failure another pass could clear, so a pass is terminal only when nothing in it could recover.

Also took `checked_duration_since` over the bare subtraction. It's guarded by the `now >= deadline` check above it so it can't underflow today, but the checked form is free.

(Written by Claude Opus 5)

kixelated:
Pushed `89edb63d`. CI was red on `cargo sort`, not on anything semantic: the `kio` dependency I added to moq-audio, moq-relay, and moq-rtmp went in at a convenient anchor rather than its alphabetical slot. I've installed `cargo-sort` and `cargo-shear` locally now and both pass, along with `cargo fmt --all --check`, so that class of CI-only failure shouldn't recur on this branch.

Also fixed Codex's finding on the same push, which was a real one and sharper than the version I'd already addressed. The import promoted a step to a failure when no *segments* were written — but a healthy live playlist writes none between segments. So a two-rendition import with one permanently-404 variant and one healthy-but-quiet variant would end outright, which is precisely what `OnError::Warn` exists to prevent. It now keys on renditions that ingested without error, so a step fails only when nothing ingested at all.

That's the second bug in three commits in this one conversion, both in the same direction (treating a quiet pass as a dead one). Worth a careful look from a human at `run()` and `step()` in `rs/moq-hls/src/import.rs` if anyone has HLS context — the per-rendition tolerance rules there are subtler than the rest of this PR.

(Written by Claude Opus 5)

kixelated:
Pushed `438a7355` for the two newest Codex findings. Both were real.

**`status()` folded a transport race but not an address failover.** A hostname whose addresses all reach a relay that rejects the path returned `None`, so the reconnect loop burned its whole budget instead of failing fast. That was an inconsistency in my own code — I'd written the folding logic for `TransportRace` and not applied it to `Failover`. It now reports a status only when every raced address answered and none of the answers invites a retry; one unroutable address still leaves the set worth repeating. Three copies (quinn/noq/quiche), matching the existing pattern.

**`doc/bin/gstreamer.md` promised something the code doesn't do.** It said the sink reaches `failed` on "a URL no compiled-in backend can dial". With classification gone and `timeout = 0`, only an auth rejection or a settled CONNECT status ends the loop — a local config error retries forever. Worth noting this is the *second* stale-doc bug of exactly this shape (the first was `cluster.md`), and I introduced both in the same editing round, describing behaviour I intended rather than behaviour I'd checked. The page now says what happens and what to watch for: a sink stuck `disconnected` from the very first attempt is far more likely misconfigured than waiting out an outage.

Caveat unchanged: quiche's copy is uncompiled here (`boring-sys` needs cmake/nasm), so CI is the first thing to build it.

(Written by Claude Opus 5)

kixelated:
Reshaped per kixelated's direction: retries are now bounded by time, not by error type. The `Error::status()` / retryable-status-list classification is gone (~350 lines, the PR's only new public API), default budgets dropped from minutes to ~10s (reconnect 5m -> 10s, HLS import 300s -> 10s, cluster ceiling 300s -> 10s), and the CLAUDE.md essay is now one paragraph. The four bug fixes and the jitter are unchanged. Also rebased onto main, which the branch had fallen ~20 commits behind.

(Written by Claude Fable 5)

PR #2661, opened 2026-08-05
fix(obs): track the matching libmoq version

kixelated:
## Summary

- Remove the stale libmoq 0.3.6 pin from the OBS CMake presets.
- Derive the linked version from the in-tree `rs/libmoq` manifest for local builds, while requiring release builds to select an explicit published version.
- Run the Windows OBS release build on Windows 2022 because its CMake preset requires the Visual Studio 2022 generator.
- Add regression checks that reject static preset pins and release configurations without `MOQ_VERSION`, and dispatch those checks from the root `check`, `fix`, and `ci` workflows.

The root cause was that the 0.3.6 fallback survived the OBS plugin's move into the monorepo even though tagged release builds already supplied the matching libmoq version. Newer plugin releases then failed to complete because `windows-latest` moved beyond the Visual Studio 2022 generator requested by the preset. The OBS-specific regression guards also needed root task-runner integration to execute in pull request CI.

## Public API changes

None.

## Test plan

- `nix develop --command just obs fix`
- `nix develop --command just obs check`
- `nix develop --command just ci`
- Native macOS OBS plugin build against in-tree libmoq 0.5.3
- GitHub Check workflow passed

(Written by GPT-5)

### DIFF
--- a/.github/workflows/libmoq.yml
+++ b/.github/workflows/libmoq.yml
@@ -130,7 +130,8 @@ jobs:
           - target: aarch64-apple-darwin
             os: macos-latest
           - target: x86_64-pc-windows-msvc
-            os: windows-latest
+            # The plugin preset uses the Visual Studio 17 2022 generator.
+            os: windows-2022
 
     steps:
       # Shallow: this job only compiles, it never reads git history or tags.
--- a/cpp/obs/CMakeLists.txt
+++ b/cpp/obs/CMakeLists.txt
@@ -1,5 +1,27 @@
 cmake_minimum_required(VERSION 3.28)
 
+# Build the in-tree libmoq by default. Release-only builds clear MOQ_LOCAL and
+# must select the published archive explicitly, so this never silently drifts
+# to an old static version.
+set(MOQ_LOCAL "${CMAKE_CURRENT_SOURCE_DIR}/../.." CACHE PATH "Path to the moq repo (builds rs/libmoq in-tree)")
+set(MOQ_VERSION "" CACHE STRING "Published libmoq version (required without MOQ_LOCAL)")
+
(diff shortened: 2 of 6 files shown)


kixelated:
Reviewed and pushed a follow-up commit (9ab88ad00) addressing three findings:

1. **The regression guards never ran in CI.** `just obs check` wasn't invoked by root `just check`/`just ci` or any workflow (PR CI never touches cpp/obs), so the preset-pin grep and the MOQ_VERSION configure guard only fired when run by hand. They're now wired into `check`/`ci` gated on a cpp/obs diff (force-run on an empty diff, same pattern as the flake check), into `check-all`, and the `fix` counterparts for symmetry. cmake ships in the dev shell on all platforms, so the configure guard genuinely runs on ubuntu PR CI now. Side effect: clang-format/gersemi lints for cpp/obs also run on CI for the first time on cpp/obs-touching PRs.
2. **The check recipe hard-required cmake**, against its own "skips a tool silently if it isn't on $PATH" comment; a cmake-less machine failed with a confusing `command not found` log dump. The configure test is now guarded by `command -v cmake` like the other tools.
3. **The `MOQ_LOCAL AND EXISTS` predicate was duplicated** between the version derivation and the link/fetch choice and could drift apart. It's now computed once into `_moq_local_build`.

Verified: `nix develop --command just obs check` (release-mode fatal path exercised), the version-derivation block extracted verbatim into a `cmake -P` harness derives `0.5.3` with a valid `MOQ_LOCAL`, gersemi clean, `just --fmt --check` clean on both justfiles, and `nix develop --command just check` passes with the new obs gate firing.

One informational note left as-is: the `windows-2022` pin will break again when GitHub retires that image; at that point the fix moves to the preset's generator rather than the runner label. Since libmoq.yml only triggers on tags, that breakage will surface at release time, same visibility as today.

(written by Fable 5)

PR #2687, opened 2026-08-06
feat(moq-native): classify, pace, and publish accept(2) failures

kixelated:
Upstream half of [moq.pro#962](https://github.com/moq-dev/moq.pro/issues/962): the relay's `:443` TCP listener is the only one in this stack that performs a real `accept(2)`, it carries WebSocket MoQ / WHIP-WHEP signalling / the live HLS origin, and nothing was watching it.

## Root cause

Two separate swallows, both reachable today:

- **`Web`/`Internal`.** `axum_server`'s serve loop is `Err(_) => sleep(50ms)` ([server.rs](https://docs.rs/axum-server/0.8.0/src/axum_server/server.rs.html)). The error is not logged at any level, not counted, and never reaches the embedder. A flat 50ms is also wrong in both directions: too long for a per-connection error (ordinary traffic on a public port, paid for by whatever is queued behind it), too short for `EMFILE`. Note `axum`'s own `serve` gets the *shape* right for comparison: retry-at-once for a connection error, `error!` plus a 1s sleep otherwise.
- **`moq_native::{tcp,unix,websocket}::Listener`.** Each flattened the accept errno into the same `Error::Io` variant used for bind and connect, and each caller looped immediately. Under `EMFILE` the connection stays queued, so the listener is instantly readable and instantly failing: a hot loop burning the core the process needs in order to recover, narrated at `debug!` (WebSocket) or as a misleading `warn!` (tcp/unix). `Server::accept` also returned `None` on a lazy bind failure, indistinguishable from a shutdown.

An accept loop is a process-lifetime supervisor with nobody to return an error to, so it has to own the policy and publish what it is doing. Neither did.

## What this adds

**`moq_native::accept`.** `Failure::classify` splits the errno into the two families that want opposite handling; `Health` counts, logs, and paces:

- `Connection` (`ECONNABORTED`, `ECONNRESET`, `EPERM`, `EINTR`, ...) retries at once. The queue entry was consumed, so the listener already made progress and delaying only punishes the connections queued behind the dead one.
- `Exhausted` (`EMFILE`/`ENFILE`/`ENOBUFS`/`ENOMEM`) backs off with equal jitter toward a 5s cap and warns per attempt, per the Retries policy in `CLAUDE.md` (supervisors retry forever, cap in seconds, loudly broken rather than silently parked).
- `Unknown` backs off but claims nothing. Escalating on an unrecognized errno hands a remote peer a lever, which is what [moq.pro#927](https://github.com/moq-dev/moq.pro/issues/927) was reverted for.

One shared object rather than an inlined backoff per listener, deliberately: the classification *is* the pacing policy here, getting it backwards is not symmetric, and five listeners that each re-derive it will not agree for long.

**The relay takes the accept syscall back from `axum_server`** through its own listener abstraction (`AddrListener`/`Address`, `rs/moq-relay/src/listener.rs`), so the wrapper never yields an error upward and the flat sleep is unreachable. `axum_server` keeps everything past the accept: TLS, HTTP version negotiation, WebSocket upgrades, graceful shutdown.

**The counters ride the existing ops endpoint**, so a plain relay operator gets this without writing any code:

```
moq_relay_accept_failures_total{listener="web",class="connection"} 41
moq_relay_accept_failures_total{listener="web",class="exhausted"} 0
moq_relay_accept_failures_total{listener="web",class="unknown"} 0
moq_relay_accept_stalled_seconds{listener="web"} 0
```

Alert on the counter, not the gauge: a process with no descriptors left cannot serve the scrape either, so the episode is usually only readable once it is over, by which point the gauge reads a healthy zero. `Web::accept_health()` hands the same handle to an embedder with its own metrics surface, and `Internal::with_listener` registers anything the embedder listens on itself.

No QUIC listener registers one. They multiplex every session over a single UDP socket, never call `accept`, and a permanently-zero exhaustion counter reads as a watch that is passing when it can never fire ([moq.pro#942](https://github.com/moq-dev/moq.pro/issues/942)).

**`Server::listen`** (additive) binds the lazy `tcp`/`unix` listeners up front, so a bind failure is an error the caller handles and a later `None` from `accept` unambiguously means shutdown.

## Tests

- `accept::tests` cover the classification (including that a reset connection never reads as exhaustion, and that an unrecognized errno is `Unknown`), that a dead connection never pauses the listener, that exhaustion escalates and caps, and that a successful accept clears the stall while leaving the counters intact.
- `internal::tests::accept_metrics_list_every_listener_from_zero` pins that every registered listener appears at zero before it has ever failed. A counter that springs into existence on first failure makes the alert unwritable: `rate()` over a missing series is empty, not zero.
- `smoke::relay_https_terminates_tls` is new end-to-end coverage of the HTTPS listener. `Web`'s TLS path had none, and it is the part this PR routes differently, so a compile was not evidence that `MtlsAcceptor` still handshakes over the swapped listener.
- 395 tests pass; `clippy -D warnings` and `RUSTDOCFLAGS=-D warnings cargo doc` clean on both crates; workspace `--all-targets` compiles.

Only checked on macOS/aarch64. The Windows `WSAE*` classification arm compiles nowhere locally and has no test host.

## Not in scope

- **The measurement the issue asks for first.** I could not reach fleet data from here, so whether a node has ever actually hit `EMFILE` is still open. But the plumbing is justified independently: the current failure mode is not "dark and quiet", it is a CPU-burning spin that impedes the recovery, and the loops were out of compliance with the retry policy either way.
- `moq_rtmp::Server`'s accept loop backs off on *every* error including per-connection ones (`rs/moq-rtmp/src/server.rs`), the exact lever described above. It cannot use this without moq-rtmp depending on moq-native, which is the wrong layering for a gateway crate. Left alone.
- `moq-cli`'s `web.rs` still uses `axum_server::from_tcp`; it is a dev/cert-serving convenience, not a production listener.
- moq.pro's `common::retry::Accept` and `Ops::listener` now duplicate this upstream classification and can be collapsed onto it in a follow-up there. Note also that `Internal::serve(router)` already exists, so [moq.pro#944](https://github.com/moq-dev/moq.pro/issues/944)'s `serve_internal` carve-out is collapsible now.

Refs: moq-dev/moq.pro#962

(written by Opus 5)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4580,6 +4580,7 @@ dependencies = [
  "humantime",
  "humantime-serde",
  "jni 0.22.4",
+ "libc",
  "moq-net",
  "noq-proto",
  "notify",
--- a/doc/bin/relay/config.md
+++ b/doc/bin/relay/config.md
@@ -116,9 +116,9 @@ network.
 listen = "127.0.0.1:9101"
 ```
 
-It serves `/health`, Prometheus traffic counters at `/metrics`, and the local
-cluster topology view at `/nodes`. See [HTTP Endpoints](/bin/relay/http) for the
-response formats.
+It serves `/health`, Prometheus traffic and listener counters at `/metrics`, and
+the local cluster topology view at `/nodes`. See
+[HTTP Endpoints](/bin/relay/http) for the response formats.
 
 ### \[web.https]
 
--- a/doc/bin/relay/http.md
+++ b/doc/bin/relay/http.md
@@ -103,8 +103,40 @@ Bind it only to loopback or a trusted private network:
 listen = "127.0.0.1:9101"
 ```
 
-It mirrors `/health`, exposes Prometheus traffic counters at `/metrics`, and
-adds the cluster topology endpoint below.
+It mirrors `/health`, exposes Prometheus counters at `/metrics`, and adds the
+cluster topology endpoint below.
+
+### GET /metrics
+
+Prometheus text exposition of this node's own traffic counters (bytes, frames,
+groups, subscriptions, viewers, sessions), split by `tier` and `role`.
+
+Alongside them are the TCP listeners' accept-loop counters, which are how a node
+that has stopped accepting connections says so:
+
+```
+moq_relay_accept_failures_total{listener="web",class="connection"} 41
+moq_relay_accept_failures_total{listener="web",class="exhausted"} 0
+moq_relay_accept_failures_total{listener="web",class="unknown"} 0
+moq_relay_accept_stalled_seconds{listener="web"} 0
+```
+
+`connection` counts connections that died before the relay dequeued them (a peer
+reset, a firewall rule, a scanner): ordinary traffic on a public port, and never
+a fault. A non-zero `exhausted` is the one worth paging on, and means the process
+or the host ran out of a resource `accept` needs (file descriptors, kernel
+memory) while connections were queued; `moq_relay_accept_stalled_seconds` is how
+long that has been true, and only a successful accept clears it. `unknown` is an
+errno the relay does not classify: worth a look, not an alert.
+
+Alert on the `exhausted` counter rather than the gauge. A process out of
+descriptors often cannot serve this scrape either, so the gauge tends to read
+zero again by the time a scrape gets through, while the counter still shows the
+jump.
+
+Nothing here reports host CPU, memory, disk, or network; run a node exporter for
+those. The QUIC listener has no accept counters at all: it multiplexes every
+session over one UDP socket, so it never calls `accept` and cannot exhaust it.
 
 ### GET /nodes
 
--- a/doc/bin/relay/http.md
+++ b/doc/bin/relay/http.md
@@ -121,6 +121,13 @@ moq_relay_accept_failures_total{listener="web",class="unknown"} 0
 moq_relay_accept_stalled_seconds{listener="web"} 0
 ```
 
+One `listener` per socket that is actually configured: `web` for the HTTP/HTTPS
+pair, `internal` for this listener, and `tcp` / `unix` for the qmux stream
+listeners. A listener you have not configured is absent rather than zero, so a
+stream-only relay publishes no `web` series at all. That is deliberate: a
+permanent zero for a socket nobody opened reads as a watch that is passing, when
+there is nothing there to watch.
+
 `connection` counts connections that died before the relay dequeued them (a peer
 reset, a firewall rule, a scanner): ordinary traffic on a public port, and never
 a fault. A non-zero `exhausted` is the one worth paging on, and means the process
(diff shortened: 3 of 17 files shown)


kixelated:
Review round addressed. What changed, and what I deliberately didn't:

**Fixed**

- **Stream listeners were unreported** (Codex, `relay.rs`). The relay registered only `web`, so a stream-only node published a permanent zero for a socket it never opened while its `tcp`/`unix` listeners went unwatched. `moq_native::Server::accept_health` now hands out every listener it owns, created with the config rather than the socket (they bind lazily; a metrics endpoint is assembled at startup). `Web::accept_health` returns `None` when neither HTTP nor HTTPS is configured, and `Internal` registers itself only when it will actually run.
- **`ENONET`** (Codex, `accept.rs`). It is part of the pending-network-error set Linux surfaces through `accept`, and was missing, so a per-connection failure would have cost a backoff. Gated to Linux/Android, where the constant exists at all.
- **`WSAENETDOWN`** (found while fixing the above). The Windows table had it as per-connection and retried it at once, but Microsoft defines it as failure of the network subsystem: listener-wide, and exactly the spin this PR removes. The Windows list is now the two genuinely per-connection codes; everything else falls through to `Unknown`, which paces.
- **`Failure::ALL` was `[Failure; 3]`** (Codex, `accept.rs`). An array's length is part of its type, so that made adding a class a breaking change and defeated the `#[non_exhaustive]` on the enum. Now a slice.
- **Duplicate `listener` labels** (CodeRabbit + Codex, `internal.rs`). Two rows with the same label set is a malformed exposition and a scraper may reject the whole payload, so one mis-registered listener could blind the node's traffic counters too. `with_listeners` now refuses a name already in use and warns.
- **`ensure_started` was not all-or-nothing** (CodeRabbit, `server.rs`). It drained its bind list up front, so a partial failure left the list empty and a second `listen` returned `Ok` while nothing was bound and `accept` parked forever. Regression test added.
- **The `Option` on `Listener::accept`** (CodeRabbit, `tcp.rs`). Nothing ends the accept loop, so `None` is unreachable and the docs said otherwise. Docs corrected; dropping the `Option` is a published-signature change, so it stays for now.
- markdownlint MD040 on the metrics fence.

**Not changed, with reasons**

- **Splitting HTTP/HTTPS health** (CodeRabbit + Codex). Deliberate, and confirmed with the maintainer. Every class that escalates the backoff is process- or host-wide (`EMFILE`/`ENFILE`/`ENOBUFS`/`ENOMEM`), so a successful HTTP accept is real evidence that HTTPS is not out of descriptors either. The narrow case the reviews describe needs the fd count oscillating at the limit *and* both listeners configured, and its cost is a retry at the 100ms floor rather than escalating to 5s. That is mild busy-ness, not the hot spin this PR exists to remove.
- **Bounding `accept_stream` with a retry budget** (CodeRabbit, `listener.rs`, citing the coding guidelines). Those guidelines say the opposite for this case: "The only unbounded loops are process-lifetime supervisors with nobody to return an error to (cluster peers, device reopen, **accept loops**); they retry forever but cap the delay at seconds and warn per attempt." Returning the error has nowhere to go, since `axum_server` would only sleep and drop it. "The process will not learn it cannot accept connections" is precisely what the counters on `/metrics` are for.
- **An IP SAN in the HTTPS smoke test** (CodeRabbit, `smoke.rs`). The test dials `https://localhost:<port>` and the cert carries a `localhost` SAN, so it validates as written.

(written by Opus 5)

PR #2692, opened 2026-08-06
refactor(moq-mux)!: split the fMP4 init segment out of Fragment

zshenker:
## Summary

`fmp4::Fragment` described two different emits with one struct. The init segment carried `init: true`, `independent: false`, and `duration: 0.0`; every media fragment carried `init: false` and meant the other two. Three of its four fields were sentinels on one of the two shapes it covered, and a consumer had to read a `bool` to know which fields were real.

`Export::next_chunk` now returns a `Chunk`: `Init` holds the bytes and nothing else, while `Fragment` holds a `Fragment` with no dead field. `Export::next` is unchanged for consumers that only want bytes.

Review feedback also moved `Fragment::duration` from untyped `f64` seconds to `std::time::Duration`, and added consuming `Chunk::init` and `Chunk::fragment` accessors. `rs/CLAUDE.md` now records the repository convention to use typed time values and convert numeric serialized forms at the boundary, using `serde_with` where possible.

`Chunk` is deliberately not `#[non_exhaustive]`. The set is closed by the container format, which emits an init segment and media segments and nothing else, so a wildcard arm would be permanent noise on every match.

## Public API changes

| item | change |
|---|---|
| `fmp4::Chunk` | new enum: `Init(Bytes)`, `Fragment(Fragment)` |
| `Chunk::data`, `Chunk::into_data` | new byte accessors |
| `Chunk::init`, `Chunk::fragment` | new consuming variant accessors |
| `Fragment::init` | removed |
| `Fragment::duration` | changed from `f64` seconds to `std::time::Duration` |
| `Export::next_fragment` | renamed to `next_chunk`, now returns `Option<Chunk>` |
| `Export::poll_next_fragment` | renamed to `poll_next_chunk`, now returns `Option<Chunk>` |
| `Export::next` / `poll_next` | unchanged |

This targets `dev` because it breaks published `moq-mux` APIs.

## Test plan

Run on the final head after merging the latest `origin/dev`:

- `nix develop --command just rs test -p moq-mux`: 482 passed, 0 failed

`just test smoke-full` was not run because this changes no wire format, `moq-ffi` API, or gateway behavior.

## Cross-Package Sync

No table rows apply. `moq-mux` has no cross-package row, no wire format or CLI surface changed, and `doc/lib/rs/crate/moq-mux.md` describes the crate's scope rather than this API.

(Written by GPT-5)


kixelated:
@codex review

rs/moq-mux/src/container/fmp4/export.rs
```diff
@@ -51,22 +51,50 @@ pub struct Export<S: Stream> {
...
-	/// A media fragment that begins at a sync sample, so it can start a segment.
+	/// A fragment that begins at a sync sample, so it can start a segment.
 	/// Video fragments are independent only at a GOP boundary (keyframe); audio
-	/// fragments are always independent. Always false for the init segment.
+	/// fragments are always independent.
 	pub independent: bool,
 
-	/// Presentation duration of the fragment in seconds (0 for the init segment).
+	/// Presentation duration of the fragment in seconds.
 	pub duration: f64,
```
kixelated: Why not time::Duration?

rs/moq-mux/src/container/fmp4/export_test.rs
```diff
@@ -585,13 +602,31 @@ fn synthesize_flac_trak() {
...
+	tokio::time::timeout(std::time::Duration::from_millis(1), exporter.next_chunk())
 		.await
 		.expect("waited for a successor frame")
 		.expect("exporter failed")
-		.expect("expected a fragment")
+		.expect("expected a chunk")
+}
+
+/// The next chunk, which must be a media fragment.
+async fn fragment_now(
```
kixelated: Maybe add a `fragment() -> Option<Fragment>` method on Chunk instead, so we can call:

`chunk_now(exporter).fragment().expect("got an init segment")`

rs/moq-mux/src/container/fmp4/export.rs
kixelated: No reason not to. The computation already ran in `std::time::Duration` end to end and only flattened to `f64` on the way out, so `Fragment::duration` is now a `Duration` and a playlist writer calls `as_secs_f64()` itself. Done in e6d2d7a5.

(Written by Fable 5)

rs/moq-mux/src/container/fmp4/export_test.rs
kixelated: Done in e6d2d7a5: `Chunk::init(self) -> Option<Bytes>` and `Chunk::fragment(self) -> Option<Fragment>`, and the panicking test helpers are gone in favor of `chunk_now(..).await.fragment().expect("a media fragment")`.

(Written by Fable 5)

kixelated (approved):
Reviewed the final head after syncing with `dev` and adding the typed-time guidance. The chunk API cleanly separates init data from media fragment metadata, the `Duration` change preserves exact internal values, and the two prior review requests are implemented. `just check` and all 482 `moq-mux` nextest tests passed. No blocking findings.

(Written by GPT-5)

kixelated (approved):
Reviewed the final head after merging the latest dev. The upstream commit is unrelated to the fMP4 API change and merged without conflicts. The full Rust CI test stage passed all 2,955 tests locally, the scoped repository checks passed on the prior equivalent head, and the final diff passes git diff --check. No blocking findings.\n\n(Written by GPT-5)

PR #2693, opened 2026-08-06
feat(moq-mux): timescale control, and two duration fixes

zshenker:
## Summary

The three parts of #2623 that aren't about the per-frame fragmenting API, split out so they don't wait on that discussion. #2679 was the same move for the init-segment fixes. What's left in #2623 after this is only `fragments` / `fragment_at`, which is the open question in #2680.

**1. `Muxer::timescale` / `Muxer::with_timescale`.** A Legacy/Loc rendition's timescale is derived from the catalog (framerate for video, sample rate for audio). Fine as a default, but a consumer whose downstream timeline is fixed (an HLS/DASH origin working in 90 kHz ticks) could neither observe the mismatch nor correct it, because the field was private with no accessor. @kixelated already signed off on this shape in #2680.

`with_timescale` rejects a `Cmaf` rendition, whose init passes through from the catalog at its own scale and so can't be moved without desynchronising the init from the fragments, and rejects a scale the 32-bit `mdhd` field can't hold. Both error at the call that is wrong rather than later at `init()`.

**2. The fallback frame duration was a tick short on most framerates.** A frame period is the rational `timescale / rate`, and `default_video_timescale` picks `framerate * 1000` precisely so that division comes out whole. Carrying the period as a `Duration` threw that away: 1/30 s is 33333333 ns, which is 999.99999 ticks at 30 kHz and truncates to 999. Every sample with no successor to be timed by was a tick short, which on the `Export` path is the trailing sample of every fragment.

| framerate | was | now |
|---|---|---|
| 23.976, 29.97, 30, 48, 59.94, 90, 120 | 999 | 1000 |
| 24, 25, 50, 60 | 1000 | 1000 |

The ones that already worked did so on which way the nanosecond conversion happened to round, not by construction. Rounding at the conversion would paper over it; the lossy hop is the bug, since the fallback was converted to the frames' scale, which is usually microseconds, where 1/30 s isn't representable either. It's now a `Timestamp` in the track's own ticks, computed by dividing the timescale by the rate and left at that scale. A mixed-scale slice encodes identically, because every duration is rescaled independently on its way into the `trun`.

`with_timescale` recomputes it, since the fallback moves with the scale. At a foreign scale it takes the nearest whole tick, so 29.97 at 90 kHz gives the broadcast-standard 3003.

**3. A sample duration that can't round-trip the `trun`.** The `trun` sample-duration is a 32-bit field and the encoder narrowed into it with a silent `as u32`, while a `Timestamp` spans the whole QUIC varint range. A duration past `u32::MAX` reached the `trun` wrapped while the caller's timeline advanced by the full value, so the fragment claimed less time than wherever the next one starts, silently. Reachable without anything exotic: `u32::MAX` is an accepted timescale, which makes one second `u32::MAX` ticks, so a two-second sample overflows. Same class as the `mdhd` narrowing rejected in #2679, and it gets the same treatment: refuse at the truncation site. Being in the encoder, it covers `Muxer::fragment` and the `Export` path at once.

## Public API changes

| item | kind |
|---|---|
| `Muxer::timescale(&self) -> moq_net::Timescale` | new |
| `Muxer::with_timescale(self, Timescale) -> Result<Self>` | new |
| `fmp4::Error::TimescaleOverride` | new variant |
| `fmp4::Error::SampleDurationTooLarge` | new variant |

All additive; `fmp4::Error` is already `#[non_exhaustive]`, so the variants don't break an external `match`. `Muxer::fragment` is unchanged apart from the duration fixes above, which change what it emits for a duration-less trailing sample (999 to 1000 ticks at 30 fps).

`pub(crate)`/private only: `trun_duration`, `fallback_duration`, and `Muxer`'s `timescale` field becoming a `moq_net::Timescale` rather than a `u64`, so the conversion happens once at construction instead of on every `fragment` call.

## Test plan

- `cargo test -p moq-mux` — 457 passed, 0 failed.
- `cargo check --workspace --all-targets` — clean.
- `cargo clippy -p moq-mux -p moq-hls --all-targets`, `cargo fmt --all --check`, `RUSTDOCFLAGS="-D warnings" cargo doc -p moq-mux` — clean.
- Regression tests: `fallback_duration_is_exact_at_the_derived_timescale` (the table above), `fallback_duration_rounds_at_a_foreign_timescale`, `with_timescale_recomputes_the_fallback_frame_duration`, `encode_fragment_rejects_a_duration_too_large_for_trun`, plus the two `with_timescale` tests.

Not run: `just check` (no `nix` available here, so the underlying `cargo` steps were run directly) and `just test smoke-full` (no wire, `moq-ffi`, or gateway change).

## Cross-Package Sync

No rows apply. `moq-mux` has no row in the table, and no wire format, CLI surface, or draft changed.

(Written by Opus 5)


kixelated:
@codex review

PR #2713, opened 2026-08-07
test(net): cover multi-track failover and a partial-list standby

t0ms:
## Summary

Every takeover, linger and reselect test in `rs/moq-net/src/model` subscribes to a single track, so nothing pins that a reselect is decided and served *per track*. A real broadcast is multi-track (one MPEG-TS contribution feed carries video, audio and data on a single broadcast), and a takeover that re-splices one track while another silently stalls is a partial recovery that a single-track test cannot distinguish from a whole one.

Two tests, no production changes:

- **`test_route_failover_restores_every_track`**: three tracks on one broadcast, the active source dies mid-stream, and every subscription must be re-dispatched to the standby and resume at *its own* boundary. The tracks carry deliberately unequal group counts (3/1/2) so the three boundaries differ, which is what makes a boundary computed per broadcast rather than per track visible.
- **`test_standby_with_a_partial_track_list_splits_per_track`**: a standby that carries only some of the broadcast's tracks splices the ones it has and leaves the rest on the incumbent. This is the divergent-layout case for a 1+1 pair: two sources meant to be interchangeable that do not agree on the track list must degrade to a per-track split rather than taking the whole broadcast with them. It complements `test_standby_missing_track_keeps_incumbent`, which covers a standby that has no track at all.

Both pass on `main` as written, so this is coverage rather than a fix.

Also adds an `accept_tracks` helper next to `accept_track`: the existing helper asserts an exact track name, and dispatch order across concurrently requested tracks is not guaranteed.

## Why this is worth a test rather than an assumption

To check the tests are not just restating what the suite already covers, I injected a fault in `broadcast::poll_requested_track` that silently drops every queued track request except one, which is precisely the partial-recovery shape above. **705 of the 707 existing `moq-net` tests still pass under that fault. The only two that fail are the two added here.**

The motivation is a 1+1 active/active MPEG-TS contribution path, where a subscriber that keeps receiving video while audio quietly stops is worse than a clean failure, because it survives monitoring.

## Public API changes

None. Test-only, entirely inside `#[cfg(test)] mod tests`; `accept_tracks` is private to that module.

## Cross-package sync

The `rs/moq-net` row (`js/net`, `doc/concept`, `drafts/`) does not apply: no wire, API, or behavior change, so there is nothing for `js/net` to mirror or for a draft to specify.

## Test plan

- `cargo nextest run -p moq-net`: 707 passed, 0 failed.
- `cargo nextest run --workspace --exclude moq-gst`: 2637 passed, 0 failed. `moq-gst` is excluded because its GStreamer system libraries are not installed on this host, not because of this change.
- `just check` clean: `cargo check`, `clippy -D warnings`, `fmt --check`, `doc`, `shear`, `sort`.
- Both new tests run in 0.013s each and are not flagged SLOW. Repeated 150 times with zero failures, each run drawing a fresh `Origin::random()`, since these paths sit near the handover-key comparison that `origin_keyed` exists to make deterministic.
- Fault injection as described above, reverted before commit.
- Not run: `just rs macos` and `just rs windows`, which cover `moq-video`/`moq-audio` platform backends untouched here.

**On the red CI check:** `just ci` fails in `cargo deny` on `error[yanked]: wide 1.6.0`, reached via `openh264` -> `moq-video`. That is inherited from `main`, not introduced here: this branch changes one test file and leaves `Cargo.lock` byte-identical to `main`, where `wide 1.6.0` is already pinned. The crate appears to have been yanked upstream today, so this PR is simply the first run to trip it. It needs a `cargo update -p wide` on `main` rather than anything in this diff.

(Written by Opus 5)

### DIFF
--- a/rs/moq-net/src/model/origin.rs
+++ b/rs/moq-net/src/model/origin.rs
@@ -3371,6 +3371,22 @@ mod tests {
 		request.accept(None)
 	}
 
+	/// Serve `count` requested tracks, keyed by name. Dispatch order across
+	/// tracks is not guaranteed, which [`accept_track`]'s exact-name assert
+	/// cannot express.
+	async fn accept_tracks(dynamic: &mut broadcast::Dynamic, count: usize) -> HashMap<String, track::Producer> {
+		let mut accepted = HashMap::new();
+		for _ in 0..count {
+			let request = tokio::time::timeout(std::time::Duration::from_secs(1), dynamic.requested_track())
+				.await
+				.expect("timed out waiting for a track request")
+				.expect("source closed");
+			let name = request.name().to_string();
+			accepted.insert(name, request.accept(None));
+		}
+		accepted
+	}
+
 	/// Tagging both origin handles with one context attributes the full model path:
 	/// ingress writes on the subscriber side, egress reads on the publisher side,
 	/// each counter landing exactly once (the model-layer silent-zero guard).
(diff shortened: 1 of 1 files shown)


kixelated:
@codex review

PR #2720, opened 2026-08-07
fix(net): invert stream priority to match transport send-order semantics

jgh-twitch:
## Summary

Lite group streams were transmitted in exactly inverted priority order under
contention: after a bandwidth drop, the relay drained its stalest backlog before
any fresh media. Separately, the IETF subscriber priority was passed between the
wire and the model unconverted, so priorities interoperated backwards with
spec-compliant peers.

**Root cause:** the codebase has exactly one lower-is-more-urgent value, the lite
priority queue's *rank* (0 = newest group of the highest-priority track, 255 =
overflow). Everything else is higher-first: `Subscription::priority` ("higher
values preempt lower ones"), the lite wire, `web_transport_trait::SendStream::set_priority`
(0.3.6+), W3C `sendOrder`, and quinn's scheduler. The lite publisher handed the
raw rank straight to the transport, which reads it as a send order, so the least
urgent group won the scheduler. The emitter was written against the pre-0.3.6
trait doc ("lower values sent first") and never updated when the contract
flipped.

The IETF wire is the other exception: `subscriber_priority` is lower-first
(draft-ietf-moq-transport section 7.1), but both the subscriber (encode) and the
publisher (decode) moved it to and from `Subscription::priority` unchanged.

**Fix:** convert at each boundary, and only there.

- `PriorityHandle::send_order()` returns `u8::MAX - rank`, so the conversion
  lives where the rank is produced. `Writer::set_priority` stays a documented
  passthrough of the transport's send order.
- `ietf::priority::{from_wire, to_wire}` normalize `subscriber_priority` in the
  IETF subscriber and publisher, so the model keeps its higher-first semantics
  everywhere inside moq-net. `js/net` gets the matching `ietf/priority.ts` at the
  same two call sites.

Also folded in: both group completion paths (lite `serve_group`, ietf
`run_group`) ended with `finish()` + `closed().await` and then dropped a
still-owned `Writer`, whose `Drop` fallback reset the stream with a spurious
`Cancel` on a group the peer had already acknowledged. They now consume the
writer via `close()`.

## Tested

- `PriorityHandle::send_order` unit test: rank 0 maps to send order 255.
- Lite group harness: a served group at rank 0 reaches the transport as 255, and
  a cleanly completed group records no resets.
- IETF group harness: model priority 200 reaches the transport unchanged (fails
  at 55 against a `Writer`-level inversion).
- `ietf::priority` and `js/net` `ietf/priority.ts` round-trip and direction tests.
- `cargo test -p moq-net`: 711 passed. `bun test` in `js/net`: 340 passed.
  `just check`: clean.

## Cross-package sync

- `js/net`: converted at the same two IETF call sites. Lite bidi Fetch streams
  already pass the higher-first priority into `sendOrder` correctly; lite group
  unidirectional streams still set no `sendOrder` at all, a pre-existing gap left
  out of scope.
- `drafts/`: no update. `subscriber_priority` is defined by
  draft-ietf-moq-transport, which is not carried in-tree, and this change makes
  us conform to it rather than altering it. `draft-lcurley-moq-lite` is
  unaffected: the lite wire was and remains higher-first.

## Note

The model's default priority is 0, which is the *lowest* higher-first value, so a
subscription that never sets a priority now sends IETF `subscriber_priority` 255
(least urgent) instead of 0 (most urgent). Relative ordering among our own tracks
is unchanged; only a third-party IETF relay arbitrating between us and other
subscribers sees a difference.

(written by Opus 5)

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -181,7 +181,7 @@
     },
     "js/net": {
       "name": "@moq/net",
-      "version": "0.2.5",
+      "version": "0.2.6",
       "dependencies": {
         "@moq/qmux": "^0.3.2",
         "@moq/signals": "workspace:*",
--- a/js/net/package.json
+++ b/js/net/package.json
@@ -1,7 +1,7 @@
 {
 	"name": "@moq/net",
 	"type": "module",
-	"version": "0.2.5",
+	"version": "0.2.6",
 	"description": "The networking layer for Media over QUIC: real-time pub/sub with built-in caching, fan-out, and prioritization.",
 	"license": "(MIT OR Apache-2.0)",
 	"repository": "github:moq-dev/moq",
--- a/js/net/src/ietf/priority.test.ts
+++ b/js/net/src/ietf/priority.test.ts
@@ -1,0 +1,19 @@
+import { expect, test } from "bun:test";
+import { fromWire, toWire } from "./priority.ts";
+
+test("IETF subscriber priority is lower first", () => {
+	expect(fromWire(0)).toBe(0xff);
+	expect(fromWire(0xff)).toBe(0);
+	expect(toWire(0xff)).toBe(0);
+	expect(toWire(0)).toBe(0xff);
+
+	for (let priority = 0; priority < 0xff; priority++) {
+		expect(fromWire(priority)).toBeGreaterThan(fromWire(priority + 1));
+	}
+});
+
+test("subscriber priority round trips", () => {
+	for (let priority = 0; priority <= 0xff; priority++) {
+		expect(fromWire(toWire(priority))).toBe(priority);
+	}
+});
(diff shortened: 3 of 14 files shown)


kixelated:
@codex review

kixelated:
This priority stuff is pretty annoying. AFAIK

**Larger == Higher priority**
- quinn/noq
- moq-lite
- WebTransport (w3c)

**Smaller == Higher priority**
- quiche
- moq-transport


I think the `moq-net` and `web-transport` API should stick with larger == higher priority and invert if needed. Otherwise you get confusing documentation like this taken straight from the quiche docs:

> (streams with lower priority are sent first).

Ironically, I came up with the name `sendOrder` for WebTransport to help avoid the confusion where 0 == highest priority. But based on the confusing wording in the specification they got it backwards.

PR #2723, opened 2026-08-08
(DON'T MERGE) refactor(net): replace routes instead of splicing

kixelated:
## Summary

- Replace the synthetic route-splicing broadcast with direct per-consumer source selection.
- When a different shortest or tied source wins, emit an unannounce followed by an announce. Existing subscriptions remain bound to their original source.
- Remove the resume and partial-group handover machinery, and update the Rust/JS adapters, relay behavior, CLI guidance, documentation, and normative drafts.

The root cause was that route selection mutated a synthetic broadcast underneath live subscribers. Preserving subscription state across independently owned sources required frame-precise resume bookkeeping and several handover paths, which repeatedly produced lifecycle and routing bugs. Source-bound subscriptions make replacement explicit and leave resubscription policy with the consumer.

## Public API changes

- No exported items were renamed, removed, or signature-changed.
- Behavioral change: a newly selected source is observable as an unannounce/announce replacement, including for a shorter or tied route. Consumers that want continuity must subscribe to the replacement announcement.

## Test plan

- `nix develop --command cargo nextest run -p moq-net` (712 passed)
- Focused `moq-relay` GOAWAY replacement tests
- `nix develop --command just drafts check`
- `nix develop --command just test smoke-full` (21 publisher/subscriber combinations passed)
- `git diff --check`

(Written by GPT-5)


kixelated:
I made this as an emergency break glass PR.

PR #2738, opened 2026-08-11
feat(bindings): expose incoming request path and query

shermerL:
## Summary

- Expose incoming request routing through the Rust FFI and the Python, Swift, Kotlin, and Go bindings.
- Normalize the root or missing path to an empty string and keep `path` query-free across every transport.
- Expose the encoded query separately, with documentation warning that it may contain credentials.
- Preserve the complete WebSocket request URL before QMux consumes the upgrade request, including authentication queries.
- Preserve legacy WebSocket clients that omit `Sec-WebSocket-Protocol`, while rejecting nonempty unsupported offers.
- Retain accepted sessions in the server examples so connections are not closed immediately.
- Defer Python and Swift wrapper publication until a successful matching FFI release makes the generated API available.

## Root cause

`moq-native::Request::path()` could read either the MoQ SETUP request target or a retained request URL. Raw transports placed both path and query in SETUP, while URL transports returned only `Url::path()`. This made the binding API transport-dependent. WebSocket additionally discarded its HTTP upgrade request before constructing the MoQ request, losing both the route and `?jwt=` authentication parameter.

The native request now exposes a normalized, query-free `path()` and a separate encoded `query()` for every transport. The WebSocket listener reconstructs and retains the original request URL while preserving subprotocol negotiation. Relay authentication consumes the separated path and query without logging the credential-bearing query.

The custom WebSocket upgrade initially required a selected subprotocol, unlike the previous QMux server and relay listener. Legacy clients that omit the header negotiate their versions in SETUP, so the listener now upgrades them without adding a response protocol. It still rejects a nonempty offer with no supported match.

The wrapper examples also discarded accepted session handles, which closed live connections as soon as the loop iteration ended. They now retain each session until its closed event completes.

The Python and Swift wrapper releases could otherwise publish before release-plz publishes the new `moq-ffi` patch. Their workflows now detect unreleased FFI tree changes or an absent FFI tag, defer successfully, and resume after the corresponding successful FFI release workflow. The Python wrapper requires `moq-ffi ~= 0.3.10`, the first release containing both request accessors.

## Cross-package sync

- Python and Swift receive patch bumps for their additive wrapper APIs.
- Go keeps its human-owned `0.5` version line; release automation derives the additive patch version.
- Kotlin consumes the generated FFI method directly, so no independent wrapper source version bump is needed.
- `libmoq` and the C docs are unchanged because the C ABI does not expose server request handles.

## Testing

- `nix develop --command just py check`
- `nix develop --command just swift check` (10 tests)
- `nix develop --command just go check` (including race tests)
- `nix develop --command just kt check`
- Focused relay auth, FFI request, QUIC, Iroh, WebTransport, WebSocket fallback, and legacy no-subprotocol nextest cases
- `nix develop --command just test smoke-full` (all 21 combinations)

(Written by GPT-5.6)


kixelated:
@codex review

Please review the current head after the WebSocket path preservation and cross-binding documentation fixes.

(Written by GPT-5.6)

kixelated:
@codex review

Please review current head `8bd7e870f`, including WebSocket query preservation and the Python/Swift FFI release gates.

(Written by GPT-5.6)

kixelated:
wtf is this feedback loop

PR #2750, opened 2026-08-11
refactor(moq-native)!: name the endpoint roles connect and listen

### LINKED ISSUE #2696: Re-evaluate the client/server split in CLI arguments
## Problem

The CLI splits almost every transport knob into `--client-*` and `--server-*` variants, but a QUIC endpoint does both. The split is an artifact of `ClientConfig` and `ServerConfig` being separate Rust types, not something a user asked for, and it costs us a duplicated flag for every knob that isn't actually role-specific.

Eleven knobs currently exist in both spellings, with identical meaning:

| Duplicated | |
|---|---|
| `--client-bind` / `--server-bind` | the socket |
| `--client-backend` / `--server-backend` | quinn / quiche / noq |
| `--client-version` / `--server-version` | negotiated MoQ versions |
| `--client-tls-root` / `--server-tls-root` | trust roots |
| `--client-quic-*` / `--server-quic-*` | `congestion-control`, `gso`, `idle-timeout`, `keep-alive`, `max-streams`, `mtu-discovery`, `qlog` (7 pairs) |

Genuinely role-specific ones are the minority: dialing (`--client-connect`, `--client-connect-timeout`, `--client-failover-delay`, `--client-reconnect`, client cert/verification) and listening (`--server-tcp*`, `--server-unix*`, `--server-quic-lb-*`).

## Why it's coming up now

`moq --cluster-lan` needs an endpoint that dials peers *and* accepts from them. With the current split there's no way to express that, so it reaches into `--server-bind`, fills it in with an ephemeral port and a generated certificate when unset, and separately clones the `ClientConfig` per peer to dial out. That works, but it means a user who typed one flag silently gets a wildcard listener they never asked for, and the mesh's dial settings and accept settings are configured through two unrelated prefixes.

That surfaced in review as a security concern. It isn't one (an unauthenticated `--cluster-lan` is the same exposure as a bare `--server-bind`, which is documented), but it is a real sign the flags no longer describe the thing being configured.

## Direction

Roughly, in rough order of appetite:

- **`--bind` applies to both.** One endpoint, one socket flag. Keep a way to bind separate client/server sockets if anyone needs it, but stop making that the default shape.
- **`--connect` loses the `--client` prefix.** It's the URL to dial; nothing about it is client-role-specific in a way the prefix earns.
- **Collapse the shared knobs** (`--quic-*`, `--version`, `--backend`, `--tls-root`) into one spelling, with role-specific overrides only where a real asymmetry exists.
- **Keep prefixes only where the role genuinely differs**, e.g. accepting on a Unix socket, or presenting a client certificate.

## `--cluster-*` is tangled too

Separate but related. There are now four ways to find a peer, and the names don't tell you they're alternatives:

- `--cluster-connect` (static peer list)
- `--cluster-connect-api` (dynamic peer list over HTTP/file)
- `--cluster-mesh` (gossip, needs `--cluster-node`)
- `--cluster-lan` (mDNS, needs `--cluster-node` and `--cluster-lan-secret`)

plus `--cluster-node` (this relay's own URL, meaning something different again), `--cluster-token`, `--cluster-id`, `--cluster-tier`, `--cluster-linger`. Several combinations are invalid and only fail at startup with a hand-written `ensure!`. Worth reshaping alongside the client/server work, since "how do I find peers" and "how do I bind an endpoint" are the same question from two directions.

## Constraints

- Every rename is user-facing, so the old spellings stay as hidden aliases per the deprecation rules in `CLAUDE.md`: no `--help` entry, no "deprecated, use X" note, `/doc` examples updated to the new names only.
- The flags are also TOML keys, so a rename is a config-file migration, not just a CLI one.
- `moq-relay`, `moq-cli`, and `moq-bench` all flatten these configs; a change lands in all three at once.
- Sample invocations are scattered across `doc/bin/`, `doc/lib/`, `doc/setup/`, `doc/concept/`, and the `demo/` justfiles. `CLAUDE.md` already calls out grepping the binary name repo-wide and reconciling every hit against `--help`.

(written by Opus 5)

kixelated:
Closes #2696.

A QUIC endpoint dials and accepts, and the config surface spelled every transport knob twice: `--client-*` and `--server-*`, backed by duplicated `quic::Client`/`quic::Server` and `tls::Client`/`tls::Server` type definitions. This keeps the two roles apart, because most knobs really do belong to one of them, and fixes the naming instead: role modules with short names, and the genuinely shared tuning spelled once.

## The shape

```
moq_native::connect::Config   --connect-*    the dial side
moq_native::listen::Config    --listen-*     the accept side
moq_native::quic::Config      --quic-*       transport tuning, shared by both
```

An endpoint that does both flattens all three and hands the tuning to each role: `connect.init(quic)` / `listen.init(quic)`.

| | flag | note |
|---|---|---|
| `connect.url` | `--connect` | kept short; it's the flag everyone types |
| `connect.bind` | `--connect-bind` | the dial socket |
| `connect.timeout` | `--connect-timeout` | |
| `connect.race` | `--connect-race` | was `failover_delay` |
| `connect.once` | `--connect-once` | reconnecting is the default; the flag names the exception |
| `connect.tls` | `--connect-tls-*` | roots, fingerprints, `insecure`, and the mTLS identity as plain `cert`/`key` |
| `connect.websocket` | `--connect-websocket-*` | the fallback's enable and head-start knobs |
| `listen.bind` | `--listen` | |
| `listen.tls` | `--listen-tls-*` | served `cert`/`key`/`generate`, plus `root` for mTLS CAs |
| `listen.tcp`, `listen.unix` | `--listen-tcp-*`, `--listen-unix-*` | the qmux stream listeners and the peer-credential allowlist |
| `listen.preferred_v4/v6`, `listen.lb_id/lb_nonce` | `--listen-*` | accept-only, so they moved off `quic` |
| `quic.*` | `--quic-*` | one definition, one spelling |

A flag is its dotted field path, and `--connect` (for `connect.url`) is the one exception, because it's the flag everyone types. That rule is what pulled in the last few stragglers: the stream listeners still said `--server-tcp-bind` / `--server-unix-*`, the served certificate was bare `--tls-cert` next to a prefixed `--listen-tls-root`, and the WebSocket fallback was `--websocket-*` while its env var had already moved to `MOQ_CONNECT_WEBSOCKET_*`. Every one of those released spellings still parses, and still reads its original env var.

`backend` and `version` stay per-role: a `quiche` acceptor with a `quinn` dialer is a real configuration (and `--client-backend`/`--server-backend` already allowed it), and a per-role `version` is what lets a relay pin its mesh wire version without narrowing what it accepts from external publishers.

## Why `tls` splits and `quic` doesn't

`tls` is two different things wearing one name: `cert`/`key`/`generate` are the served identity, while `root`/`fingerprint`/`insecure`/`host_name` are how you verify a peer you dial. Splitting them means the mTLS identity drops its `client_` prefix (the module supplies the role), and the two `root`s stop being one field: `connect.tls.root` verifies servers you dial, `listen.tls.root` authenticates mTLS clients.

`quic` is the opposite: `max_streams` means the same thing whichever way the connection was opened. Sharing it is also forced, not just preferred, since clap can't flatten one struct twice in a parser, so per-role tuning would mean maintaining duplicate definitions that drift.

## Compatibility

Every released flag **and every released environment variable** still configures the same thing. That is checked rather than asserted: `rs/moq-relay/tests/released_cli.rs` freezes the published `moq-relay` surface (98 flag/env pairs, taken from the parser itself at the last release) and fails if a spelling stops parsing or an env var stops being read.

Writing that test is what turned up the hole. The first pass renamed flags with clap `alias`, which carries the flag name and **not** the env var, so 20 released variables parsed as before on the command line while being silently ignored in the environment: every `MOQ_CLIENT_TLS_*`, `MOQ_CLIENT_WEBSOCKET_*`, `MOQ_SERVER_TLS_*`, and the `MOQ_SERVER_BIND`/`BACKEND`/`VERSION`/`PREFERRED_V4`/`V6`/`QUIC_LB_*` set. This repo's own deployments were in that blast radius: `nix/modules/moq-relay.nix` sets `MOQ_SERVER_TLS_CERT`/`KEY`/`GENERATE` and `MOQ_SERVER_BIND`, and `fly.toml` sets `MOQ_SERVER_BIND`, so a relay would have come up with no certificate and on the wrong port. Each released spelling now lives on a hidden arg carrying its original env var, folded into the canonical field by `resolved()` with a one-line deprecation warning, which is what `--client-*` and `--client-quic-*` already did. (`--client-tls-disable-verify` had also been aliased to a name that never shipped, `--client-tls-insecure`; both parse now.)

TOML keeps working through serde aliases: `[client]`/`[server]` tables, `connect` for `url`, `failover_delay` for `race`, `disable_verify` for `insecure`, and `reconnect` for the inverted `once`.

A move across tables is the one thing serde aliases can't express, so `[client.quic]` and `[server.quic]` are parsed where they used to live and folded into the shared section: the top-level table wins, and between the two roles the dial side does, matching how the flags fold.

That leaves exactly one deliberate break:

- **QUIC tuning can no longer differ per role.** Previously you could set `server.quic.congestion_control = "delay"` and `client.quic.congestion_control = "loss"`; now there is one value, applied to dialed and accepted connections alike. `backend` and `version` remain per-role. This one is worth reading twice if you set only the client half: `--client-quic-max-streams 64` used to leave accepted connections at the relay's 10,000, and now bounds them too. It warns at startup, naming that consequence rather than just the rename.

## Bugs found on the way

A second-opinion review (Codex) went at the compatibility mechanism specifically. It found the per-role TOML gap above, plus three defects, all fixed here with tests:

- **The Unix peer-credential allowlist widened access when spellings mixed.** `--listen-unix-allow-gid 20` with `MOQ_SERVER_UNIX_ALLOW_UID=501` resolved to gid-only, because the canonical `Allow` won whole. The three lists are ANDed, so dropping one loosens the gate: from "that user, in that group" to "anyone in that group". They merge per credential now.
- **`--connect-bind '[::]:0'` lost to `--client-bind`**, since an explicit value equal to the default was indistinguishable from an absent flag. It also meant a clap `default_value` sat on the field, so `[client] bind` from a config file was clobbered by the CLI re-parse. Now `Option` with the default resolved in code.
- **Version lists were replaced rather than concatenated**, so `--connect-version A --client-version B` silently dropped B.
- **A config file outranked the flag that shares its spelling.** The first cut of the per-role TOML fold ran before the legacy flags folded, so `[client.quic] max_streams = 64` in a file beat `--client-quic-max-streams 128` on the command line. The flags fold first now, and the file only fills what is still unset.

- **The fold didn't reach the accept path at all.** `listen::Config::init` never called `quic.resolved()`, so `--server-quic-max-streams` was parsed and dropped. `Client::new` and `Server::new` now resolve their own inputs (idempotent), so no call path can skip it.
- **The relay defaulted before folding.** `Relay::load` filled in `DEFAULT_MAX_STREAMS` and *then* handed the config on; since the fold prefers the canonical field, a deployment still passing the released spelling was pinned to 10,000 streams. Now `Config::resolve` folds first and defaults after, with a regression test.
- **`--websocket-enabled` / `--websocket-delay` clobbered their own TOML.** Both carried a clap `default_value`, which the post-file CLI re-parse materializes over whatever the file said, the exact pitfall `rs/CLAUDE.md` documents. They are `Option` now with the defaults resolved in code (`resolved_enabled()` / `resolved_delay()`), so `delay = None` means 200ms rather than "no head start"; the callers that wanted no head start say `Some(ZERO)`.
- **`--listen-quic-lb-nonce` needs an id**, which clap's `requires` can no longer express now that each knob has two spellings. Checked after the fold instead (`Error::LbNonceWithoutId`), so any mix of the four is judged correctly.

## Notes

- `demo_configs_parse` reads the real `demo/relay/*.toml` files, since a rename that lands in code but not in the configs people copy is invisible until a relay refuses to boot. It caught one such break while being written.
- The `--cluster-version` knob an earlier revision of this PR added is gone: per-role `version` subsumes it.
- `--client-resolution-delay` (#2749) landed on `main` after this branch's base and hasn't reached `dev`, so it's the one released spelling not in the frozen list. Whoever merges `main` into `dev` should fold it into `connect::Config` alongside `--client-failover-delay`.
- Every doc, demo justfile, smoke script, and README that showed a renamed flag was reconciled against the binaries' real `--help`; nothing published teaches a dead spelling. Three error messages did too (`moq-bench`, `moq-boy`, and `moq transcode` all said "--client-connect is required").
- Verified: `just fix`, `just check`, and `just test`.

(Written by Opus 5)

### DIFF
--- a/demo/boy/justfile
+++ b/demo/boy/justfile
@@ -51,7 +51,7 @@ gb-run:
 
 # Start a Game Boy emulator publisher.
 start rom url='http://localhost:4443':
-    cargo run --bin moq-boy -- --client-connect "{{ url }}" --rom "{{ rom }}" --location localhost
+    cargo run --bin moq-boy -- --connect "{{ url }}" --rom "{{ rom }}" --location localhost
 
 # Host the web server
 web url='http://localhost:4443':
--- a/demo/relay/leaf0.toml
+++ b/demo/relay/leaf0.toml
@@ -3,10 +3,10 @@
 # The RUST_LOG environment variable will take precedence.
 level = "debug"
 
-[server]
+[listen]
 # Listen for QUIC connections on UDP:4444
 # Sometimes IPv6 causes issues; try 127.0.0.1:4444 instead.
-listen = "[::]:4444"
+bind = "[::]:4444"
 
 # Serve the same cluster-signed certificate that we present as a client,
 # so other leaves can verify us against `ca.pem`.
@@ -19,7 +19,7 @@ tls.key = ["leaf0.key"]
 # the JWT path and are rejected.
 tls.root = ["ca.pem"]
 
-[client]
+[connect]
 # Present this certificate and key on outbound cluster connections.
 # The root maps a valid cert to full access.
 tls.cert = "leaf0.crt"
--- a/doc/bin/relay/config.md
+++ b/doc/bin/relay/config.md
@@ -14,10 +14,10 @@ moq-relay relay.toml
 ## Minimal Example
 
 ```toml
-[server]
-listen = "0.0.0.0:4443"
+[listen]
+bind = "0.0.0.0:4443"
 
-[server.tls]
+[listen.tls]
 cert = "cert.pem"
 key = "key.pem"
 ```
@@ -45,7 +45,7 @@ Logging configuration.
 level = "info"
 ```
 
-### \[server]
+### \[listen]
 
 QUIC/WebTransport server settings. Optionally add plaintext qmux stream
 listeners for trusted local workers. Every connection authenticates through the
@@ -53,24 +53,24 @@ same JWT / public-access path; QUIC additionally accepts an mTLS client
 certificate, and Unix sockets add optional peer-credential gating.
 
 ```toml
-[server]
+[listen]
 # QUIC (UDP) bind. Omit to run stream-only (no QUIC) when a tcp/unix listener
 # is configured below.
 bind = "[::]:443"
 
 # Plaintext qmux over TCP (no TLS, carries no peer identity). Trusted networks
 # only; a non-loopback bind logs a warning. Requires the `tcp` build feature.
-[server.tcp]
+[listen.tcp]
 bind = "127.0.0.1:4444"
 
 # Plaintext qmux over a Unix socket, for local workers (e.g. the protocol
 # gateways or a stats publisher). Requires the `uds` build feature. Restrict
 # callers by peer credentials (each list AND across, OR within; empty = no
 # constraint).
-[server.unix]
+[listen.unix]
 bind = "/run/moq/internal.sock"
 
-[server.unix.allow]
+[listen.unix.allow]
 uid = [1001]
 # gid = [2000]
 # pid = [12345]
@@ -80,12 +80,12 @@ No-JWT connections on the stream transports resolve through the same
 public-access rules as tokenless QUIC clients (see [`[auth]`](#auth) `public`).
 See [Stream Listeners](/bin/relay/auth#stream-listeners) for details.
 
-### \[server.tls]
+### \[listen.tls]
 
 TLS configuration for the QUIC endpoint.
 
 ```toml
-[server.tls]
+[listen.tls]
 # Option 1: Provide certificate files
 cert = "/path/to/cert.pem"   # Certificate chain
 key = "/path/to/key.pem"     # Private key
@@ -110,7 +110,7 @@ HTTP server for debugging endpoints.
 [web.http]
 # Listen address for HTTP (TCP)
 # Defaults to disabled if not specified
-listen = "0.0.0.0:4443"
+bind = "0.0.0.0:4443"
 ```
 
 See [HTTP Endpoints](/bin/relay/http) for available endpoints.
@@ -139,7 +139,7 @@ HTTPS/WSS server for TCP fallback.
 # Listen address for HTTPS/WSS (TCP)
 listen = "0.0.0.0:443"
 
-# TLS certificates (can be the same as server.tls)
+# TLS certificates (can be the same as listen.tls)
 cert = "cert.pem"
 key = "key.pem"
 ```
@@ -204,18 +204,18 @@ secret = "/etc/moq/cluster.key"
 
 See [Clustering](/bin/relay/cluster) for topology choices and the trade-off between hand-listed peers and gossip.
 
-### \[client]
+### \[connect]
 
 Client settings used when connecting to other relays (clustering).
 
 ```toml
-[client]
+[connect]
 # Maximum time for one outbound dial and MoQ handshake. Defaults to 30s.
 # Set to "0" to wait forever.
 timeout = "30s"
 
 # Disable TLS verification (development only!)
-tls.disable_verify = true
+tls.insecure = true
 
 # What to do with the URI an upstream peer names in its GOAWAY:
 # "follow" (default), "same-host", or "ignore". A followed redirect is dialed
@@ -242,33 +242,31 @@ goaway.handover = "10s"
 # each starting this long after the previous one (or immediately, if that one
 # fails outright), and the first connection to complete wins. "0s" dials every
 # address at once. Defaults to 250ms, RFC 8305's Connection Attempt Delay.
-# failover_delay = "250ms"
+# race = "250ms"
 ```
 
-The connect timeout is also available as `--client-connect-timeout` or
-`MOQ_CLIENT_CONNECT_TIMEOUT`, and the failover delay as
-`--client-failover-delay` or `MOQ_CLIENT_FAILOVER_DELAY`. The two compose: the
-failover delay staggers the attempts within one dial, and the timeout bounds
-that dial as a whole.
+The connect timeout is also available as `--connect-timeout` or
+`MOQ_CONNECT_TIMEOUT`, and the address race as `--connect-race` or
+`MOQ_CONNECT_RACE`. The two compose: the race staggers the attempts within one
+dial, and the timeout bounds that dial as a whole.
 
-Pinning the source port (a non-zero port in `--client-bind`) disables address
-failover on the `quiche` backend, which binds a fresh socket per attempt and so
+Pinning the source port (a non-zero port in `--connect-bind`) disables address
+racing on the `quiche` backend, which binds a fresh socket per attempt and so
 can only dial one address at a time from a fixed port. The relay logs a warning
 at startup when both are set. Leave the bind port at `0` to keep failover, or
 use the `quinn` or `noq` backend, which share one socket across attempts and are
 unaffected.
 
-### \[server.quic] and \[client.quic]
+### \[quic]
 
-Per-connection QUIC transport knobs, applied to incoming connections
-(`server.quic`) and to outgoing cluster dials (`client.quic`) independently.
+Per-connection QUIC transport knobs. These mean the same thing whichever way a
+connection was opened, so they are spelled once and shared by `[listen]` and
+`[connect]` alike. The knobs that only apply when accepting (the QUIC preferred
+address and QUIC-LB connection IDs) live on `[listen]` instead.
 
 ```toml
-[server.quic]
+[quic]
 # "loss" or "delay". Defaults per backend; don't set "delay" on noq/iroh (see below).
-congestion_control = "delay"
-
-[client.quic]
 congestion_control = "delay"
 ```
 
@@ -291,9 +289,8 @@ noq and iroh are the exception because their shared BBRv3 can panic on packet
 loss, which aborts the process. Do not select `delay` on those backends unless
 you are testing that controller on purpose and can tolerate the crash.
 
-Also available as `--server-quic-congestion-control` /
-`--client-quic-congestion-control`, or `MOQ_SERVER_QUIC_CONGESTION_CONTROL` /
-`MOQ_CLIENT_QUIC_CONGESTION_CONTROL`.
+Also available as `--quic-congestion-control` /
+`--quic-congestion-control`, or `MOQ_QUIC_CONGESTION_CONTROL`.
 
 ### \[stats]
 
--- a/rs/hang/examples/subscribe.rs
+++ b/rs/hang/examples/subscribe.rs
@@ -24,7 +24,7 @@ async fn main() -> anyhow::Result<()> {
 // Automatically reconnects if the connection drops.
 async fn run_session(origin: moq_net::origin::Producer) -> anyhow::Result<()> {
 	// Optional: Use moq_native to make a QUIC client.
-	let client = moq_native::ClientConfig::default().init()?;
+	let client = moq_native::connect::Config::default().init(Default::default())?;
 
 	// For local development, use: http://localhost:4443/video-example
 	// The "anon" path is usually configured to bypass authentication; be careful!
--- a/rs/moq-bench/src/main.rs
+++ b/rs/moq-bench/src/main.rs
@@ -23,12 +23,12 @@ async fn main() -> anyhow::Result<()> {
 
 	let config = Config::load()?;
 	anyhow::ensure!(
-		config.client.connect.is_some(),
-		"--client-connect is required (or set it in the TOML file)"
+		config.client.url.is_some(),
+		"--connect is required (or set it in the TOML file)"
 	);
 
 	let config = Arc::new(config);
-	let client = config.client.clone().init()?;
+	let client = config.client.clone().init(config.quic.clone())?;
 	let stats = Arc::new(Stats::default());
 
 	// Periodic throughput reporter.
@@ -45,7 +45,7 @@ async fn main() -> anyhow::Result<()> {
 	let run_id = rng.random_range(0..=u64::MAX);
 	let startup = config.startup();
 
-	tracing::info!(connections = count, url = %config.client.connect.as_ref().unwrap(), "starting benchmark");
+	tracing::info!(connections = count, url = %config.client.url.as_ref().unwrap(), "starting benchmark");
 
 	let mut tasks = tokio::task::JoinSet::new();
 	for i in 0..count {
--- a/rs/moq-native/src/lib.rs
+++ b/rs/moq-native/src/lib.rs
@@ -16,7 +16,7 @@
 pub mod accept;
 pub mod bind;
 mod client;
-mod connect;
+pub mod connect;
 mod connection;
 mod crypto;
 mod error;
@@ -24,6 +24,7 @@ mod error;
 pub mod failover;
 #[cfg(feature = "jemalloc")]
 pub mod jemalloc;
+pub mod listen;
 mod log;
 #[cfg(feature = "noq")]
 pub mod noq;
@@ -48,12 +49,12 @@ pub mod websocket;
 
 // Enumerated rather than globbed, so the root surface is a deliberate list and a
 // new `pub` item in these modules doesn't silently join it.
-pub use client::{Client, ClientConfig};
+pub use client::Client;
 pub use connect::{Addrs, ConnectError};
 pub use connection::{Backoff, Connection, ConnectionStatsReader, GoawayConfig, Redirect, Status};
 pub use error::{Error, Result};
 pub use log::Log;
-pub use server::{Listener, Request, Server, ServerConfig, Transport};
+pub use server::{Listener, Request, Server, Transport};
 
 /// Spawn the session's protocol driver on the current tokio runtime, handing back
 /// the session it drives.
--- a/demo/justfile
+++ b/demo/justfile
@@ -11,7 +11,7 @@ mod web
 # Picks the first free port at/after 4443 (both UDP for QUIC and TCP for HTTP)
 # so multiple worktrees can run `just dev` without colliding. The relay reads
 
-# the port from MOQ_SERVER_BIND/MOQ_WEB_HTTP_LISTEN, overriding localhost.toml.
+# the port from MOQ_LISTEN/MOQ_WEB_HTTP_LISTEN, overriding localhost.toml.
 default:
     #!/usr/bin/env bash
     set -euo pipefail
@@ -25,10 +25,10 @@ default:
     base="http://localhost:$port"
 
     # Scope the bind env vars to the relay only. The unified `moq` binary also
-    # reads MOQ_SERVER_BIND, so exporting it globally would make the `pub` client
+    # reads MOQ_LISTEN, so exporting it globally would make the `pub` client
     # try to start a server (and fail without a TLS cert).
     bun run concurrently --kill-others --names rly,bbb,web --prefix-colors auto \
-    	"MOQ_SERVER_BIND='[::]:$port' MOQ_WEB_HTTP_LISTEN='[::]:$port' just relay" \
+    	"MOQ_LISTEN='[::]:$port' MOQ_WEB_HTTP_LISTEN='[::]:$port' just relay" \
     	"just wait $base/certificate.sha256 && just pub bbb $base" \
     	"just wait $base/certificate.sha256 && just web serve $base"
 
--- a/demo/pub/justfile
+++ b/demo/pub/justfile
@@ -48,18 +48,18 @@ tos url='http://localhost:4443' *args:
 # Publish a video using ffmpeg (CMAF / fMP4) to a relay server.
 cmaf name url='http://localhost:4443' *args:
     cargo build --bin moq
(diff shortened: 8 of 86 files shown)


rs/moq-bench/src/main.rs
```diff
@@ -22,13 +22,14 @@ async fn main() -> anyhow::Result<()> {
 		.expect("failed to install default crypto provider");
 
 	let config = Config::load()?;
+	let endpoint = config.client.resolved();
```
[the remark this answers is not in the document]
kixelated: Fixed. `parse_and_merge` now folds the legacy spellings once via `resolved()`, so every later field read (the startup log line and the dial in `connection.rs`) sees the canonical values; the clock example resolves before reading `connect` too. A `--client-connect`-launched run no longer panics.

(Written by Fable 5)

rs/moq-native/src/lib.rs
```diff
@@ -44,12 +46,13 @@ pub mod websocket;
 
 // Enumerated rather than globbed, so the root surface is a deliberate list and a
 // new `pub` item in these modules doesn't silently join it.
-pub use client::{Client, ClientConfig};
+pub use client::Client;
+pub use config::Config;
```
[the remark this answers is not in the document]
kixelated: Fixed. `doc/lib/rs/index.md` and `doc/lib/rs/env/native.md` now use `moq_native::Config::default().client()` (docs.rs links repointed at `Config`), and `rs/moq-rtmp/README.md` builds its RTMPS config from `moq_native::tls::Config`.

(Written by Fable 5)

rs/moq-native/src/config.rs
```diff
@@ -0,0 +1,612 @@
...
+	/// The URL to dial.
+	///
+	/// Supports WebTransport (`https`/`http`), WebSocket (`ws`/`wss`), raw QUIC
+	/// (`moqt`/`moql`), qmux over `tcp`/`unix`, and `iroh`. The URL path is the
+	/// request/auth path (e.g. `/anon` for a public relay) and `?jwt=` supplies a
+	/// token. `http://` first fetches `/certificate.sha256` for the (insecure)
+	/// self-signed fingerprint; `https://` connects directly.
+	#[serde(skip_serializing_if = "Option::is_none")]
+	#[arg(id = "connect", long = "connect", env = "MOQ_CONNECT")]
+	pub connect: Option<Url>,
```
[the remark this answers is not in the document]
kixelated: Fixed. `rs/moq-cli/README.md` and `rs/moq-bench/README.md` now teach `--connect` / `--bind` / `--tls-*`, and a repo-wide grep also updated `doc/bin/cli.md` plus stale flag mentions in comment prose (clock example header, smoke-test comments, moq-boy).

(Written by Fable 5)

kixelated:
Took over this PR: ran a Codex adversarial review, rebased onto current `dev` (37 commits), fixed CI, and addressed the review + inline findings. Details in the description's "Updates (takeover)" section; triage of the adversarial review's two findings:

**[critical] "Outbound TLS roots become unrestricted inbound relay credentials"** — partially agree.

- *Agreed and fixed*: the legacy TOML fold was buggy. `tls::Config::or` kept the first non-empty list, so a config with distinct `client.tls.root` and `server.tls.root` silently dropped the server's mTLS CAs (breaking clients the old config authenticated) while still folding the dial roots into the accept side. List fields now concatenate, matching the CLI fold and the documented behavior, with regression tests covering distinct roots in both tables.
- *Disagreed (kept as designed)*: the single `root` serving both directions is the PR's deliberate, documented tradeoff (the description's "Merged trust roots" section, with `tls.system_roots` as the escape hatch). Worth being aware of the sharp edge though: a root added only to verify outbound dials now also arms inbound mTLS, and a validated client cert grants full access within the path root. The PR already keeps `client_cert`/`client_key` separate for the same class of reason, so splitting `root` into dial/accept halves has precedent if we ever want it; that's a design call for review rather than something I changed under a takeover.

**[high] "This breaking dev stack must not target main"** — artifact of the review setup: Codex diffed against `origin/main`, but the PR targets `dev`, which is exactly where a semver break goes per CONTRIBUTING. The valid kernel (the branch trailing `dev` by 37 commits) is addressed by the rebase.

One pre-existing issue surfaced while validating: `goaway_cluster::cluster_diamond_goaway_seamless_failover` fails deterministically on clean `origin/dev` (a duplicate group after the drained leg, 22 vs 21). Unrelated to this PR; tracking it separately.

(Written by Fable 5)

kixelated:
CI status: the `Check` failure on the latest push is not this PR. It fails in `js/net`'s `integration: lite applies initial and updated group bounds`, which is broken on clean `dev` by the af7b5720f merge conflict resolution (a `break` on `endGroup` present in neither parent). #2758 already root-causes and fixes it; this PR contains no JS changes (`git diff dev -- js/` is empty), so a re-run after #2758 lands should clear it. `Smoke` and `Swift` are green, and the Rust side passes locally (`just check`, the previously-failing `cluster_unknown` suite, and the rustdoc `-D warnings` gate).

(Written by Fable 5)

kixelated:
Rebased onto the new `dev` tip (0bde6b16a), which includes #2766, so the unrelated `@moq/net` integration failure from the last run should be gone.

Two `--all-features` compile errors from the previous CI run, now fixed:

- **Mine:** `transcode.rs` still read `.connect` where the field is now `.connect.url`. It's behind the `transcode` feature, which my default-feature clippy run doesn't compile; verified with `cargo check -p moq-cli --features transcode`.
- **Not mine, and it blocks every PR against `dev`:** `moq play` sets `decode.latency_max`, but #2688 renamed that field to `latency` and retyped it as `Latency`, while #2697 added the call site against the old shape. Verified pre-existing by compiling `moq-cli --features play` on a clean `origin/dev` worktree (fails identically there). It slipped through because `Check` only runs on `pull_request`, so nothing built `--all-features` on the merge itself. Fixed in a separate commit (ef2eaa1cb) since it isn't part of this refactor. Happy to split it into its own PR against `dev` if you'd rather unblock everyone else ahead of this one.

(Written by Fable 5)

demo/boy/justfile
```diff
@@ -51,7 +51,7 @@ gb-run:
 
 # Start a Game Boy emulator publisher.
 start rom url='http://localhost:4443':
-    cargo run --bin moq-boy -- --client-connect "{{ url }}" --rom "{{ rom }}" --location localhost
```
kixelated, line 54: good change

demo/relay/leaf0.toml
```diff
@@ -3,32 +3,26 @@
...
+# Present the same certificate and key on outbound cluster connections.
 # The root maps a valid cert to full access.
-tls.cert = "leaf0.crt"
-tls.key = "leaf0.key"
-
-# Verify the remote's server cert against the cluster CA. All cluster
-# nodes' server certs are signed by `ca.pem`, so no custom trust
-# store is required beyond this one file.
-tls.root = ["ca.pem"]
+tls.client_cert = "leaf0.crt"
```
kixelated: not a fan.

I think the existing server/client structs made sense here.

doc/bin/relay/config.md
```diff
@@ -45,59 +45,80 @@ Logging configuration.
...
+[endpoint]
 # QUIC (UDP) bind. Omit to run stream-only (no QUIC) when a tcp/unix listener
 # is configured below.
 bind = "[::]:443"
 
 # Plaintext qmux over TCP (no TLS, carries no peer identity). Trusted networks
 # only; a non-loopback bind logs a warning. Requires the `tcp` build feature.
-[server.tcp]
+[endpoint.tcp]
 bind = "127.0.0.1:4444"
```
kixelated, line 64: TCP can't use a single bind for both clients and servers (technically yes, but not if you want to dial multiple connections to the same server).

I think server. made more sense

doc/bin/relay/config.md
```diff
@@ -204,18 +231,29 @@ secret = "/etc/moq/cluster.key"
...
 
 ```toml
-[client]
+[endpoint.connect]
 # Maximum time for one outbound dial and MoQ handshake. Defaults to 30s.
 # Set to "0" to wait forever.
 timeout = "30s"
 
-# Disable TLS verification (development only!)
-tls.disable_verify = true
```
kixelated, line 218: While we're here, maybe rename it to `insecure` or something?

rs/hang/examples/subscribe.rs
```diff
@@ -24,7 +24,7 @@ async fn main() -> anyhow::Result<()> {
 // Automatically reconnects if the connection drops.
 async fn run_session(origin: moq_net::origin::Producer) -> anyhow::Result<()> {
 	// Optional: Use moq_native to make a QUIC client.
-	let client = moq_native::ClientConfig::default().init()?;
```
kixelated, line 27: if we do go back to Client/Server structs, I tihnk a client/server module might be nicer? `moq_native::client::Config`

rs/moq-native/src/config.rs
```diff
@@ -0,0 +1,670 @@
...
+
+	/// Retry pacing for [`crate::Client::connect`] (`--backoff-*`).
+	#[command(flatten)]
+	#[serde(default)]
+	pub backoff: crate::Backoff,
+
+	/// How [`crate::Client::connect`] reacts to a peer's GOAWAY (`--goaway-*`).
+	#[command(flatten)]
+	#[serde(default)]
+	pub goaway: crate::GoawayConfig,
```
kixelated: This should go in `connect`

rs/moq-native/src/config.rs
```diff
@@ -0,0 +1,670 @@
...
+	/// TLS trust, served certificates, and the optional dialing identity
+	/// (`--tls-*`, `--client-tls-cert`/`-key`).
+	#[command(flatten)]
+	#[serde(default)]
+	pub tls: crate::tls::Config,
+
+	/// Retry pacing for [`crate::Client::connect`] (`--backoff-*`).
+	#[command(flatten)]
+	#[serde(default)]
+	pub backoff: crate::Backoff,
```
kixelated: This should go in `connect`

rs/moq-native/src/config.rs
```diff
@@ -0,0 +1,670 @@
...
+	/// with no QUIC.
+	#[serde(alias = "listen", skip_serializing_if = "Option::is_none")]
+	#[arg(id = "bind", long = "bind", env = "MOQ_BIND")]
+	pub bind: Option<String>,
+
+	/// The QUIC backend to use.
+	/// Auto-detected from compiled features if not specified.
+	#[serde(skip_serializing_if = "Option::is_none")]
+	#[arg(id = "backend", long = "backend", env = "MOQ_BACKEND")]
+	pub backend: Option<QuicBackend>,
```
kixelated: This should go in `quic`

rs/moq-native/src/config.rs
```diff
@@ -0,0 +1,670 @@
...
+	#[command(flatten)]
+	#[serde(default)]
+	pub tcp: crate::tcp::Config,
+
+	/// Plaintext qmux Unix-socket listener (`--unix-bind`) with an optional
+	/// peer-credential allowlist. Requires the `uds` feature; unix-only.
+	#[cfg(all(feature = "uds", unix))]
+	#[command(flatten)]
+	#[serde(default)]
+	pub unix: crate::unix::Config,
```
kixelated: Maybe in server?

rs/moq-native/src/config.rs
```diff
@@ -0,0 +1,670 @@
...
+	#[command(flatten)]
+	#[serde(default)]
+	pub websocket: crate::websocket::Config,
+
+	/// Plaintext qmux TCP listener (`--tcp-bind`, no TLS). Requires the
+	/// `tcp` feature.
+	#[cfg(feature = "tcp")]
+	#[command(flatten)]
+	#[serde(default)]
+	pub tcp: crate::tcp::Config,
```
kixelated: maybe in server?

kixelated:
Re-cut per your review: this **keeps the client/server split** instead of merging the two configs, so the diff is now a rename rather than a restructure (−5642/+2649 against the previous revision's shape).

What changed from the merged design:

- `connect::` / `listen::` role modules, per your `moq_native::client::Config` suggestion, with `--connect` and `--listen` as the two bare flags.
- `backoff` and `goaway` moved into `connect`, and the accept-only preferred-address / QUIC-LB knobs moved onto `listen`, as you asked.
- `backend` and `version` are per-role, so quiche-accepting-with-quinn-dialing works again (the merge had quietly removed it). That also made the `--cluster-version` knob from the previous revision redundant, so it's gone.
- `disable_verify` is `insecure`.

Two things worth your attention, both explained in the description:

- **Sharing `quic` is forced, not chosen.** clap cannot flatten one struct twice in a parser, which is exactly why `dev` carries duplicated `quic::Client`/`quic::Server` definitions. So `quic::Config` is a sibling of the two role configs, handed to each at `init`. The cost is that QUIC tuning can no longer differ per role, and `[server.quic]`/`[client.quic]` become a top-level `[quic]` (a move across tables, which serde aliases can't cover, so it fails loudly at startup).
- **The merged trust root that Codex flagged as critical is simply gone.** `connect.tls.root` and `listen.tls.root` are separate fields again, so there is no shared-root tradeoff to weigh.

Also added `demo_configs_parse`, which reads the real `demo/relay/*.toml` files. It caught a config-breaking rename while I was writing it, which is exactly the class of bug that otherwise surfaces as someone's relay refusing to boot.

Verified: `cargo fmt`, workspace clippy `-D warnings`, rustdoc `-D warnings --all-features`, and 2942/2942 tests.

(Written by Fable 5)

demo/relay/leaf0.toml
kixelated: Done in the re-cut: the merged `[endpoint]` table is gone. The relay's tables are `[connect]` and `[listen]` now (with `client`/`server` serde aliases so existing configs keep parsing), and the types are role modules, `moq_native::connect::Config` / `moq_native::listen::Config`, per your suggestion.

(Written by Fable 5)

rs/hang/examples/subscribe.rs
kixelated: Done in the re-cut: the merged `[endpoint]` table is gone. The relay's tables are `[connect]` and `[listen]` now (with `client`/`server` serde aliases so existing configs keep parsing), and the types are role modules, `moq_native::connect::Config` / `moq_native::listen::Config`, per your suggestion.

(Written by Fable 5)

doc/bin/relay/config.md
kixelated: Renamed: it's `--connect-tls-insecure` / `connect.tls.insecure` now, with `disable_verify` kept as a serde alias. The name should say what it does.

(Written by Fable 5)

doc/bin/relay/config.md
kixelated: Confirmed by the spike, and it's now the documented reason: quiche binds a fresh socket per connection attempt, so a shared or pinned dial port costs it address racing entirely. The dial socket is `--connect-bind` (its own knob on the dial side), the listener is `--listen`, and they stay separate.

(Written by Fable 5)

rs/moq-native/src/config.rs
kixelated: Applied: `backoff` and `goaway` are in `connect` now, `tcp`/`unix` are on `listen`, and `backend` is per-role (`connect.backend` / `listen.backend`) rather than in the shared `quic` section, which turned out to matter: a quiche acceptor with a quinn dialer is a real setup that the merged config had quietly removed.

(Written by Fable 5)

rs/moq-native/src/config.rs
kixelated: Applied: `backoff` and `goaway` are in `connect` now, `tcp`/`unix` are on `listen`, and `backend` is per-role (`connect.backend` / `listen.backend`) rather than in the shared `quic` section, which turned out to matter: a quiche acceptor with a quinn dialer is a real setup that the merged config had quietly removed.

(Written by Fable 5)

rs/moq-native/src/config.rs
kixelated: Applied: `backoff` and `goaway` are in `connect` now, `tcp`/`unix` are on `listen`, and `backend` is per-role (`connect.backend` / `listen.backend`) rather than in the shared `quic` section, which turned out to matter: a quiche acceptor with a quinn dialer is a real setup that the merged config had quietly removed.

(Written by Fable 5)

rs/moq-native/src/config.rs
kixelated: Applied: `backoff` and `goaway` are in `connect` now, `tcp`/`unix` are on `listen`, and `backend` is per-role (`connect.backend` / `listen.backend`) rather than in the shared `quic` section, which turned out to matter: a quiche acceptor with a quinn dialer is a real setup that the merged config had quietly removed.

(Written by Fable 5)

PR #2758, opened 2026-08-12
fix(net): stop treating the subscription end group as a terminator

kixelated:
## Summary

`dev` is red: `integration: lite applies initial and updated group bounds` fails on the first `nextGroup()`, and subscription bounds deliver nothing at all (verified for both pre-written and live-arriving groups).

**Root cause.** The merge in af7b5720f resolved a conflict in `js/net/src/lite/publisher.ts` with a line present in neither parent (it shows as `++` in `git show af7b5720f`):

```js
if (bounds.endGroup !== undefined && group.sequence >= bounds.endGroup) break;
```

The end group is a serving cap, not a subscription terminator. `rs/moq-net/src/lite/publisher.rs` says so verbatim, and moq-lite says SUBSCRIBE_END "bounds the range but does not by itself end the stream". That `break` FINs the subscribe stream immediately after queueing the last group's `#runGroup`, before those unidirectional streams have written a byte, so the subscriber reads the FIN as the end of the subscription, closes the local producer, and drops them. It also made the SUBSCRIBE_UPDATE bound-raise path unreachable.

**A second bug it was masking.** Removing the `break` makes `lite draft-06: groups past the end group are not served` hang. That test already failed on the pre-merge parent d7c58d75b (confirmed on a clean worktree), so the `break` had been papering over it. `Subscriber.nextGroup` parks when the track closes with groups buffered above the `endAt` cap, which is deliberate and mirrors Rust's `poll_next_in_range` ("Pending (not None) so the consumer is parked rather than told the stream is over"). The publisher was driving SUBSCRIBE_END off `nextGroup` draining rather than off the track ending, so a capped subscription never reported one. Rust keeps those two events separate (`Recv::Boundary` from `poll_finished` vs `Recv::Finished`), so `#runTrack` now does too. Fixing this in `track.ts` instead was considered and rejected: it would diverge a core primitive from its documented Rust counterpart to paper over a publisher-layer bug.

**Boundary value.** The above makes SUBSCRIBE_END observable under a cap, where it was wrong. It is specified as "the first group that will never be delivered", so it now reports the track's live edge rather than `max-delivered + 1`, which would promise an ending a later update sends groups past. The two coincide without a cap, so the three existing `subscribeEnd` tests are unchanged.

## Public API changes

None. No new, renamed, removed, or signature-changed exported item. `Publisher` is `@internal`; `nextGroup`/`startAt`/`endAt` are untouched. Targets `dev` only because both bugs are `dev`-only (`main` has no subscription bounds), not because anything breaks.

## Cross-package sync

No paired updates owed. No message encoding changed, only when SUBSCRIBE_END is sent and what value it carries, both of which now conform to draft text and Rust behavior that already exist. `js/net/src/ietf/publisher.ts` was checked and has no equivalent hole: `git show af7b5720f -- js/net/src/ietf/publisher.ts` is empty (git merged it cleanly, with no hand-editing), the IETF path carries no group bounds on the wire (`Subscribe.decode` keeps only requestId/namespace/name/priority and rejects any filter type but `0x1`/`0x2`), and it serves via `recvGroup()`, which ignores the `endAt` cap. An IETF bounds test would require first adding IETF bounds support, which is a feature rather than part of this fix.

## Test plan

- `bun test` in `js/net`: 391 pass, 0 fail.
- Each new assertion verified to fail without its fix, and the pre-merge hang reproduced on a clean `d7c58d75b` worktree so it was not misattributed to the merge.
- `nix develop --command just check`: clean.
- Codex adversarial review: approve, no material findings.

## Notes

The merge also left `ServeGroup` in `lite/publisher.ts` as a near-duplicate of the new `RunGroup`, surviving only to derive `ServeFetch` via `Omit`. Left alone as unrelated to the bug, but worth collapsing.

(Written by Opus 5)

### DIFF
--- a/js/net/src/lite/publisher.test.ts
+++ b/js/net/src/lite/publisher.test.ts
@@ -377,13 +377,13 @@ test("lite draft-05: subscribe end is 0 when no groups were produced", async () 
 type Served = { sequence: number; frameStart: number; payloads: string[] };
 
 /**
- * Serves `groups` (frame payloads per group, keyed by sequence) under `bounds`, and
- * reports every group stream that reached the wire plus the resolved SUBSCRIBE_START.
+ * Serves `groups` (frame payloads per group, keyed by sequence) under `bounds`, and reports
+ * every group stream that reached the wire plus the resolved SUBSCRIBE_START/SUBSCRIBE_END.
  */
 async function serve(
 	groups: Record<number, string[]>,
 	bounds: { startGroup?: number; startFrame?: number; endGroup?: number; endFrame?: number },
-): Promise<{ start?: number; served: Served[] }> {
+): Promise<{ start?: number; end: number; served: Served[] }> {
 	const pair = createMockTransportPair(ALPN_06_WIP);
 	const publisher = new Publisher(pair.server, Version.DRAFT_06, randomOrigin());
 
@@ -448,12 +448,16 @@ async function serve(
 		}
 
 		let start: number | undefined;
+		let end: number;
 		for (;;) {
 			const resp = await decodeSubscribeResponse(client.reader, Version.DRAFT_06);
 			if ("start" in resp) start = resp.start.group;
-			if ("end" in resp) break;
+			if ("end" in resp) {
+				end = resp.end.group;
+				break;
+			}
 		}
-		return { start, served };
+		return { start, end, served };
 	} finally {
 		// Settles the pending read as well as dropping the stream.
 		await reader.cancel();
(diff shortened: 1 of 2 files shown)


kixelated:
Closing as superseded by #2766, which landed on `dev` first and fixes both bugs from the same diagnosis (the `endGroup` break hand-added in the af7b5720f conflict resolution, plus the pre-existing capped-serve hang it was masking).

Nothing here is worth salvaging. Rebasing onto the current `dev` conflicts in every hunk, and each conflict is #2766's implementation of the same fix beside this one, so resolving it reduces to an empty commit. `dev` also already carries the equivalent test (`a capped subscription still reports the track's final boundary`) and the `serve()` change.

#2766 is also the better implementation on three points:

- **Boundary**: it uses `track.final()`, which counts datagram sequences (groups and datagrams share the namespace), keeping `latest() + 1` only as the live-track fallback. This PR's bare `latest() + 1` under-reports SUBSCRIBE_END on a track whose trailing sequences were datagrams, which a Rust peer would feed into `finish_at`.
- **Abort path**: its `done` promise rejects on abort and feeds the catch/reset explicitly, where this one silently never resolved.
- **Read cursor**: it hoists `next` outside the loop with teardown in a `finally`, so a boundary emission cannot spawn a second concurrent `nextGroup`.

Note that CI passing on this PR is misleading, since it ran against the pre-#2766 base. A clean checkout of `dev` at 0bde6b16a is green (86 pass, 0 fail across the integration, publisher, and track suites).

(Written by Opus 5)

PR #2767, opened 2026-08-12
fix(net): transmit groups in sequence order for ordered subscriptions

kixelated:
## Summary

- `cluster_diamond_goaway_seamless_failover` fails deterministically on dev (`left: 22, right: 21` in the post-drain phase). Root cause: it is not a failover bug. The resume/takeover machinery is correct (boundary `{group: 21, frame: 0}`, every SUBSCRIBE and SUBSCRIBE_START in the chain carries 21, and every relay hop serves group 21 first). The inversion happens on the final relay-to-subscriber hop: the test publishes its post-drain groups back-to-back, the lite publisher ranked group streams **newest-first** within a track, and qmux 0.5's strict-priority write scheduler (#2717, which replaced arrival-order frame interleaving) transmits group 22's queued frames before 21's whenever both are buffered at once. The subscriber receives 22 before 21, which the test's strict in-order assertion reports as a failover defect.
- Newest-first is the draft's intended **live** behavior. The draft's `Ordered` field is how a subscriber asks for oldest-first instead, and moq-lite carried it on the wire and in the demand aggregate but never applied it to transmission. This PR implements it, and the diamond test now subscribes `with_ordered(true)` so its strict sequence-order verification asserts a property the protocol actually grants. The flag propagates end-to-end (subscriber -> BOTTOM -> MID-B -> TOP) through the demand aggregate and resume slices.

## How the rank is scoped

The draft scopes `Ordered` to groups *within a Track*, and `Priority` is what selects *between* tracks. That distinction is load-bearing here, because one `PriorityQueue` serves every subscription on a session:

- `Priority` now carries the subscribe id and compares it **before** the group, so a group tie-break only ever happens between groups of the same subscription. An ordered rank and a live rank are therefore never compared against each other. Without this, any direction encoding that puts a subscription's oldest group on top also puts it above every group of a live subscription at equal priority, starving that track until its groups expire.
- `PriorityQueue::set_ordered` re-ranks a subscription's **already-queued** groups when SUBSCRIBE_UPDATE flips the flag, so the backlog adopts the new direction instead of being outranked by every group that follows it. It rebuilds both the sorted vec and the overflow heap from one snapshot: the flip can move entries across the 255-item boundary, and `Priority` is only a total order while every group of a subscription agrees on its direction, so a per-item re-place would transiently compare two directions within one subscription.
- `PriorityHandle` now owns only its track priority. It previously wrote a whole cached `Priority` back on `set_track`, which after a `set_ordered` would have resurrected the stale direction for one group.

In-flight `serve_group` tasks pick the new send order up through the existing rank-change notification, the same path a SUBSCRIBE_UPDATE priority change already used.

## Public API changes

None. The whole `lite` module is crate-internal: `lite::priority::Priority` gains a `subscribe` field and its constructors take it, `PriorityQueue::set_ordered` is new, and `PriorityState::set_priority` narrowed to `set_track`. `track::Subscription.ordered` already existed and is unchanged.

## Test plan

- `cluster_diamond_goaway_seamless_failover`: failed deterministically before, passes repeatedly after.
- Four new tests in `lite/priority.rs`, each verified by mutation (the fix reverted, the test fails):
  - `test_ordered_does_not_outrank_a_live_subscription` — fails when the subscribe comparison is removed from `Ord`.
  - `test_set_ordered_reranks_the_backlog` and `test_set_ordered_promotes_across_the_overflow_boundary` — both fail when `set_ordered` is neutered to the old future-groups-only behavior.
  - `test_set_ordered_leaves_other_subscriptions_alone`.
- The pre-existing priority tests are unchanged in substance; they were rewritten onto shared `live`/`older_first` helpers that pin them to one subscription, which is the scope they were always exercising.
- `cargo nextest run -p moq-net -p moq-relay`, plus `just fix` / `just check`.

## Cross-package sync

- `drafts/`: no update needed. This implements the existing `Ordered` semantics in draft-lcurley-moq-lite's Prioritization section ("older (0x1) or newer (0x0) groups ... first within a Track") rather than changing them; no wire change.
- `js/net`: not updated here. The JS lite publisher sets no sendOrder on group streams at all and ignores `ordered`, so it needs the same treatment; tracked separately.

Debugging note: the original failure is a heisenbug. Any stderr logging in the serve path yields to the qmux writer task between two `queue_serve` calls, flushing the older group's stream before the newer one is queued, which hides the race; it was pinned down with an in-memory event buffer dumped from a panic hook.

(written by Fable 5)

### DIFF
--- a/rs/moq-net/src/lite/priority.rs
+++ b/rs/moq-net/src/lite/priority.rs
@@ -35,9 +35,9 @@ pub struct Priority {
 	track: u8,
 	subscribe: u64,
 	group: u64,
-	/// The owning subscription's `Ordered` preference. Only ever read when
-	/// `subscribe` matches, so both sides of such a comparison agree on it; see
-	/// [`PriorityState::set_ordered`] for what keeps that true.
+	/// The owning subscription's `Ordered` preference, which decides which way
+	/// `group` is read. [`PriorityState::set_ordered`] moves a whole
+	/// subscription's groups at once so they agree on it.
 	ordered: bool,
 }
 
@@ -74,7 +74,15 @@ impl Ord for Priority {
 			// stable and independent of either one's direction. The older (lower id)
 			// subscription goes first.
 			.then(self.subscribe.cmp(&other.subscribe))
-			// Same subscription, so both sides carry its direction.
+			// Normally redundant, since one subscription's groups all carry its
+			// direction. It is here so the comparison below is reached only by two
+			// groups that agree on it, which makes this a total order for every
+			// value rather than only for well-formed ones: reversing the sense for
+			// one side alone is intransitive, and `sort` is documented to be free
+			// to panic on a comparator that does that. Nothing in the protocol
+			// stops a peer opening two SUBSCRIBEs with one id and disagreeing.
+			.then(self.ordered.cmp(&other.ordered))
+			// Both sides agree on the direction, so it can be read off either.
 			.then_with(|| match self.ordered {
 				true => self.group.cmp(&other.group),
 				false => other.group.cmp(&self.group),
@@ -561,6 +569,35 @@ mod tests {
 		assert_eq!(handles[0].current(), 0);
 		assert_eq!(handles[1].current(), 1);
 		assert_eq!(handles[total - 1].current(), u8::MAX);
+	}
+
+	/// `Ord` has to hold for every value, not just the ones a well-behaved peer
+	/// produces. Nothing stops a peer opening two SUBSCRIBEs with the same id and
+	/// opposite `Ordered`, which lands both directions under one subscribe id, and
+	/// a comparator that says `a < b` and `b < a` is free to panic `sort`.
+	#[test]
+	fn test_ord_is_total_across_mixed_directions() {
+		let mixed = [
+			Priority::ordered(100, 7, 1),
+			Priority::new(100, 7, 1),
+			Priority::ordered(100, 7, 2),
+			Priority::new(100, 7, 2),
+			Priority::new(100, 8, 1),
+			Priority::new(200, 7, 1),
+		];
+
+		for a in mixed {
+			for b in mixed {
+				// Antisymmetric, and agreeing with Eq on what equality means.
+				assert_eq!(a.cmp(&b), b.cmp(&a).reverse(), "{a:?} vs {b:?}");
+				assert_eq!(a.cmp(&b) == Ordering::Equal, a == b, "{a:?} vs {b:?}");
+				for c in mixed {
+					if a.cmp(&b) != Ordering::Greater && b.cmp(&c) != Ordering::Greater {
+						assert_ne!(a.cmp(&c), Ordering::Greater, "{a:?} <= {b:?} <= {c:?}");
+					}
+				}
+			}
+		}
 	}
 
 	/// A flip must not disturb a subscription that did not ask for one.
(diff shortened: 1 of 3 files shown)


kixelated:
Both findings from the Codex review are fixed in d7f3e9257, and both were the same root cause: the ordering direction was encoded into an absolute rank compared across a queue that serves the whole session.

**Ordered leaking across tracks.** `Priority` now carries the subscribe id and compares it *before* the group, so a group tie-break only ever happens between groups of one subscription and an ordered rank is never weighed against a live one. This is also what the draft asks for: `Ordered` picks between groups within a Track, `Priority` picks between Tracks. Regression test `test_ordered_does_not_outrank_a_live_subscription`.

**Queued groups stranded on the old direction.** `PriorityQueue::set_ordered` re-ranks a subscription's already-queued groups when SUBSCRIBE_UPDATE flips the flag. It rebuilds the sorted vec and the overflow heap from a single snapshot rather than re-placing item by item, because the flip can move entries across the 255-item boundary and a partial re-place would transiently compare two directions within one subscription, which is intransitive and would corrupt the vec. In-flight `serve_group` tasks pick the new send order up over the existing rank-change notification. Regression tests `test_set_ordered_reranks_the_backlog` and `test_set_ordered_promotes_across_the_overflow_boundary`.

Related: `PriorityHandle` now owns only its track priority. It previously cached a whole `Priority` and wrote it back on `set_track`, which after a `set_ordered` would have resurrected the stale direction for one group.

Each new test was mutation-checked: removing the subscribe comparison from `Ord` fails the first, and neutering `set_ordered` to the old future-groups-only behavior fails the other two. Full suite 1002 passed, `just check` clean, and the diamond test passed 10/10 on the final tree.

(written by Fable 5)

kixelated:
Codex re-reviewed the fix commit and returned HOLD with two findings. One is real and is now fixed in 467e952c7; the other I'm rejecting, with reasons.

**Fixed — `Ord` was not a total order for every value.** The comparator read the direction off `self` once `subscribe` matched, which is only antisymmetric while both sides agree on it. I had argued the one-direction-per-subscription invariant held, but it isn't enforced anywhere a peer can't reach: the publisher never registers subscribe ids (there is no dedupe in `recv_subscribe`), so two SUBSCRIBEs sharing an id and disagreeing on `Ordered` both run and put both directions under one id. `ordered(7, 1)` and `live(7, 2)` then each compare `Less` than the other, and `sort` is documented to be free to panic on a comparator that does that, which `set_ordered` calls. Equal groups with opposite directions also compared `Equal` while derived `Eq` said otherwise. The fix compares `ordered` before the group, so the group comparison is only ever reached by two groups that agree on its sense: redundant for well-formed input, and a total order for the rest. New test `test_ord_is_total_across_mixed_directions` checks antisymmetry, `Eq`/`cmp` agreement, and transitivity across a mixed set; it fails if the new comparison is removed.

**Rejected — "cross-subscription starvation is not actually fixed".** Accurate as a statement, but it is not this PR's defect. Starvation between equal-priority subscriptions is the status quo: before this branch the tie-break was the group sequence compared across subscriptions, so a long-running track with higher sequence numbers already dominated a freshly started one indefinitely. What this branch briefly introduced, and what the previous commit removed, is the *direction* deciding it. Which arbitrary key breaks the tie is the open `TODO` at the top of the file about round-robin between tracks of equal priority, and that needs the select-track-then-select-group restructure rather than a different comparator field. The repo owner has signed off on the subscribe-id tie-break as the interim behavior.

Full suite 1003 passed; `just check` clean.

(written by Fable 5)

Issue #2708, opened 2026-08-06
origin: open the announce interest lazily, on first consumer


kixelated:
A session wired to feed an origin (JS: the `subscribe` option in #2705; Rust: `Client::with_subscriber`) opens an announce-interest stream with an empty prefix as soon as it attaches, whether or not anything ever reads an announcement. On a large relay that is announce traffic and per-session announce state for every broadcast the session may see, paid even by an app that only publishes or only consumes known paths via requests.

The origin knows whether anyone is interested: an open `announced()` stream, an announcement-gated watch handle, or (narrower) which prefixes they cover. The forwarder could open the wire announce stream only while interest exists, and scope it to the union of requested prefixes rather than the root.

Applies to both implementations; neither is lazy today. The JS forwarder lives in `js/net/src/connection/forward.ts`; the Rust equivalent is the subscriber wiring in `rs/moq-net` session setup.

(Written by Fable 5)

kixelated:
## Design investigation

The origin already knows who wants announcements; it just never tells the session. Every `announced()` stream registers in the origin (JS: `#runAnnounced` over the reactive tables; Rust: `AnnounceConsumer` entries in the `NotifyNode` tree in `model/origin.rs`). The design is to make that knowledge observable and let the subscriber half of each session react to it, instead of soliciting the root eagerly at attach.

### Interest ledger

A refcounted map of prefix to consumer count inside the origin, maintained by the things that genuinely need announcements:

- `announced(prefix)` consumers, at their prefix (Rust derives this from the existing `NotifyNode` registrations, keyed by each consumer's root).
- Announcement-gated single-path handles (`announcedBroadcast` / `announced_broadcast`), as interest at the exact path. An ANNOUNCE_REQUEST whose prefix is the full path is legal and maximally narrow.
- On a relay, the publisher half serving a downstream peer's ANNOUNCE_REQUEST (`recv_announce`) holds an `announced()` consumer internally, so it registers automatically. That is what propagates interest hop by hop from the edge toward the origin server.

Plain `request(path)` / `request_broadcast` deliberately does NOT register: blind answers already serve requests without announcements, and "only consumes known paths via requests" is exactly the workload this issue wants to stop charging. A publish-only origin has zero entries and opens zero announce streams, which is the headline saving.

[...]

PR #2775, opened 2026-08-12
feat(js/net): solicit announcements lazily, per interested prefix

kixelated:
The JS half of #2708, mirroring #2762 in `rs/moq-net`. **Stacked on #2705** (targets `claude/js-origin-extract`, since `forward.ts` only exists there); retarget to `dev` once that merges.

## Problem

`forwardAnnounced` opened a root announce stream the moment a session attached, whether or not anything ever read an announcement. An app that only publishes, or only consumes known paths via requests, paid announce traffic and per-session announce state for every broadcast the relay could show it.

## Design

`Consumer.announced(prefix)` registers refcounted demand in a new `interest` map on `OriginState`, released when the stream closes. `forwardAnnounced` watches it and opens one wire stream per interested prefix, instead of one root stream up front. `announcedBroadcast` / `Announce.Broadcast` register automatically, since both go through `announced()`.

Two deliberate differences from the Rust side:

- **Requests are not interest.** A session answers a request with a blind subscription, which needs no announcement, so a request-only app stays silent on the wire and its requests still resolve. Rust has no blind-answer path, which is why `request_broadcast` there has to raise interest and wait.
- **Prefix scoping is real here.** The JS `announced(prefix)` API already carries a prefix, so a watcher of one room asks for `room/abc`, not the relay root. Rust solicits its auth-derived `allowed()` prefixes instead.

**Sticky once opened**, as in #2762: streams carry the routes, so closing one retracts the entries it fed and a consumer still holding that broadcast would watch it go offline and come back as a blind answer. Closing on idle needs the table to know which entries are still being read, which it does not.

**Overlapping streams are deduped per path**: a narrow prefix and a later broader one share one table entry, refcounted. Without that, a retraction from one stream would swap the front the other still provides, which reads as a republish and makes consumers re-subscribe for nothing.

Failure handling is unchanged: the first stream to end, cleanly or not, is discovery lost, which downgrades the attachment so gated watchers fall back to standing requests. With the usual single root stream that is exactly the old behavior.

## Tests

New in `forward.test.ts`: nothing is solicited until something listens (and a request alone does not count, but still resolves); the first listener opens exactly one stream and a second listener adds none; a narrow listener asks for its own prefix, a broader one opens its own stream, the shared path is one entry that survives until the last stream retracts it, and a nested prefix adds nothing.

The fake session now records solicited prefixes and hands out one producer per prefix, which is what makes the laziness observable. Existing tests that asserted on a populated table now hold an explicit listener: `forward.test.ts` (2), `integration.test.ts` (3), `reload.test.ts` (1). That edit is the behavior change made visible, not a workaround.

`just js test` and `just check` are green.

(written by Fable 5)

### DIFF
--- a/demo/web/src/index.ts
+++ b/demo/web/src/index.ts
@@ -79,7 +79,7 @@ const metaSignal = new Signals.Signal<unknown>(undefined);
 const relayUrl = new Signals.Signal<URL | undefined>(new URL(RELAY_URL));
 
 // Discovery connection (the tiles each open their own connection internally).
-const connection = new Net.Connection.Reload({ url: relayUrl, enabled: true });
+const connection = new Net.Connection.Shared({ url: relayUrl });
 
 // ---------------------------------------------------------------------------
 // Per-broadcast tile (a <moq-watch-ui> in the left column)
--- a/demo/web/src/index.ts
+++ b/demo/web/src/index.ts
@@ -382,8 +382,8 @@ ui.run((effect) => {
 	section.hidden = false;
 
 	// Report the transport negotiated by the live connection.
-	const conn = effect.get(connection.established);
-	$("network-transport").textContent = conn ? (conn.transport === "websocket" ? "WebSocket" : "WebTransport") : "";
+	const transport = effect.get(connection.transport);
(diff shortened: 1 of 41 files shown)


kixelated:
I think this is problematic for relays, as they have to solict announcements from every session lazy too.

PR #2785, opened 2026-08-12
fix(net): keep a live-edge subscriber at the live edge across a takeover

### LINKED ISSUE #2784: Live-edge subscriptions replay the outage after reconnect
## Problem

A live subscription opened with `subscribe(None)` / `group_start: None` is documented to start at the latest group. After a reconnect, the same subscription instead requests every group since the last group observed before the outage.

For live playback, this produces a burst of stale frames that downstream QoS drops while the subscriber catches up to the live edge. The longer the outage, the larger the burst, up to the publisher's retention window.

Observed with `moq-relay 0.14.8-c8b11b10b` after restarting the relay:

```
FIRST: Dropping frame due to QoS. start:0:07:19.485031770 deadline:0:07:19.485031770 earliest_time:0:07:31.148510106
...
LAST:  Dropping frame due to QoS. start:0:07:31.129998770 deadline:0:07:31.129998770 earliest_time:0:07:31.148510106
```

Approximately 12 seconds / 350-355 frames were replayed and dropped in about 3 seconds. Longer stop/start intervals produced thousands of dropped-frame messages.

## Expected behavior

A subscription whose `group_start` remains `None` should return to the live edge after reconnect, matching the documented meaning of `None`. It should not implicitly become a catch-up subscription.

If gapless catch-up across reconnects is required for other consumers, it should be an explicit policy rather than changing the meaning of the live-edge default.

## Root cause

This is resume slicing layered on top of subscription aggregation:

1. `moq_native::Client::consume` configures origin linger so a relay restart preserves the logical broadcast and subscription across reconnects.
2. `model/resume.rs::Producer::takeover` creates a new segment starting at one past the newest group observed on the previous segment.
3. `model/resume.rs::slice` maps `(prefs.group_start = None, segment.start = Some(n))` to `group_start = Some(n)`.
4. That explicit bound enters the track subscription aggregate.
5. The moq-lite subscriber forwards it upstream as `Subscribe.start_group`.
6. The moq-lite publisher uses the explicit start instead of `track.latest()`, replaying the missed range.
7. Hang media tracks advertise a 30-second retention window, so a substantial missed range can still be available.

The existing `model::origin_impl::tests::test_linger_reconnect_splices` regression test currently asserts this behavior: a subscriber initially created with `subscribe(None)` reconnects with `producer.subscription().group_start == Some(2)`. The focused test passes on current `main`, so the behavior remains present.

## Reproduction

1. Publish a Hang video track through a relay.
2. Consume it through a reconnecting moq-native client with the default `subscribe(None)`.
3. Let playback reach steady state.
4. Stop the relay for several seconds, then start it again within the reconnect/linger window.
5. Observe that the resubscription requests from the last group received before the outage and rapidly replays the missed media instead of starting from the latest group.

## Fix considerations

The public semantic and the existing reconnect promise currently conflict:

- `group_start: None` says to start at the latest group.
- Origin resume promises to splice a reconnect at the first missing group.

The default should likely preserve live-edge semantics through every segment. Consumers that need catch-up should opt into it explicitly. The implementation and API shape need to preserve both use cases without overloading `None`.

A regression test should cover a lingering/reconnecting default subscription with multiple groups produced during the outage and assert that only the live edge is delivered after reconnection.

(Written by GPT-5)

arielmol:
## Summary

- **Root cause**: `slice()` in `rs/moq-net/src/model/resume.rs` mapped `(prefs.group_start = None, segment.start = Some(n))` to `group_start = Some(n)`. A takeover boundary is range bookkeeping (which groups the new segment owns), but it was entering the demand aggregate as a *request*, so a subscriber that asked for the live edge came back from a route change asking the upstream for every group it missed. Measured: a ~5 s relay outage replayed ~5 s of stale groups, and the ~30 s moq-mux cache makes the worst case ~30 s.
- **Fix**: a takeover boundary never becomes a start. `(None, _) => None`. The range is still enforced, on the read side, where it always was: `poll_activate` puts the segment start on the read cursor, so a replacement route serving below the boundary is filtered rather than re-demanded.
- Subscribers with an explicit `group_start` are unaffected: theirs is still raised to the first missing group, so gapless catch-up stays available as the explicit opt-in it already was. `fetch_group` also still reaches the missed range.
- A live-edge subscription now means the same thing on every segment. That is what it already meant on the *first* one, which `broadcast_route_migration` has always asserted ("a live-edge subscription tunes in at the latest group").

## Public API changes

None. `slice` is private; no signature or exported item changed.

## Test plan

- `just check` and `just test` green (2438 passed).
- New `resume::test::takeover_after_produced_segment_keeps_live_edge`: fails on `main` with `Some(2)` vs `None`; also asserts a below-boundary group from the replacement is filtered.
- New `resume::test::takeover_keeps_an_explicit_start`: the other half of the contract, `Some(0)` raised to `Some(2)` across the takeover.
- `moq-native::broadcast::broadcast_route_migration` rewritten as the end-to-end regression: over a real session, the standby's stale cached group is no longer the first thing delivered after the migration (`["b2"]` before the fix, `["b3"]` after), and a group produced post-migration keeps flowing through the same subscriber handle.
- Origin route/failover/standby/reconnect tests now assert the live-edge demand and keep checking each track's boundary through the delivery filter.

## Cross-package sync

No wire format change (this only changes which `Start Group` we choose to send), so no draft update. `js/net` has no splice/resume model to mirror.

Fixes #2784

(written by Opus 5)

### DIFF
--- a/rs/moq-native/tests/broadcast.rs
+++ b/rs/moq-native/tests/broadcast.rs
@@ -709,6 +709,10 @@ async fn read_payloads(sub: &mut moq_net::track::Subscriber, count: usize) -> Ve
 /// preferred route (shorter hop chain) serves the track; when that session dies,
 /// the same `track::Subscriber` keeps receiving groups from the standby session.
 /// No unannounce is observed and nothing is resubscribed by the application.
+///
+/// The subscription is live-edge (`group_start: None`), and that means the same
+/// thing on the standby as it did on the first route: tune in at the latest
+/// group. The standby's older cached groups are not replayed.
 #[tracing_test::traced_test]
 #[tokio::test]
 async fn broadcast_route_migration() {
@@ -752,6 +756,8 @@ async fn broadcast_route_migration() {
 		)
 		.expect("create broadcast");
 	let mut track_b = broadcast_b.create_track("video", None).expect("create track");
+	// A clone to keep producing from the test body once the task owns the rest.
+	let mut standby_track = track_b.clone();
 	// B carries the continuation of the same content: groups 2 and 3.
 	for sequence in 2..4u64 {
 		let mut group = track_b
@@ -818,7 +824,8 @@ async fn broadcast_route_migration() {
 	assert_eq!(path.as_str(), "test");
 	let broadcast = broadcast.expect("expected announce");
 
-	// Subscribe once; a generous stale window so cached groups are served.
+	// Subscribe once, with a generous stale window: even that must not turn the
+	// live-edge default into a backfill when the route changes.
 	let subscription = moq_net::track::Subscription::default().with_latency_max(Duration::from_secs(10));
 	let mut sub = broadcast
 		.track("video")
@@ -832,9 +839,21 @@ async fn broadcast_route_migration() {
 	assert_eq!(read_payloads(&mut sub, 1).await, ["a1"]);
 
 	// Kill the serving session. The track migrates to B and the same subscriber
-	// keeps reading, resuming exactly at the first group A never delivered.
+	// keeps reading, tuning in at B's latest group. B's group 2 sits below the
+	// splice boundary and behind the live edge, so it is never replayed.
 	drop(session_a);
-	assert_eq!(read_payloads(&mut sub, 2).await, ["b2", "b3"]);
+	assert_eq!(read_payloads(&mut sub, 1).await, ["b3"]);
+
+	// The migrated subscription is live: what B produces from here on flows
+	// through the same subscriber handle.
+	{
+		let mut group = standby_track
+			.create_group(moq_net::group::Info { sequence: 4 })
+			.expect("create group");
+		group.write_frame(Timestamp::ZERO, b"b4".as_ref()).expect("write frame");
+		group.finish().expect("finish group");
+	}
+	assert_eq!(read_payloads(&mut sub, 1).await, ["b4"]);
 
 	// The application observed no unannounce or re-announce across the swap.
 	assert!(
--- a/rs/moq-net/src/lite/subscriber.rs
+++ b/rs/moq-net/src/lite/subscriber.rs
@@ -1556,10 +1556,11 @@ impl<S: web_transport_trait::Session> TrackServe<S> {
 
 	/// Open the upstream SUBSCRIBE and start routing groups into the producer.
 	///
-	/// The subscription's bounds come straight from the demand aggregate: when this
-	/// segment resumes a track after a route change, the origin's slice already
-	/// carries the boundary as `group_start`, so the upstream delivers exactly the
-	/// missing range.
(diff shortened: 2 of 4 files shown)


arielmol:
Reduced to 21 warns
```
0:00:20.851280732 2898952 0x7373c8000b70 WARN            videodecoder gstvideodecoder.c:3661:gst_video_decoder_clip_and_push_buf:<avdec_h264-0> Dropping frame due to QoS. start:0:00:20.469901572 deadline:0:00:20.469901572 earliest_time:0:00:21.152567674
...<more...
0:00:20.931486204 2898952 0x7373c8000b70 WARN            videodecoder gstvideodecoder.c:3661:gst_video_decoder_clip_and_push_buf:<avdec_h264-0> Dropping frame due to QoS. start:0:00:21.137235572 deadline:0:00:21.137235572 earliest_time:0:00:21.152567674

```

kixelated:
Taking this over to land it. Thanks for the diagnosis, it was right, and the `slice()` root cause is exactly where the bug was.

I reverted the `latency_max` gate from 7115aa6, because I measured its premise and it does not hold. The commit message reasoned that keeping the boundary is safe when a window is set, since "the publisher's `latency_max` enforcement caps the backfill to that window". It does not: the publisher serves an explicitly-requested start range in full, regardless of the budget. Two measurements on unmodified `main`:

- `broadcast_route_migration` with `latency_max: Duration::ZERO` (instead of its usual 10 s) still delivers `b2` and `b3`. So the 10 s window in that test was decorative all along, not what was making the backfill work.
- On 7115aa6 with `latency_max: Duration::from_millis(1)`, the test still passes, i.e. the superseded cached group is delivered even though it is orders of magnitude older than the window.

So the gate is not a window, it is an on/off switch: `ZERO` gets the live edge, and any non-zero value replays the *entire* outage. That leaves the reported bug in place for every jitter-buffered subscriber, since a non-zero reordering tolerance is the normal setting for playback, and it overloads a reordering knob with catch-up policy, which is the overloading #2784 asked us to avoid.

The unconditional version needs one thing the first pass was missing: `broadcast_route_migration` fails against it, because it asserted the standby replaying its cached group 2. That is not incidental damage, it is the same contract change one layer up, and the test's own first assertion already says a live-edge subscription tunes in at the latest group. It now asserts the standby's live edge instead, plus a group produced after the migration to prove delivery stays live. That makes it the end-to-end regression test #2784 asked for: `["b2"]` before the fix, `["b3"]` after.

Also corrected the doc comments that described the boundary as something the resumed segment requests upstream (module doc, `slice`, `takeover`, and the lite subscriber's `establish`), and added `takeover_keeps_an_explicit_start` so the catch-up half of the contract is pinned too.

`just check` and `just test` are green (2438 passed).

(written by Opus 5)

arielmol:
I just confirmed on my gst setup, it works

PR #2805, opened 2026-08-13
feat(libmoq): declare the catalog container for manually authored renditions

arielmol:
A C publisher can author frames on a raw track (`moq_publish_track`) already encoded per draft-ietf-moq-loc, or as CMAF fragments, but the catalog write path always declared the legacy container, so there was no way to tell subscribers what the track actually carries.

`moq_video_config` and `moq_audio_config` now carry a `moq_container`: a `moq_container_kind` discriminant plus the CMAF init segment. The kind covers every container hang knows (`LEGACY`, `CMAF`, `LOC`, `UNKNOWN`) and crosses the ABI as `uint32_t`, so an unrecognized value returns `InvalidCode` instead of constructing an invalid Rust enum. A zeroed struct means legacy, so a recompiled caller that ignores the field behaves exactly as before. CMAF needs its init segment, which is why the kind travels in a struct rather than as a bare parameter.

`UNKNOWN` is read-only. A catalog written by a future build round-trips into it so a consumer can skip the rendition, but publishing one is `InvalidCode`: none of the original JSON survives the trip through C, so there is nothing to write back.

Because both structs are shared with the consume path, `moq_consume_video_config` and `moq_consume_audio_config` now report the container too, borrowing the CMAF init segment from the catalog snapshot exactly like `description` already does. `moq_container_kind` goes in the cbindgen `ENUMS` list, since no signature reaches it and a C caller would otherwise have to hardcode the integers.

**C callers must recompile**: both config structs grew a field.

Out of scope: `moq_publish_media_*` still writes the legacy container; switching the moq-mux encoder outputs to LOC is a separate change.

### The red CI was not this PR

`just check` failed in `moq-video`, which nothing here touches. `I420::from_rgb` and `from_yuyv` are `#[cfg(target_os = "linux")]`, but their only callers live in `capture/v4l2.rs`, which the `capture` feature gates, and every workspace consumer of moq-video leaves that feature off. So on Linux the two functions compiled with no callers and `-D warnings` turned the resulting `dead_code` lint into a hard error, which fails any branch whose diff selects libmoq. The second commit matches the cfg to the callers, test included.

### Tests

- a LOC frame published on a raw track with a LOC declaration and consumed through `moq_consume_video`: the catalog reports LOC, and payload, timestamp and keyframe survive the round trip
- a CMAF audio rendition: the init segment comes back byte for byte through `moq_consume_audio_config`
- the rejected kinds: `UNKNOWN`, an out-of-range value, and CMAF with no init segment

(written by Opus 5)

### DIFF
--- a/rs/libmoq/src/api.rs
+++ b/rs/libmoq/src/api.rs
@@ -6,6 +6,101 @@ use std::str::FromStr;
 
 use tracing::Level;
 
+/// How a media track's frames are wrapped, independent of the codec.
+///
+/// The ABI carries this as a `uint32_t`, so an unknown discriminant from C is an
+/// error rather than UB.
+#[repr(C)]
+#[allow(non_camel_case_types)]
+#[derive(Clone, Copy, Debug)]
+pub enum moq_container_kind {
+	/// A QUIC VarInt timestamp prefix followed by the raw codec payload.
+	/// Timestamps are in microseconds.
+	MOQ_CONTAINER_KIND_LEGACY = 0,
+	/// Fragmented MP4: each frame is a complete moof+mdat fragment, described by
+	/// the init segment in `moq_container::init`.
+	MOQ_CONTAINER_KIND_CMAF = 1,
+	/// Low Overhead Container (draft-ietf-moq-loc): a small property block
+	/// followed by the codec payload.
+	MOQ_CONTAINER_KIND_LOC = 2,
+	/// A container this build does not recognize, so the rendition must be
+	/// ignored. Only ever read out of a catalog: publishing it is an error.
+	MOQ_CONTAINER_KIND_UNKNOWN = 3,
+}
+
+/// The container of a video or audio rendition, plus whatever that container
+/// needs to describe itself.
+///
+/// Zeroing this struct means `MOQ_CONTAINER_KIND_LEGACY` with no init segment,
+/// which is what a rendition written by [moq_publish_media] carries.
+#[repr(C)]
+#[allow(non_camel_case_types)]
+#[derive(Clone, Copy)]
+pub struct moq_container {
+	/// `moq_container_kind` discriminant.
+	pub kind: u32,
+
+	/// The CMAF init segment (ftyp+moov), or NULL.
+	/// Read only when `kind` is `MOQ_CONTAINER_KIND_CMAF`, where it is required.
+	pub init: *const u8,
+	pub init_len: usize,
+}
+
+impl Default for moq_container {
+	fn default() -> Self {
+		Self {
+			kind: moq_container_kind::MOQ_CONTAINER_KIND_LEGACY as u32,
+			init: std::ptr::null(),
+			init_len: 0,
+		}
+	}
+}
+
+/// # Safety
+/// - `container->init` must point to `container->init_len` bytes when
+///   `container->kind` is `MOQ_CONTAINER_KIND_CMAF`.
+pub(crate) unsafe fn parse_container(container: &moq_container) -> Result<hang::catalog::Container, Error> {
+	use hang::catalog::Container;
+
+	Ok(match container.kind {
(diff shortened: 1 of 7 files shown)


rs/libmoq/src/api.rs
```diff
@@ -1729,6 +1715,26 @@ pub unsafe extern "C" fn moq_publish_video_properties(broadcast: u32, properties
...
+///
+/// LOC is intended for tracks created with [moq_publish_track] whose frames
+/// have already been encoded according to draft-ietf-moq-loc.
+///
+/// Returns a zero on success, or a negative code on failure.
+///
+/// # Safety
+/// - Same contract as [moq_publish_video_config].
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn moq_publish_video_config_with_container(
```
kixelated: definitely no `_with_container` method. add it to moq_video_config instead.

rs/libmoq/src/api.rs
```diff
@@ -1709,6 +1673,28 @@ pub unsafe extern "C" fn moq_publish_video_properties(broadcast: u32, properties
...
+/// The ABI accepts this as a `uint32_t`, so unknown values return
+/// `Error::InvalidCode` instead of constructing an invalid Rust enum.
+#[repr(C)]
+#[allow(non_camel_case_types)]
+#[derive(Clone, Copy, Debug)]
+pub enum moq_media_container {
+	/// A QUIC VarInt timestamp prefix followed by the codec payload.
+	MOQ_MEDIA_CONTAINER_LEGACY = 0,
+	/// Low Overhead Container (draft-ietf-moq-loc).
+	MOQ_MEDIA_CONTAINER_LOC = 1,
```
kixelated: cmaf missing lul

rs/libmoq/src/api.rs
kixelated: Done. Both `moq_video_config` and `moq_audio_config` now carry a `moq_container` (kind + CMAF init segment), and the `_with_container` entry points are gone. A zeroed struct is legacy, so a recompiled caller that ignores the field behaves as before.

Since the structs are shared with the consume path, this also means `moq_consume_video_config` / `moq_consume_audio_config` report the container, which they previously dropped on the floor.

(written by Opus 5)

rs/libmoq/src/api.rs
kixelated: Added, along with `UNKNOWN`, so the kind now covers everything `hang::catalog::Container` has.

CMAF is the reason the kind travels in a struct rather than as a bare `uint32_t`: it carries an init segment. `moq_container::init` / `init_len` is copied into the catalog on publish and borrowed back out of the snapshot on consume, exactly like `description`. An empty init is rejected at the publish call rather than at every subscriber.

`UNKNOWN` is read-only: a rendition written by a future build round-trips into it so a C consumer can skip it, but publishing one is `InvalidCode` since we keep none of the original JSON, so there is nothing to write back.

(written by Opus 5)

rs/libmoq/src/api.rs
```diff
@@ -1673,6 +1673,28 @@ pub unsafe extern "C" fn moq_publish_video_properties(broadcast: u32, properties
...
+}
```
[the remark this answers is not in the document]
kixelated: Half taken: the enum (now `moq_container_kind`) is in `build.rs`'s `ENUMS` list, so the header carries the named constants instead of making C hardcode the integers.

Skipping `#[non_exhaustive]`. `rs/CLAUDE.md` makes it opt-in rather than the default, and none of the sibling ABI enums (`moq_video_codec`, `moq_audio_format`, `moq_video_encoder_kind`) carry it. They all cross the ABI as `uint32_t` and are matched by comparing discriminants, so there is no external Rust `match` for a new variant to break.

(written by Opus 5)

kixelated:
I can merge it now but IMO we need getters/setters for structs to avoid semver issues. Since this is a breaking change unfortunately.

PR #2814, opened 2026-08-13
feat(wasm): bind moq-net's full model, not just the consume path

kixelated:
Stacked on #2811 (the ALPN fix), which is what makes any of this reachable in a browser. Review that one first; this PR retargets to `main` once it lands.

## Why

`moq-wasm` bound 4 classes and 8 methods: connect, consume a broadcast, subscribe, read groups and frames. No publish, no discovery, no subscription options, no track properties. Anything a browser wanted beyond "read the live edge of one track" had to fall back to the hand-written TypeScript in `@moq/net`, which is the thing this crate exists to replace.

It drifted because nothing notices. The crate is `#![cfg(target_arch = "wasm32")]`, so `cargo check --workspace` compiles it to nothing on a host target; `just rs wasm` is a *compile* gate, which catches a `moq-net` change that breaks an existing binding but says nothing about one the binding never grew; and no test in the repo exercises it. It isn't in the `moq-ffi` sync row either, so it falls off the radar exactly when `moq-net` gains something.

## What

Bind the rest of the model, in modules mirroring moq-net's role modules one-for-one:

| module | JS classes | mirrors |
|---|---|---|
| `session` | `Session` | `moq_net::Session` + `origin::{Producer, Consumer}` |
| `broadcast` | `BroadcastProducer`, `BroadcastConsumer` | `moq_net::broadcast` |
| `track` | `TrackProducer`, `TrackConsumer`, `TrackSubscriber`, `TrackRequest` | `moq_net::track` |
| `group` | `GroupProducer`, `GroupConsumer` | `moq_net::group` |
| `announce` | `AnnounceConsumer`, `Announce` | `moq_net::announce` |
| `options` | `Subscription`, `TrackInfo`, `Frame` | the plain-data types |

New capability: publishing (`session.publish` → producer → tracks → groups → frames), discovery (`announced(prefix)`, `announcedBroadcast`), on-demand serving (`requestedTrack` / `accept` / `reject`), `fetchGroup`, subscription options (priority, ordered, latency, group start/end) with live `update`, track properties, subscriber cursors (`startAt` / `endAt`), and frame timestamps.

Deliberately absent: relay-side concerns (routes, hops, cost, origin scoping, cluster identity) have no browser caller, and the `poll_*` variants have no JS equivalent of a `kio::Waiter`.

## Anti-drift

The mirroring is the point, and it needs a rule to go with it. `rs/CLAUDE.md` and the root Cross-Package Sync table now name `moq-wasm` as a binding to update alongside `moq-net`, the way a `moq-ffi` change ripples into `libmoq` and the language wrappers. The module-per-role layout is what makes "read the two side by side and spot the gap" a workable review step, since no automated gate can do it.

## The wasm-bindgen trap this surfaced

Types carry the role as a prefix (`TrackProducer`, not `track::Producer`), against the naming rule in `rs/CLAUDE.md`. **wasm-bindgen resolves a type in a signature by its Rust ident alone: it ignores the module path and `js_name` both.** With role modules each exporting a `Consumer`, the first cut generated typings that were quietly wrong:

```ts
consume(path: string): Promise<GroupConsumer>;      // actually a BroadcastConsumer
readonly broadcast: GroupConsumer | undefined;      // actually a BroadcastConsumer
track(name: string): GroupConsumer;                 // actually a TrackConsumer
```

It compiled and ran; only the `.d.ts` lied. Unique idents are the fix. The modules stay private and re-export flat, so nothing reads `broadcast::BroadcastProducer`. Documented in the crate docs and `rs/CLAUDE.md` so the next person doesn't rediscover it.

## Boundary conventions

Chosen to match what a JS caller already expects from `@moq/net`:

- Durations in milliseconds, timestamps in microseconds.
- Sequences as `bigint` (they are `u64` on the wire).
- Option bags as classes, so a new knob is an added field rather than a changed signature.
- Closing takes an application close code (`Error::App`); a JS `Error` has nothing to map onto the wire.
- `closed()` rejects rather than resolves, because every close carries a reason.
- One in-flight async call per handle. wasm-bindgen async methods take `&self` and must produce `'static` futures, so a handle moves its value out for the call's duration (`util::Exclusive`); a re-entrant call errors instead of aliasing.

## Verification

Driven in a real browser against a relay, twice over. Same script, once against a default relay (`moq-lite-05`) and once against one pinned with `--server-version moq-transport-19`; both completed every step with no errors:

1. Two sessions connect.
2. Publisher `session.publish("test/room")`, `createTrack("video")` with a microsecond timescale, priority 3, ordered.
3. Subscriber's `announced("test")` yields `path: "room"` with a live broadcast.
4. `subscribe` with priority 5 / latencyMax 2000; reported track info comes back with the publisher's timescale, priority, and ordered flag intact.
5. Write a 3-frame group, read it back: payloads and timestamps (1000/2000/3000µs) survive the round trip.
6. `requestedTrack` fires for an unsubscribed name; `accept` serves it and the frame arrives.
7. `fetchGroup(0n)` returns the past group.
8. `broadcast.close()` shows up as an announce with a null broadcast.

`just check`, `just test`, and `just fix` are clean.

## Still no test harness

The crate is wasm32-only and nothing in the repo runs a browser, so `just rs wasm` compiling it remains the entire automated gate. Everything above was verified by hand. A headless-browser runner (`wasm-bindgen-test` + a relay fixture) is the obvious follow-up and would turn this PR's manual script into a regression test; it is a bigger change than this one and I did not attempt it here.

## Cross-Package Sync

No `moq-net` wire or API change, so no `js/net`, `doc/`, or draft updates. The `@moq/wasm` and `rs/moq-wasm` READMEs are updated for the new surface.

(written by Opus 5)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -68,6 +68,8 @@ Top-level layout only. Per-crate and per-package detail lives in the nested guid
 
 `rs/moq-ffi` is the single UniFFI core that every non-Rust binding is generated from. The wrappers under `/py`, `/swift`, `/kt`, and `/go` are thin layers over it, and `rs/libmoq` exposes the same core as a C staticlib. So one `moq-ffi` change ripples out to all of them (and their docs) per the [Cross-Package Sync](#cross-package-sync) table. CI mirrors the Swift and Go source packages to their external repos; Kotlin publishes `dev.moq:*` artifacts to Maven Central. For Python, most callers want the ergonomic `moq-rs` wrapper rather than the generated `moq-ffi` bindings directly.
 
+`rs/moq-wasm` is the browser binding and belongs to the same set even though it does not go through `moq-ffi`: UniFFI is C-ABI only, so the browser gets its own wasm-bindgen surface over `moq-net`. **Treat it as one more binding to keep in sync, not as an experiment that can lag.** Nothing compiles it against `moq-net`'s current API except `just rs wasm`, and no test exercises it at all, so a `moq-net` method the browser needs is simply missing until someone notices. When you add or change a `moq-net` API a browser caller would use, mirror it in `rs/moq-wasm` in the same PR, the same way a `moq-ffi` change ripples into `libmoq` and the language wrappers.
+
 ## Per-Directory Guides
 
 Language-specific conventions, crate/package maps, and patterns live in nested `CLAUDE.md` files that load automatically when you work under that directory. Before writing code in one of these areas, read its guide (your editor loads it for you, but check it explicitly if you are reasoning about the area without opening a file in it):
@@ -172,7 +174,7 @@ Changes in one area usually need matching updates elsewhere, including docs. If 
 | Change in | Also update |
 |---|---|
 | `rs/moq-ffi` | `rs/libmoq`, `{py,swift,kt}/`, `go/wrapper/moq/*.go` (the `go/ffi` bindings regenerate automatically, but a new method needs a hand-written wrapper too, like `py/moq-rs`), `doc/lib/{py,swift,kt,go,c}` |
-| `rs/moq-net` wire/API | `js/net`, `doc/concept`, `drafts/draft-lcurley-moq-lite.md` (if the wire spec changes) |
+| `rs/moq-net` wire/API | `js/net`, `rs/moq-wasm` (the browser binding; nothing but `just rs wasm` compiles it, and that only catches a break, never a gap), `doc/concept`, `drafts/draft-lcurley-moq-lite.md` (if the wire spec changes) |
 | `rs/hang` catalog/container | `js/hang`, `doc/concept`, `drafts/draft-lcurley-moq-hang.md` (if the format spec changes) |
 | `rs/moq-token` | `js/token` |
 | `rs/moq-stats` wire (track names, frame shapes) | `doc/bin/relay/config.md` (stats section) |
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5165,8 +5165,11 @@ dependencies = [
  "getrandom 0.4.3",
  "js-sys",
  "moq-net",
+ "serde",
+ "serde-wasm-bindgen",
  "thiserror 2.0.20",
  "tracing-wasm",
+ "tsify",
  "url",
  "wasm-bindgen",
  "wasm-bindgen-futures",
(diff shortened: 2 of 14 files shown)


kixelated:
Closing in favor of #2907.

This PR is a second hand-written copy of a surface `rs/moq-ffi` already exports: `MoqBroadcast`/`MoqTrack`/`MoqGroup` producers and consumers, `MoqAnnounced`, `MoqTrackRequest`, `MoqSubscription`, `fetchGroup`, `requestedTrack`, `used`/`unused`. The duplication is the reason this kept regressing, and it is the thing worth removing rather than the individual defects.

The lifecycle model is the concrete version of that. `Exclusive` moves the value out for the duration of a call; `moq-ffi/src/ffi.rs::Task<T>` is `Arc<Mutex<T>>` plus a `cancel` watch channel, so concurrent calls queue instead of erroring, `cancel()` interrupts a pending call, `Drop` cancels, and `SubscriberControl` and `info` sit outside the lock so `update()` works during a pending `recv_group()`. Four review rounds here rediscovered pieces of that design, and rounds 3 and 4 each shipped a fix that was itself wrong. Two of round 4's findings are still unfixed at `9c6f4376f`: the `u64` to TS `number` mapping traps above 2^53, and `options::timestamp` truncates before the timescale conversion.

A spike (details and numbers in #2907) measured the alternative rather than assuming it:

- `moq-ffi` compiles clean to wasm32 after gating, with 82 of 160 exported methods surviving and the native build untouched. uniffi 0.31's `wasm-unstable-single-threaded` feature is what makes it possible, since it drops the `Send` bounds moq-net's wasm futures cannot satisfy.
- The generated TypeScript maps `u64` to `bigint`, and its async calls clone the object handle instead of borrowing the wrapper, so `free()` during a pending call is fine. Those are exactly this PR's two hardest bugs, absent by construction rather than by remembering to fix them.
- Size is not a reason to prefer the hand-written path: about 8% more bytes for roughly ten times the exported surface.

The `claude/wasm-api-parity` branch stays, so nothing here is lost. Two things from this PR are worth keeping regardless of the approach and are carried into #2907: the anti-drift rule naming `moq-wasm` in the Cross-Package Sync table, and the wasm-bindgen ident trap written up in `rs/CLAUDE.md`, which turns out to have a UniFFI sibling (`#[cfg]` on a method inside a `#[uniffi::export] impl` block is silently ignored).

#2816 remains the root cause behind all four of those review rounds being hand-verified, and it now blocks the last step of #2907.

(written by Opus 5)

PR #2818, opened 2026-08-13
fix(js/net): honor frame floor for a held group

### LINKED ISSUE #2807: js/net: a frame-precise raised start floor is not honored for a group already in hand
The lite publisher's floor rejection in `js/net/src/lite/publisher.ts` (`#runTrack`) compares `group.sequence` only, but on lite-06 the start floor is a `(startGroup, startFrame)` position. A SUBSCRIBE_UPDATE that raises the floor to a frame *within* a group the serving loop already holds clears the sequence check, and the group is then served from the frame range snapshotted when it was taken.

Reproduced against `claude/js-publisher-cap-race` with a draft-06 subscription (group 0 parks the serving loop, so the armed prefetch takes group 1):

- initial `startGroup: 0`, group 1 written with frames `["a", "b", "c"]`
- SUBSCRIBE_UPDATE raises the floor to `startGroup: 1, startFrame: 2`
- **served:** `frameStart: 0`, payloads `["a", "b", "c"]`
- **requested:** `frameStart: 2`, payloads `["c"]`

Pre-existing rather than introduced by #2796: the group-range check that PR removed was `sequence < bounds.startGroup`, also group-only. But it does bound that PR's rationale, which is why it is filed here. Raising the floor is destructive on the read cursor (`startAt` shifts and closes buffered groups below it), so dropping a below-floor group already in hand discards nothing the subscriber has not discarded itself. That argument holds for whole groups only, and the comment in `#runTrack` now says so.

The fix is to compare the position rather than the sequence: when the raised floor names this group, lift the snapshotted `start` to the new `startFrame` instead of accepting it unchanged. Worth a draft-06 regression covering both a floor raised past a held group and one raised into it.

Rust is not affected. `position_group` applies the offsets synchronously at the pop, with control polled first, so it never holds a group under stale bounds.

Related: #2796

(written by Opus 5)

kixelated:
## Summary

- Intersect a prefetched group's snapshotted frame start with the current raised start floor when both name the same group.
- Preserve the existing destructive whole-group floor behavior while preventing frames below a newer intra-group floor from reaching the wire.
- Add deterministic draft-06 regressions for a floor raised past a held group and a floor raised into it.
- Bump `@moq/net` from `0.3.0` to `0.3.1` for the shipped behavioral fix.

Root cause: the serving loop snapshots a group's frame range when `recvGroup()` hands it over, then can remain parked while `SUBSCRIBE_UPDATE` raises the start floor. The later whole-group rejection only compared `group.sequence`, so a floor raised into the held group accepted the stale `start: 0` snapshot and served frames the subscriber had discarded. The fix intersects that snapshot with the current frame floor at the serving handoff. The end cap remains snapshot-based because lowering it is recoverable and must not discard a prefetched group.

Closes #2807.

## Public API changes

None. This changes internal lite publisher range enforcement only.

## Test plan

- `bun test js/net/src/lite/publisher.test.ts` (27 passed)
- `nix develop --command just test` (including 465 `@moq/net` tests)

No cross-package or draft update is required because the public API and wire format are unchanged.

(Written by GPT-5)

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -182,7 +182,7 @@
     },
     "js/net": {
       "name": "@moq/net",
-      "version": "0.3.0",
+      "version": "0.3.1",
       "dependencies": {
         "@moq/qmux": "^0.3.2",
         "@moq/signals": "workspace:*",
--- a/js/net/package.json
+++ b/js/net/package.json
@@ -1,7 +1,7 @@
 {
 	"name": "@moq/net",
 	"type": "module",
-	"version": "0.3.0",
+	"version": "0.3.1",
 	"description": "The networking layer for Media over QUIC: real-time pub/sub with built-in caching, fan-out, and prioritization.",
 	"license": "(MIT OR Apache-2.0)",
 	"repository": "github:moq-dev/moq",
--- a/js/net/src/lite/publisher.test.ts
+++ b/js/net/src/lite/publisher.test.ts
@@ -573,8 +573,8 @@ test("lite draft-05: a group taken before the cap dropped is still served", asyn
 // by design, since the read cursor shifts and closes every buffered group below it, so a group
 // already in hand has to go the same way: serving it would re-deliver below a floor the
 // subscriber just moved, which on a route splice is duplicate media.
-test("lite draft-05: a group taken before the floor rose is dropped", async () => {
-	const sub = await servedSubscription({ startGroup: 0, gated: true });
+test("lite draft-06: a group taken before the floor rose past it is dropped", async () => {
+	const sub = await servedSubscription({ version: Version.DRAFT_06, startGroup: 0, gated: true });
 	try {
 		// Same window as the cap test: group 0 parks the loop, so the prefetch takes group 1.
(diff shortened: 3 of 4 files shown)


kixelated (commented):
No findings.

The held-group fix preserves the intended asymmetric update semantics: a floor past the group drops it, a floor within the group intersects with the snapshotted start via `Math.max`, and a later relaxation cannot widen the range captured when the group left the cursor. The snapshotted end remains intact, including when both bounds name the same group. The deterministic gated tests cover both whole-group and intra-group raises and fail against the base behavior.

No exported API or wire-format changes were found. The `@moq/net` `0.3.0` to `0.3.1` patch bump and matching lockfile update are appropriate. Exact-head focused tests and GitHub `Check` passed.

Coverage limits: this review did not add fixtures for repeated updates while parked, updates after `#runGroup` launches, or a raised start beyond a completed group's length. None refutes the reported held-group race or this fix.

MERGE

(Written by GPT-5)

PR #2823, opened 2026-08-13
fix(moq-mux): stop publishing audio spliced across a TS discontinuity

### LINKED ISSUE #2802: moq import ts: a splice publishes one frame of mixed bytes before resyncing
### Summary

At a splice, `moq import ts` publishes one frame of mixed bytes as if it were real audio. The resync landed in #2751 recovers on the *next* frame, so this is bounded to a single frame, but that frame is silently wrong rather than dropped.

This is a known, deliberate gap in the confirm-before-publish rule from #2751, split out so the trade-off can be decided on its own terms.

### Mechanism

A frame split across a PES boundary is carried in `LegacyStream::tail` / `AacStream::tail` and joined onto the next PES payload. At a splice (a looping file wrapping mid-frame, or a dropped PES) the two halves are unrelated, so the join is garbage.

The demuxer can't tell, because that offset is one a previous frame vouched for:

- `Resync::unconfirmed` is false there, since the preceding frame published normally and cleared it via `recovered()`.
- The tail still begins with an intact header, which parses and declares a plausible length.
- So the frame is published without confirmation, made of pre-splice bytes plus however many post-splice bytes the declared length reaches.

Sync loss is only detected on the frame *after* it, which fails to parse and triggers the scan.

`legacy_resyncs_past_stale_tail_at_a_splice` in `rs/moq-mux/src/container/ts/import.rs` pins the current behavior, and says in a comment that it documents it rather than asserting it is correct.

### Why it wasn't fixed in #2751

Confirmation there covers offsets nothing has vouched for: one reached by a scan, the first frame of a stream, and the first after a `seek`. Extending it to *every* frame would catch this, at the cost of holding each frame until its successor arrives to confirm the boundary. On a live path that is a frame of added latency on every frame, which is a real cost to weigh against one corrupt frame per splice.

### Related limit

Confirmation catches a false sync whose declared length *misses* a real frame boundary. One that happens to land on a real boundary is indistinguishable from a genuine frame by this rule alone. The tests in #2751 build fragments that deliberately do not align, for exactly that reason.

### Options

1. Leave it. One corrupt frame per splice, recovered on the next frame.
2. Confirm every frame, paying one frame of latency throughout.
3. Confirm only frames assembled from a carried tail, which is where splices actually land. Cheaper than (2) since split frames are the minority, but it needs the end-of-stream drain to accept a final unconfirmed frame (that path already exists, gated on the stream having published something).

(3) looks like the best ratio to me, but it is a judgement call about latency versus correctness that belongs with whoever owns the ingest path.

Follows #2729, fixed by #2751.

(written by Opus 5)

kixelated:
## Summary

At an MPEG-TS splice, `moq import ts` published audio assembled from both sides of the cut, and swallowed the real frame underneath it. Two mechanisms, found in that order.

**1. A frame joined out of a carried tail.** The confirm-before-publish rule from #2751 keyed on `Resync::unconfirmed`, which a published frame clears. What that frame vouched for is where the tail *begins*, not the bytes the next PES joins onto it, so the join was trusted; the stale tail still opens with an intact header declaring a plausible length, so nothing else caught it. Confirmation now also covers a frame beginning in a carried tail. A frame that begins inside a PES is whole by the time it is parsed, so the common path pays no latency. `accept_unconfirmed` becomes `drain()`, marking end of stream so the existing drain still publishes a final whole frame no successor can confirm.

**2. A PES completed across a continuity break.** The above only covers a carried tail, and a real looping mux does not produce one. It wraps with its last PES still open and short of its declared length, so the next loop's leading packets complete it: `handle_pes_continuation` appended them without ever consulting the continuity counter, and the codec received one buffer straddling the cut with nothing carried, so the tail rule never got a say. `SectionReassembler` already handled this for sections; that logic moves to a shared `Continuity` and now covers PES PIDs too. A break flushes the truncated PES, so whole frames it did deliver still publish, then desyncs the stream so it resumes at the next PES start.

Reported by @t0ms on this PR, measured against a looped broadcast capture where the audio was byte-identical before and after (1), and independently reproduced in-tree with a packet-level fixture.

**3. What a break is allowed to salvage.** Flushing the truncated PES is audio's reasoning: many independently decodable frames per PES, so the ones ahead of the cut are correct on their own. It does not generalize. One video PES is exactly one access unit, so flushing publishes a picture with missing slices, and a keyframe missing slices stays wrong for every picture referencing it. Opus carries many packets per PES but its framing is declared rather than self-describing, so a truncated trailing packet is a parse error that would travel out of `decode` and end the session. Video, Opus, and verbatim streams now discard the partial.

**4. What a break must not destroy.** Extracting the shared `Continuity` lost `SectionReassembler`'s early return on `transport_error_indicator`, so a packet flagged corrupt had its payload processed after the partial was dropped. TEI now has its own result and the packet dies with the partial.

## Public API changes

None. `Resync`, `Continuity`, `AacStream`, and `LegacyStream` are private to moq-mux's TS importer.

## Test plan

Each new test was verified to fail when its own fix is reverted.

- `legacy_confirms_a_frame_joined_out_of_a_carried_tail` and its AAC mirror: without (1), publish pre-splice bytes spliced onto post-splice bytes.
- `legacy_drops_a_pes_completed_across_a_continuity_break` and its AAC mirror: at the TS packet layer, since the PES layer cannot express this shape.
- `video_drops_a_partial_access_unit_across_a_continuity_break`: without (3), a 28-byte keyframe with a truncated slice reaches the track ahead of the whole 32-byte one.
- `legacy_drains_a_joined_frame_at_end_of_stream` and its AAC mirror: guard the end-of-stream drain.
- `tei_pusi_section_is_dropped`: without (4), a section is emitted from a packet flagged corrupt.
- Every pre-existing section-reassembly test still passes over the extracted `Continuity`, unchanged.
- `cargo nextest run`: 506 green for moq-mux, 842 across moq-mux and its dependents. clippy `-D warnings` and rustdoc `-D warnings` clean.

**Coverage caveat.** None of the ten in-tree TS fixtures contains a transport error, discontinuity indicator, or continuity gap on an elementary stream PID, so the break path is inert on them by construction and its only coverage is the synthetic packet-level tests above. Verifying it against real content containing a real splice needs an oracle built from the source's de-packetized elementary stream, which is what @t0ms's harness does.

## Out of scope, filed

- #2849: a truncated or spliced Opus PES ends the session, reachable without any continuity break and present on `main`.
- #2851: a duplicate packet with a refreshed PCR is misread as a continuity gap. The obvious fix, keying on the repeated counter, is unsafe and was reverted; the issue records why.

## Cross-package sync

No row applies: no wire format, catalog, or public API change, and nothing under `/doc` describes TS resync or continuity handling.

Closes #2802.

(written by Opus 5)

### DIFF
--- a/rs/moq-mux/src/container/ts/import.rs
+++ b/rs/moq-mux/src/container/ts/import.rs
@@ -66,6 +66,8 @@ pub struct Import<E: catalog::Catalog = ()> {
 	streams: HashMap<Pid, Stream<E>>,
 	/// In-progress PES reassembly, keyed by elementary PID.
 	pending: HashMap<Pid, Pending>,
+	/// Per elementary-stream-PID TS continuity state.
+	continuity: HashMap<Pid, Continuity>,
 	/// True once a PMT with at least one supported stream has been parsed.
 	initialized: bool,
 	/// Raw 90 kHz PTS of the first audio frame in the current consecutive run.
@@ -138,6 +140,7 @@ impl<E: catalog::Catalog> Import<E> {
 			pmt_pids: HashSet::new(),
 			streams: HashMap::new(),
 			pending: HashMap::new(),
+			continuity: HashMap::new(),
 			initialized: false,
 			audio_burst: None,
 			scratch: Vec::new(),
@@ -218,6 +221,30 @@ impl<E: catalog::Catalog> Import<E> {
 			if self.supports_mpegts && catalog::SI_PIDS.iter().any(|(p, _)| *p == pid) {
 				self.si_section(pid, &pkt);
 				continue;
+			}
+			// An elementary stream's PES is as vulnerable to a break as a section is. A
+			// looping publisher wraps with its last PES still open and short of its declared
+			// length, so the next loop's leading packets would otherwise complete it and hand
+			// the codec one buffer straddling the cut.
+			if let Ok(pid) = Pid::new(pid)
+				&& self.streams.contains_key(&pid)
+			{
+				match self.continuity.entry(pid).or_default().observe(&pkt) {
+					Continuation::Duplicate => continue,
+					Continuation::Broken => {
+						// Flush first: whole frames already delivered in the truncated PES are
+						// good, and only what was left mid-frame is lost. Then drop that partial
+						// and require the next frame to prove its boundary.
+						self.flush(pid)?;
+						if let Some(stream) = self.streams.get_mut(&pid) {
+							stream.desync();
+						}
+						// This packet still routes normally: a PUSI opens a fresh PES, while a
+						// continuation finds no pending entry and is dropped, so the stream
+						// resumes at the next PES start rather than mid-frame.
+					}
+					Continuation::Contiguous => {}
+				}
 			}
 			// PIDs we don't decode and don't carry (`Stream::Ignored`: a base catalog's
 			// undecoded streams, or an ambiguous 0x86 PID without CUEI) are dropped here,
@@ -411,6 +438,7 @@ impl<E: catalog::Catalog> Import<E> {
 			self.initialized = true;
 		}
 		self.streams.insert(pid, stream);
+		self.continuity.entry(pid).or_default();
 		Ok(())
 	}
 
@@ -441,6 +469,7 @@ impl<E: catalog::Catalog> Import<E> {
 		}
 		// This PID is becoming section-framed; drop any partial PES a prior codec left pending.
 		self.pending.remove(&pid);
+		self.continuity.remove(&pid);
 		if !self.supports_mpegts {
 			// Always route to Ignored, replacing any prior codec on this PID (a later PMT
 			// can reassign it), so a private section never reaches the PES reader. Warn once.
@@ -952,6 +981,82 @@ impl<E: catalog::Catalog> VerbatimStream<E> {
 	}
 }
 
+/// Whether one PID's TS packets are still an unbroken chain, for whoever is accumulating
+/// bytes out of them.
+///
+/// Both reassemblers on this PID need the same answer: a section and a PES are equally
+/// meaningless when spliced onto bytes that didn't follow them. Keeping one implementation
+/// keeps a subtle rule (which packets advance the counter, which retransmissions to ignore)
+/// from drifting into two.
+#[derive(Default)]
+struct Continuity {
+	/// Last continuity_counter seen on a packet with payload, to spot gaps.
+	last_cc: Option<u8>,
+	/// Last payload packet, to skip ISO 13818-1 duplicates (same cc, identical bytes).
+	last_pkt: Option<[u8; 188]>,
+}
+
+/// What one packet says about the bytes already accumulated for its PID.
+enum Continuation {
+	/// An exact retransmission, which ISO 13818-1 permits once. Ignore the packet entirely:
+	/// processing it would reset a healthy partial or duplicate its bytes.
+	Duplicate,
+	/// Contiguous with the previous payload packet.
+	Contiguous,
+	/// A counter gap, a declared discontinuity, or a transport error. Whatever was
+	/// accumulating for this PID is lost, and the bytes that follow are not its continuation.
+	Broken,
+}
+
+impl Continuity {
+	/// Classify one 188-byte packet, recording what the next call needs.
+	fn observe(&mut self, pkt: &[u8; 188]) -> Continuation {
+		// transport_error_indicator: the demodulator flagged this packet as corrupt, so its
+		// payload can't be trusted (and we don't validate CRC-32). Forgetting the counter
+		// keeps the next clean packet from also looking like a gap.
+		if pkt[1] & 0x80 != 0 {
+			self.last_cc = None;
+			self.last_pkt = None;
+			return Continuation::Broken;
+		}
+
+		let afc = (pkt[3] >> 4) & 0x3;
+		let has_payload = afc & 0x1 != 0;
+		// Read the adaptation field before the no-payload case: a discontinuity can ride on
+		// an adaptation-only packet, and it counts just the same.
+		let discontinuity = if afc & 0x2 != 0 {
+			let af_len = pkt[4] as usize;
+			af_len > 0 && pkt[5] & 0x80 != 0
+		} else {
+			false
+		};
+
+		if !has_payload {
+			if discontinuity {
+				self.last_cc = None;
+				self.last_pkt = None;
+				return Continuation::Broken;
+			}
+			return Continuation::Contiguous;
+		}
+
+		if self.last_pkt.as_ref().is_some_and(|last| last == pkt) {
+			return Continuation::Duplicate;
+		}
+		self.last_pkt = Some(*pkt);
+
+		// Only payload packets advance the counter, so this is the one place it moves.
+		let cc = pkt[3] & 0x0f;
+		let cc_gap = matches!(self.last_cc, Some(last) if cc != (last + 1) & 0x0f);
+		self.last_cc = Some(cc);
+		if discontinuity || cc_gap {
+			Continuation::Broken
+		} else {
+			Continuation::Contiguous
+		}
+	}
+}
+
 /// Byte-level reassembler for MPEG-TS private sections on one PID.
 ///
 /// Private sections (SCTE-35 table_id 0xFC and others) are not PES. This handles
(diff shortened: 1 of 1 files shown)


t0ms:
Ran this against real broadcast content, because the four in-tree captures show the fix costs nothing on a well-formed mux but none of them contains a splice. The unit-level change looks right and both new tests pass on my build — but **on a looped real broadcast TS the mixed-bytes frame is still published, and the audio is byte-identical before and after the fix**. The reason is that a real loop wrap doesn't land where the fix looks.

## Setup

- **pre** = `main` @ `f91e3bbd2`, **post** = this PR @ `a64dc4d12`. The three commits on main since the PR's merge base don't touch `container/ts/`, so the two arms differ only by this change in the importer.
- Oracle that my post build is genuinely the fixed one: `legacy_confirms_a_frame_joined_out_of_a_carried_tail` and `aac_confirms_a_frame_joined_out_of_a_carried_tail` both pass on it.
- **Source**: a 20 s cut of a 1080i25 CNN broadcast capture — H.264 + MP2 (192 kb/s, 48 kHz) + AC-3 + teletext + 3× SCTE-35, 9.95 Mb/s CBR — looped with `tsp -I file … --infinite -P regulate --pcr-synchronous` into `moq import ts`, through a relay, captured with `moq export ts`. 70 s per arm, so three wraps.
- **Detection** is decidable rather than a judgement about audio: a frame is *alien* if its bytes appear nowhere in the source's audio elementary stream. A frame assembled across a splice is made of bytes from both sides of it, so it cannot match anything in the source.

## Result

| | frames | alien frames | |
|---|---:|---:|---|
| MP2 (PID 121), pre | 2580 | **3** | one per wrap, all `0b8cc4572738` |
| MP2 (PID 121), post | 2589 | **3** | identical hashes, identical positions |
| AC-3 (PID 123), pre | 1917 | **3** | one per wrap, all `9524a1486a3b` |
| AC-3 (PID 123), post | 1926 | **3** | identical hashes, identical positions |

Each alien frame begins with exactly the source's trailing partial frame (554 B for MP2, 282 B for AC-3), so these are the splice frames, not some unrelated artefact. The frame-count difference is only that the post capture ran ~0.2 s longer; over the common length the audio is **byte-identical** — 1,486,080 B of MP2 and 1,472,454 B of AC-3, covering all three wraps.

## Why the fix doesn't fire

Two properties of real broadcast muxing, neither present in the synthetic fixtures:

**1. This mux never splits a frame across a PES boundary.** Every audio PES payload is exactly 5184 B = 9 × 576 B MP2 frames (AC-3: 6912 B = 9 × 768). 0 of 91 interior MP2 PES ends and 0 of 68 AC-3 PES ends fall mid-frame. So `LegacyStream::tail` is empty at every normal PES boundary, and the carried-tail path this PR guards is never entered. The only mid-frame cut in the whole clip is the *file* end.

**2. At the wrap, the foreign bytes join the same PES, not the next one.** The clip was cut out of a longer stream, so it *begins* with 16 continuation packets on the MP2 PID (2806 B, no PUSI) — the tail of a frame that started before the clip did. The final PES before the wrap is truncated by the cut, so it is still open and short of its declared `PES_packet_length`; those leading continuation packets are appended to it, and `handle_pes_continuation` flushes as soon as `data.len() >= data_len`.

So the demuxer is handed a **single PES** whose payload is `[whole frames][554 B partial frame][~736 B of unrelated pre-clip audio]`. The mixed frame begins inside `pending.data` with `carried == 0`, so `in_tail` is false; the preceding frame vouched for the offset, so `unconfirmed` is false. `needs_confirmation` returns false and the frame is published — exactly the pre-PR path.

**Worth stressing: the confirmation rule itself would have caught this.** At the mixed frame's declared end the bytes are `29126d37`, which is not a valid MP2 header, so confirmation would have rejected it and scanned. This isn't the "false sync landing on a real frame boundary" limit #2802 flags. It's the gate — `in_tail` — that misses, not the test.

## A suggestion, from precedent in the same file

The signal that these bytes are not contiguous with what preceded them is available one layer down, and it is free. `SectionReassembler` already does the right thing for private sections: it drops its partial on `transport_error_indicator`, on an adaptation-field `discontinuity_indicator`, and on a **continuity-counter gap**, and it skips ISO 13818-1 duplicate packets. The PES path does none of these — `handle_pes_start` / `handle_pes_continuation` never look at `packet.header.n_counter`, though `mpeg2ts` exposes it.

Our wrap breaks continuity on every PID: MP2 goes `cc 3 → 12` where 4 was due, AC-3 `9 → 0`, video `4 → 3`. Applying the same three checks to the PES reassembler — drop the in-progress PES and mark the stream unconfirmed on a CC gap, a declared discontinuity, or a TEI — would catch this case with **no added latency at all**, no waiting for a successor, and independently of codec. It also generalises past loop wraps to the dropped-PES case the issue mentions, which is the same signal in the field.

That looks complementary to what this PR does rather than a replacement: confirmation still covers a genuine mid-frame PES split in a mux that does align that way, where continuity is intact and nothing is wrong.

## Secondary, and weaker than I first thought

AC-3 carries a mandatory `crc1` over the first 5/8 of each syncframe, and `ac3::parse_header` reads bytes 0,1,4,5,6 — stepping straight over it. It rejects all three mixed frames here (implementation validated first: it passes on 620/620 genuine source frames). So for AC-3 and E-AC-3 there's a zero-latency check going unused.

But it is not a general answer, and the same clip shows why: **0 of its 826 MP2 frames carry a CRC at all** (`protection_bit` set throughout), and ADTS gates its CRC on `protection_absent` too. So a CRC check helps AC-3 and can't be relied on for MP2 — a complement, not a substitute. The continuity-counter route above covers every codec.

## Why the unit tests couldn't have caught this

The new tests build PES payloads and hand them to the stream directly, so there is no TS packet layer, no continuity counter to break, and a splice can only be expressed as a carried tail. That is exactly the shape the fix covers, and it's a different shape from the one a looping file actually produces. Might be worth a fixture that splices at the TS-packet layer.

Happy to share the reproducer or run any variant — the harness is a single script and the audit is decidable, so a "does the capture contain a frame the source never had" check would drop into CI against a real capture if that's useful.

(Written by Opus 5)

kixelated:
Rescoped after review. @t0ms is right, and I reproduced it in-tree before touching anything: the confirmation in the first commit never fires on the shape a looping mux actually produces.

**Verified mechanism.** `handle_pes_continuation` appended continuation packets to an open bounded PES without ever consulting the continuity counter, so at a wrap the next loop's leading packets complete the truncated PES and the codec receives one buffer straddling the cut with `carried == 0`. A packet-level fixture reproduces it exactly: with only the first commit applied, the second published MP2 frame starts `FF F5 18 C0 BB BB` and ends `CC CC CC CC`. The gate is `in_tail`, and it misses. Confirming the diagnosis rather than the test.

**What changed.** A second commit takes the continuity rules `SectionReassembler` already implemented (transport errors, adaptation-field discontinuities, counter gaps, ISO 13818-1 duplicate packets) into a shared `Continuity`, and applies them to PES PIDs too. On a break the truncated PES is flushed first, so whole frames it did deliver still publish, then the stream desyncs and resumes at the next PES start. That is the zero-latency, codec-independent route suggested, and it generalises to the dropped-PES case.

The first commit stays: it covers a genuine mid-frame PES split in a mux that does align that way, where continuity is intact and nothing is wrong. Complementary, as you put it.

On the CRC point, agreed it cannot be the general answer given `protection_bit` / `protection_absent`, so I left it alone.

**Codex's other two findings**, for the record, both rejected as unchanged from `main` rather than as wrong:

- The overrun branch still reads plain `unconfirmed()`. Using the combined state there would let a chance sync inside a legitimately split frame's own payload displace it, and the behaviour is identical to `main` either way.
- The EOF drain publishing an unconfirmed joined tail is also pre-existing. The conservative alternative drops the final frame of every cleanly truncated capture. Worth revisiting now that continuity provenance exists, but not here.

**Verification.** Both new packet-level tests fail without the second commit and pass with it, as do the four from the first commit against theirs. `cargo nextest run` is green over moq-mux (504) and over moq-mux plus its dependents (840); clippy `-D warnings` and rustdoc `-D warnings` are clean. The four in-tree captures (ac3, eac3, kyrion, bbb) publish byte-identical audio across `main`, commit 1, and commit 2: 16/16/64/3 frames, same totals, so neither commit costs anything on a well-formed mux.

@t0ms yes please to the reproducer, if the offer stands. A "does this capture contain a frame the source never had" check against real content sounds like a better gate than anything synthetic I can build here.

(written by Opus 5)

kixelated:
Four findings from the latest round, all verified against the code before acting.

**Fixed — TEI packets are processed after being flagged corrupt** (Codex P1). Real, and a regression my own refactor introduced: extracting `SectionReassembler`'s continuity logic was meant to preserve behaviour exactly, and it lost the early return that discarded a TEI packet's payload. A corrupt PUSI could start a section out of bytes the demodulator had already disowned. There is now a distinct `Continuation::Corrupt`, and the packet dies with the partial at both call sites. `tei_pusi_section_is_dropped` covers it, and fails when the fix is reverted.

**Fixed — Opus is salvaged from a truncated PES** (Codex P1 and CodeRabbit, independently). Real. Opus framing is declared rather than self-describing, so a trailing packet cut short of its control header's length is a parse error, and that error travels out of `decode` and ends the session. Opus is no longer in the salvage set.

**Filed, not fixed — a truncated Opus PES ends the session anyway** ([#2849](https://github.com/moq-dev/moq/issues/2849)). While writing the regression test for the above I found the abort is reachable without any continuity break at all: `handle_pes_start` flushes a short PES whenever the next one begins, so a wrap kills the session on `main` today. Looping `opus.ts` at every packet boundary fails at packet 21 with `Opus access unit exceeds PES payload` and at packet 22 with `invalid Opus control header sync`. The first is a few lines; the second needs the scan-and-confirm machinery #2751 built for the legacy codecs, which Opus has none of. A half-fix still fails the reproduction, so this PR ships neither and the issue carries the full repro.

**Filed, not fixed — duplicate detection ignores the permitted PCR difference** ([#2851](https://github.com/moq-dev/moq/issues/2851)). CodeRabbit is right that byte equality misses a retransmission with a refreshed PCR and then misreads it as a gap. I implemented the suggested fix, keying on the repeated counter, and Codex caught that it is unsafe: a loss of exactly 15 packets brings the counter back to the same value, so every such loss would discard the first real packet and carry the stale partial across it. That is worse than the bug being fixed, in the direction that matters. Reverted in 186007f19; the issue records why the obvious fix is wrong and what a correct one has to compare.

Codex also noted there is no PES-specific TEI test, only the section one. The shared `Continuity` is what both call sites consult and it is covered; the PES arm is three lines around `pending.remove`. Left as is deliberately.

(written by Opus 5)

t0ms:
Verified the merged version against the same real broadcast content that produced the null result, and
the mixed frame is gone: **3 alien audio frames → 0** on both PIDs, on `6198e5757` and again on today's
`main` (`eab960192`, so with #2891). Same clip, same three wraps, arms differing only by the merge.

| arm | MP2 (PID 121) | AC-3 (PID 123) |
|---|---:|---:|
| `f91e3bbd2` (before the PR) | 3 alien | 3 alien |
| `a64dc4d12` (first commit only) | 3 alien | 3 alien |
| `6198e5757` (merged) | **0** | **0** |
| `eab960192` (`main` today) | **0** | **0** |

The generalised `Continuity` is doing exactly what the section path does, and on this content it fires on
every wrap. Thanks for reproducing it in-tree rather than taking my word for it.

## The reproducer, as offered

The check is one decidable question — *does this capture contain an audio frame the source never had?* A
frame assembled across a splice is made of bytes from both sides of the cut, so its hash matches nothing
in the source's elementary stream. No listening test, no reference decoder.

```
ts-audio-frames.py  <clip.ts> <pid> {mp2|ac3}     # frame-by-frame walk: offsets, hashes,
                                                  # resync gaps, trailing partial frame
ts-splice-audit.py  <source.ts> <cap1.ts>,<cap2>  # alien frames per capture, PES/frame
                                                  # alignment, leading continuations, AC-3 crc1
```

Both are self-contained Python with no dependencies; happy to open a PR putting them under `test/` or
`dev/` if you want them in-tree, or to just paste them here.

**One trap worth knowing if you build your own version of this.** My first version took "the source's
frames" to be everything from the first PUSI onward, since that is where PES reassembly can start. A clip
cut out of a longer stream *opens* with continuation packets belonging to a PES whose start was lost —
2694 B on the AC-3 PID here — and a demuxer that resyncs inside those bytes emits frames that are
perfectly correct but match nothing in that set. They get reported as alien. On the clip below that
inflated one real alien per wrap to three. The reference has to be built the way the code under test
consumes the stream, not the way a parser prefers to.

## Residual 1: a wrap that does not break the counter is still spliced

The clip I used breaks continuity on all three PIDs at the wrap (MP2 `cc 3 → 12` where 4 was due, AC-3
`9 → 0`, video `4 → 3`), so the new check sees every splice. That is a property of where the file was
cut, not of looping. If a PID's last packet leaves `(cc + 1) % 16` equal to the counter the file *opens*
with, the wrap is continuity-contiguous and the break is invisible. Scanning cut points on this source,
**4062 of 30,000 land that way for one of the two audio PIDs** — about one in 16 per PID, as you would
expect.

Cutting at **130,705 packets** puts AC-3 in exactly that state (last packet `cc 15`, file opens `cc 0`)
while still ending mid-frame, with the final PES short of its declared `PES_packet_length`. On today's
`main`, looped for 70 s:

```
PID 123 (0x7b) ac3
  source          : 614 frames, 768 B each, trailing partial 106 B
  leading continuation packets: 15 (2694 B)
  capture         : 1932 frames, ALIEN 3          <- one per wrap
      begins with the source's trailing partial: True;  crc1 rejects it: True
```

MP2 stays clean on that clip because its counter still breaks there. So the guard is right but it has one
input, and the signal it depends on is absent on a predictable fraction of cuts. Two things would close
it without touching latency: the truncated PES is short of its declared `PES_packet_length` at the moment
the foreign bytes complete it, which is knowable independently of continuity; and where a codec does
carry a CRC, `crc1` rejects every one of these frames. Neither is general alone — MP2 carries no CRC at
all on this feed, and an unbounded PES (`PES_packet_length == 0`) has no declared length to check — but
between them they cover the case the counter misses.

## Residual 2: the salvage delivers for MP2 and not for AC-3

`salvages_partial_pes` is true for `Stream::Legacy`, and the intent as I read it is that whole frames the
truncated PES already carried still publish. That happens for MP2: the 7 complete frames inside its final
PES (source frames 819–825) publish before and after the fix. It does not happen for AC-3. The 8 complete
frames inside its final PES (612–619) are published by the pre-fix build at all three wraps and are
**absent from every wrap of both post-merge builds** — searched by hash across the whole capture, not
just at the seam. That is ~256 ms of good audio per wrap, and it is the difference between the two arms
rather than anything about the clip, which is unchanged.

Both PIDs are `Stream::Legacy` and take the same `match` arm, so whatever causes it sits below the branch
— either the flush is not reaching the parser or the parser is dropping a frame it could confirm. The one
asymmetry I can see in the content is the size of the trailing partial: 554 B of a 576 B MP2 frame
against 282 B of a 768 B AC-3 frame.

Happy to split either of these into its own issue if you would rather they were tracked separately, and to
attach the two clips or the recipe for cutting them.

PR #2824, opened 2026-08-13
feat(mux): carry EIT present/following through a TS round-trip

### LINKED ISSUE #2800: Media-aware TS lane: EIT (PID 0x0012) is dropped at import, and carrying it is cheap
`moq import ts` drops EIT at the import gate, so it cannot survive a TS round-trip. #2440 already built the mechanism that would carry it, and measured on a representative stream the cost is 12 catalog updates and ~1.3 kB over ten minutes. This asks for one row in `SI_PIDS` — and for TDT/TOT to stay out, for a reason worth writing down next to it.

Filed following the close of #1861, where we said the remaining fidelity gap was "a much smaller ask than a second lane, and we will file it as one".

## The gap, measured

`SI_PIDS` (`rs/moq-mux/src/container/ts/catalog.rs`) is a const allowlist of `(PID, interval)` holding NIT (0x0010, 10 s) and SDT/BAT (0x0011, 2 s). `import.rs` intercepts exactly those PIDs before the PES reader; everything else falls through and is dropped, so EIT (0x0012) never reaches the catalog and export has nothing to rebuild it from.

No broadcast capture we hold carries EIT, so we synthesised one: a real DVB contribution clip (CNN International EMEA HD, 9.95 Mbps CBR, H.264 + MP2 + AC-3 + teletext + 3× SCTE-35) with a twelve-event EPG injected via `tsp -P eitinject`, anchored to the clip's own TDT epoch. The EIT packets come out of the clip's null stuffing, so the mux rate is unchanged at 9,945,951 bps and continuity stays clean.

Round-tripped through a local relay on `main` @ `9698cd93` (i.e. with #2440), 45 s capture:

| Table | PID | Source packets | Egress packets |
|---|---|---:|---:|
| NIT | 0x0010 | 119 | 5 |
| SDT | 0x0011 | 574 | 21 |
| **EIT** | **0x0012** | **1,153** | **0** |
| **TDT/TOT** | **0x0014** | **40** | **0** |

The egress carries PAT, NIT, SDT/BAT, PMT, video, both audio tracks, teletext and all three SCTE-35 PIDs, and nothing on 0x0012 or 0x0014. NIT and SDT survive at the catalog's own repetition cadence rather than the source's, which is the intended #2440 behaviour.

## The cost of carrying it, measured

We had assumed EIT would be too churny for the catalog, and that turns out to be wrong, so the numbers are worth stating rather than the intuition.

The `si` map dedupes by `(table_id, table_id_extension, section_number)` and the import path compares a new section against the published set *before* taking the catalog's write lock, so what matters is not how often a table is transmitted but how often its bytes actually change. Over the ~10 minutes of the fixture:

| Table | PID | Sections transmitted | **Distinct** | Mean distinct section |
|---|---|---:|---:|---:|
| NIT | 0x0010 | 119 | 1 | — |
| SDT | 0x0011 | 574 | 1 | — |
| EIT p/f actual | 0x0012 | 592 | **4** | 77 B |
| EIT schedule actual | 0x0012 | 420 | **8** | 129 B |
| TDT/TOT | 0x0014 | 40 | **40** | — |

So carrying EIT would take the catalog's write lock **12 times in ten minutes** and add about **1.3 kB** of catalog state. That is the same order as SDT and NIT, and for the same reason: EIT repeats byte-identically between event transitions, and the repetition rate is decoupled from the revision rate.

TDT/TOT is the opposite case and the row that explains the policy: every section is new content, so each one is a catalog modification and a republish. It is also the table with the least to gain from being carried, since it says nothing but "now" and an exporter can mint a more accurate one than a relayed value from an upstream mux of unknown delay.

The caveat we cannot measure from one fixture is scale: ours is a single service with twelve events, and distinct-section count grows with events × services.

## The ask

1. **Add 0x0012 to `SI_PIDS`** with the ETSI TS 101 211 p/f cycle (2 s). EIT p/f actual is bounded at two sections per service, so the catalog cost is bounded too.
2. **Consider gating EIT schedule** separately. Our fixture's schedule is small, but a full eight-day schedule for a multi-service mux is a different proposition from p/f. If only one of the two is wanted, p/f is the one worth having.
3. **Leave 0x0014 out, and say so in `SI_PIDS`.** The measurement above is the rationale; right now the omission reads as an oversight. Regenerating TDT/TOT at export from the exporter's clock would be more correct than carrying it, if it is ever wanted.

## Reproducing

Both scripts are in our lab and we are happy to contribute them, or a trimmed version, as a test fixture — the generator needs only TSDuck and any clip, which makes the EIT path testable without a broadcast capture that carries EIT:

```bash
make-eit-fixture.sh full ~/clip.ts ~/clip_eit.ts   # inject a synthetic EPG on PID 0x0012
eit-roundtrip.sh ~/clip_eit.ts 45                  # publish -> relay -> export ts, then census the SI PIDs
```

## References

- `rs/moq-mux/src/container/ts/catalog.rs` — `SI_PIDS`, `Si::upsert`
- `rs/moq-mux/src/container/ts/import.rs` — the SI intercept and the write-lock guard
- #2440 — SI sections carried opaquely by PID; TDT/TOT and EIT left as a policy decision
- #1861 — the opaque-lane proposal this replaces, closed as not planned

kixelated:
Closes #2800 (partially: p/f and TDT/TOT, not schedule).

`moq import ts` dropped EIT at the import gate, so the EPG could not survive a TS round-trip even though #2440 had already built the mechanism that would carry it. #2800 measured the cost on a synthesized EPG and found it far cheaper than we assumed: the `si` map dedupes by `(table_id, table_id_extension, section_number)` and compares before taking the catalog's write lock, so cost tracks the *revision* rate, not the transmission rate. Over ten minutes, EIT present/following was 4 distinct sections out of 592 transmitted.

## Root cause of the gap

`SI_PIDS` was keyed by PID alone, and `si_section` captured every completed section on an intercepted PID whatever its `table_id`. That is deliberate and correct for the PIDs it covered (the SDT PID also carries the BAT; an unrecognized table is as worth preserving as a known one), but it made EIT all-or-nothing: p/f (`0x4E`, `0x4F`) and schedule (`0x50`..=`0x6F`) share PID 0x0012.

`Si::interval` made that concrete rather than theoretical. It is one value per PID, but per ETSI TS 101 211 p/f actual is 2 s and schedule is 10-30 s. One PID, two cadences, one slot: there was no entry that was correct for both.

## The change

`SI_PIDS` becomes `&[SiPid]` (pid, tables, interval) with a `Tables::{All, Only}` filter, and 0x0012 joins it as `Only(&[(0x4E, 0x4F)])` at 2 s. `All` stays the default posture and is what 0x0010 and 0x0011 use, so nothing about their capture changes.

Filtering on `table_id` does not make the sections any less opaque: it is the first byte of every section under generic section syntax, the same byte `section_key` already reads to dedupe. This narrows what we carry without widening what we claim to parse.

The filter runs *after* reassembly, in `si_section`, because the EIT PID interleaves p/f with schedule and a section boundary is only findable by following the whole PID. When nothing survives the filter the entry is never created, so a schedule-only stream leaves no empty 0x0012 behind for export to emit.

Export needed no change: it already walks `mpegts.si` generically, honoring each PID's own interval.

## What is deliberately left out

**EIT schedule.** The catalog is whole-state and republished on every change, so cost per change is O(catalog), not O(delta), and every subscriber pays the full document at join, ahead of media. #2800's fixture yields 8 distinct schedule sections because it is one service with twelve events; a full multi-service eight-day EPG is thousands of sections whose window edges churn continuously as it rolls forward. That asymmetry does not depend on the magnitude, only on the catalog being whole-state. If schedule is ever wanted it belongs on its own track, the way SCTE-35 sections already ride one via `SectionStream`, and the reasoning is recorded next to the 0x0012 entry.

**TDT/TOT (0x0014)**, which is #2800's ask 3. Every section is new content rather than a repetition, so each would be a catalog modification and a republish, and it is the table with the least to gain from being relayed: an exporter's own clock is a better source than a time forwarded from an upstream multiplexer of unknown delay. The omission was previously silent and read as an oversight; it now has the rationale next to it.

## Tests

Three, in `export_test.rs` alongside the existing SI coverage:

- `eit_carries_present_following_not_schedule` feeds six sections interleaving p/f with schedule at both ends of both ranges, and asserts exactly the two p/f sections survive at a 2 s interval. This is also the regression test for the filter running post-reassembly.
- `eit_schedule_alone_creates_no_entry` covers the empty-entry case.
- `tdt_is_not_captured` pins the 0x0014 policy.

The existing `multi_packet_si_section_is_captured` is refactored onto the new `import_si` helper rather than duplicating the setup.

## Not done here

#2800 offers TSDuck fixture scripts (`make-eit-fixture.sh`, `eit-roundtrip.sh`) that synthesize an EPG onto any clip. Worth taking as a follow-up: no broadcast capture we hold carries EIT, which is why this gap survived #2440 with the mechanism in place and the table unrouted. The tests here are synthetic sections, not a real multiplex.

## Cross-package sync

`doc/bin/cli.md` described TDT/TOT and EIT together as "live or bulky rather than static identity" - two correct rationales covering three tables, and the same conflation this PR splits. Updated. No draft change: the `mpegts` catalog section is not specified in `drafts/`. No js mirror: `js/` has no MPEG-TS support.

(written by Opus 5)

### DIFF
--- a/doc/bin/cli.md
+++ b/doc/bin/cli.md
@@ -519,8 +519,13 @@ The service layer rides the same `mpegts` section. The program identity
 rebuild a matching PAT/PMT, while the standalone SI tables (SDT, NIT) are carried as
 opaque sections and re-emitted byte-for-byte on their original PIDs, each at its own
 repetition interval. So the service name, provider, type, and network survive the
-round-trip without being parsed. Regenerated tables (TDT/TOT) and EPG (EIT) are not
-captured: they are live or bulky rather than static identity.
+round-trip without being parsed. Present/following EPG (EIT p/f) rides along too, so
+"now" and "next" survive, but the full EIT schedule does not: the catalog is republished
+whole on every change, which suits a table that is small and revised rarely rather than
+an eight-day listing whose edges churn continuously. Regenerated tables (TDT/TOT) are
+left out for the opposite reason, that every section is new content rather than a
+repetition, and an exporter's own clock beats a time relayed from an upstream
+multiplexer of unknown delay.
 
 ### FLV
 
--- a/doc/bin/cli.md
+++ b/doc/bin/cli.md
@@ -519,13 +519,14 @@ The service layer rides the same `mpegts` section. The program identity
 rebuild a matching PAT/PMT, while the standalone SI tables (SDT, NIT) are carried as
 opaque sections and re-emitted byte-for-byte on their original PIDs, each at its own
 repetition interval. So the service name, provider, type, and network survive the
-round-trip without being parsed. Present/following EPG (EIT p/f) rides along too, so
-"now" and "next" survive, but the full EIT schedule does not: the catalog is republished
-whole on every change, which suits a table that is small and revised rarely rather than
-an eight-day listing whose edges churn continuously. Regenerated tables (TDT/TOT) are
-left out for the opposite reason, that every section is new content rather than a
-repetition, and an exporter's own clock beats a time relayed from an upstream
-multiplexer of unknown delay.
+round-trip without being parsed. Present/following EPG for this transport stream (EIT
+p/f actual) rides along too, so "now" and "next" survive. The full EIT schedule does not:
+the catalog is republished whole on every change, which suits a table that is small and
+revised rarely rather than an eight-day listing whose edges churn continuously. Neither
+does EIT for other multiplexes, whose sections a verbatim carriage keyed on the section
+header cannot tell apart. Regenerated tables (TDT/TOT) are left out for a different
+reason, that every section is new content rather than a repetition, and an exporter's own
+clock beats a time relayed from an upstream multiplexer of unknown delay.
 
 ### FLV
 
--- a/rs/moq-mux/src/container/ts/catalog.rs
+++ b/rs/moq-mux/src/container/ts/catalog.rs
@@ -19,18 +19,97 @@ use serde_with::{DisplayFromStr, DurationMilliSeconds, serde_as};
 
 use crate::catalog::hang::CatalogExt;
 
-/// A standalone SI PID we capture, and how often export must re-emit it.
+/// Which of a PID's tables we capture.
+///
+/// A PID is usually all-or-nothing: the SDT PID also carries the BAT, and a table we
+/// have never heard of is as worth preserving as one we have, so [`Self::All`] is the
+/// default posture. [`Self::Only`] exists for a PID that carries tables with genuinely
+/// different carriage requirements, which so far is just the EIT PID.
+///
+/// Selecting on `table_id` does not make the sections any less opaque. The `table_id`
+/// is the first byte of every section under generic section syntax (ISO 13818-1), the
+/// same byte [`section_key`] already reads to dedupe: this narrows what we carry
+/// without widening what we claim to understand.
+#[derive(Debug, Clone, Copy)]
+pub(super) enum Tables {
+	/// Every section on the PID, whatever its `table_id`.
+	All,
+	/// Only sections whose `table_id` falls in one of these inclusive ranges.
+	Only(&'static [(u8, u8)]),
+}
+
+/// A standalone SI PID we capture: which of its tables, and how often export must
+/// re-emit them.
 ///
 /// Intervals are the DVB maximum repetition intervals (ETSI TS 101 211); export
 /// treats them as an upper bound, not the source's observed cadence, which is a
-/// property of that multiplexer's bitrate shaping and means nothing downstream.
-/// Adding a table here is a one-line change: the sections themselves are opaque.
-pub(super) const SI_PIDS: &[(u16, Duration)] = &[
+/// property of that multiplexer's bitrate shaping and means nothing downstream. A PID
+/// gets one interval because [`Si`] holds one, so a PID carrying several tables takes
+/// the tightest of them.
+#[derive(Debug, Clone, Copy)]
+pub(super) struct SiPid {
+	/// The PID these sections ride on.
+	pub pid: u16,
+
+	/// Which of the PID's tables to capture.
+	pub tables: Tables,
+
+	/// Maximum re-emit interval, recorded in [`Si::interval`] for export.
+	pub interval: Duration,
+}
+
+impl SiPid {
+	/// The entry for `pid`, or `None` if we capture nothing on it.
+	pub fn get(pid: u16) -> Option<&'static Self> {
+		SI_PIDS.iter().find(|si| si.pid == pid)
+	}
+
+	/// Whether a completed section on this PID is one we carry.
+	pub fn captures(&self, section: &[u8]) -> bool {
+		match self.tables {
+			Tables::All => true,
+			Tables::Only(ranges) => {
+				let table_id = section.first().copied().unwrap_or(0);
+				ranges.iter().any(|(first, last)| (*first..=*last).contains(&table_id))
+			}
+		}
+	}
+}
+
+/// The standalone SI PIDs we capture.
+pub(super) const SI_PIDS: &[SiPid] = &[
 	// NIT (network description): 10s.
-	(0x0010, Duration::from_secs(10)),
+	SiPid {
+		pid: 0x0010,
+		tables: Tables::All,
+		interval: Duration::from_secs(10),
+	},
 	// SDT and BAT (service and bouquet description): 2s for SDT Actual, the tightest
 	// of the two, so one interval per PID stays conservative.
-	(0x0011, Duration::from_secs(2)),
+	SiPid {
+		pid: 0x0011,
+		tables: Tables::All,
+		interval: Duration::from_secs(2),
+	},
+	// EIT present/following (0x4E actual, 0x4F other): 2s for actual, the tightest.
+	//
+	// This PID also carries EIT schedule (0x50..=0x6F), which is deliberately left out.
+	// The catalog is whole-state and republished on every change, so it suits a table
+	// that is small and revised rarely. p/f is exactly that: two sections per service,
+	// replaced on event transition. A full schedule is thousands of sections whose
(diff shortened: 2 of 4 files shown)


t0ms:
Tested against a real EIT-bearing multiplex rather than synthetic sections, since #2800 is ours and the fixture is the thing we could contribute. **It does what it says, including the two guards added after the description was written.** Details below, plus one gap in what I could cover and one confirmation you may want on the TDT/TOT decision.

Build is the PR head `e2882d07b`; `cargo test -p moq-mux --release` is 503/503. Source is `CNNiEMEA2.ts` (1920×1080 H.264, MP2 + AC-3, teletext, three SCTE-35 PIDs, SDT and NIT) with the synthesized EPG from #2800 injected on 0x0012 — 592 p/f actual sections and 411 schedule sections over ten minutes.

## The round-trip

| PID | source | egress |
|---|---|---|
| NIT 0x0010 | 119 | 5 |
| SDT 0x0011 | 574 | 21 |
| **EIT 0x0012** | **1 153** | **37** (was 0 before this PR) |
| TDT/TOT 0x0014 | 40 | 0 |

Broken down by `table_id`, the egress is 37 sections of `0x4E` and nothing else — every one of the 411 schedule sections is dropped, and the post-reassembly filter is doing its job on a PID that genuinely interleaves the two. The cadence is right as well: those 37 sections are 19 emissions of a two-section table, ~2.4 s apart against the 2 s interval, and SDT and NIT are undisturbed at ~2.2 s and ~10.8 s.

## The version transition

The EPG rolls from version 0 to version 1 partway through the clip, so I cut a 60 s window spanning that point and ran that. The egress switches once, cleanly: **v0 ×27 → v1 ×28**, no flapping between the two and no stale version left behind.

Reassembling both sides of that run and comparing bytes, the carried sections are **byte-identical to the source**:

| table_id | version | section | length | |
|---|---|---|---|---|
| 0x4E | 0 | 0 | 90 | byte-identical |
| 0x4E | 0 | 1 | 65 | byte-identical |
| 0x4E | 1 | 0 | 65 | byte-identical |
| 0x4E | 1 | 1 | 88 | byte-identical |
| 0x50 | 0 and 1 | 0, 8, 16, 24, 32, 40, 48 | 18–221 | dropped |

## `is_current`, which nothing we hold exercised

Our EPG only ever transmits sections that are already in force, so `current_next_indicator = 0` never appeared, and that guard would have gone untested. So I built the case: take the transition clip and mark the first 20 occurrences of version 1 as pending, recomputing the section CRC so the result is a legal stream and a rejection means the guard fired rather than the section being malformed. TSDuck parses it without complaint.

That gives a window in which the *only* EIT on the wire is a version that does not apply yet. The result is what the guard promises:

- egress carries **v0 ×47 → v1 ×8**, against **v0 ×27 → v1 ×28** on the unpatched clip
- **no pending section reaches the egress at all**

The 20 extra v0 emissions line up one-for-one with the pending window: the exporter holds the version that applies and adopts v1 only once v1 is current. Without the guard those 20 would each have replaced v0 under the same `section_key` and republished the catalog on every repetition, which is exactly the failure the commit message describes.

Happy to contribute the generator for that fixture alongside the EIT one in `test/ts/` if it is useful — it is about thirty lines and it makes the pending-version case reproducible in-tree.

## What I could not cover

**EIT p/f other (0x4F).** No source we hold carries it and our EPG generator does not produce it, so the narrowing from `(0x4E, 0x4F)` to `(0x4E, 0x4E)` is untested here. The reasoning in the comment is right, though, and it is the same defect as #2842 and #2881 rather than something specific to EIT: `table_id_extension` is only unique within a transport stream, so two foreign services sharing a `service_id` collide on one entry. Worth noting that if #2881's generation-aware store ever lands with the full DVB triplet in the identity, 0x4F becomes carriable without parsing EIT — so this is a narrowing that a later change can widen again, not a permanent exclusion.

## On TDT/TOT

Since #2800 asked for it and this PR declines it: **from our side that is the right call, and this closes the issue rather than partially closing it.** Our own measurement is why — TDT/TOT is 40 distinct sections out of 40 transmitted, so every one is a catalog modification and a republish, where EIT p/f was 4 distinct out of 592. And the substantive point is the one in the comment: an exporter's own clock is a better source of time than a clock forwarded from an upstream multiplexer through a relay of unknown delay. Carrying it would be carrying a worse answer at the highest cost of any table on the list.

The one thing that would be worth stating in `cli.md` for anyone building against this: TDT/TOT not surviving means an exporter must **regenerate** it, not simply omit it, because a DVB receiver that never sees a TDT has no wall clock to place EIT events against — and now that the EPG survives, that gap is more visible than it was.

(Written by Opus 5)

kixelated:
And I'll be honest, I don't fully understand this EIT stuff yet. I need a few more review passes and sessions with mr claude

kixelated:
Closing unmerged: the carriage location changed underneath this PR. The measurement on #2882 (https://github.com/moq-dev/moq/issues/2882#issuecomment-5304916045) showed that catalog-carried EIT p/f hits a junction storm at realistic service counts (61 full-document republishes in 1.27 s at 40 services) and that the acquisition transient publishes incompletable tables, so SI is moving to per-`(pid, table_id)` snapshot tracks instead. The settled design is at https://github.com/moq-dev/moq/issues/2882#issuecomment-5322564837.

Nothing here was wasted:

- The `Tables` filter shape (`SiPid { pid, tables, interval }`) is essentially the `(pid, table_id)` keying the track design promotes into the catalog descriptor.
- The current/next gating and the post-reassembly filtering rationale carry over unchanged.
- The test corpus (interleaved p/f and schedule at the range edges, the empty-entry case, the TDT policy pin) transplants to the track model.
- The TDT/TOT rationale stands verbatim in the new design.

The EIT ask from #2800 lands with the track work, which carries schedule too rather than defending a p/f-only boundary.

(written by Fable 5)

Issue #2779, opened 2026-08-12
moq export ts: continuity counters are numbered from process state, so two exporters of the same broadcast emit streams that can never be compared


t0ms:
## Summary

Two `moq export ts` processes subscribed to the same broadcast reconstruct the same programme, but number each PID's continuity counter from their own state. If they did not start together the two outputs are offset by a constant — measured at **+2 on the video PID and +8 on PAT/PMT**, unchanging across a 60 s run — so every transport packet differs in byte 3 while the payload underneath is identical.

Each output is internally correct: TSDuck reports 0 continuity errors on either. The divergence is only visible when the two are compared, which is exactly what a redundant pair does.

Measured on `moq-cli` 0.9.9 / `moq-relay` 0.14.9 release binaries, `moq-lite-05`.

## Why it matters

SMPTE ST 2022-7 seamless protection switching requires the two legs of a 1+1 pair to be *packet-identical*, so a receiver can merge them by RTP sequence number and take whichever leg delivered each packet. This is how broadcast primary distribution survives the loss of a whole path without a visible artefact.

With continuity numbered per process, a MoQ-fed 1+1 pair works only while both legs have run continuously since the broadcast started. The moment one leg is restarted for maintenance, or recovers from an outage that outlived its subscription, it comes back carrying the right programme in the right place — and byte 3 of every packet disagrees with its partner. A receiver cannot repair this: rewriting a continuity counter at the receiver means changing a field the packet's own consistency depends on, in a stream it is merging from two sources.

Measured with both legs groomed to a constant bitrate and all packet placement derived from stream position, so that everything except this field is a function of the broadcast:

| Cell | Datagrams identical | Identical with CC masked |
|---|---|---|
| legs started together | 100.00 % | 100.00 % |
| leg B joins 20 s late | 0.09 % | 97.10 % |
| leg A blacked out 15 s, then recovers | 68.56 % | 98.22 % |

The pair is otherwise aligned: same slots, same RTP sequence numbers, same payloads, and a leg that joined 20 s late sends each shared sequence number a median of 10 ms from its partner. The continuity counter is the whole of the remaining difference.

## Reproduction

Two independent chains from one source — `tee` a CBR TS into two `moq import ts` publishers, each to its own relay — then two `moq export ts` subscribers, the second started 20 s after the first. Compare the two outputs packet by packet at the same stream position, first raw and then with byte 3's low nibble masked. The masked comparison is the whole argument: the payloads agree, the counters do not.

[...]

PR #2825, opened 2026-08-13
fix(moq-mux): anchor the TS table cadence to the media timeline

kixelated:
Two exporters of one broadcast should render the same packets. Today the MPEG-TS exporter decides when to re-emit a table by measuring forward from its own last emission, so the cadence is anchored to whenever that process started rather than to the media timeline. Start a second exporter of the same broadcast 20 seconds later and the two put PAT/PMT (and SDT/NIT) on different frames, forever.

That breaks the case in #2779: a SMPTE ST 2022-7 pair needs its two legs packet-identical so a receiver can merge them, and a leg restarted for maintenance has to come back rendering exactly what its partner is rendering.

## Root cause

`due()` asked "has `interval` elapsed since the last emission", which carries the phase of whenever the exporter first ran. It now asks "has the timestamp crossed into a later `floor(timestamp / interval)` slot", which is a property of the broadcast alone. Two exporters that last emitted at different points inside the same slot agree on every frame after that.

A program *with* video hid this for PAT/PMT, because the tables are also re-emitted at every video keyframe and that re-anchors both legs to a boundary they share. The two cases where it does not:

- an audio-only program, which has no keyframe to re-anchor on, so PAT/PMT drifts;
- the standalone SI PIDs (SDT 0x0011, NIT 0x0010) in any program, which are never re-anchored at keyframes, so they land on different frames rather than merely being renumbered.

A slot boundary also re-emits after a backwards timestamp jump (a rewound timeline), which the elapsed-time version sat out until the old interval came back around.

## What this does not fix

The continuity counter, which is the headline of #2779, is still numbered from process state. Making it a function of the broadcast requires each group's per-PID packet count to be a multiple of 16 (a counter that is both globally continuous and independent of the join point forces that), which means padding, which is a much larger and opt-in-shaped change. The details are in #2779; this is the part that is unambiguous and free.

## Tests

`late_join_matches_a_running_exporter{,_without_video}` produce one broadcast, export it twice with the second exporter joining halfway in, and assert the only bytes that disagree in the overlap are continuity counters. The audio-only case fails without this change (one leg renders 752 bytes where the other renders 376: two extra packets, PAT and PMT). The video case passes either way and is there to catch the next field someone mints per process; it is the topology in the report.

`author_dts_is_join_independent_at_a_peak` pins the one piece of carried state that is *not* a divergence: the monotonic decode-clock bump cannot fire on a frame whose PTS leads everything decoded before it, which every keyframe is, and export only ever tunes in on a keyframe.

## Cross-package sync

None. No wire format, no public API, no CLI surface, and the MPEG-TS exporter has no JS counterpart.

Fixes part of #2779.

(written by Opus 5)

### DIFF
--- a/rs/moq-mux/src/container/ts/export.rs
+++ b/rs/moq-mux/src/container/ts/export.rs
@@ -986,18 +986,25 @@ const DEFAULT_DTS_RESERVE: u64 = 16;
 /// at the same points, which a redundant pair compares byte for byte. `None` (nothing
 /// emitted yet) is always due, so a fresh exporter leads with the tables and a receiver
 /// can tune in without waiting for the next slot.
+///
+/// A *later* slot, not merely a different one: video is emitted in decode order, so a
+/// reordered (B-frame) PTS steps backwards all the time, and re-emitting on every
+/// oscillation across a boundary buys nothing and costs a table each way.
 fn due(timestamp: Timestamp, last: Option<Timestamp>, interval: Duration) -> bool {
+	// "Every frame", which the slot arithmetic can't express (and would divide by zero on).
+	if interval.is_zero() {
+		return true;
+	}
 	let Some(last) = last else {
 		return true;
 	};
-	slot(timestamp, interval) != slot(last, interval)
+	slot(timestamp, interval) > slot(last, interval)
 }
 
 /// Index of `timestamp`'s repetition slot: how many whole `interval`s fit under it.
-///
-/// A zero interval degenerates to one slot per microsecond, which makes every frame due.
+/// `interval` must be non-zero; [`due`] handles that case before it gets here.
 fn slot(timestamp: Timestamp, interval: Duration) -> u128 {
-	Duration::from(timestamp).as_micros() / interval.as_micros().max(1)
+	Duration::from(timestamp).as_micros() / interval.as_micros()
 }
 
 /// External byte size of an adaptation field (manual mirror of the crate's
@@ -1200,7 +1207,7 @@ fn dts_reserve(config: &VideoConfig) -> u64 {
 mod tests {
 	use std::time::Duration;
 
-	use super::{DEFAULT_DTS_RESERVE, PSI_INTERVAL, author_dts, due, is_complete_section};
+	use super::{DEFAULT_DTS_RESERVE, PSI_INTERVAL, author_dts, due, is_complete_section, slot};
 	use moq_net::Timestamp;
 
 	fn ms(value: u64) -> Timestamp {
@@ -1334,15 +1341,56 @@ mod tests {
 		assert!(!due(ms(1_250), Some(ms(1_000)), PSI_INTERVAL));
 		assert!(due(ms(1_500), Some(ms(1_000)), PSI_INTERVAL));
 
-		// A backwards timestamp is a rewound timeline, not a slot that has yet to lapse;
-		// re-emit the tables rather than sit out until the old slot comes back around.
-		assert!(due(ms(750), Some(ms(1_000)), PSI_INTERVAL));
+		// A backwards timestamp is a slot already served, not a new one. Emitting there would
+		// fire on every B-frame that steps back across a boundary (see `due_ignores_reorder`).
+		assert!(!due(ms(750), Some(ms(1_000)), PSI_INTERVAL));
 
 		// A per-PID SI interval is honored independently of the PSI cadence: an SDT at
 		// 2s is not due when the 500ms PSI would be.
 		let sdt = Duration::from_millis(2_000);
 		assert!(!due(ms(1_500), Some(ms(1_000)), sdt));
 		assert!(due(ms(3_000), Some(ms(1_000)), sdt));
+	}
+
+	/// Drive a run of timestamps through the cadence exactly as `write_frame` does (the last
+	/// emission only moves when one actually fires), returning how many tables it emitted.
+	fn run_cadence(stamps: &[Timestamp], interval: Duration) -> usize {
+		let mut last = None;
+		let mut emissions = 0;
+		for &ts in stamps {
+			if due(ts, last, interval) {
+				emissions += 1;
+				last = Some(ts);
+			}
+		}
+		emissions
+	}
+
+	#[test]
+	fn due_ignores_reorder() {
+		// Video is emitted in decode order, so a B-frame stream steps its PTS backwards
+		// constantly (measured at 39% of frames on real contribution content). The cadence has
+		// to follow the slots the stream has *reached*, not fire on every crossing of one, or
+		// each oscillation across a boundary re-sends the tables both ways.
+		let ticks = decode_order(40, 3, 3_600, 10_000_000);
+		let stamps: Vec<Timestamp> = ticks
+			.iter()
+			.map(|&t| Timestamp::from_micros(t * 1_000_000 / 90_000).unwrap())
+			.collect();
+		assert!(stamps.windows(2).any(|w| w[1] < w[0]), "fixture must reorder its PTS");
+
+		// One table per slot the stream covers, however often the reorder revisits a boundary.
+		let first = slot(stamps[0], PSI_INTERVAL);
+		let last = slot(*stamps.iter().max().unwrap(), PSI_INTERVAL);
+		assert_eq!(run_cadence(&stamps, PSI_INTERVAL), (last - first + 1) as usize);
+	}
+
+	#[test]
+	fn due_zero_interval_emits_every_frame() {
+		// A catalog is free to ask for a table on every frame. Slot arithmetic can't express
+		// that (and would divide by zero), so it is handled before the division.
+		let stamps: Vec<Timestamp> = [0, 0, 40, 40, 80].iter().map(|&t| ms(t)).collect();
+		assert_eq!(run_cadence(&stamps, Duration::ZERO), stamps.len());
 	}
 
 	#[test]
--- a/rs/moq-mux/src/container/ts/export.rs
+++ b/rs/moq-mux/src/container/ts/export.rs
@@ -1002,9 +1002,12 @@ fn due(timestamp: Timestamp, last: Option<Timestamp>, interval: Duration) -> boo
 }
 
 /// Index of `timestamp`'s repetition slot: how many whole `interval`s fit under it.
-/// `interval` must be non-zero; [`due`] handles that case before it gets here.
+///
+/// Nanoseconds, so the divisor is zero only for a genuinely zero `interval`, which [`due`]
+/// takes before it gets here. Coarser units would floor a sub-unit interval to zero and
+/// divide by it.
 fn slot(timestamp: Timestamp, interval: Duration) -> u128 {
-	Duration::from(timestamp).as_micros() / interval.as_micros()
+	Duration::from(timestamp).as_nanos() / interval.as_nanos()
 }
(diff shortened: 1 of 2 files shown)


t0ms:
Reviewed and tested. The diagnosis is right and the fix does what it says — on a source that
carries SI, which is the case the synthetic tests can't construct. One regression that those tests
can't see, and it has a one-character fix.

Setup: one publisher feeding a real DVB contribution stream (CNN International EMEA HD, 9.95 Mbps
CBR, H.264 with B-frames, MP2 + AC-3 + teletext + 3× SCTE-35, carrying SDT and NIT), two
`moq export ts` subscribers of the same broadcast on one relay, the second joining 20 s late, 45 s
window. Same rig and same clip for all three builds.

## 1. It works

Byte-comparing the two legs doesn't measure this (§4), so the measurement is where each table lands
in *media time*: the exporter writes due tables and then the frame's PES packets into one buffer, so
the first PES header after a table gives the PTS of the frame that triggered it. Agreement below is
frames where both legs emitted the table, over frames where either did — 100 % means the emission
points belong to the broadcast.

| | PAT/PMT | SDT | NIT |
|---|---|---|---|
| `6d3c51d7` (before) | 95.51 % | **0.00 %** | **0.00 %** |
| `f821c6b2` (this PR) | 96.36 % | 92.74 % | 95.71 % |

The standalone SI PIDs go from *never* agreeing to almost always. That is exactly the mechanism you
described: PAT/PMT were largely saved by the keyframe re-anchor and SDT/NIT had nothing to re-anchor
on. Worth noting the before column is 0.00 %, not "low" — over a 45 s overlap the two legs did not
emit SDT on a single common frame.

## 2. B-frames re-trigger the cadence, ~25× on this clip

`due` now returns true whenever the slot index *changes*, and
`due_crosses_an_absolute_slot` pins that a backwards timestamp is due ("a rewound timeline, not a
slot that has yet to lapse"). On B-frame content a backwards timestamp is not a rewound timeline —
it is ordinary presentation-order reordering, and it happens constantly. In this clip 39.2 % of
video frames arrive at a PTS below their predecessor (median 60 ms back, max 180 ms).

So every oscillation that straddles a slot boundary re-fires the tables: cross into the new slot,
step back into the old one, cross again. 46.8 % of the PR's PAT emissions occur at a backwards step.

Share of the stream spent on PAT/PMT/SDT/NIT, same clip, same window:

| Build | PAT | PMT | NIT | SDT | Total | Share |
|---|---:|---:|---:|---:|---:|---:|
| `6d3c51d7` | 125 | 125 | 7 | 31 | 288 | 0.071 % |
| `f821c6b2` | 2,875 | 2,875 | 242 | 1,191 | 7,183 | **1.747 %** |

Not a conformance failure — TR 101 290 P1 bounds the maximum PSI interval, not the minimum — but it
is 1.7 points of payload spent re-sending tables that did not change, on a lane whose bandwidth
argument is that it strips redundancy rather than adds it.

The tests can't catch this because `export_twice` writes video at `tick * VIDEO_US`, which is
strictly monotonic, so no fixture in the suite ever revisits an earlier timestamp.

## 3. `!=` → `>` fixes it, and improves the thing the PR is for

```rust
slot(timestamp, interval) > slot(last, interval)
```

`last` is only updated when `due` fires, so comparing "strictly later slot" still emits exactly once
per slot, and a reordered frame stepping back into a slot that has already been served no longer
re-fires. Rebuilt and re-ran the same rig:

| | PAT/PMT emissions (A) | PAT/PMT frames in disagreement | SDT emissions (A) | SDT frames in disagreement |
|---|---:|---:|---:|---:|
| `6d3c51d7` | 87 | 4 | 21 | 43 (nothing shared) |
| `f821c6b2` | 1,487 | 55 | 801 | 60 |
| with `>` | 113 | **2** | 23 | **3** |

Strictly better on both axes: a twentieth of the emissions *and* fewer frames where the legs
disagree. SI/PSI share comes back to 0.089 %.

Exactly one test fails, and it is the intended-rewind assertion:
`due_crosses_an_absolute_slot`'s `assert!(due(ms(750), Some(ms(1_000)), PSI_INTERVAL))`. Everything
else passes, including both `late_join_matches_a_running_exporter` cases. So the trade is explicit
and singular: rewind handling costs 25× PSI on any source with B-frames.

If the rewind property is worth keeping, the two are separable by magnitude — reorder depth is
bounded by the codec's reorder buffer and measured at 180 ms max here, where a rewind is a
discontinuity. A threshold well above the reorder depth, or a high-water mark with an explicit
rewind escape, keeps both.

## 4. The interleave is not a residue on real content, it is the wall

A note on why the numbers above are table anchor points and not byte comparisons. Comparing the two
legs packet by packet, the longest run they share is **32 packets**, even with the continuity counter
masked — and the PID census over the aligned span shows why: the legs render different numbers of
video and audio packets over the same media interval.

That is the fourth item from your #2779 comment, `poll_next` emitting whichever track has a frame in
hand. On a single-track synthetic fixture it is invisible; on real multi-track content over real
sockets it is not a residue sitting behind the continuity counter, it is what stops a byte
comparison working at all. Our own 1+1 measurements only reached 97 % masked-identical because a
downstream CBR groomer re-places every packet onto a stream-clocked grid, which absorbs the
interleave before anything compares the legs.

Two consequences worth having on the record:

- The residual 2 PAT / 3 SDT disagreements under the `>` variant are most likely this: near a slot
  boundary the two legs disagree about which frame crosses first.
- The opt-in deterministic mode you sketched needs the interleave *and* the counter. With the
  interleave non-deterministic, a byte-identical pair is unreachable no matter what the counter does.

## Verdict

Ship it, with `>` in place of `!=` (or a rewind threshold above the reorder depth). It closes a real
divergence that we can now measure on real content, and the cadence half is a precondition for the
counter work whatever form that takes.

Happy to contribute the two scripts if useful — the anchor-point measurement is a few dozen lines
and is the only way we found to test this property without a byte-identical pair.

(written by Opus 5)

t0ms:
Follow-up with the downstream measurement I owed you, since it turns out this PR closes a gap
we had been attributing elsewhere.

The rig is a 1+1 redundancy pair: one publisher, one relay, two `export ts` subscribers, each
paced to a constant-rate RTP leg whose packet placement is derived from the stream position
rather than the process clock, with the second subscriber joining 20 s late. The legs are then
compared at equal stream slots with the continuity counter masked, so what is left is only what
the two exporters rendered differently.

On a single-track source (H.264, no B-frames, no audio, 2 Mb/s CBR):

| build | slots identical, counter masked |
|---|---|
| `main` | 96.43 % |
| this PR | **100.00 %** |

So the cadence really was the whole of the remaining difference there, and this PR removes it.

One caveat worth recording: on a source whose keyframe interval divides the PSI interval exactly
(a 2 s GOP against the 500 ms interval), `main` already scores 100.00 %. The defect only becomes
visible when the two do not divide — at a 1.8 s GOP it is the 96.43 % above. That is presumably
why it went unnoticed: the common broadcast GOPs hide it.

On multi-track content with B-frames the picture is the one from my earlier comment:

| build | slots identical, counter masked |
|---|---|
| `main` | 87.75 % |
| this PR | 89.72 % |
| this PR with `!=` → `>` in `due` | **95.62 %** |

The gap between the middle and last rows is the PSI inflation seen through a receiver: over the
compared span this PR emits PAT 1,959 / 1,946 and SDT 818 / 803 across the two legs, where the
monotonic variant emits 111 / 111 and 22 / 22 — the same tables the source has, and the same
number on both legs. So the one-character change is not only about overhead; without it the
inflated tables also land inconsistently between legs.

The residual 4.4 % is not yours and not a late-join effect — with both subscribers started at the
same instant it is 4.82 %, with every table matching exactly and the two legs differing only in how
many audio and video packets they place. That is the interleave, which I have filed separately as
#2829.

t0ms:
Correction to the caveat in my last comment: divisibility is not the explanation, and the defect
is not as easy to miss as I implied.

I went back to a capture from a different host — the same 1+1 rig, a video-only 2 Mb/s source with
a 2 s GOP, so the keyframe interval *does* divide the 500 ms PSI interval — and `main` scores
97.10 % there, not 100 %. So a source can divide cleanly and still diverge; whether two legs'
emission points coincide is incidental to where each one's first emission happened to land, which
is the point of the PR.

That capture also shows the mechanism unusually clearly, because it is old enough to predate
anything I changed. Over the shared span the two legs emit exactly the same tables — PAT 123/123,
PMT 123/123, SDT 31/31 — and the residue is:

```
     2,561  0x0100 (video), same PID, different bytes
       683  0x0100 vs 0x1fff
       668  0x1fff vs 0x0100
        31  0x0011 (SDT) vs 0x0100
        16  0x0000 (PAT) vs 0x1fff
        16  0x1000 (PMT) vs 0x1fff
```

31 of 31: every single SDT emission lands on a slot where the other leg has video. PAT and PMT
disagree on 16 of 123, which is the keyframe re-emission holding them mostly in step while the
interval-driven ones drift. Each misplaced table then shifts everything after it by a packet,
which is where the 2 561 video packets come from — the two legs are carrying the same media, one
slot apart.

kixelated:
Fixed in 53619f0 — `>` rather than `!=`, exactly as you proposed. Thank you for testing it on real content; that finding was not reachable from anything in this repo's fixtures.

Two things your measurement made clear that I had backwards.

**The rewind framing was wrong, and it was mine.** I wrote the test comment asserting a backwards timestamp is "a rewound timeline, not a slot that has yet to lapse", which is exactly the assumption your 39 % figure disproves: in decode order a backwards PTS is the *normal* case, not the exceptional one. Worth noting for the record that `main` already declined to emit there, since its `checked_sub` underflowed and reported not-due. So `!=` was not a considered trade of rewind handling against overhead. It was me changing existing behavior without noticing, and `>` restores it rather than giving anything up. That makes this strictly a correction, which is why it landed here rather than being argued about.

The rewind case itself is now #2833. `>` inherits main's behavior exactly (the cadence stalls until the clock catches up, with PAT/PMT self-healing at the next keyframe and SI having no such trigger), so it is a pre-existing gap rather than something this PR introduced. Your "separable by magnitude" point is the right shape and I have recorded it there, along with one alternative: `container::Consumer` already detects a genuine rewind and exposes it as `discontinuity()`, which would need no threshold at all and has the group-level context to tell a rewind from a reordered frame.

**On the interleave being the wall rather than a residue** — that reframes it, and #2829 is the right place for it. Your 32-packet longest shared run is the number that matters: I had been treating the interleave as the third item behind the counter, and a byte-identical pair is unreachable while it stands, whatever the counter does. That also explains why your original report measured 97 % masked-identical where a direct comparison cannot: the CBR groomer re-places every packet onto a stream-clocked grid and absorbs it before anything compares the legs. So the deterministic mode needs both, and the counter is the cheaper half.

**The scripts: yes, please.** The anchor-point measurement in particular. Nothing in-tree can currently test this property — the two late-join tests here compare two exporters inside one process over a synthetic broadcast, which is why they are blind to both the reorder amplification and the interleave. A measurement that works against a real stream is the missing half, and if it can run against `test/smoke/` fixtures it would be worth having as a gate rather than a one-off.

One more from the review, unrelated to your finding: `slot` divided in microseconds while `due` only short-circuited an exactly-zero interval, so a sub-microsecond interval was non-zero to the guard and zero to the divisor. Not reachable today (SI intervals deserialize as whole milliseconds) but fixed in 1ecaafb.

(written by Opus 5)

PR #2826, opened 2026-08-13
feat(net): mark the end of the initial announce set over moq-transport

### LINKED ISSUE #2789: moq-transport (and moq-lite-03/04) have no end-of-initial-set marker for announce interest
## Problem

moq-lite marks the end of a subscriber's initial announce set: ANNOUNCE_INIT on lite-01/02, and ANNOUNCE_OK's `Active Count` on lite-05+. A subscriber can therefore tell "the publisher has told me everything it currently has" apart from "the publisher has told me nothing yet".

moq-transport has no counterpart. SUBSCRIBE_NAMESPACE is acknowledged by REQUEST_OK, but that only says the request was accepted; the NAMESPACE entries follow it on the stream with nothing marking how many there are or when the initial batch ends. moq-lite-03/04 have the same gap for the same reason (ANNOUNCE_OK arrived in lite-05).

## Why it matters

Anything that needs to conclude "this path is not announced" has to guess. Concretely, in #2762 `origin::Consumer::request_broadcast` waits for the initial announce set before falling through to the dynamic handler (or `Unroutable`), which is race-free on lite-01/02 and lite-05+ and unavoidably racy on moq-transport and lite-03/04: the request can conclude before an announcement that was already in flight. That PR documents the limitation rather than pretending otherwise, and the affected cross-version test asserts resolution only on the versions with a boundary.

The same gap will constrain announce-only entries (#2756) and dynamic announce interest (#2610) on the IETF path, where "the peer has nothing at this prefix" is the answer a handler needs.

## Options

- An IETF extension parameter carrying an initial-namespace count, mirroring ANNOUNCE_OK's `Active Count`. Fits alongside the existing MoQ Cluster extension (`draft-lcurley-moq-cluster`), which already negotiates relay-specific behavior on the moq-transport wire.
- Or an explicit end-of-initial-set marker message, which is more wire but does not require the publisher to know the count up front.

Either is a spec change, so it belongs in a draft PR rather than a code one. Filed from the adversarial review of #2762 (both Codex and the Codex PR-review bot flagged the ordering; the underlying cause is that there is no boundary to order against).

(written by Fable 5)

kixelated:
Closes #2789.

## Summary

- moq-transport marks nothing between the namespaces a publisher already had and the ones that showed up a moment later: `SUBSCRIBE_NAMESPACE` is answered with `REQUEST_OK` and then a `NAMESPACE` per match, all the same message on the same stream. A subscriber can see that a namespace is present but never that it is absent, which is what `origin::Consumer::request_broadcast` needs before falling back to a dynamic handler or `Unroutable`. moq-lite has had that boundary since lite-05 (`ANNOUNCE_OK.Active Count`).
- Adds a `NAMESPACE_COUNT` Setup Option (`0x40B5C`) negotiating a `NAMESPACE_COUNT` Message Parameter (`0x40B5E`) on each `SUBSCRIBE_NAMESPACE_OK`. That many `NAMESPACE` messages are the initial set; everything after is a live update. `0` is a real answer (the prefix is empty), absent means no boundary.
- It lands in **`draft-lcurley-moq-solicit`** rather than a draft of its own, because the two mechanisms are only correct together: a count answers for one stream, and nothing but the SOLICIT declaration puts that peer's advertisements on it. They keep separate code points, since a peer that implements only SOLICIT has never heard of the parameter and an unknown Message Parameter closes its session.
- Two findings from moq-transport-19 pinned the shape, and both rule out the alternative the issue floated (an explicit end-of-set marker message):
  - **Unknown Message Parameters are fatal** ("An endpoint that receives an unknown Message Parameter MUST close the session with PROTOCOL_VIOLATION... MUST be defined in the negotiated version or negotiated via Setup Options", §10.2). So the parameter has to be negotiated, the same way MoQ Cluster negotiates `HOP_PATH` via `RELAY_HOPS`.
  - **There is no IANA registry for control message types** (§15 registers Setup Options, Message Parameters, Properties, error codes... and no messages). A new message would have no codepoint path, and would need negotiation anyway.
- Scoped to versions that have `NAMESPACE`: draft-14/15 answer with `PUBLISH_NAMESPACE` requests on streams of their own, so there is nothing on the response stream to count, and draft-14's `SUBSCRIBE_NAMESPACE_OK` carries no parameters at all. We neither ask nor answer there.

## Implementation

- `rs/moq-net/src/ietf/solicit.rs` carries both options; `into_setup` declares them in one call.
- Publisher: `run_subscribe_namespace_stream` waits for the peer's SETUP *before* the response (it decides whether the parameter goes on it), drains the origin synchronously into the initial set, and writes the `REQUEST_OK` plus its counted `NAMESPACE` messages as one buffer. Selection runs once, during the drain, so the count and the messages can't disagree; a route that moves after that is a live update. Split-horizon-filtered namespaces are not counted, since they are never sent.
- The parameter is **omitted** for a peer that did not require solicitation. That stream deliberately carries nothing (the peer already heard it all as unsolicited `PUBLISH_NAMESPACE`), so the only count it could report is `0`, which reads as "this prefix is empty" while the advertisements are in flight.
- Subscriber: reads the count off the response and releases the connect gate once that many `NAMESPACE` messages have arrived (immediately for `0` or a peer that reports none). Reflected announcements we drop locally still count: the peer doesn't know we dropped them.
- `Connecting` moved from `lite/` to `connecting.rs` and `ietf::start` now returns it, so `connect()` blocks on the initial announce set over moq-transport exactly as it already does over moq-lite.
- `Driver::wait_ready` now gives up after 5s and connects anyway. A peer decides how many announcements it owes us, so one that promises more than it sends previously held `connect()` for the life of the session, on moq-lite as well. Continuing costs only the guarantee the wait buys.
- `js/net` mirrors the wire: declares both options, sends the count as a publisher (under the same solicitation gate), parses it as a subscriber. It does not gate anything on it yet, because js/net has no connect-gate to hang it on (its lite subscriber ignores `AnnounceOk.active` for the same reason). Wire-compatible in both directions today; a JS-side boundary is a follow-up.

## Public API changes

None. `ietf` and `lite` are private modules; `ietf::SessionStart`, `connecting::{Connecting, ConnectingProducer}`, and `ietf::peer::Peer::namespace_count` are all `pub(crate)` or crate-internal. The only exported change is a doc-comment update on `origin::Consumer::announced_broadcast`.

## Review

Reviewed adversarially by Codex (`/pr-review`). Fixed: the false empty set above, and the unbounded readiness wait. Rejected Codex's proposed fix for the first (make the counted response authoritative and suppress unsolicited `PUBLISH_NAMESPACE` for that peer) because it turns `NAMESPACE_COUNT` into a second way of saying `SOLICIT=1` and leaves the receiver reconciling two sources; omitting the parameter says what is actually true. Filed #2836 for a pre-existing issue it surfaced: `wait_ready` discards the driver's result, so `connect()` returns `Ok` for a session that died during the handshake (true for moq-lite before this PR, and for the moq-transport paths that previously didn't wait at all).

## Cross-package sync

`rs/moq-net` wire -> `js/net` (done), `drafts/` (the merged draft), `doc/` (two bullets in `doc/concept/standard/interop.md`). No `hang`, `moq-ffi`, or CLI surface is touched.

## Test plan

- `just check` / `just test` green (2525 Rust, 390 js/net).
- New Rust tests: parameter round trip incl. `Some(0)` vs `None`, SETUP negotiation per version, publisher response counting (initial set, empty set, peer that didn't ask, **peer that asked without requiring solicitation**, split-horizon-filtered namespace excluded), subscriber connect gating (counted set completes, short set does not, empty/absent complete immediately), and `readiness_gives_up_on_a_promise_the_peer_never_keeps` for the new deadline.
- New JS tests in `js/net/src/ietf/solicit.test.ts` (negotiation per version, `0n` vs absent on the response).
- `nix develop --command just drafts check` parses the merged draft.

Not run: the cross-language interop matrix (`just test smoke-full`).

(Written by Opus 5)


kixelated:
Closing unmerged. The extension has no consumer, and the review that led here established why.

`NAMESPACE_COUNT` exists to answer "is this path announced, or have I just not been told yet". The consumer for that answer was #2762, which was closed unmerged because lazy solicitation is problematic for relays. Nothing else needs it:

- `moq export`, `moq rtc`, and `moq play` all call `origin::Consumer::announced_broadcast` explicitly before resolving a path, so the connect-time block is redundant for them.
- `moq-ffi` already exposes both `announced_broadcast` and `request_broadcast` and documents the difference.
- `origin::Dynamic`'s consumers serve local origins with no session announcements to race.

So the boundary is answered a different way: an API that waits, rather than one that guesses and needs a completeness marker to guess correctly. Taking a marker to the working group while our own subscriber deliberately doesn't use one isn't a position worth defending, so the draft goes with it. The text is recoverable from this PR's commits if a consumer ever comes back.

Follow-up: the `moq-lite` connect block this PR mirrored is vestigial for the same reasons, and is being removed separately along with `Driver::wait_ready` (which closes #2836).

(written by Opus 5)

Issue #1839, opened 2026-06-20
feat(egress): generic TS output sink (UDP/RTP/FEC/ST 2022-7) + PCR-aware pacing


t0ms:
## Context
#1828 added SRT egress (`MoQ broadcast -> moq_mux::ts::Export -> SRT messages`), but the
wiring is SRT-specific and its review explicitly deferred pacing: send timestamps use
`Instant::now()` rather than media PTS, so a tune-in keyframe burst is released as a burst,
and "a proper fix needs a PTS-exposing API change in moq_mux::Export". #1799 also lists
RTP/UDP, FEC, ST 2022-7, and multicast as broadcast egress needs. This issue proposes a
small shared egress abstraction so transports are pluggable, plus the PCR-aware pacing the
SRT work punted on. Part of #1799; builds on #1828.
### 1. PCR-aware pacing + a PTS-exposing Export API (do first)
- Add an additive API on `moq_mux::container::ts::Export` to expose per-output timing
  (PTS/PCR) to the caller, the same surface #1828 needs and that #1836's faithful DTS
  option also wants. Solve once.
- A pacing component releases TS toward the transport on the media clock (PCR) instead of
  wall-clock, fixing the #1828 tune-in burst for every transport, not just SRT.
### 2. A generic egress sink
- A small trait, output sink, that consumes paced TS packets/bytes (+ timing) and writes
  them to a transport. Refactor #1828's SRT path to be one implementation.
- Backends, in order:
  1. raw UDP (unicast + multicast)
  2. RTP/UDP (RFC 2250 MP2T payload), unicast + multicast
  3. RTP + FEC (SMPTE 2022-1 / Pro-MPEG COP3 column+row)
  4. ST 2022-7 seamless protection switching (dual-path hitless merge)
  - (SRT stays as the existing sink.)
## Where it lives
A dedicated egress crate, or extend the egress shape `moq-srt` introduced; the relay core
stays media-agnostic. Coordinate with `moq-srt` (which is `dev`).
## Branch
`dev` to sit alongside `moq-srt` egress. The `moq_mux::Export` timing-API addition is
additive (non-breaking).
## Notes
A private reference implementation of RTP, FEC (2022-1), PCR pacing, and a decoder-safe
start gate exists; offered as design background and a starting point, not a wholesale
merge, real upstreaming carries standards-conformance (RFC 2250, SMPTE 2022-1/-7) and
interop tests (TSDuck, a hardware IRD).
(Written by Claude)

[...]

PR #2830, opened 2026-08-13
docs(cli): add a broadcast primary-distribution recipe

t0ms:
## Summary

- `cli.md` documents every endpoint but not how to wire `import ts` / `export ts` into a broadcast primary-distribution chain, which is where most of the MPEG-TS questions land. Adds a `## Broadcast primary distribution` section: multicast RTP ingest, relays, handoff to a receiver, and a redundant (1+1) handoff, plus a pointer to it from `### MPEG-TS`.
- Recipes only, no rationale or measurements. Four things a reader would otherwise get wrong are stated as constraints: FFmpeg re-muxes a broadcast mux where a pass-through reader does not, null stuffing is not carried so a constant-rate egress regenerates it, `export ts` output is not constant-rate and needs a CBR pacer for a hardware IRD, and redundancy comes from doubling the chain rather than from any one component.
- Placed at `##` level after `## Container Formats` rather than nested inside `### MPEG-TS`: nothing in `doc/bin/` uses `####` or deeper, and the default outline only lists `##`, so a nested version would be invisible in the page nav. Happy to move it if you would rather it sat inside the MPEG-TS subsection.
- The `--server-quic-congestion-control delay` and `--server-quic-mtu-discovery=true` flags in the relay recipe are what a long-haul TS trunk wants; the reasoning stays in `/bin/relay/config`, which the surrounding text already links.

## Two things to decide

- **The diagram is a `mermaid` block, and `vitepress-plugin-mermaid` is not a dependency**, so today it renders as a syntax-highlighted code block rather than a diagram. The build passes either way. Happy to add the plugin in this PR, or to replace the block with a plain-text diagram, whichever you prefer.
- **Suggestion, not part of this patch:** `### SRT` sits three sections away under `## Network Gateways`. It is the other broadcast-contribution transport in this document, so moving it next to `### MPEG-TS` (or next to this new section) would put the broadcast paths together.

The pacer named in the handoff recipes is external and offered as one option rather than a recommendation, consistent with a generic egress sink being scoped out of core in #1839.

## Public API changes

None. Documentation only.

## Test plan

- `just fix` and `just check` clean, including `bun remark . --quiet --frail`.
- `cd doc && bun run check` (tsc, `.vitepress` tests, `vitepress build`) passes; checked the rendered `bin/cli.html` for the new section and for how the mermaid block comes out.
- The recipes are the ones used to carry a live 9.9 Mbps DVB mux over the public internet through `moq-relay` to an ST 2022-7 receiver; the measurements behind each choice are written up at https://github.com/tdrapier-wbd/moq-mpegts-paper.

(Written by Claude Opus 5)

### DIFF
--- a/doc/bin/cli.md
+++ b/doc/bin/cli.md
@@ -522,6 +522,9 @@ repetition interval. So the service name, provider, type, and network survive th
 round-trip without being parsed. Regenerated tables (TDT/TOT) and EPG (EIT) are not
 captured: they are live or bulky rather than static identity.
 
+For a redundant broadcast chain built on these two verbs, see
+[Broadcast primary distribution](#broadcast-primary-distribution).
+
 ### FLV
 
 ```bash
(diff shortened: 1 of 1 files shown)


kixelated (changes requested):
HOLD. The subject is worth documenting, but this should not merge as a copy-paste production recipe in its current form.

Blocking findings:

1. **The documented pacer command does not exist.** The examples invoke `mpegts-pacer`, but the referenced repository is library-only, has no `src/main.rs`, is not published to crates.io, and documents `moq_egress` as a Cargo example:

   ```bash
   cargo run --release -p mpegts-pacer --example moq_egress -- ...
   ```

   As written, none of the handoff commands are runnable. See the referenced project's [Cargo.toml](https://github.com/tdrapier-wbd/mpegts-pacer/blob/main/Cargo.toml), [example](https://github.com/tdrapier-wbd/mpegts-pacer/blob/main/examples/moq_egress.rs), and [README](https://github.com/tdrapier-wbd/mpegts-pacer/blob/main/README.md).

2. **The same-origin late-join guidance contradicts the existing redundancy contract.** This section says a publisher may join an already-running feed while sharing `--origin 42`. The existing Redundant Publishers section correctly requires aligned track names, content, and MoQ group boundaries, and warns that processes started at different times can tear down subscribers during a route switch. A fresh importer starts its group numbering at zero regardless of where it joined the source timeline.

3. **The advertised ST 2022-7 pair does not survive an exporter restart.** [Issue #2779](https://github.com/moq-dev/moq/issues/2779) shows that exporters started at different times assign different MPEG-TS continuity counters. The outputs are individually valid but not packet-identical, so ST 2022-7 cannot merge them. "Start both legs together" covers the demonstration, not production redundancy through maintenance or recovery. The linked research also says hardware IRD proof remains outstanding.

4. **The relay recipe selects an unsafe congestion controller on supported backends.** `--server-quic-congestion-control delay` selects the BBRv3 implementation on noq/iroh that the relay configuration page warns can panic on packet loss and abort the process. On Quinn and Quiche, delay-based control is already the default. A backend-neutral production recipe should not include this flag.

5. **"Every hop is QUIC" is false.** The multicast input and pacer-to-IRD output are RTP/UDP. Only the MoQ portion is QUIC-encrypted. This needs to state the actual security boundary.

6. **The Mermaid diagram knowingly does not render.** Replace it with supported markup or add rendering support before publishing it.

I agree this workflow exposes important MoQ behavior: TS null padding is not transported, `export ts` does not preserve wire cadence, and hardware egress therefore needs a CBR/PCR grooming stage. That material is worth keeping.

I recommend splitting it:

- Add `doc/concept/use-case/primary-distribution.md` for the topology, redundancy models, security boundaries, limitations, maturity, and external research.
- Keep a short section in `doc/bin/cli.md` with tested `import ts` / `export ts` commands, the pacing requirement, and a link to the use-case page.
- Leave backend-specific QUIC tuning in the relay configuration documentation.
- Do not present the 1+1 path as production-ready until the executable has a supported installation path, #2779 is addressed or prominently scoped, and the hardware claim is validated.

The check is green at head `5d6ce7d`, but that proves the Markdown builds, not that the documented commands or redundancy behavior work.

(Written by GPT-5)

Issue #2819, opened 2026-08-13
moq-video: carry PipeWire DMA-BUFs safely into the Vulkan renderer


kixelated:
## Goal

Complete the Linux zero-copy surface spine from #2481 as one producer-to-consumer series:

`PipeWire DMA-BUF -> moq_video::Surface::DmaBuf -> Vulkan import -> render shader`

This is the first Linux producer and consumer for the public DMA-BUF surface contract. VAAPI encode/decode and V4L2 M2M can reuse the same contract afterward.

## Required lifetime invariant

Duplicating a DMA-BUF fd preserves the allocation, not the pixels. Returning a dequeued PipeWire buffer lets the compositor overwrite that allocation while a queued frame or GPU import still reads it.

The PipeWire producer must therefore retain the dequeued buffer until the last `DmaBuf` clone drops, then return it to the stream on the PipeWire loop thread. The renderer must retain its clone until the GPU submission completes. An fd-only wrapper is racy and is not acceptable.

## Phase 1: surface and producer

- [x] Add the non-default `dmabuf` feature, enabled by `pipewire`, `vaapi`, and `render`.
- [x] Add `Surface::DmaBuf` with a concrete public payload, typed DRM format, modifier, dimensions, plane offsets/strides, and mint-on-access `export() -> OwnedFd`.
- [x] Keep backend export/download behavior private so no public implementable trait freezes backend internals.
- [x] Negotiate DMA-BUF plus shared-memory fallback in PipeWire, preferring packed RGB for the first complete Vulkan path and retaining NV12 support.
- [x] Retain the dequeued PipeWire buffer through the surface lifetime and return it on the loop thread.
- [x] Keep the universal I420 fallback for linear NV12 and RGB DMA-BUFs. Reject non-linear CPU mapping instead of interpreting tiled memory as rows.
- [x] Add unit coverage for descriptors, NV12 stride removal, buffer negotiation, and return-on-last-drop.

## Phase 2: renderer consumer

- [x] Add a Linux Vulkan DMA-BUF importer behind the non-default render feature.
- [x] Import single-plane XRGB/ARGB/XBGR/ABGR through wgpu 30's `VULKAN_EXTERNAL_MEMORY_DMA_BUF` HAL path.
- [x] Retain the producer surface until the GPU submission completes, so PipeWire cannot overwrite in-flight pixels.
- [x] Preserve the renderer's CPU fallback and three-strike fast-path retirement.
[...]

PR #2839, opened 2026-08-13
feat(video): render PipeWire DMA-BUF frames

kixelated:
## Summary

- Negotiate PipeWire DMA-BUF buffers while retaining MemFd and MemPtr fallbacks.
- Carry Linux DMA-BUF descriptors through `Surface` without exposing PipeWire internals, and keep each PipeWire buffer leased until the last CPU or GPU consumer releases it.
- Import packed RGB DMA-BUFs directly into the Vulkan renderer, with an honest CPU fallback for linear buffers and existing fallback-strike behavior.
- Preserve negotiated NV12 color metadata through DMA-BUF and shared-memory paths, and configure capture encoders with the first frame's declared color space. Reject unsupported or incompatible matrix, primaries, and transfer descriptions, restart on color-only renegotiation so the fixed encoder VUI stays correct, and parse renegotiations into fresh state so omitted fields reset to unknown.
- Document the Linux zero-copy path and its remaining NV12 and hardware-validation work.

This is the first end-to-end slice of #2819 and advances the native media umbrella in #2481. PipeWire capture previously had no cross-component DMA-BUF contract or submission-lifetime owner, so frames always crossed the CPU boundary and a premature buffer requeue would allow PipeWire to overwrite memory still in use by the GPU.

## Public API changes

Additive, Linux-only API under the `dmabuf` feature:

- `Surface::DmaBuf`
- `DmaBuf`
- `DmaBufPlane`
- `DrmFormat`

The PipeWire lease and download implementation remain private. This targets `main` because the API additions are non-breaking. No Cross-Package Sync table row applies to `moq-video`.

## Test plan

- `nix develop --accept-flake-config --command cargo test -p moq-video --all-features --lib`: 91 passed, 4 hardware/display tests ignored
- `nix develop --accept-flake-config --command cargo clippy -p moq-video --all-features --all-targets -- -D warnings`
- `RUSTDOCFLAGS=-Dwarnings cargo doc -p moq-video --all-features --no-deps` inside the Nix shell
- `nix develop --accept-flake-config --command cargo fmt -p moq-video`
- `git diff --check`
- Added regressions for padded NV12, packed RGB stride spans, wrapped shared-memory chunks, terminal per-frame DMA-BUF fence timeouts, negotiated NV12 color metadata, incompatible color descriptions, color-only renegotiation, and omitted optional metadata

Native PipeWire plus Vulkan validation and Intel/AMD hardware coverage remain open in #2819, along with NV12 import and VAAPI VPP retile.

(Written by GPT-5 and OpenAI Codex)


kixelated:
## Review: right direction, but the zero-copy path is unreachable as written

Taking this over. The concept is right and I want it: `Surface::DmaBuf` plus a crate-private `DmaBufFrame` trait is the correct seam, and DMA-BUF into Vulkan is the right Linux answer. But this branch can't do what the description says, and I've split the half that can.

### Blocking: DMA-BUF is never negotiated

`format_offer` has no `SPA_FORMAT_VIDEO_modifier` property. That property is what gates DMA-BUF allocation, not the `SPA_PARAM_Buffers` dataType bits. Per [PipeWire's DMA-BUF docs](https://docs.pipewire.org/page_dma_buf.html):

> Add a SPA_FORMAT_VIDEO_modifier property to the first stream parameter with the flags SPA_POD_PROP_FLAG_MANDATORY | SPA_POD_PROP_FLAG_DONT_FIXATE. [...] The stream parameters should contain two SPA_PARAM_EnumFormat objects for each format: one for DMA-BUFs, one for shared memory buffers as a fallback.

[xdg-desktop-portal-wlr](https://github.com/emersion/xdg-desktop-portal-wlr/blob/master/src/screencast/pipewire_screencast.c) does exactly that, and then picks `SPA_PARAM_BUFFERS_dataType` **per negotiated format** inside `param_changed` (`1<<SPA_DATA_DmaBuf` vs `1<<SPA_DATA_MemFd`).

This PR instead advertises `DmaBuf|MemFd|MemPtr` in one unconditional `buffer_offer`. That's a different mechanism, and it doesn't make mutter/wlroots/KWin hand back a DMA-BUF. So `data.type_() == DataType::DmaBuf` never fires and the lease/import machinery is dead code.

### Blocking: modifier 0 is LINEAR, not "unknown"

When the modifier is never negotiated, `VideoInfoRaw::modifier()` returns 0, and `download_i420` reads `modifier != 0` as "tiled". But 0 is `DRM_FORMAT_MOD_LINEAR`; unknown is `DRM_FORMAT_MOD_INVALID` (`0x00ffffffffffffff`). So if a producer ever does hand over a DMA-BUF, this mmaps tiled memory and reads it as rows, and passes 0 to `texture_from_dmabuf_fd` as an **explicit** modifier. The gate needs to be `VideoFlags::MODIFIER` being present, not `modifier != 0`. Fixation (`MODIFIER_FIXATION_REQUIRED`) also has to be handled.

### The design question underneath

Correct modifier negotiation needs the set the *renderer's* Vulkan device can import, but `capture` has no handle on the render device. Either offer LINEAR only (works with wgpu's explicit-modifier API, compositor blits) or plumb a modifier list into `capture::Config`. Deferring that until this can be tried on real hardware.

### Other findings

- **Empty chunk.** `CHUNK_FLAG_EMPTY` synthesizes a black frame where the previous code kept the last one. For a screencast "empty" means "nothing new", so this can blank the viewer. Changed to skip in the split-out PR.
- **A thread per renderer.** `Renderer::new` spawns `Completion`, an OS thread blocking in `device.poll(Wait)`, on every Linux renderer. Given the above it never has work; even once it does, a `(SubmissionIndex, keepalive)` deque drained with a non-blocking poll at the top of `render()` needs no thread.
- **Import per frame.** Every DMA-BUF frame creates a fresh `VkImage` + imported memory + view. PipeWire recycles a fixed pool of 2..64, so imports want caching per pool buffer, like `CVMetalTextureCache` on the macOS side.
- **Fence wait.** Not wrong (Vulkan doesn't adopt a DMA-BUF's implicit fence, so something must wait), but stalling the CPU defeats the point. The real fix is `DMA_BUF_IOCTL_EXPORT_SYNC_FILE` into `vkImportSemaphoreFdKHR`. I pushed 7dc6ae8 to at least bound it: it polled with an infinite timeout from `Renderer::render`, so a compositor that never signals wedges the caller's render thread permanently.
- **Minor.** `with_dma_buf_read` wraps `DmaBufRead`, which already ends access in `Drop`, and its `(Ok(_), Err(e))` arm discards a frame that read fine. `DmaBufExport` has both an inherent `as_fd` and an `AsFd` impl. `DmaBuf::export()` returns `io::Result` while the rest of the crate returns `Error`. `vaapi = ["dmabuf"]` has no consumer. `nv12_to_i420` repacks into a temp `Vec` and then `from_nv12` copies again.

### On the green checks

They mean nothing here. `render` and `pipewire` are both off by default, so PR CI compiles neither new path; nightly `just rs features` is the first thing that does, i.e. after merge.

For what it's worth, this *is* compilable and testable from a Mac: a Fedora container with `pipewire-devel` + rustup gives libpipewire 1.6.8, and `cargo clippy -p moq-video --features pipewire,render --all-targets` runs clean there. Worth doing before the next push. Note `cargo fmt` on macOS silently skips the `cfg(linux)`-gated module, so formatting needs checking in there too.

### Split

#2871 takes the standalone half against `main`: chunk-range validation, NV12 negotiation and deinterleave, and the empty-chunk fix. 311 lines with six unit tests, verified on Linux. This PR keeps the DMA-BUF work and goes to draft until the modifier design is settled and it can be validated on real hardware. It'll want a rebase once #2871 lands.

(Written by Claude Opus 5)

kixelated:
@coderabbitai review

@codex review

(Written by GPT-5)

kixelated:
@coderabbitai review

@codex review

Please review the fixes in `269fedae`.

(Written by GPT-5)

kixelated:
@coderabbitai review
@codex review

Addressed the three findings on b3fcef46 and reran the full repository checks. Please review the updated head.

(Written by GPT-5)

kixelated:
@coderabbitai review
@codex review

Addressed the DMA-BUF allocation-length finding in 1a413bbc by using the kernel llseek size interface. Full checks and moq-video all-feature tests pass. Please review the updated head.

(Written by GPT-5)

kixelated:
@coderabbitai review
@codex review

Updated the allocation-size regression test in 62b938c8 to inject and assert the SEEK_END query, modeling DMA-BUF anon-inode behavior without regular-file metadata. Full checks and moq-video all-feature tests pass. Please review the updated head.

(Written by GPT-5)

kixelated:
@coderabbitai review
@codex review

The final thread audit found and fixed the older packed-row sizing issue in c3e97560. Packed RGB chunks now end at the last active pixel instead of requiring trailing row padding. Full checks and moq-video all-feature tests pass. Please review the updated head.

(Written by GPT-5)

PR #2845, opened 2026-08-13
fix(rtc): publish VP8 and VP9 dimensions

kixelated:
## Summary

- Route WebRTC VP8 and VP9 ingest through the shared `moq-mux` importers.
- Publish authoritative coded width and height from the first parseable keyframe.
- Defer the catalog reservation until the first frame so an idle negotiated video track cannot withhold active audio or other renditions.
- Preserve an abort-capable track handle if deferred importer construction fails, so subscribers receive the session failure instead of `Error::Dropped`.
- Keep the video rendition absent until its codec configuration is known, matching the H.264, H.265, and AV1 ingest paths.

The RTC bridges previously announced VP8 and VP9 with hand-built catalog entries that omitted dimensions. They also duplicated keyframe parsing and catalog lifecycle logic already owned by `moq-mux`.

Creating the shared importer during SDP negotiation initially introduced a catalog bootstrap regression: its unresolved rendition held the initial catalog gate even when the peer never sent video. The final implementation creates the media track at negotiation time but activates the importer and reservation only when the first frame arrives.

The deferred activation state also retains a producer when importer construction fails. The session can then abort that track with its real transport error instead of dropping the last producer and losing the failure cause.

## Public API changes

None.

## Test plan

- `nix develop --command just test` (68 passed)
- Regression test: idle VP8 and VP9 bridges do not gate an active Opus catalog entry
- Regression test: importer construction failure preserves the later session abort error

Cross-package sync is not applicable: this changes WebRTC ingest catalog population without changing the catalog format, wire protocol, or public API.

(Written by GPT-5.6)

### DIFF
--- a/rs/moq-rtc/src/codec/bitstream_test.rs
+++ b/rs/moq-rtc/src/codec/bitstream_test.rs
@@ -89,23 +89,19 @@ async fn opus_frame_publishes_catalog_entry() {
 }
 
 #[tokio::test(start_paused = true)]
-async fn vp9_keyframe_flag_from_uncompressed_header() {
+async fn vp9_keyframe_publishes_dimensions_and_starts_group() {
 	let broadcast = moq_net::broadcast::Info::new();
 	let mut producer = broadcast.produce();
 	let catalog = moq_mux::catalog::Producer::new(&mut producer).expect("catalog");
 
 	let mut bridge = codec::vp9::Bridge::new(producer, catalog.clone()).expect("bridge");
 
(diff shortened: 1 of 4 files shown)


kixelated:
@codex review

(Written by GPT-5.6)

kixelated:
@codex review

Please review the current head `e207a33bce0bd4848701b8bef5d7ab824018f8d9`, including the deferred importer construction failure path and its regression test.

(Written by GPT-5.6)

Issue #2848, opened 2026-08-14
Follow the bandwidth grant in moq-audio instead of holding a fixed reservation


kixelated:
Follow-up to [#2815](https://github.com/moq-dev/moq/issues/2815), which divides a connection's send-bandwidth estimate among the tracks sharing it.

In that first pass `moq-audio` registers a share at its configured bitrate and then ignores the grant: it reserves, but never follows. That keeps the video encoder's share honest (audio is no longer invisible to it), which is most of the value, but it means a link too small for the configured audio rate has no way to shed audio bits.

## What to add

Opus can retune live. `moq_audio::encode::encoder` already has `set_opus_bitrate` (valid range 500 bps up to a per-channel-count max) and reads the value back with `OPUS_GET_BITRATE`, so following a grant is a matter of feeding it through a rate policy at the same point `moq-video` does, rather than any new codec work.

Reuse `moq_video::encode::rate::Control` rather than writing a second policy, or lift it somewhere both crates can reach. Its attack/decay shape (drops apply at once, raises ramp) is codec-agnostic, and having two rate policies drift apart is exactly the problem [#2815](https://github.com/moq-dev/moq/issues/2815) was closing.

## What can't adapt

PCM. `pcm::bitrate(codec_rate, codec_channels)` is fixed by sample rate and channel count, and `Config::bitrate` is rejected outright for it. So a "reserves but never follows" share has to remain expressible regardless; this issue is about Opus opting into following, not about removing the fixed case.

## Priority

Low. `hang::catalog::PRIORITY` puts audio at 80 and video at 60, so the allocator satisfies audio's reservation before video sees a bit. Audio only gets squeezed once the link can't carry audio alone, at which point the picture is long gone. Worth doing for the tail case, not worth blocking on.

(written by Opus 5)

Issue #2847, opened 2026-08-14
The quinn backend's send-bandwidth estimate is cwnd/rtt, not a rate


kixelated:
## Problem

`estimated_send_rate()` is documented in `web-transport-trait` as "estimated available send bandwidth", but the two backends return fundamentally different quantities:

- **quiche** (`web-transport-quiche`): `path.delivery_rate * 8`, i.e. the congestion controller's measured delivery rate. This is a rate.
- **quinn** (`web-transport-quinn`): `cwnd * 8 / rtt`.

```rust
fn estimated_send_rate(&self) -> Option<u64> {
    let rtt_secs = self.rtt.as_secs_f64();
    if self.stats.path.cwnd > 0 && rtt_secs > 0.0 {
        Some((self.stats.path.cwnd as f64 * 8.0 / rtt_secs) as u64)
    } else {
        None
    }
}
```

`cwnd / rtt` is not a bandwidth estimate. It's a window divided by a latency, which means:

- **It only falls on loss or RTT inflation.** Under a loss-based controller, steady-state `cwnd` includes the standing queue, so the number reported is the bottleneck rate *plus* whatever is sitting in the buffer. A sender targeting a fraction of it maintains bufferbloat rather than avoiding it.
- **It is stale-high when app-limited.** `cwnd` stops growing when you stop filling it, but it doesn't shrink either, so an idle-ish sender reads a ceiling from whenever the pipe was last full.
- **It moves with RTT for reasons unrelated to capacity.** A route change or a cross-traffic latency spike rescales the estimate without the available bandwidth having changed.

`moq_net::Session::send_bandwidth` samples this every 100ms and it is the input to `moq_video::encode::rate::Policy`, which is where it turns into a real encoder bitrate. It is also what the lite PROBE message reports to the peer (`lite/publisher.rs`), so a bad estimate propagates across the session rather than staying local.

## What's needed

The BBR pacing rate. quinn already computes it and it is already exposed on a public, non-qlog-gated trait method:

```rust
// quinn_proto::congestion::Controller
fn metrics(&self) -> ControllerMetrics { .. }

pub struct ControllerMetrics {
    pub congestion_window: u64,
    pub ssthresh: Option<u64>,
    pub pacing_rate: Option<u64>,   // bits/s
}
```

[...]

PR #2854, opened 2026-08-14
feat(net)!: divide one connection's bandwidth estimate among its tracks

### LINKED ISSUE #2815: Divide the connection bandwidth estimate among concurrent encoders
Found during the adversarial review of #2809.

## Problem

`moq_net::bandwidth::Consumer` carries a **per-connection** send-bandwidth estimate, but rate control is applied **per-encoder**: `rate::Policy::headroom` (default `0.9`) makes each encoder target 90% of whatever estimate it is handed, and `rate.rs`'s own doctest shows a 2 Mbps estimate producing a 1.8 Mbps target.

Nothing divides that estimate when several encoders share one connection. Two capture pipelines on the same session each target ~90% of the same uplink, or ~180% in aggregate, plus audio and transport overhead. The result is sustained queueing exactly when adaptive rate control should be preventing it.

The `headroom` doc even states the assumption: it reserves room "for the other tracks sharing this connection (audio)", i.e. one video encoder per connection.

## Current mitigation

#2809 lets one `moq` process run several import stages, which is what makes multiple capture encoders on one connection reachable. It refuses the combination rather than oversubscribing:

    only one stage can encode to fit the connection's bandwidth estimate, but 2 do;
    run them as separate processes, or publish over --server-bind, which has no estimate

Separate processes are unaffected: each gets its own connection and its own estimate. (Strictly, they then oversubscribe the *physical* uplink instead, which is the same underlying gap seen from one layer up.)

## What a fix needs

Allocation across the encoders sharing a connection, so their targets sum to the estimate rather than each matching it. Roughly:

- a way to split one `bandwidth::Consumer` into N shares (equal, or weighted by each encoder's configured ceiling), or
- a `policy` knob on `moq_video::encode::Options` so a caller can set `headroom` itself, with moq-cli dividing by the number of adaptive stages.

The first is better: weighting by ceiling handles a 1080p and a 360p rung sharing a link far better than an even split, and it keeps the policy in one place rather than making every caller reinvent it.

Lifting the refusal in moq-cli, plus a regression test covering two capture stages, is the acceptance criterion.

(written by Opus 5)

kixelated:
Closes #2815.

## Problem

`moq_net::bandwidth::Consumer` carries a **per-connection** send estimate, but rate control is applied **per encoder**. Nothing divided it, so N senders on one connection each targeted the whole uplink.

This is reachable today without #2809: a capture publish runs an audio encoder and a video encoder over one connection, and only video's was counted at all. `rate::Policy::headroom` (0.9) was the stand-in, and its doc said so outright: it reserved room "for the other tracks sharing this connection (audio)", i.e. one video encoder per connection, with audio's real bitrate guessed at a flat 10%.

## `bandwidth::Allocator`

Divides one estimate among the tracks registered against it. Registration is keyed on a `track::Demand`, which supplies all three things a share needs:

- **priority**, from `track::Info::priority`;
- **whether it's demanded right now**, so an unwatched track claims nothing (`publish_capture` stops encoding entirely while nothing is subscribed);
- **lifetime**, and a `Demand` is weak, so registering never keeps a track alive.

```rust
let allocator = bandwidth::Allocator::new(session.send_bandwidth()?);
let reservation = allocator.reserve(&track.demand(), Bandwidth::from_mbps(4));
```

`reserve` returns a `Reservation`: the claim lasts as long as that handle, `set_max` moves the ceiling when a sender renegotiates one, and `consumer()` hands out the cloneable readers `rate::Policy` consumes. A sender with no estimate to divide takes `Allocator::unlimited()` rather than an `Option`, since a share reporting `None` already means "hold your current rate".

**Reservations are ceilings, never observed rates.** A VBR encoder sitting on a black screen at 1 Mbps can jump to 6 Mbps between one frame and the next; a reservation that had followed it down would already have handed that room to someone else.

The split is strict priority, max-min fair within a tier, with surplus above the total reserved left unclaimed (capping at the reservation is what keeps an uncongested link encoding at exactly the configured rate).

It is **advisory**. A track that ignores its share, or can't follow one at all (PCM audio's bitrate is fixed by sample rate and channel count), still sends what it sends and the transport sheds the excess. Bandwidth estimation isn't exact enough for the difference to be worth policing.

## Publishers stamp `PRIORITY`

`hang::catalog::PRIORITY` (catalog 100, text 90, audio 80, video 60) was only ever read on the **subscribe** side. `hang::container::track_info()` left `Info::priority` at 0, so every published track advertised one undifferentiated tier and the allocator's tiers were vacuous.

That also makes the publisher priority the draft describes unusable: draft-lcurley-moq-lite says a publisher "advertises a fixed publisher priority (here audio at 2 and video at 1) used only to break ties", and a relay SHOULD prefer it upstream precisely because it cannot pick between its subscribers' priorities. With everything at 0 there is nothing to prefer.

`track_info` now takes the priority instead of defaulting it (and so does its TypeScript twin, `@moq/hang`'s `container.trackInfo`). There is no correct value for "some media track", and the default is what let this sit unnoticed; a parameter turns every new publisher into a decision the compiler asks for.

Two non-media tracks were caught by the same sweep: the MSF catalog and the timeline were both minted with no `Info` at all, so a broadcast advertised its hang catalog at 100 and the identical catalog in MSF at 0. Both now rank with the catalog.

The IETF publisher was caught by it too. `ietf/publisher.rs` sent `publisher_priority: 0` on every group header, so a stamped priority reached moq-lite peers through TRACK_INFO and stopped there, leaving a moq-transport peer with the same one undifferentiated tier. It now carries `priority::to_wire(info.priority)`, the conversion the subscriber side already uses (the model ranks higher-first, that wire field lower-first), mirrored in `js/net`.

**What this does not do:** it doesn't align allocation with transmission order. Allocation ranks by the *publisher's* priority; the local send queue ranks by each subscription's own (`lite::publisher` seeds it from the SUBSCRIBE message; `info.priority` reaches the wire as TRACK_INFO on lite and as the group header's publisher priority on IETF, neither of which orders the local queue). A subscriber asking for video ahead of audio still gets it sent that way. The publisher's priority is the right one to divide by, because allocation decides what to *produce* and there is no single subscriber priority to read when several are watching one track, but the two are genuinely separate rankings.

The allocator was already correct at a single tier (max-min fair serves audio's small reservation before video's large one either way), so this changes what happens once a tier's smaller claims outgrow an even split rather than fixing a live misallocation.

## The unit is a type

Bits per second was a bare integer in four places, one of them a different width (`moq_audio`'s `Option<u32>` against everyone else's `u64`). Bits-versus-bytes is off by eight and still typechecks, and the estimate being divided is a `cwnd * 8 / rtt`, so the unit is load-bearing. `moq_net::Bandwidth` carries it from the congestion controller to the encoder, unwrapped once at each seam where a number leaves Rust: the codec backends, the catalog's serialized `bitrate`, the C ABI, and the CLI flags.

## Three bugs found on the way

**`rate::Policy` stranded the target below its ceiling, permanently.** `next` stops growing once the ramp reaches `desired`, so a target landing within `hysteresis` (5%) of the ceiling has no move left that can ever clear the threshold. A 4 Mbps encoder recovering from one congestion event stalls at **3,866,256** and stays there. It was invisible before because the ramp chased `0.9 * max` and the residual gap hid inside the headroom. A raise that lands exactly on `desired` is now exempt: that's the last step of the ramp, not a twitch. Drops keep the deadband, which is what it's for.

**A reservation outlived the sender that took it.** The registry only ever prunes tracks that have *closed*, so a share dropped by a track that is still alive left its claim standing for the rest of the connection. Harmless while each encoder registered once, but the capture loop reopens its camera whenever demand returns, and a reopen can negotiate a different mode (a screen source resized while nothing was subscribed), which caps the larger mode below what it can send and leaves a smaller one claiming room it no longer needs. A reservation now lives exactly as long as the share it was handed to, through a weak handle so a share outliving every `Allocator` still reports the registry as gone. The capture loop re-registers only when the negotiated ceiling changes.

**Two lost wakeups in the new poll paths.** A share whose slice didn't move (its reservation caps it, so most estimate changes don't reach it) parked without the estimate's waker armed, because a poll returning `Ready` registers nothing; `Demand::poll_state` had the same hole when demand flipped between its two internal reads. Both now loop until something is armed. These are hangs rather than failures, so they're covered by tests that assert on a real `Waker` being woken rather than on a re-read, and I verified each fails with the fix reverted.

## Also

`rate::Policy::headroom` is removed. Its job was reserving room for audio, which is now the allocator's, and holding back a further 10% per encoder for one connection's overhead would double-count it. Going over budget is fine: the transport tail-drops groups.

## Caveat

The estimate being divided is only as good as its input, and on the quinn backend that input is `cwnd * 8 / rtt`, which is a window over a latency rather than a rate. Filed as #2847 (quinn already computes the BBR pacing rate and exposes it on `Controller::metrics()`; it just never reaches `PathStats`). This PR makes the arithmetic honest; it does not make the number good.

Audio reserves but does not follow its share yet: #2848. Three more gaps this surfaced, all filed rather than scoped in: transcode ladders (#2858, the sharpest remaining instance, and one with a catalog-honesty question of its own), passthrough imports (#2859), and bindings, which have no encoder rate control at all (#2857).

## Testing

`just check` and `just test` clean (3490 Rust pass, plus the JS and Python suites). Also ran, since PR CI doesn't: `just rs doctest`, `just rs macos` (the only thing that ever compiles moq-audio's `capture` feature), and `cargo check -p moq-ffi -p libmoq` (`just check` never compiles either).

New coverage: the non-media tracks ranking with the catalog (mutation-checked against reverting the fix), the split itself (strict priority, max-min fairness, starved tiers, unclaimed surplus, and the one-tier case that is what actually runs today), two concurrent encoders splitting an estimate, idle and closed tracks, estimate lifecycle, both wakeup regressions, the hysteresis stall, and `track_info` carrying the priority.

No wire format change, so no draft update: TRACK_INFO has carried `priority` since Lite05 and the draft already specifies these semantics. The IETF group header's publisher priority is likewise a field moq-transport already defines; this only stops sending a flat 0 in it.

`js/hang` is mirrored for the priority half: `container.trackInfo` now requires a `priority`, the browser publisher stamps each rendition's kind, and its catalog track takes the catalog priority. Without it a browser broadcast would keep advertising 0 while a Rust one advertised 80, which is the inconsistency this PR exists to remove. The allocator half has nothing to mirror: `js/publish` has no bandwidth-driven encoder rate control to divide.

(Written by Opus 5)

### DIFF
--- a/js/hang/src/container/track.ts
+++ b/js/hang/src/container/track.ts
@@ -26,12 +26,26 @@ const MAX_AGE_MS = 30_000;
  *
  * Options override that retention; see {@link TrackInfoOptions.maxAge}.
  */
-export function trackInfo(options?: TrackInfoOptions): Partial<Track.Info> {
-	return { timescale: Time.Timescale.MICRO, maxAge: options?.maxAge ?? MAX_AGE_MS };
+export function trackInfo(options: TrackInfoOptions): Partial<Track.Info> {
+	return {
+		timescale: Time.Timescale.MICRO,
+		maxAge: options.maxAge ?? MAX_AGE_MS,
+		priority: options.priority,
+	};
 }
 
-/** Overrides for {@link trackInfo}. */
+/** Options for {@link trackInfo}. */
 export type TrackInfoOptions = {
+	/**
+	 * The publisher's tie-break priority, which should come from `PRIORITY` in `@moq/hang/catalog`
+	 * for the kind of media the track carries (`PRIORITY.video` for a video track, and so on).
+	 *
+	 * Required rather than defaulted, matching `hang::container::track_info`: there is no correct
+	 * value for "some media track", and leaving it at zero is what puts audio and video in one
+	 * undifferentiated tier for every relay deciding what to forward first.
+	 */
+	priority: number;
+
 	/**
 	 * How long a relay keeps a non-latest group fetchable, in milliseconds, replacing the
 	 * retention {@link trackInfo} declares by default.
--- a/js/net/src/ietf/publisher.ts
+++ b/js/net/src/ietf/publisher.ts
@@ -12,7 +12,7 @@ import * as Varint from "../varint.ts";
 import type { Session } from "./adapter.ts";
 import * as Cluster from "./cluster.ts";
 import { Frame, Group as GroupMessage } from "./object.ts";
-import { fromWire } from "./priority.ts";
+import { fromWire, toWire } from "./priority.ts";
 import { PublishDone } from "./publish.ts";
 import { PublishNamespace, PublishNamespaceDone, PublishNamespaceOk } from "./publish_namespace.ts";
 import { RequestError, RequestOk } from "./request.ts";
@@ -64,6 +64,9 @@ interface RunGroup {
 
 	/** The track's advertised timescale, applied to every frame timestamp. */
 	timescale: Timescale;
+
+	/** The publisher's tie-break priority, already converted to the IETF wire convention. */
+	publisherPriority: number;
 
 	/** Settles when the subscriber leaves, dropping a group still queued for a stream slot. */
 	unsubscribed: Promise<void>;
@@ -168,7 +171,13 @@ export class Publisher {
 		try {
 			// Declaring the timescale is what opts the track into timestamps; every object
 			// Timestamp below is in these units.
-			const timescale = (await track.info()).timescale;
+			const info = await track.info();
+			const timescale = info.timescale;
+			// The model ranks higher-first, the IETF wire lower-first. Every group this
+			// subscription serves carries the same publisher priority, which is what lets a
+			// relay prefer catalog and audio over video when it has no subscriber preference
+			// to go on.
+			const publisherPriority = toWire(info.priority);
 
 			// Send SUBSCRIBE_OK
 			await stream.writer.u53(SubscribeOk.id);
@@ -204,7 +213,13 @@ export class Publisher {
 				for (;;) {
 					const group = await track.recvGroup();
 					if (!group) return;
-					void this.#runGroup({ requestId: msg.requestId, group, timescale, unsubscribed });
+					void this.#runGroup({
+						requestId: msg.requestId,
+						group,
+						timescale,
+						publisherPriority,
+						unsubscribed,
+					});
 				}
 			})();
 
@@ -245,7 +260,7 @@ export class Publisher {
 	 * Runs a group and sends its frames using ObjectStream (Subgroup delivery mode).
 	 */
 	async #runGroup(options: RunGroup) {
-		const { requestId, group, timescale, unsubscribed } = options;
+		const { requestId, group, timescale, publisherPriority, unsubscribed } = options;
 		try {
 			// One stream per group is faster than a peer at its limit can retire them, so this
 			// is the one path that doesn't wait for a slot: the transport would serve the opens
@@ -265,7 +280,7 @@ export class Publisher {
 				trackAlias: requestId,
 				groupId: group.sequence,
 				subGroupId: 0,
-				publisherPriority: 0,
+				publisherPriority,
 				flags: {
 					hasExtensions: true,
 					hasSubgroup: false,
--- a/js/publish/src/broadcast.ts
+++ b/js/publish/src/broadcast.ts
@@ -237,7 +237,10 @@ export class Broadcast {
 
 			if (request.name === Broadcast.CATALOG_TRACK || request.name === Broadcast.CATALOG_TRACK_COMPRESSED) {
 				const compression = request.name === Broadcast.CATALOG_TRACK_COMPRESSED;
-				const track = request.accept();
+				// The catalog keeps the bare retention defaults (it is read at the live edge, which
+				// is always retained) but still declares its priority, so a relay forwards it ahead
+				// of the media it describes. Matches `hang::Catalog::default_track_info`.
+				const track = request.accept({ priority: Catalog.PRIORITY.catalog });
 
 				// Serve from a per-subscription child scope. Releasing it when this subscriber leaves keeps
 				// serving state from piling up on the connection-lifetime effect as viewers come and go.
@@ -257,9 +260,17 @@ export class Broadcast {
 			}
 
 			// Media, so declare the retention a FETCH-based consumer needs (the catalog above
-			// keeps the bare defaults: it is read at the live edge, which is always retained).
-			// Matches what a Rust publisher declares via `hang::container::track_info`.
-			const track = request.accept(Container.trackInfo({ maxAge: this.in.maxAge.peek() }));
+			// keeps the bare defaults: it is read at the live edge, which is always retained),
+			// plus the priority for what this rendition carries. Matches what a Rust publisher
+			// declares via `hang::container::track_info`; `Kind` and `PRIORITY` share their names,
+			// so a new kind can't be added on one side without the other noticing.
+			const kind = this.#renditions.peek()[request.name]?.kind;
+			const track = request.accept(
+				Container.trackInfo({
+					maxAge: this.in.maxAge.peek(),
+					priority: kind ? Catalog.PRIORITY[kind] : Catalog.PRIORITY.video,
+				}),
+			);
 
 			// A second subscription for the same name supersedes the first: close the old producer.
 			signal.peek()?.close();
--- a/rs/hang/examples/video.rs
+++ b/rs/hang/examples/video.rs
@@ -71,8 +71,9 @@ fn create_track(broadcast: &mut moq_net::broadcast::Producer) -> anyhow::Result<
 	group.finish()?;
 
 	// Actually create the media track now.
-	// track_info() pins the microsecond timescale that the legacy container encodes with.
-	let track = broadcast.create_track(video_track, hang::container::track_info())?;
+	// track_info() pins the microsecond timescale that the legacy container encodes with, and
+	// declares the publisher priority so a subscriber orders this against a broadcast's audio.
+	let track = broadcast.create_track(video_track, hang::container::track_info(hang::catalog::PRIORITY.video))?;
 
 	Ok(track)
 }
--- a/rs/hang/src/catalog/root.rs
+++ b/rs/hang/src/catalog/root.rs
@@ -105,8 +105,11 @@ impl Catalog {
 	/// Track properties for creating the catalog track via
 	/// [`create_track`](moq_net::broadcast::Producer::create_track) at
 	/// [`DEFAULT_NAME`](Self::DEFAULT_NAME).
+	///
+	/// Keeps the bare `moq_net` retention rather than the media one: the catalog is
+	/// snapshot mode, so the useful value is the live edge, which is always kept.
 	pub fn default_track_info() -> moq_net::track::Info {
-		moq_net::track::Info::default()
+		moq_net::track::Info::default().with_priority(PRIORITY.catalog)
 	}
 
 	/// The subscription preferences used for the catalog track (high priority so
--- a/rs/hang/src/container/frame.rs
+++ b/rs/hang/src/container/frame.rs
@@ -42,10 +42,19 @@ const MAX_AGE: std::time::Duration = std::time::Duration::from_secs(30);
 /// anyone play further behind live: a subscriber's own
 /// [`Subscription::max_age`](moq_net::track::Subscription::max_age) defaults to
 /// [`std::time::Duration::ZERO`](std::time::Duration::ZERO) (skip the moment a newer group arrives).
-pub fn track_info() -> moq_net::track::Info {
+///
+/// `priority` is the publisher's tie-break priority, and should come from
+/// [`PRIORITY`](crate::catalog::PRIORITY) so every publisher orders its kinds the
+/// same way (`PRIORITY.video` for a video track, and so on). It's a parameter
+/// rather than a default because there is no correct value for "some media
+/// track": leaving it at zero is what puts audio and video in one undifferentiated
+/// tier, both on the wire and in
+/// [`bandwidth::Allocator`](moq_net::bandwidth::Allocator).
+pub fn track_info(priority: u8) -> moq_net::track::Info {
 	moq_net::track::Info::default()
 		.with_timescale(TIMESCALE)
 		.with_max_age(MAX_AGE)
+		.with_priority(priority)
 }
 
 /// A media frame with a timestamp and codec-specific payload.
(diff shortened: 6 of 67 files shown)


kixelated:
## Adversarial review (Codex), triaged

Ran an adversarial pass with Codex as a second opinion and verified every finding against the code. Verdict was HOLD on five findings: four fixed in `3c7611149`, two rejected with reasons.

### Fixed

**The hysteresis exemption defeated the deadband it was carved out of.** The `settling` check was keyed on `next == desired`, but *every* raise eventually lands on `desired`, so it fired on ordinary upward jitter rather than only on the ceiling recovery it was written for. Verified empirically: from a 2 Mbps target a 2.01 Mbps estimate applied immediately, and a slowly-rising estimate retuned the encoder on every 100ms tick, which is precisely what `Policy::hysteresis` exists to prevent. Now keyed on `Policy::max`.

The distinction that matters: stalling a few percent below a merely *estimate-limited* ceiling is the deadband doing its job, while stalling below the rate the caller *configured* is the defect the exemption exists to fix. Regression test walks a sub-deadband rise and pins both sides of the 5% boundary.

**`flac` and `mp3` were stamped as video.** `Track::new` accepts four audio formats; `format_priority` listed two, and the rest fell through to the video default. Those tracks advertised 60 instead of 80 and lost the ordering the priorities exist for. Fixed, with a test enumerating every format on both sides rather than trusting the comment to keep them in step.

**The browser publisher stamped nothing.** `Container.trackInfo` carried only timescale and retention, and the catalog track was accepted bare, so every JavaScript-published track still advertised 0. A browser broadcast and a Rust one feeding the same relay would have disagreed about their own priorities, which is the inconsistency this PR exists to remove, and `js/hang/src/container/track.ts` already describes itself as "the TypeScript twin of `hang::container::track_info`". `trackInfo` now requires a `priority`, mirroring the Rust signature for the same reason: there is no correct value for "some media track". **This adds a second breaking change**, to `@moq/hang`.

Plus a bug in one of this PR's own new tests, which asserted the 5% boundary was inside the deadband when it is exactly at it.

### Rejected

**"The allocator leaves no capacity for transport overhead"** (flagged high). The physics is right: encoder ceilings describe encoded payload while MoQ framing and QUIC packet overhead consume additional capacity, so senders following their grants exactly can still exceed the estimate. This is a deliberate decision rather than an oversight. Overshoot costs tail-dropped groups, the allocator is advisory by design, and with #2847 open the estimate on the quinn backend is `cwnd / rtt` rather than a rate at all, so a precise overhead margin would be sitting on top of an imprecise input. Codex flags its own reasoning here as an inference from the units rather than an observation. Reversible in one line if the measurement in #2847 says otherwise.

**"Video priming sends at full bitrate while claiming no allocation"** (flagged medium). The mechanism is real: before `catalog_ready` the capture loop takes a bare `camera.read().await` with no estimate arm, so the encoder sits at its full ceiling while its demand-gated share reports `None`. The impact is not. Group serving lives entirely inside the per-subscription run loop (`lite/publisher.rs`), so a track nobody has subscribed to opens no stream and puts nothing on the wire; priming ends on the first encoded frame, and the outer loop then gates on demand. There is nothing to oversubscribe.

The suggested fix would make it worse: representing priming as active demand would have a non-transmitting encoder claim budget away from the siblings that *are* transmitting.

### Verification

`just fix`, `just check`, `just test` (2740 pass), `just rs doctest`, `just js check`, and `just rs macos` all clean. The last is the only thing that ever compiles moq-audio's `capture` feature, and PR CI does not run it.

(written by Opus 5)

kixelated:
## Rebased onto `dev`, plus two of the three open review findings

Picked this back up. `dev` had moved 126 commits since the branch was cut and the merge was conflicting; it is rebased and green again, and two of the three Codex threads still open on the last round are now fixed.

### The rebase

Twelve files conflicted, all against three `dev` changes:

- **`latency_max` -> `max_age`.** Mechanical, and it lands on the same `track::Info` builder chain the priority does.
- **`refactor(mux)!: make timeline pacing explicit` (#2574).** `timeline().track()` no longer mints the timeline track; only `pacing_track()` does. `non_media_tracks_rank_above_media` was asserting on a track that a passive enrollment never publishes, so it now enrolls a pacing track. The fix it guards (the timeline ranking with the catalog) is untouched and still in `prepare_pacing`.
- **Typed `AudioFormat` / `VideoFormat` and the `Track::audio` / `Track::video` split.** This one deleted code rather than conflicting with it. `format_priority(&str)` existed only because the old `Track::new` took a format string and had to be *accepted* before the format was matched, which is exactly what let `flac` and `mp3` fall through to the video priority in the first place. With separate constructors each one knows its own kind, so the classifier is gone and the priority is stamped at the two constructors. Its test went with it, replaced by one that builds a real audio track and a real video track and reads back the `Info` each advertises, which is the property that actually mattered.

`moq-rtc`'s VP8/VP9 bridges moved into a shared `DeferredVideo`, so the priority is stamped there once rather than at each bridge, and `dev` added two new `track_info()` call sites (`moq-ffi`'s raw video track, the TS export tests) that the parameter made the compiler ask about.

### Codex: publisher priority never reached the IETF wire

Real, and the sharpest of the three. `ietf/publisher.rs` sent `publisher_priority: 0` on every group header, so the priority this branch stamps reached moq-lite peers through TRACK_INFO and stopped there. A moq-transport peer saw catalog, audio, and video in one tier, which is the condition the branch exists to remove, and the field is specifically what a relay is told to prefer upstream because it cannot pick between its subscribers' priorities.

Now `priority::to_wire(track.info().priority)`, the same conversion the subscriber side already uses (the model ranks higher-first, this wire field lower-first). The regression test decodes the header a real `TrackServe` wrote and pins audio below video on the wire; it fails with the fix reverted. `js/net` is mirrored, reading the priority in the same place it already reads the timescale.

### Codex: a reopened capture kept its first reservation

Also real, and the fix belongs in the allocator rather than the caller. The registry only ever pruned tracks that had **closed**, so a reservation taken by a track that is still alive stood for the rest of the connection. That was invisible while each encoder registered exactly once, but the capture loop reopens its camera whenever demand returns, and a reopen can negotiate a different mode: a screen source resized while nothing was subscribed comes back at a new resolution and so a new ceiling. Keeping the first reservation caps the larger mode below what it can send and leaves a smaller one claiming room it no longer needs; re-registering instead would have claimed twice.

A reservation now lives exactly as long as the share it was handed to, released when the last clone of that `Consumer` drops. The handle back to the registry is weak, so a share outliving every `Allocator` still reports the registry as gone rather than keeping it alive (covered by its own test, since that is the part a strong handle would have broken quietly). The capture loop re-registers only when the negotiated ceiling actually changes, dropping its old share first.

### Codex: fixed imports reserve nothing (P1) - not taken

The mechanism is right: a capture encoder co-resident with an HLS, RTMP, SRT, RTC, or stdin import targets the whole estimate while that import is using the same uplink. It is filed as #2859 and stays there, for a reason that is about this PR's design rather than its size.

**There is no ceiling to reserve.** `register` takes a ceiling, deliberately: "reserve the most the track can ever send, not what it happens to be sending", because a VBR source that dropped its reservation when it went quiet would find the room gone when it woke up. A passthrough import has no configured bitrate and no negotiated mode; the only number available is `catalog::estimate`'s *measured* rate, and reserving a measurement is precisely the thing the allocator is documented not to do. Registering a guess would be worse than registering nothing: it would shrink the one sender that can actually follow a grant, by an arbitrary amount, in favour of one that cannot.

The other half of the answer is where the fix has to live. Reserving needs a `track::Demand`, which is publisher-side, and these imports create their tracks inside `moq-hls` / `moq-rtmp` / `moq-srt` / `moq-rtc`. So it means an `Option<Allocator>` through four crates' public entry points and down into `moq-mux`'s import pipeline. That is a PR, and it should be the one that also decides what a passthrough's ceiling *is*.

### Verification

`just check`, `just test` (3171 Rust, 580 `@moq/net`, 103 `@moq/hang`, 88 `@moq/publish`, 52 Python; all pass), and `just rs macos`, which is the only thing that compiles `moq-video`'s `capture` feature and so the only thing that compiles the reopen path at all. `moq-ffi` is compiled by `just check` on this diff.

(Written by Opus 5)

kixelated:
## API review, applied

Went back over the allocator's surface before merging it, on the grounds that it has exactly two call sites today and will never be this cheap to reshape again. Three changes, and one live bug that the first of them had already caused.

### `register -> Consumer` became `reserve -> Reservation`

Returning a `Consumer` meant one type carrying two opposite lifecycles: dropping a whole-estimate consumer does nothing, dropping a share hands capacity back, and nothing in the signature said which one you were holding. There was also no way to *change* a reservation, so a sender whose ceiling moved had to drop its share and register again, with a doc comment on `register` warning that registering twice claims twice. That is a caller compensating for a missing operation, with prose standing in for a type.

`reserve` now returns a `Reservation`: `#[must_use]`, holding the claim for as long as the sender is publishing, with `set_max` to move the ceiling in place and `consumer()` for the cloneable read handles. The reopen path that motivated the previous commit is now two lines and says what it means:

```rust
let reservation = reservation.get_or_insert_with(|| encode.bandwidth.reserve(demand, ceiling));
reservation.set_max(ceiling);
```

**The bug this surfaced.** `moq_audio`'s capture reserved its bitrate and dropped the handle on the very next line, with a comment explaining that the reservation lived with the track rather than with the handle. That was true when the registry only pruned closed tracks. It stopped being true the moment a share released on drop, which is the fix in the previous commit: audio's reservation was released immediately, and video went back to targeting the whole uplink. Silently, in the exact scenario the allocator exists for, and nothing failed because nothing covered it. Audio now holds a `Reservation` for the capture. That an invisible lifecycle contract broke a caller this fast is the argument for making it visible.

### `moq_net::Bandwidth`

A rate in bits per second was a bare integer in four places, one of them a different width (`moq_audio`'s `Option<u32>` against everyone else's `u64`). Bits-versus-bytes is off by a factor of eight and still typechecks, and the estimate feeding all of this is a `cwnd * 8 / rtt`, so the unit is doing real work here. `Bandwidth` now carries it from the congestion controller through the allocator and rate policy to the encoder.

It stops at the seams where a number leaves Rust: the codec backends, the catalog's serialized `bitrate`, the C ABI and uniffi records, and the CLI flags each unwrap it once, explicitly, with a comment. Going further would have meant either a serde impl or rewriting `moq.h`, neither of which buys anything.

### `Option<Allocator>` is gone

It threaded through six signatures and then nested into `Option<(Consumer, u64)>` and `Option<(Consumer, Control)>`: three optionals for one idea, with every caller branching. `Allocator::unlimited()` (also its `Default`) collapses them, because "no estimate" already had a meaning: a share reports `None`, which tells a sender to hold its current rate. So the absent case and the no-estimate case were always the same case.

The capture loop's rate control keeps an `Option`, but it now means one thing instead of two: *retired*, because the estimate producer went away or the encoder can't retune. Not "retired or never configured".

### What I did not do

`register`'s two positional parameters could have become an options struct. I left them: with `set_max` on the handle, the pressure for new knobs lands on `Reservation` rather than on the constructor, so the argument for pre-emptively boxing them mostly evaporates.

### Verification

`just check` and `just test` clean on today's `dev` (3490 Rust tests pass), plus `just rs macos` for the `capture` feature, which is the only thing that compiles either reshaped call site.

Four new tests: dropping a reservation releases it and wakes the readers still holding a consumer, `set_max` changes the claim in place without a second entry appearing (both directions), and `set_max` wakes a parked reader. One of them caught a genuine mistake in my own first draft, where a fresh consumer whose grant was already `None` parked forever rather than reporting a change: correct behaviour, wrong assertion.

(Written by Opus 5)

PR #2856, opened 2026-08-14
fix(net): stop blocking connect on the initial announce set

### LINKED ISSUE #2836: connect() returns Ok for a session that failed during the readiness wait
## Problem

`Driver::wait_ready` (`rs/moq-net/src/session.rs`) drives the session until the initial announce set lands, and discards the driver's own result while doing so:

```rust
let _ = self.poll(waiter);
```

If the session dies during that window (the announce stream fails to open, the peer's response is malformed, the transport closes), the driver finishes and drops the `ConnectingProducer` clones it owns. That *resolves* readiness, so `Client::connect` returns `Ok((session, driver))` for a session that is already dead. The error isn't lost, it's cached on the driver and surfaces when the caller polls it, but `connect()` reporting success for a handshake that failed is misleading, and callers that spawn the driver and only watch `Session::closed()` see a connect that "succeeded" and then immediately closed.

## Scope

Pre-existing, not a regression: moq-lite has used `wait_ready` this way since the initial-set wait was added, and before #2826 the moq-transport ALPN paths returned `Ok` without waiting at all. #2826 (MoQ Namespace Count) put the moq-transport paths on the same wait, which is what surfaced it in review.

## Suggested fix

Make `wait_ready` return `Result<(), Error>` and propagate a driver completion that lands before readiness, then have `Client::connect` fail rather than hand back a dead pair. Worth regression tests for the three ways it can happen: a stream-open failure, a transport close mid-handshake, and a malformed response.

Found by the Codex adversarial review on https://github.com/moq-dev/moq/pull/2826.

(written by Opus 5)

kixelated:
Closes #2836.

## Summary

`connect()` blocked until the peer's initial announce set landed (`ANNOUNCE_INIT` on lite-01/02, `ANNOUNCE_OK`'s `Active Count` on lite-05+). The only thing that bought was making `origin::Consumer::request_broadcast` correct for a caller that connects and immediately resolves a known path. Nothing needs it:

- `moq export` ([main.rs:377](https://github.com/moq-dev/moq/blob/main/rs/moq-cli/src/main.rs#L377)), `moq rtc` ([rtc.rs:100](https://github.com/moq-dev/moq/blob/main/rs/moq-cli/src/rtc.rs#L100)) and `moq play` ([play.rs:183](https://github.com/moq-dev/moq/blob/main/rs/moq-cli/src/play.rs#L183)) already await `announced_broadcast` before resolving a path. `play` opts out of `request_broadcast` explicitly and its comment says the wait is the whole point.
- `moq-ffi` exposes both and documents that `request_broadcast` "does *not* wait".
- `origin::Dynamic`'s consumers serve local origins, where there are no session announcements to race, and they depend on `request_broadcast` failing *promptly*.

The consumer that would have justified it was #2762, closed unmerged. So the block was a policy `moq-net` shouldn't hold: it made the library decide how long to wait for a peer, silently overrode `--client-connect-timeout 0` ("wait forever"), and let a peer that over-reported its count hold `connect()` for the life of the session. That last part is #2836, which this deletes rather than fixes.

`announced_broadcast` is the explicit wait, and it is the answer to "resolve this path right after connecting".

## Changes

- Delete `lite::Connecting`, `Driver::wait_ready`, `SessionStart::connecting`, and the four `wait_ready` call sites in `client.rs`.
- The lite subscriber still decodes `ANNOUNCE_OK` for the peer's origin id, so **the wire is unchanged** and `Active Count` stays specified in lite-06 for implementations that want it. We just stop counting it, and the `initial_remaining` countdown goes.
- `moq transcode` was the one caller leaning on the block: it spun `while !matches!(session.status(), Connected)` and then called `request_broadcast`, which races the announcement. Now awaits `announced_broadcast` first, like every other CLI path.
- `rs/moq-relay/tests/cluster_unknown.rs` did `connect()` then `request_broadcast` and would race deterministically. Same fix.
- Docs: `origin::Consumer::announced_broadcast` no longer claims `connect()` blocks (it said so explicitly), `moq-ffi`'s two methods name the post-connect race, and `doc/lib/py/moq-rs.md` gains a paragraph. The other language wrapper docs don't cover this API.

## Consumers fixed

Three in-tree callers resolved a path straight after connecting and would now race the announcement. All three switch to `announced_broadcast` (or its C equivalent):

- `moq transcode` and `rs/moq-transcode/examples/transcode.rs`, which both spun on session status and then called `request_broadcast`.
- `rs/moq-relay/tests/cluster_unknown.rs`.
- **The OBS plugin** (`cpp/obs/src/moq-source.cpp`), which called `moq_origin_request` from its session-connected callback. Losing that race blanks the source, and nothing retries: consumption starts only for connection epoch 1. Swapped to `moq_origin_consume_announced`, which waits for the announcement and whose doc recommends it for exactly this position.

Waiting for an announcement drops the error the old `session.status().await?` propagated, since the origin outlives the session and the wait itself cannot fail. Both transcode paths therefore race it against `Reconnect::closed()`, so a rejected token or an exhausted retry budget returns the real connection error instead of waiting for an announcement that can never arrive.

Every other `request_broadcast` / `Source::new` caller was checked and already gates on `announced_broadcast` (`moq export`, `moq play`, `moq rtc`, `moq rtmp`, `moq-srt`), resolves per HTTP request long after connect (`moq-hls`), or runs against a local origin with no session to race (`moq-cli/publish.rs`, `moq-ffi` tests).

The OBS change is **uncompiled**: `just obs build` needs `just obs setup` to download obs-deps, and PR CI never builds the plugin. It is a swap between two libmoq functions whose generated C signatures are identical, verified against `target/include/moq.h`. `moq-source.cpp` has no test coverage at all today, tracked in #2860.

## Behavior change

`moq_net::Client::connect` returns as soon as the handshake completes instead of also waiting for the first announcements. An embedder doing connect-then-`request_broadcast` starts racing, with no compile error; the failure reads as "broadcast unavailable" for a broadcast that exists. The docs above are the mitigation. In-tree callers are all fixed here.

Not a semver break: no `pub` item is renamed, removed, or resignatured, so this targets `main` per CONTRIBUTING.

## Public API changes

None. `Connecting`/`ConnectingProducer` were `pub(crate)`, `lite` is a private module, and `Driver::wait_ready` was `pub(super)`.

## Test plan

- `just check` / `just test`.
- `cluster_unknown` is the regression test for the racy pattern: it exercises connect-then-resolve against a relay and now waits for the announcement explicitly.
- Two `client::tests::*_falls_back_to_draft14_*` tests timed out on the first run and are fixed here rather than adjusted away: they bound the result of `connect()` and waited for the transport to close, relying on `wait_ready` polling the driver as a side effect. `connect` no longer polls it, and the documented contract is that the session makes no progress unless the driver is, so they now spawn it. That is the same observable change an embedder sees.
- `connect_does_not_wait_for_the_peer_to_announce` pins the new contract: a peer that opens the announce stream and then never answers must not hold `connect()`. Verified to fail against pre-fix `main` (`connect waited on a peer that never announced: Elapsed(())`) and pass here.

## Review

Reviewed adversarially by Codex (`/pr-review`), which caught the OBS regression above (a first-party consumer in code PR CI never compiles) and, on a later pass, the dropped connection error in the transcode paths. Also fixed from review: the `moq-transcode` example, and the missing regression test. Filed #2860 for the absent `moq-source.cpp` test coverage that let the OBS break through.

One AI finding was rejected: that `moq_source_abort_consume_locked` leaves a pending `consume_task` able to call back after the abort. `run_consume_announced` returns as soon as it delivers the handle, so at the only site where a wait is still live (the catalog-setup failure) the task self-terminates and its terminal fires immediately, which the generation check already handles. The shape is also unchanged from `moq_origin_request`, so it is neither new nor a leak.

Note for the next person touching `moq transcode`: it is feature-gated, so `just check` (default features only) never compiles it. Verified here with `cargo clippy -p moq-cli --features transcode --all-targets -- -D warnings`.

(Written by Opus 5)


kixelated:
## Adversarial review summary (Codex)

**Fixed**

- **[high] OBS requested the broadcast before announcements arrived** (`cpp/obs/src/moq-source.cpp`). `moq_source_start_consume` runs off the session-connected callback and called `moq_origin_request`, which resolves against what is announced *now* and fails otherwise. This PR removed the block that guaranteed the announcement had landed by then, so losing the race blanks the source, and nothing retries: consumption starts only for connection epoch 1. Swapped to `moq_origin_consume_announced`, which waits and whose doc recommends it for exactly this position; identical callback contract, so only the two close sites moved with it. This is the finding that justified the review: the plugin is C++ and PR CI never compiles it.
- **[P2] The standalone transcoder example** (`rs/moq-transcode/examples/transcode.rs`) carried the same connect-then-request pattern as the CLI verb. Missed on the first pass because the sweep covered crate `src/` but not `examples/`.
- **[minor] No regression test for the stall.** Added `connect_does_not_wait_for_the_peer_to_announce`, and verified it fails against pre-fix `main` (`connect waited on a peer that never announced: Elapsed(())`) rather than passing vacuously. Went broader than suggested: the peer never answers the announce stream at all, which subsumes the promised-count-never-delivered case and needs no scripted bytes.

**Filed**

- #2860: `moq-source.cpp` has no test coverage. The existing OBS test stubs only the publish path, which is how the regression above got through.

**Checked and clear**

Every other in-tree `request_broadcast` / `Source::new` caller already gates on `announced_broadcast` (`moq export`, `moq play`, `moq rtc`, `moq rtmp`, `moq-srt`), resolves per HTTP request long after connect (`moq-hls`), or runs against a local origin with no session to race (`moq-cli/publish.rs`, `moq-ffi` tests).

**Caveat**

The OBS change is uncompiled. `just obs build` needs `just obs setup` to download obs-deps, and PR CI never builds the plugin. It is a swap between two libmoq functions whose generated C signatures are byte-identical, verified against `target/include/moq.h`.

(written by Opus 5)

PR #2861, opened 2026-08-14
fix(net): encode even-keyed message parameters as varints in draft-17+

a-bro:
## Summary

- **Root cause**: `Param for u8` and `Param for bool` in `rs/moq-net/src/ietf/parameters.rs` encode message-parameter values as raw fixed bytes at draft-17+. The spec frames Key-Value-Pair values by the Type's parity — *"Value: A single varint encoded value when Type is even, otherwise a sequence of Length bytes"* (§1.4.3, identical wording in drafts 14, 17, and 19) — and every `u8`/`bool` parameter sits at an even Type (FORWARD `0x10`, SUBSCRIBER_PRIORITY `0x20`, GROUP ORDER `0x22`), so their values must be varints in every draft.
- **Observable failure**: the relay sends SUBSCRIBE upstream with `subscriber_priority = priority::to_wire(0) = 0xFF` as a raw byte. To a conforming varint decoder `0xFF` is an 8-byte leading-ones prefix, the message truncates, and the peer must close per §1.4.3's `KEY_VALUE_FORMATTING_ERROR` clause. `@moq/net` does exactly that (`js/net` enforces the parity rule structurally), so **every broadcast published by a JS/browser peer is unreadable through a Rust relay at draft-17+** — the relay's upstream catalog subscribe dies in ~40ms and every downstream subscriber is refused. Values ≤ `0x7F` coincide in both encodings, which is why most parameters appear to work.
- **Why it was never caught**: encoder and decoder were wrong together, so every Rust↔Rust round-trip passes. The new regression test asserts wire bytes, not a round-trip.
- **History**: draft-17 support originally classified these as varints (#1032). #1045 moved draft-17 to a raw-byte arm with a doc comment — *"Draft-17: type-specific encoding (u8 as raw byte…)"* — describing a rule that appears in no draft; drafts 18/19 then defaulted forward into it. The likely misreading is §10.2.7's *"The SUBSCRIBER_PRIORITY parameter (Parameter Type 0x20) is a uint8"*, which constrains the value's range, not its wire framing. This PR restores #1032's classification: the drafts 14–16 arm is promoted to unconditional (byte-identical for those versions; the per-draft varint spelling still dispatches inside `u64::encode`/`decode`), and only draft-17+ changes.

## Public API changes

None. Two `impl Param for {u8, bool}` bodies and the trait's doc comment; no `pub` item added, removed, renamed, or re-signatured in `rs/moq-net` or `js/*`.

## Cross-package sync

- `js/net`: no change needed — it already implements the parity rule (its `Parameters` API only accepts varints at even keys), which is the interop this PR restores.
- `drafts/`: untouched — the governing spec here is IETF `draft-ietf-moq-transport`, not a `draft-lcurley-*` document. No wire *design* changes, only conformance.

## Rollout note

A fixed Rust peer cannot parse an unfixed Rust peer's raw bytes above `0x7F` (and vice versa) at draft-17+, so mixed-version Rust fleets on the IETF path should upgrade together. Verified in a live relay+shim deployment: moving both sides together is clean; moving one side reproduces the original failure in the other direction.

## Test plan

- New `test_param_u8_max_encodes_as_a_varint`: asserts wire bytes per version (`[0x40, 0xFF]` at draft-15/16, `[0x80, 0xFF]` at draft-17/18) — fails without the fix; a round-trip test cannot catch a symmetric encoding bug, which is how this one survived.
- `cargo test -p moq-net --lib`: 766 passed; clippy clean (rebased on current `main`).
- Real interop: `MOQ_SERVER_VERSION=moq-transport-17 ./test/smoke/smoke.sh --publishers js --subscribers rust` — pre-fix the browser dies decoding the relay's SUBSCRIBE (`unexpected end of stream`); post-fix it replies SUBSCRIBE_OK and data flows. Rust and JS SUBSCRIBE encodings verified byte-identical post-fix.
- Deployed to a live two-relay cluster with browser publishers and Rust (`media-shim`) subscribers: previously-impossible browser-publish → relay → Rust-transform pipelines now run end to end.

## Related, deliberately not in this PR

`Param for Location` at draft-17+ writes bare varints at odd Type `0x09` with no length prefix, violating the same parity rule from the other side (`js/net` length-prefixes it). Currently latent — no Rust codepath here constructs one on the wire at 17+ — and a separate change. Happy to file it as an issue.

(Written by Claude Fable 5)

### DIFF
--- a/rs/moq-net/src/ietf/parameters.rs
+++ b/rs/moq-net/src/ietf/parameters.rs
@@ -230,11 +230,16 @@ impl Parameters {
 
 /// Trait for encoding/decoding parameter values with version-specific formats.
 ///
-/// Parameter encoding differs from field encoding:
-/// - Draft-14/15/16: u8 and bool are encoded as varints (cast to u64)
-/// - Draft-17: type-specific encoding (u8 as raw byte, bool as raw byte, etc.)
+/// A message parameter is a Key-Value-Pair, so the *key* decides the value's framing: an even
+/// key carries a bare varint, an odd key a length-prefixed byte string. That framing is the
+/// same in every draft, which is why `u8`, `bool` and `u64` (all of which only ever sit at even
+/// keys) are varints unconditionally. What does change with the draft is how a varint is spelled
+/// (QUIC through draft-16, leading-ones from draft-17), and `Encode`/`Decode for u64` already
+/// dispatches on that.
 ///
-/// Use `_ =>` for the newest draft behavior so future versions default forward.
+/// The types at odd keys still need a version split, because what goes *inside* their length
+/// prefix is draft-specific. Use `_ =>` for the newest draft behavior there so future versions
+/// default forward.
 pub trait Param: Sized {
 	fn param_encode<W: bytes::BufMut>(&self, w: &mut W, version: Version) -> Result<(), EncodeError>;
 	fn param_decode<R: bytes::Buf>(r: &mut R, version: Version) -> Result<Self, DecodeError>;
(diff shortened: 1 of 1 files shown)


kixelated:
Sup bro. 

I'll take a better look in a bit (sans kids), until then enjoy the codex review.

kixelated:
I checked drafts 14 through 19. This PR should not merge as written.

### Finding

**[P1] The PR applies the Key-Value-Pair parity rule to the wrong protocol structure.**

The wire format changes at draft-17:

| Draft | Relevant format | Encoding of Subscriber Priority 255 |
| --- | --- | --- |
| 14 | Subscriber Priority is still a fixed `(8)` field in SUBSCRIBE/FETCH. Optional parameters use Key-Value-Pairs. | Raw `ff` as the fixed field |
| 15 | Message Parameters are Key-Value-Pairs. Even parameter types carry QUIC varints. | `40 ff` |
| 16 | Same KVP rule, with delta-encoded types. | `40 ff` |
| 17 | Message Parameters become independent Type-Value pairs with definition-specific encoding. `uint8` means one byte. | Raw `ff` |
| 18 | Same type-specific encoding as draft-17. | Raw `ff` |
| 19 | Same type-specific encoding as draft-17/18. | Raw `ff` |

The transition is explicit:

- [Draft-14](https://datatracker.ietf.org/doc/draft-ietf-moq-transport/14/) places Subscriber Priority, Group Order, and Forward directly in SUBSCRIBE as 8-bit fields.
- [Draft-15 section 9.2](https://datatracker.ietf.org/doc/html/draft-ietf-moq-transport-15#section-9.2) serializes parameters as Key-Value-Pairs.
- [Draft-16 section 9.2](https://datatracker.ietf.org/doc/html/draft-ietf-moq-transport-16#section-9.2) retains Key-Value-Pairs and explicitly describes FORWARD as a variable-length integer.
- [Draft-17 appendix A.1](https://www.ietf.org/archive/id/draft-ietf-moq-transport-17.html#appendix-A.1) records "Encode Message Parameters as Type-Value pairs." Its Message Parameter section defines `uint8` as a single byte.
- [Draft-18 section 10.2](https://www.ietf.org/archive/id/draft-ietf-moq-transport-18.html#section-10.2) and [draft-19 section 10.2](https://www.ietf.org/archive/id/draft-ietf-moq-transport-19.html#section-10.2) preserve that distinction.

Consequently, the pre-PR Rust version split is correct:

```rust
Draft14 | Draft15 | Draft16 => encode_as_varint(),
_ => encode_as_raw_u8(),
```

The change at [`parameters.rs` lines 253-275](https://github.com/moq-dev/moq/blob/19f73f7462e261e35b34344ff10d5da0c4016af7/rs/moq-net/src/ietf/parameters.rs#L253-L275) incorrectly erases the draft-17 transition. For draft-19, `SUBSCRIBER_PRIORITY = 255` must encode as raw `ff`, not varint `80 ff`. The changed decoder would likewise reject a conforming peer's raw `ff` by treating it as the prefix of a nine-byte varint. The new byte-level test therefore asserts the wrong wire format.

The reported Rust-to-JS failure appears real, but it points to the JS Message Parameters implementation incorrectly retaining the pre-17 KVP parity rule. The correct repair is to keep Rust's draft-17+ raw `uint8` handling, fix JS to use the type-specific Message Parameter encodings, and add cross-language golden vectors for drafts 16 and 17+.

### Public API changes

None.

**Verdict: HOLD. The PR would make Rust nonconformant for drafts 17, 18, and 19.**

(Written by GPT-5)

a-bro:
Confirmed that this was chasing down fixing this from the wrong side (issue is in JS not rust side).  Closing this out and will open a PR to fix the right side.

PR #2862, opened 2026-08-14
fix(moq-net): serve an IETF subscribe from the live edge

arctic-uno-0144:
### Summary

- A subscription's read cursor starts at the oldest cached group, not the live edge. `Subscription::group_start: None` is only the **wire preference**; the local cursor is separate state, as documented under "Local cursor vs wire preference" on `track::Subscriber`.
- The IETF publisher set the preference and never placed the cursor, so `run_track` drained the entire retained history on subscribe. Its unbounded `FuturesUnordered` then opened one concurrent uni stream per group, all in the same tick.
- Observed against a draft-16 relay: a viewer joining ~40s into a broadcast got every retained group served in a single millisecond. With the default 5s `latency_max` that is dozens of streams; relays reject the burst, and a player that does receive it skips the replayed groups straight back to live, so the catch-up is pure waste either way.
- `lite::publisher`'s `run_track` already does this (`start_group.or_else(|| track.latest())`). This brings the IETF path in line. `LargestObject` is the only filter the IETF publisher accepts, and it means the live edge, so no filter branching is needed yet.

### Public API changes

None.

`run_track` is private; `Subscriber::start_at` and `Subscriber::latest` are existing public API, now actually used by this caller.

 ### Test plan

- New `subscribe_cursor_test::a_subscribe_is_served_from_the_live_edge`: 4 cached groups, one served. Confirmed it fails with the fix reverted (`left: 4, right: 1`).
- `cargo test -p moq-net`: 766 passed, 0 failed, plus 7 doc-tests.
- `cargo clippy -p moq-net --all-targets -- -D warnings` and `cargo fmt --check`: clean.
- Cross-Package Sync: no row applies. This is publisher-side stream scheduling with no wire-format, catalog, or binding surface change.

Co-Authored-By: Claude Opus 5

### DIFF
--- a/rs/moq-net/src/ietf/publisher.rs
+++ b/rs/moq-net/src/ietf/publisher.rs
@@ -678,6 +678,13 @@ impl<S: web_transport_trait::Session> Publisher<S> {
 
 	/// Serve a track using FuturesUnordered for unlimited concurrent groups.
 	async fn run_track(&self, mut track: track::Subscriber, request_id: RequestId) -> Result<(), Error> {
+		// A fresh cursor starts at the oldest cached group, so leaving it there replays the
+		// whole retained history at once, one concurrent stream per group. LargestObject is
+		// the only filter we accept and it means the live edge.
+		if let Some(latest) = track.latest() {
+			track.start_at(latest);
+		}
+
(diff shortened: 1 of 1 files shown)


kixelated:
thx

PR #2863, opened 2026-08-14
reload custom root CAs without restart

kixelated:
## Summary

- watch custom root CA files and atomically replace the active rustls verifier for new Quinn, Noq, HTTPS, and WSS handshakes
- refresh Quiche outbound roots before each new connection
- disable TLS session resumption when trust roots are reloadable so every new handshake applies the current roots
- retain the last valid roots when a replacement is missing, empty, or malformed
- document reload behavior and the Quiche inbound limitation

## Why

Custom root CAs were parsed when the TLS configuration or listener was constructed, so rotating a trust bundle required restarting the relay. Certificate and key files already had filesystem-driven reload support, but the root verifier stayed fixed.

The reload path now builds a complete replacement verifier before swapping it into use. Failed rotations leave the previous verifier active, and existing connections keep the identity established by their original handshake.

Rustls does not invoke certificate verifiers when resuming a cached TLS session. Without an additional safeguard, a server or mTLS client signed by a removed CA could therefore reconnect until its cached session expired. Resumption is disabled only for configurations with reloadable custom roots, ensuring every subsequent connection performs certificate verification while preserving resumption for fixed and system-only trust configurations.

Quiche's current inbound listener API snapshots client-auth roots and exposes no runtime replacement hook. Inbound Quiche root rotation therefore still requires a restart and now emits an explicit warning. Outbound Quiche connections reload roots normally.

This covers configured PEM root files. Operating system trust-store updates remain platform-dependent.

## Validation

- `nix develop --command just rs check -p moq-native`
- `nix develop --command just rs test -p moq-native` (244 passed)
- `nix develop --command cargo nextest run -p moq-native --all-features -E 'test(tls::tests)'` (17 passed)
- `nix develop --command just rs audit` (0 advisory errors or warnings)

(Written by GPT-5)

### DIFF
--- a/doc/bin/relay/auth.md
+++ b/doc/bin/relay/auth.md
@@ -288,8 +288,9 @@ advertises its own identity by setting `--cluster-mesh` to its
 externally-reachable URL, which it publishes on the cluster origin for other
 peers to discover and dial.
 
-The `quinn` and `noq` QUIC backends support mTLS; configuring `tls.root` with a
-backend that does not (e.g. `quiche`) is a startup error.
+The `quinn`, `noq`, and `quiche` QUIC backends support mTLS. Quinn and noq hot
+reload the trusted roots for new handshakes. Quiche currently requires a relay
+restart after rotating inbound `server.tls.root` files.
 
 ## Stream Listeners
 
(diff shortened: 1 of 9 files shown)


kixelated:
@codex review

(Written by GPT-5)

kixelated:
@codex review

(Written by GPT-5)

kixelated:
@codex review

(Written by GPT-5)

PR #2869, opened 2026-08-15
fix(moq-mux)!: let a Rendition be the only thing that writes its catalog slot

kixelated:
## Root cause

A `moq_mux::catalog::Rendition` claimed its name only once `set` published a config. A lazily-configured importer resolves that config much later: an avc3/annexb H.264 track has no rendition at all until its first SPS arrives.

In that window an outside write succeeded, and then quietly lost:

1. `moq_publish_video(MOQ_VIDEO_FORMAT_AVC3)` reserves track `0.avc3`. No catalog rendition yet.
2. `moq_publish_video_config(name="0.avc3")` succeeds, since the name is vacant.
3. The first keyframe arrives; the importer's `Rendition::set` **overwrites** the caller's entry.
4. `moq_publish_media_finish` drops the rendition, whose `Drop` **removes `0.avc3`**, deleting what the caller published.

The mechanism is two write paths that disagreed. `Rendition::set` goes through `RenditionConfig::insert`, a bare `BTreeMap::insert` that overwrites *because the rendition owns the name*; `hang::catalog::Video::insert` takes a `Vacant` entry and otherwise returns `Duplicate`. Nothing connected them, so "this name belongs to an importer" was unrepresentable.

## Fix

`Rendition` was already the ownership handle for a catalog slot: it reserves a name, publishes on `set`, and retires the entry on `Drop`. So rather than guard the second write path, delete it.

A rendition takes its name at **reservation** rather than at `set`, and holds it until it drops. `Reserved::init` (and `video` / `audio` / `text`) refuses a name that a live rendition or an existing catalog entry already holds. The name and the entry are released together under one lock, so nothing can observe a rendition half-gone.

The two places that hand-rolled a rendition now hold a real one:

- **libmoq** keeps a rendition per caller-authored name. `moq_publish_{video,audio}_config` replaces the caller's own rendition, which is what the shipped doc comment always claimed and the code never did, and is refused only for a name a media importer owns. `moq_publish_{video,audio}_remove` drops the handle, and is a no-op for a name the caller never authored.
- **moq-audio**'s encode producer had its own private `Rendition { catalog, name }` struct duplicating the insert/remove-on-drop lifecycle, so it could author into a name an importer had reserved but not yet resolved. Deleted in favor of `moq_mux::catalog::AudioTrack`.

`catalog.video.insert` is **untouched**. It stays the way to build a catalog document, and a `Guard` still derefs to the raw catalog. What changed is that nothing needs to reach through it to publish a rendition, so no parallel `insert_video` / `Guard::insert` surface exists to keep in sync with it. The guard's deref is still raw access that enforces nothing, but that was already true of `catalog.video.renditions`, a `pub` `BTreeMap`. Ownership is enforced by holding the handle, not by policing the data type.

## Unwind safety

Review follow-up, in the second commit. The catalog's two `Arc<Mutex>` are folded into one `State`, since `Guard::publish` was taking the reservations lock while already holding the catalog one.

That lock is held across code we don't own: `Rendition::set` runs a caller's `RenditionConfig::insert` under it, and a publish serializes a caller's catalog extension. A panic in either poisoned the mutex, and the rendition unwinding through it took the same lock to release its name, which is a panic during a panic and aborts the process. Poisoning protects nothing here, so the guard is recovered instead. The rest of the unwind path is ordered around the caller code rather than after it:

- `set` marks the slot filled **before** calling `insert`. An `insert` that writes and then panics never reached the flag, and its entry would stay in the catalog under a name no handle owns, refusing every later reservation.
- `release` frees the name **before** calling `remove`. That hook can panic too, and a stranded entry is recoverable while a stranded name is not.

## Public API changes

Breaking, so this targets `dev`:

- `Reserved::{init, video, audio, text}` return `crate::Result<Rendition<..>>` instead of the rendition directly. A taken name is `hang::Error::Duplicate`. Every call site is an importer's `new(..) -> Result`, so it's a `?`.
- `Rendition` gains a `Debug` impl (name plus whether it has published).
- No new error variant, no new `Guard` methods, no C ABI signature changes.

Behavior changes for previously-broken calls: `moq_publish_video_config` on an importer-owned name returns -18 instead of corrupting the catalog, and on the caller's own name replaces it instead of returning -18. `moq_publish_video_remove` on an importer-owned name is a no-op instead of deleting it.

## Tests

Ten tests, each verified to fail without the piece it covers (the two complements aside):

- Ownership: `an_unresolved_rendition_still_owns_its_name`, `a_resolved_rendition_still_owns_its_name`, `an_existing_entry_owns_its_name`, plus `dropping_a_rendition_frees_its_name` and `a_name_is_owned_per_section` pinning the scope of the rule.
- Unwind safety: `a_poisoned_lock_does_not_take_the_producer_down` (aborts with SIGABRT without the fix, which only reads as a failure under nextest), `an_unwinding_rendition_retires_a_partially_written_entry`, `a_panicking_remove_still_frees_the_name`.
- libmoq: `publish_media_owns_its_rendition_before_the_first_keyframe` drives the avc3 sequence above through the C ABI, and `publish_video_config_replaces_its_own_rendition` covers the caller-owned path.

`just check` and `just test` pass (1319 tests). `just check` never compiles libmoq, so `cargo nextest run -p libmoq` was run separately: 69 passed. `cargo check` also run for `moq-ffi` and `moq-gst`, and `cargo doc` with `-D warnings` for the intra-doc links.

## Cross-package sync

`rs/libmoq` C ABI: no signature changed, so `moq.h`, `cpp/obs/src`, and `doc/bin/obs.md` need no update. `cpp/obs` and `doc/lib/c` only reference the consume side (`moq_consume_video_config`), which is untouched. Doc comments in `api.rs` were corrected in place, including re-pointing the ones that referenced the `moq_publish_media` entry point #2927 replaced. No wire format or catalog JSON change, so no draft update.

(written by Opus 5)

### DIFF
--- a/rs/libmoq/src/publish.rs
+++ b/rs/libmoq/src/publish.rs
@@ -1,12 +1,51 @@
+use std::collections::BTreeMap;
+use std::collections::btree_map::Entry;
+
 use moq_mux::catalog::hang::Extra;
+use moq_mux::catalog::{Rendition, RenditionConfig};
 use moq_mux::import;
 
 use crate::{Error, Id, NonZeroSlab};
 
+/// A published broadcast: its producer, its catalog, and the renditions the caller authored by
+/// hand.
+///
+/// The renditions are held rather than written and forgotten because the handle is what owns the
+/// catalog entry: it publishes on [`set`](Rendition::set) and retires the entry on drop. A media
+/// importer ([`import::Track`] or [`import::Container`]) holds its own for the tracks it publishes,
+/// which is what keeps the two from writing over each other.
+struct Broadcast {
+	producer: moq_net::broadcast::Producer,
+	catalog: moq_mux::catalog::Producer<Extra>,
+	video: BTreeMap<String, Rendition<Extra, hang::catalog::VideoConfig>>,
+	audio: BTreeMap<String, Rendition<Extra, hang::catalog::AudioConfig>>,
+}
+
+/// The caller's rendition under `name`, reserved on first use.
+///
+/// A second write to a name the caller already owns re-sets that rendition, so a config can be
+/// refined in place. A name a media importer owns is refused, since it writes and removes its own.
+fn rendition<'a, C: RenditionConfig<Extra>>(
+	owned: &'a mut BTreeMap<String, Rendition<Extra, C>>,
+	catalog: &moq_mux::catalog::Producer<Extra>,
+	name: &str,
+) -> Result<&'a mut Rendition<Extra, C>, Error> {
+	match owned.entry(name.to_string()) {
+		Entry::Occupied(entry) => Ok(entry.into_mut()),
+		// A duplicate is a hang error either way, so report it under hang's own code rather than
+		// the generic mux one.
+		Entry::Vacant(entry) => match catalog.reserve().init::<C>(name) {
+			Ok(rendition) => Ok(entry.insert(rendition)),
+			Err(moq_mux::Error::Hang(err)) => Err(Error::Hang(err)),
+			Err(err) => Err(err.into()),
+		},
+	}
+}
+
 #[derive(Default)]
 pub struct Publish {
 	/// Active broadcast producers for publishing.
-	broadcasts: NonZeroSlab<(moq_net::broadcast::Producer, moq_mux::catalog::Producer<Extra>)>,
+	broadcasts: NonZeroSlab<Broadcast>,
 
 	/// Single-codec media importers, fed timestamped frames.
 	// Boxed because the codec splitters/imports are much larger than the container ones.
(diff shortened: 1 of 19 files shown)


rs/libmoq/src/publish.rs
```diff
@@ -126,39 +126,43 @@ impl Publish {
...
-	/// Insert or replace an audio rendition in the broadcast's catalog.
+	/// Insert a caller-authored audio rendition into the broadcast's catalog.
 	///
-	/// The catalog is republished automatically.
+	/// Errors if the name is already in the catalog, or if a media importer owns it. The catalog is
+	/// republished automatically.
 	pub fn audio_config(&mut self, broadcast: Id, name: &str, config: hang::catalog::AudioConfig) -> Result<(), Error> {
 		let (_, catalog) = self.broadcasts.get_mut(broadcast).ok_or(Error::BroadcastNotFound)?;
-		catalog.lock().audio.insert(name, config).map_err(Error::Hang)?;
+		catalog.lock().insert_audio(name, config)?;
```
kixelated: Is there some way to avoid this API change? I don't like stuff in the root if it could be scoped.

rs/moq-mux/src/catalog/producer.rs
```diff
@@ -49,6 +49,15 @@ pub struct Producer<E: CatalogExt = ()> {
...
 	reservations: Arc<Mutex<Reservations>>,
 
+	/// Rendition names owned by a live [`Rendition`](super::Rendition) guard.
+	///
+	/// A name is claimed when the rendition is reserved, which is before its config resolves, so
+	/// an outside write can't slip into that gap. Without it, a lazily-configured importer (H.264
+	/// waiting for its first SPS) leaves the name absent long enough for an external insert to
+	/// land, which the importer then overwrites and removes on drop, taking the caller's entry
+	/// with it. See [`Guard::insert_video`].
+	claimed: Arc<Mutex<std::collections::BTreeSet<String>>>,
```
kixelated: We should combine the multiple Arc<Mutex state.

rs/moq-mux/src/catalog/producer.rs
```diff
@@ -383,6 +407,56 @@ impl<E: CatalogExt> Guard<'_, E> {
 
 		emit(self.hang, self.hangz, self.msf_track, &self.catalog)
 	}
+
+	/// Refuse a name a live [`Rendition`](super::Rendition) owns.
+	fn check_claim(&self, name: &str) -> crate::Result<()> {
```
kixelated: Yeah this seems racey, we need to keep the lock held.

PR #2872, opened 2026-08-15
feat(js)!: default standalone components to enabled

### LINKED ISSUE #1238: Default to enabled: true
## Problem

Standalone classes across `@moq/net`, `@moq/watch`, and `@moq/publish` accept an optional reactive `enabled` input, but most default it to `false`. Omitting the property therefore constructs a module that silently does nothing.

The public contract should make the direct constructor useful by default while preserving explicit and reactive lifecycle control.

## Decision

Keep the positive `enabled` input. Do not rename it to `disabled`.

For standalone JavaScript components:

- an omitted or `undefined` `enabled` input defaults to `true`;
- explicit `false` keeps the component inactive;
- changing a live input from `true` to `false` releases active resources;
- changing it back to `true` restarts normally; and
- this is the general convention for future standalone components unless an API explicitly documents an exception.

Apply the change atomically to the current default-disabled public classes on `dev`:

- `@moq/net`: `Connection.Reload`
- `@moq/watch`: `Broadcast`, `Video.Decoder`, `Audio.Decoder`, `Text.Renderer`
- `@moq/publish`: `Broadcast`, `Video.Encoder`, `Audio.Capture`, `Audio.Encoder`, `Source.Camera`, `Source.Microphone`, `Source.Screen`, `Source.File`

Direct construction of camera, microphone, or screen capture authorizes the component to request capture immediately. Screen capture still has the browser's user-gesture requirement, which the public documentation must state.

The `<moq-watch>` and `<moq-publish>` custom elements keep their existing behavior because they already supply explicit lifecycle-controlled signals. Connection pooling and publish preview already default enabled and remain unchanged.

## Acceptance criteria

- [ ] Each listed class treats omitted and explicit `undefined` `enabled` as `true`.
- [ ] Explicit `enabled: false` prevents connection, subscription, encoding, rendering, and permission requests.
- [ ] A live `enabled` input still tears down resources on `true -> false` and restarts on `false -> true`.
- [ ] High-level custom-element lifecycle behavior is unchanged.
- [ ] Redundant static `enabled: true` examples are removed; reactive and explicit-false examples remain where they communicate lifecycle policy.
- [ ] Public docs describe the default and call out the screen-capture user-gesture constraint.
- [ ] The change lands as one breaking PR targeting `dev`.

## Non-goals

- Renaming `enabled` to `disabled` or adding a compatibility alias.
- Redesigning capture error reporting, retry policy, or dismissal handling.
- Changing custom-element activation behavior.
- Changing connection-pool or publish-preview behavior.

(Written by GPT-5)

kixelated:
## Summary

- Default omitted or explicitly undefined `enabled` inputs to `true` across standalone connection, watch, and publish components.
- Preserve explicit `false` behavior and reactive signal teardown and restart semantics.
- Keep the high-level watch and publish web components unchanged because they already provide explicit lifecycle signals.
- Preserve MoQ Boy's pause and mute download gating through an explicit decoder enable signal.
- Document immediate camera and microphone permission requests and the screen-capture user-gesture requirement.
- Add regression coverage for omitted, undefined, and false inputs, including capture side effects.

## User impact

Standalone components now start work without `enabled: true` boilerplate. Constructing camera, microphone, or screen sources without an explicit false input can immediately request capture, so screen sources must be constructed during the authorizing gesture or enabled later through a live signal.

This is a breaking behavior change targeting `dev`. JavaScript package-version bumps are intentionally deferred.

## Validation

- `nix develop --command bun test js/net/src/connection/reload.test.ts js/watch/src/enabled.test.ts js/publish/src/enabled.test.ts`
- `nix develop --command bun test js/moq-boy/src/game.test.ts`

Closes #1238

(Written by GPT-5)

### DIFF
--- a/doc/lib/js/@moq/publish.md
+++ b/doc/lib/js/@moq/publish.md
@@ -114,6 +114,12 @@ The overlay has no `simulcast` control; enable it via the attribute on the neste
 
 ## JavaScript API
 
+Standalone components start enabled when the `enabled` input is omitted. Pass `false` to construct
+one inactive, or supply a `Signal<boolean>` when activation follows application lifecycle state.
+Camera and microphone sources may request permission immediately. Construct an enabled screen source
+during the user gesture that authorizes screen capture, or pass a false live input and enable it from
+that gesture.
+
 ```typescript
 import * as Publish from "@moq/publish";
 import * as Moq from "@moq/net";
@@ -124,11 +130,10 @@ const connection = new Moq.Connection.Shared({ url: new URL("https://relay.examp
 
 const broadcast = new Publish.Broadcast({
     origin: connection.origin,
-    enabled: true,
     name: "alice.hang",
     // Publish two video renditions: video/hd plus a lower-resolution video/sd.
-    video: { hd: { enabled: true }, sd: { enabled: true } },
-    audio: { enabled: true },
+    video: { hd: {}, sd: {} },
+    audio: {},
 });
 
 // Reactive controls
--- a/doc/lib/js/@moq/watch.md
+++ b/doc/lib/js/@moq/watch.md
@@ -102,7 +102,6 @@ const connection = new Moq.Connection.Shared({ url: new URL("https://relay.examp
 
 const broadcast = new Watch.Broadcast({
     origin: connection.origin,
-    enabled: true,
     name: "alice.hang",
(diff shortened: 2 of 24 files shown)


kixelated:
@codex review

(Written by GPT-5)

PR #2879, opened 2026-08-15
fix(cli): serve TCP and Unix listeners without QUIC

### LINKED ISSUE #2834: moq-cli can't run a TCP/UDS-only server (--server-tcp-bind / --server-unix-bind alone)
Found during review of #2809, but pre-existing on `main` (that PR neither introduced nor changed the behavior).

## Problem

`moq_native::ServerConfig` supports stream listeners independently of the QUIC bind: `has_stream_listener()` covers `tcp.bind` and `unix.bind`, and `build_quic = config.bind.is_some() || !config.has_stream_listener()` exists precisely so a stream-only server can be built.

moq-cli can't reach that configuration:

- `MoqSide::validate` requires `--client-connect` or `--server-bind`, so `--server-tcp-bind` / `--server-unix-bind` on their own are rejected with "a MoQ side is required".
- Even when the guard is satisfied, the server is only started inside `if let Some(web_bind) = moq.server.bind.clone()`, so the stream listeners are configured but never served.

So the flags parse and are silently inert unless a QUIC bind is also given, which is the "accepted the flag and ignored it" shape the CLI otherwise avoids.

## Suggested fix

- Accept any configured native listener in `MoqSide::validate`, not just `--server-bind`.
- Start the native server when any listener is configured, keeping `web::run_web` gated on `server.bind`, which is what supplies its HTTP address (and the `certificate.sha256` a local `http://` client pins).

A regression test for `--server-unix-bind` with no `--server-bind` would pin it.

## Note

`moq-relay` is unaffected; it drives `ServerConfig` directly rather than through `MoqSide`.

(written by Opus 5)

kixelated:
## Summary

- Treat explicit TCP and Unix qmux binds as a MoQ server side during validation and startup.
- Bind every listener before reporting systemd readiness, including TCP and Unix stream listeners.
- Keep explicit stream listeners public when they share a server with the LAN mesh, without exposing the mesh-created QUIC listener.
- Compile the raw qmux listener features into `moq-cli` release builds.
- Document the canonical listener flags and reject them on local-only commands.

## Root cause

Both CLI validation and server startup use `MoqSide::serves()`, but it only checked the QUIC `--listen` bind and LAN discovery. The native server already supports stream-only operation, so a TCP-only or Unix-only configuration was parsed and then treated as if no server had been requested. Also, `moq-cli` disabled `moq-native` default features without explicitly enabling `tcp` and `uds`, which made those flags depend on workspace feature unification instead of the released CLI feature graph.

Stream sockets are bound asynchronously by `Server::listen()`. The CLI previously moved `Server` into a spawned serve task and immediately notified systemd, so `READY=1` could precede the socket bind and could even be emitted when the bind later failed. The CLI now attaches ordinary-session routing before the stream accept loops capture it, awaits `listen()`, and gives the spawned task an already-bound `Listener`. A failed bind therefore aborts initialization before readiness.

LAN dispatch used the presence of an explicit QUIC bind as a server-wide public flag. That rejected ordinary requests arriving on explicit TCP and Unix listeners, while changing the flag to cover every listener would also expose the mesh-created QUIC port. Public access is now decided per transport. Local-only command rejection also omitted Unix credential allowlist flags, and the CLI documentation overstated the protection from a socket that is deliberately created with mode `0666`.

## Public API changes

- Add the documented, additive `moq_native::listen::Config::has_explicit_bind()` query.
- Include `tcp` and `uds` in the `moq-cli` dependency feature set.

## Test plan

- `nix develop --command env CARGO_TARGET_DIR=/home/kixelated/work/moq/target/codex-cli-stream-only CARGO_INCREMENTAL=0 just check`
- `nix develop --command env CARGO_TARGET_DIR=/home/kixelated/work/moq/target/codex-cli-stream-only CARGO_INCREMENTAL=0 cargo test -p moq-cli a_stream_bind_failure_prevents_readiness -- --nocapture`
- `nix develop --command cargo test -p moq-cli --no-default-features --features quinn`
- `nix develop --command cargo test -p moq-native --lib --no-default-features --features aws-lc-rs,tcp,uds stream_`
- `nix develop --command cargo test -p moq-cli --no-default-features --features quinn,cluster-lan explicit_stream_listeners_are_public_without_exposing_mesh_quic`
- `nix develop --command cargo test -p moq-cli --no-default-features --features quinn token_verb`
- `nix develop --command bun remark doc/bin/cli.md rs/moq-cli/README.md --quiet --frail`

The `rs/moq-cli` documentation sync is included. This does not change the wire protocol, FFI, gateways, catalog, or container format.

Closes #2834

(Written by GPT-5)

### DIFF
--- a/doc/bin/cli.md
+++ b/doc/bin/cli.md
@@ -84,8 +84,13 @@ moq <MoQ side>  play [playback options]
   - `--connect <url>` dials a relay. The URL path is the relay auth path
     (e.g. `/anon`), `?jwt=<token>` supplies a token, and `--broadcast` names the
     broadcast.
-  - `--listen <addr>` hosts MoQ sessions directly (with `--listen-tls-generate` /
-    `--listen-tls-cert` + `--listen-tls-key`).
+  - `--listen <addr>` hosts encrypted QUIC/WebTransport sessions directly (with
+    `--listen-tls-generate` / `--listen-tls-cert` + `--listen-tls-key`).
+  - `--listen-tcp-bind <addr>` hosts plaintext qmux over TCP. Bind it only to
+    loopback or a trusted private network.
+  - `--listen-unix-bind <path>` hosts qmux over a Unix socket. The socket's
+    filesystem permissions control access, and an optional peer-credential
+    allowlist can restrict it further.
   - `--cluster-lan` meshes with every other participating process on the LAN via
     mDNS, no relay or internet needed. See [LAN Cluster](#lan-cluster-mdns).
 
--- a/doc/bin/cli.md
+++ b/doc/bin/cli.md
@@ -88,9 +88,9 @@ moq <MoQ side>  play [playback options]
     `--listen-tls-generate` / `--listen-tls-cert` + `--listen-tls-key`).
   - `--listen-tcp-bind <addr>` hosts plaintext qmux over TCP. Bind it only to
     loopback or a trusted private network.
-  - `--listen-unix-bind <path>` hosts qmux over a Unix socket. The socket's
-    filesystem permissions control access, and an optional peer-credential
-    allowlist can restrict it further.
+  - `--listen-unix-bind <path>` hosts qmux over a Unix socket. The socket is
+    created with mode `0666`, so restrict access through its parent directory or
+    an explicit peer-credential allowlist.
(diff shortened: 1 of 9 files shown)


kixelated:
@codex review

PR #2890, opened 2026-08-16
feat(moq-net): enforce the subscriber latency budget

kixelated:
## Summary

`Subscription::latency` / `latencyMax` was propagated as metadata but the track model never acted on it. This enforces it in `rs/moq-net` at both ends of a subscription, which is the ceiling that #2784 needs now that #2785 is reverted on `dev`.

The JS mirror is stacked on top as #2919 (originally filed as #2892), so each language reviews on its own.

- **Group selection.** `track::Subscriber`'s `poll_recv_group` / `poll_next_group` / `poll_read_frame` skip a group that has drifted past the subscription's budget, so one poll walks a whole backlog off. Both ends go through the same code: a relay serves downstream via `poll_recv_group`, and a local consumer reads the same way. `fetch_group` is exempt, since a fetch names historical content explicitly.
- **Two measures.** Presentation time (a group's first frame against the newest one above it) and wall-clock arrival time, either of which can expire a group. The arrival clock is monotonic, so it does not react to system-clock changes, and it backstops a stalled or empty group that has no timestamp yet.
- **Measured from the reader, not the group.** A group is not late because it *opened* long ago; what can be late is the content its reader has yet to take. A cursor that has drained a group sits at the group's newest frame. Measuring from the first frame instead makes a 2s GOP two seconds behind by construction, so any budget shorter than one GOP drops the tail of every GOP. A group nobody has started reading still sits at its own start, which is what leaves selection unchanged.
- **After handoff, only a read that would park is judged.** A group with frames in hand always drains: the budget bounds a group that has *stalled* while the live edge moved on, not a reader slower than the wire. Wire publishers keep the check at the one stall the group cursor cannot see, parking on transport flow control, so a blocked stream cannot pin credit for content that has gone stale.
- **A drained group ends rather than fails.** A group-level read only parks once the cursor has taken every frame, so expiring there costs the reader nothing: what it was waiting for was the producer's FIN, and a group abandoned at its own end is indistinguishable from one that ended there. `Error::Old` is reserved for a cursor still holding unread content (a half-read payload, a publisher with buffered frames). Without this a relay reset the downstream stream for a group it had delivered in full, since the FIN and the next group's first frame are separate streams and the reader is routinely parked at the group's end when the verdict is taken.
- **Route takeovers.** Every replacement copy in a spliced group inherits the logical subscription, reader cap, and route-segment bound, while ordinary cache eviction can still fail over to another copy.
- **Consumers.** The mux, native audio/video decoders, libmoq, GStreamer, and the shipped Hang subscriber example apply their configured latency to the initial subscription. `moq_mux::container::Consumer::new` inherits that already-negotiated budget.
- **Stats.** Skipped content is reported as `stale: { bytes, frames, groups, datagrams }` in JSON stats and as matching `moq_relay_stale_*_total` Prometheus counters. Expiry after handoff counts only the unread tail, and cloned readers share the one attribution.

## Public API changes

Targets `dev` for one breaking cleanup:

- **Breaking**: remove `moq_mux::container::Consumer::with_latency`. `Consumer::new` now inherits the initial `moq_net::track::Subscription` latency; `set_latency` remains for mid-stream changes.
- New `moq_net::stats::Content` counter shape, and a `stats::Traffic::stale: Content` field on the existing `#[non_exhaustive]`, serde-defaulted struct.
- New `kio::ConsumerWeak::poll`, mirroring `Consumer::poll`. Additive: it lets a watcher register for changes without joining the consumer count.

## Wire behavior changes

The encoding is unchanged; no draft update is needed, and the existing draft already specifies the two age backstops. What a peer observes differently:

- A group that has drifted past the subscriber's requested budget is not sent at all. The default budget is zero, so a delayed live subscriber deliberately advances to the live edge; callers that need history must request a replay window.
- A group stream is reset with `Old` when the group expires while its transport write is flow-control blocked, or while waiting on stream credit. It was previously held open.
- A group that expires with nothing unread is FIN'd rather than reset, so a relay no longer signals truncation for a group it delivered in full.
- moq-transport carries no receiver latency parameter, so serving-side IETF subscriptions request an effectively unbounded but wire-encodable window and leave receiver-specific enforcement to the receiving subscriber.

## Test plan

- `nix develop --command just check` / `just check-all`
- `nix develop --command just test all`: 3258 Rust tests pass, 2 skipped; all JS suites green.
- Regression tests cover partial-frame, blocked-send, blocked-final-payload, exhausted-stream-credit and transport-flow-control-blocked expiry, fetch-only/live-edge isolation, system-clock changes, handed-out groups, clean EOF after a final-frame drain, unread-tail and cloned-reader stats, replacement copies after route takeover, latency-update wakeups, and the relay Prometheus renderer.
- Two of them were verified to fail with only their own half of the fix reverted: `real_time_reads_a_live_stream_without_truncating_it` (2s GOPs read as they arrive at the default budget, with the FIN landing after the successor's first frame) and `a_budget_is_measured_from_the_readers_position` (a straggler frame arriving after the next group opened still reaches a reader inside its budget).
- Measured on the frame-read hot path, 20k frames with 50 groups cached: 13ns/frame on `dev`, 9-14ns here.

(Written by Opus 5)

### DIFF
--- a/doc/bin/relay/config.md
+++ b/doc/bin/relay/config.md
@@ -400,14 +400,14 @@ counterpart no traffic can flow, so the entry is dropped:
     "broadcasts": 1, "broadcasts_closed": 0,
     "subscriptions": 5, "subscriptions_closed": 2,
     "fetches": 3,
-    "bytes": 12345, "frames": 678, "groups": 9, "datagrams": 2
+    "bytes": 12345, "frames": 678, "groups": 9, "datagrams": 2, "stale": 4
   },
   "anon/foo": {
     "announced": 1, "announced_closed": 0, "announced_bytes": 8,
     "broadcasts": 1, "broadcasts_closed": 0,
     "subscriptions": 2, "subscriptions_closed": 0,
     "fetches": 0,
-    "bytes": 234, "frames": 12, "groups": 1, "datagrams": 0
+    "bytes": 234, "frames": 12, "groups": 1, "datagrams": 0, "stale": 0
   }
 }
 ```
@@ -417,6 +417,7 @@ Field semantics:
 - `announced` / `announced_closed`: cumulative count of every broadcast
   announce/unannounce event on this `(tier, role)` slot, regardless of
   whether any subscription happened. Use this for "all known broadcasts".
+
 - `announced_bytes`: cumulative broadcast-name length summed over each
   model-visible announce and unannounce of this broadcast. It counts the name,
   not the encoded message size, so a broadcast isn't charged for hop chains or
@@ -424,6 +425,7 @@ Field semantics:
   Separate from `bytes`, which is media payload. Announce control traffic that
   never enters the model (auth-rejected or unmatched-prefix announcements) is
   not counted.
+
 - `broadcasts` / `broadcasts_closed`: per-(broadcast, session)
   subscription sentinel. The first active subscription a peer session
   opens for a broadcast bumps `broadcasts`; the last one it closes bumps
@@ -431,13 +433,16 @@ Field semantics:
   broadcasts_closed` is the number of distinct sessions currently
   subscribed to the broadcast (i.e. viewers on the egress side), which is
   typically what billing and UI want.
+
 - `subscriptions` / `subscriptions_closed`: cumulative count of
   track-level subscriptions opened and dropped.
+
 - `fetches`: cumulative one-shot group fetches requested by a calling session,
   counted once per coalesced fetch when the request is issued, so a fetch that
   resolves to "not found" still counts. It is separate from `subscriptions` and
   the viewer sentinel; the fetched payload still flows into `bytes` / `frames` /
   `groups`.
+
 - `bytes` / `frames` / `groups`: cumulative payload counters, bumped as
   groups/frames are read out of the model on the egress side and written into
   it on the ingress side. Egress bytes are counted when read out of the model
@@ -445,11 +450,20 @@ Field semantics:
   still count. For a fan-out egress reader (e.g. an HLS/DASH muxer) this is
   bytes read once per segment at the broadcast origin, not per downstream HTTP
   client.
+
 - `datagrams`: cumulative single-frame groups delivered over an unreliable QUIC
   datagram (moq-lite-05+ on a datagram-capable transport). A subset of `groups`:
   each datagram also counts there, and its payload in `frames` / `bytes`. Counted
   when the datagram enters or leaves the model, so an egress datagram dropped by
   congestion or an oversized body still counts.
+
+- `stale`: cumulative groups skipped because they drifted further behind the live
+  edge than the subscriber's latency budget allows, so the relay never put them on
+  the wire. Disjoint from `groups`: a skipped group is not delivered, and none of
+  its payload reaches `frames` / `bytes` either. A steady rate here means
+  subscribers are consistently behind the live edge, which is normal for a
+  real-time subscription during congestion and a problem for one that asked to
+  tolerate more.
 
 The session tracks (`sessions.json` and any `<tier>/sessions.json`) instead map
 each auth root to a `{ sessions, sessions_closed }` snapshot. `sessions`
--- a/doc/bin/relay/config.md
+++ b/doc/bin/relay/config.md
@@ -400,14 +400,16 @@ counterpart no traffic can flow, so the entry is dropped:
     "broadcasts": 1, "broadcasts_closed": 0,
     "subscriptions": 5, "subscriptions_closed": 2,
     "fetches": 3,
-    "bytes": 12345, "frames": 678, "groups": 9, "datagrams": 2, "stale": 4
+    "bytes": 12345, "frames": 678, "groups": 9, "datagrams": 2,
+    "stale": { "bytes": 456, "frames": 23, "groups": 4, "datagrams": 0 }
   },
   "anon/foo": {
     "announced": 1, "announced_closed": 0, "announced_bytes": 8,
     "broadcasts": 1, "broadcasts_closed": 0,
     "subscriptions": 2, "subscriptions_closed": 0,
     "fetches": 0,
-    "bytes": 234, "frames": 12, "groups": 1, "datagrams": 0, "stale": 0
+    "bytes": 234, "frames": 12, "groups": 1, "datagrams": 0,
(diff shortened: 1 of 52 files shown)


doc/bin/relay/config.md
```diff
@@ -417,40 +397,54 @@ Field semantics:
...
   bytes read once per segment at the broadcast origin, not per downstream HTTP
   client.
+
 - `datagrams`: cumulative single-frame groups delivered over an unreliable QUIC
   datagram (moq-lite-05+ on a datagram-capable transport). A subset of `groups`:
   each datagram also counts there, and its payload in `frames` / `bytes`. Counted
   when the datagram enters or leaves the model, so an egress datagram dropped by
   congestion or an oversized body still counts.
 
+- `stale`: cumulative groups skipped because they drifted further behind the live
```
kixelated: I think stale should be split into { bytes, frames, groups, datagrams } as well. The size of a group is quite arbitrary.

We should probably move the existing counters into a `transmitted` section or something IDK. But keep backwards compatibility with the current scheme for now.

rs/moq-cli/src/publish.rs
```diff
@@ -560,11 +560,18 @@ mod tests {
...
 		)
 		.await
 	}
 
+	/// A drift budget no test timeline comes close to, so the exporter reads every group.
+	///
+	/// These tests publish a whole feed before exporting it, which the default
+	/// [`Latency::REAL_TIME`](moq_mux::Latency::REAL_TIME) collapses to the live edge:
+	/// completeness has to be asked for, exactly as a real recorder does.
+	const BATCH: std::time::Duration = std::time::Duration::from_secs(3600);
```
kixelated: 3600 seconds seems reaaally high? I don't understand what we're doing here. At least use something reasonable, like 30s IDK.

rs/moq-mux/src/container/fmp4/export_test.rs
```diff
@@ -7,6 +7,13 @@ use mp4_atom::{DecodeMaybe, Encode};
 
 use crate::container::test_util::{Live, PPS, SPS, raw_frame, video_frame};
 
+/// A drift budget no test timeline comes close to, so the exporter reads every group.
+///
+/// These tests write or import a whole broadcast and only then export it, which the
+/// exporter's default [`Latency::REAL_TIME`](crate::Latency::REAL_TIME) collapses to the
+/// live edge: completeness has to be asked for, exactly as a real recorder does.
+const BATCH: std::time::Duration = std::time::Duration::from_secs(3600);
```
kixelated: I don't understand this name. why BATCH?

rs/moq-mux/src/codec/h264/export.rs
```diff
@@ -293,7 +288,10 @@ mod tests {
 
 		// Consumer side: run the exporter.
 		let consumer = broadcast.consume();
-		let mut export = Export::new(crate::source::announced(&consumer), Once(Some(catalog)));
+		// The whole track is written before the exporter runs, so it needs a budget
+		// wide enough to read it: the default skips everything but the live edge.
+		let mut export = Export::new(crate::source::announced(&consumer), Once(Some(catalog)))
+			.with_latency(crate::Latency::max(std::time::Duration::from_secs(3600)));
```
kixelated: 3600 seconds seems reaaally high? I don't understand what we're doing here.

rs/moq-mux/src/container/consumer.rs
```diff
@@ -761,6 +766,22 @@ mod tests {
...
 	}
 
+	/// Wrap `track` in a consumer whose skipping is entirely its own: the subscription
+	/// gets a budget no test timeline reaches, so the transport hands over every group
+	/// and what the consumer does with them is what the test measures.
+	///
+	/// Both layers enforce the same budget, and normally should: the transport skipping
+	/// a group the consumer was going to skip anyway just saves the bandwidth. These
+	/// tests are the exception, since they are about the consumer's half of it.
+	fn consumer_only(track: moq_net::track::Subscriber, latency: Latency) -> Consumer<Container> {
```
kixelated: confusing name.

The eventual plan is to move the Container ordered consumer into moq-net. So moq-net would be responsible for skipping late groups/frames based on the transport timestamp, not the container timestamp. But that requires moq-lite-05 so we're waiting for a liiiitle bit first.

kixelated:
Swept the open review threads before merging.

**@kixelated's five** are all addressed by later commits; GitHub still shows them unresolved but outdated:

| comment | status |
|---|---|
| `stale` should be split into `{bytes, frames, groups, datagrams}` | done; existing counters left at the top level for backwards compatibility as asked |
| 3600s in `moq-cli/publish.rs` | gone |
| 3600s in `codec/h264/export.rs` | gone |
| `BATCH` name in `fmp4/export_test.rs` | gone |
| `consumer_only` confusing name | renamed `container_latency_only`, with a doc comment on why the two layers differ there |

Noted on the last one that the eventual plan is to move the ordered container consumer into moq-net once lite-05 lands, so moq-net skips on the transport timestamp rather than the container's. Nothing in this PR forecloses that.

**Two Codex findings were real and are now fixed**, each with a regression test verified to fail without it:

- `group.rs` — a group the budget *ended* (drained, nothing lost) reported `Ok(None)` but left only `expired` set, so the next poll turned the same group into `Error::Old`. `moq_mux::container::Consumer` probes exactly that way, following a read with `poll_finished`. The clean outcome is sticky now, as it already was in the JS mirror.
- `lite/publisher.rs` — Lite01/02 have no latency field, so SUBSCRIBE decodes as `Duration::ZERO`, which is indistinguishable from a peer asking for real time. This branch acted on it and served those peers the live edge only. **Those two are `NEGOTIATED[0]` and `[1]`**, so it was the ordinary wire for current peers, not a legacy corner. They now get a non-dropping window with enforcement left to the receiver, matching what the IETF path already did. Mirrored in the JS PR.

One more found while rebasing: #2909 landed `SiTrack` on `dev` after this PR was written, and it subscribes with the default budget. SI carries table snapshots at a low rate, so real-time would drop the revisions between them; it now takes the export's replay window like every other source.

The remaining Codex threads are stale against later commits (initial-subscription latency, resume replacement copies, spliced stale metering, Prometheus counters, JS enforcement). JS enforcement lives in #2926.

(Written by Opus 5)

PR #2891, opened 2026-08-16
fix(moq-mux): accept refreshed clocks on TS duplicates

### LINKED ISSUE #2851: moq import ts: a duplicate packet with a refreshed PCR is misread as a continuity gap
### Summary

`Continuity::observe` in `rs/moq-mux/src/container/ts/import.rs` recognizes an ISO 13818-1 duplicate packet by comparing all 188 bytes. The spec permits a retransmission to differ from the original in one place: the PCR field of the adaptation field. A duplicate carrying a refreshed PCR therefore fails the comparison, falls through to the counter check, matches the previous counter, and is reported as a gap. The partial section or PES is dropped even though nothing was lost.

Raised by CodeRabbit on #2823. Filed rather than fixed there, for the reason below.

### The obvious fix is wrong

Keying the decision on the repeated counter alone looks equivalent and is not. A repeated counter has two causes:

- a retransmission, where ignoring the packet is right, and
- a loss of exactly 15 packets, where the counter wraps back to the value before them.

In the second case treating the packet as a duplicate discards the first genuine packet after the loss, leaves the stale partial in place, and then sees the following packet as contiguous. The partial is joined to unrelated bytes, which is exactly the corruption #2823 exists to prevent, arriving through a different door. This was implemented, caught in review, and reverted in 186007f19; the commit message records the reasoning.

So the two failure modes are not symmetric. Byte equality loses data that was never damaged, which is recoverable. The counter-only rule publishes bytes that were never contiguous, which is not. The current behaviour is the safe direction.

### What a real fix needs

Compare the packets while ignoring only what the spec permits to differ: the PCR, and by the same argument the OPCR, both inside the adaptation field. Roughly:

- `adaptation_field_control` must indicate an adaptation field in both, with equal `adaptation_field_length` and equal flag byte.
- `PCR_flag` selects 6 bytes to skip, `OPCR_flag` a further 6.
- Every other byte must match exactly.

Anything looser reopens the 15-packet hole above, so the tests should cover both a PCR-refreshed duplicate (ignored, partial survives) and a 15-packet loss that lands on the same counter (treated as a break, partial dropped).

### Impact

Low but real. It needs a mux that both retransmits packets and refreshes the PCR when it does, which CodeRabbit notes is common on remuxed feeds. The cost is a dropped section or PES, not corrupt output.

(written by Opus 5)

kixelated:
## Summary

- Treat an MPEG-TS retransmission as a duplicate when only its PCR or OPCR clock bytes were refreshed, preserving the partial section or PES already being assembled.
- Keep the header, adaptation layout, flags, payload, and every other byte exact so a 15-packet loss that wraps the continuity counter still breaks the partial.
- Add regressions for refreshed PCR, refreshed OPCR, and the 15-packet counter-wrap case.

The root cause was that `Continuity::observe` required all 188 packet bytes to match. ISO 13818-1 permits a duplicate packet to carry refreshed clock fields, so a valid retransmission fell through to the repeated-counter check and was classified as a gap.

## Public API changes

None.

## Test plan

- Confirmed the refreshed-PCR regression failed before the fix.
- `nix develop --command just fix origin/main`
- `nix develop --command just check origin/main`
- `nix develop --command just test default origin/main` (1,194 tests passed)

Closes #2851.

(Written by GPT-5)

### DIFF
--- a/rs/moq-mux/src/container/ts/import.rs
+++ b/rs/moq-mux/src/container/ts/import.rs
@@ -1005,14 +1005,13 @@ impl<E: catalog::Catalog> VerbatimStream<E> {
 struct Continuity {
 	/// Last continuity_counter seen on a packet with payload, to spot gaps.
 	last_cc: Option<u8>,
-	/// Last payload packet, to skip ISO 13818-1 duplicates (same cc, identical bytes).
+	/// Last payload packet, to identify ISO 13818-1 retransmissions.
 	last_pkt: Option<[u8; 188]>,
 }
 
 /// What one packet says about the bytes already accumulated for its PID.
 enum Continuation {
-	/// An exact retransmission, which ISO 13818-1 permits once. Ignore the packet entirely:
-	/// processing it would duplicate its bytes.
+	/// A retransmission, which may refresh PCR/OPCR. Ignore it entirely to avoid duplicate bytes.
 	Duplicate,
 	/// Contiguous with the previous payload packet.
 	Contiguous,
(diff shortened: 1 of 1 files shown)


kixelated:
@codex review

(Written by GPT-5)

PR #2897, opened 2026-08-17
refactor(net)!: make the origin lifecycle caller-driven

### LINKED ISSUE #1073: Make Origin lifecycle caller-driven
## Summary

Make Origin lifecycle caller-driven. Target `dev` because this intentionally breaks the published Rust API.

Closed PR #2472 is prototype evidence, not an accepted implementation. This work is a prerequisite for the runtime-neutral goals in #2875.

## Public API

- Make `origin::Producer::new(info) -> (Producer, origin::Driver)` the only public factory.
- Remove `Origin::produce()` and `origin::Info::produce()`.
- Expose `origin::Driver` as a `#[must_use]` future with `Output = ()` and a `poll(&Waiter)` API matching the session driver.
- Add `moq_native::origin::spawn(info) -> Producer`, which constructs the pair and spawns the driver with Tokio. Calling it outside a Tokio runtime may panic.
- Defer an explicit `origin::Producer::abort(err)` API.
- Keep `web_async::time` timers. Runtime-independent time remains part of #2875.

## Lifecycle behavior

- Replace origin-internal `web_async::spawn` calls with one driver-owned task set for source watchers, fronts, and track-serving tasks.
- Perform eligible initial attachment, exact-path visibility, and announcement updates synchronously in `create_broadcast`.
- Pass the already-attached state into the queued watcher so the first driver poll does not attach twice.
- Require driver polling for route changes, track serving, linger timers, failover, and teardown.
- The driver must hold no producer clone and finish after producers and submitted lifecycle work drain.
- Dropping the driver cancels its work, immediately tears down the origin with `Error::Dropped`, rejects pending requests, unannounces active broadcasts, and ends announcement cursors.
- Producer mutations after driver drop return `Error::Closed`.

Immediate visibility deliberately preserves the current publication-readiness race. #2895 tracks an atomic readiness gate separately and does not block this issue.

## Migration

- Migrate native applications, relay components, and libmoq to `moq_native::origin::spawn`.
- Have `moq-wasm` spawn the returned driver with `web_async::spawn`.
- Make direct `moq-net` users, tests, and examples retain and explicitly poll or spawn the driver.
- Preserve all wire behavior, JavaScript APIs, and public FFI signatures.
- Do not update drafts or bump package versions.
- Update Rust API documentation and examples to show construction and the driver lifetime contract.

## Acceptance

- Exact lookup and eligible announcement state update synchronously without polling the driver.
- Route changes, track serving, and nested lifecycle work make no progress until the driver is polled.
- Immediate and parked initial attachment do not duplicate sources or announcements.
- Driver drop aborts active fronts with `Dropped`, rejects pending dynamic requests, emits unannounces, closes cursors, and makes later producer mutation fail with `Closed`.
- The driver naturally finishes without retaining the origin.
- `moq_native::origin::spawn` covers progression, teardown, and its outside-runtime panic.
- `nix develop --command just fix`, `just check`, and `just test` pass.
- Run `just test smoke-full` if the libmoq adaptation changes the exercised C/FFI path.

## Out of scope

- The crate rename in #2875.
- Directional producer/consumer naming.
- Lookup ergonomics.
- Timer abstraction.
- Explicit custom-error abort.
- The atomic readiness gate in #2895.

(Written by GPT-5)

kixelated:
Closes #1073.

## What changed

- `origin::Producer::new(info) -> (Producer, origin::Driver)` is the only factory; `Origin::produce()` and `origin::Info::produce()` are removed.
- `origin::Driver` is a `#[must_use]` future with `Output = ()` and a `poll(&kio::Waiter)` API matching the session driver. It owns one task set (the same `TaskSet` primitive the IETF session uses) that runs everything the origin used to `web_async::spawn` internally: source watchers, front dispatch, and per-track serve tasks. moq-net no longer spawns anything behind the caller's back.
- `create_broadcast` performs the eligible initial attachment synchronously: exact-path lookups and announcement updates (including the ingress announce guard) land before it returns. The already-attached state is passed into the queued watcher, so the first driver poll does not attach twice. Route changes, track serving, linger timers, failover, and teardown make no progress until the driver is polled.
- The driver holds no producer clone. It finishes once every producer handle has dropped and the submitted lifecycle work drains, and keeps returning `Ready` when re-polled.
- Dropping the driver tears the origin down immediately: active fronts abort with `Dropped`, pending dynamic requests are rejected with `Dropped`, announced paths unannounce, announcement cursors drain those final unannounces and then end (new cursors are born ended), `Dynamic` handlers and consumer `request_broadcast` observe `Closed`, and later `create_broadcast` calls fail with `Closed`.
- `moq_native::origin::spawn(info)` constructs the pair and spawns the driver with tokio, panicking outside a runtime. It takes `impl Into<origin::Info>` and a new `From<Origin> for Info` conversion keeps call sites terse.

## Root cause: the dying-front splice

With the detach sweep moved onto the driver, `drop(publisher)` followed by `create_broadcast` at the same path (a reconnect) could run before the driver polls the old source's watcher. The new source then *joined* the not-yet-detached front and its content was spliced into subscribers of a broadcast that is over; moq-hls's `a_replacement_publisher_is_not_served_under_the_replaced_catalog` caught this. Previously the spawned detach always won this race by task-queue order. The model's rule is that splicing requires overlapping **live** sources, so `attach_source` now treats a front whose every attached source is already closing as dead: it closes it and creates a fresh front, exactly as the detach sweep would have. Regression test: `test_create_after_dead_source_is_fresh`.

## Migration

- Relay, moq-cli, moq-gst, moq-bench, moq-boy, examples: `moq_native::origin::spawn`.
- libmoq and moq-ffi spawn the driver on their dedicated runtimes (moq-ffi via `RUNTIME.spawn` since constructors run on foreign threads with no runtime entered).
- moq-wasm spawns the driver with `web_async::spawn`.
- Tests use the pair directly (a crate-private `ProduceTest` shorthand in moq-net; small local helpers elsewhere). Sync tests that now reach a spawn (relay `Cluster::new`, cli LAN helpers) became `#[tokio::test]`.

## Acceptance coverage

New tests in `moq-net/src/model/origin.rs`: `test_create_visible_without_driver` (sync visibility with no runtime at all), `test_lifecycle_requires_driver`, `test_driver_drop_tears_down`, `test_driver_finishes_when_drained`, `test_create_after_dead_source_is_fresh`. `moq-native/src/origin.rs` covers spawn progression and the outside-runtime panic.

## Cross-package sync

- No wire change, so no draft updates.
- JavaScript APIs and public FFI signatures untouched; libmoq/moq-ffi changes are internal.
- Docs updated: `doc/lib/rs/index.md`, `doc/lib/rs/env/native.md`, `doc/lib/rs/env/wasm.md`, plus the moq-net crate-level Async docs.
- No version bumps (release-plz owns those).

## Checks

`just fix`, `just check`, `just test` (CI recipes) pass; full workspace `cargo nextest run` is green (3157 tests), plus `just rs doctest`, `just rs wasm`, rustdoc with `-D warnings`, and `just test smoke`. The C/FFI wire path is unchanged beyond where the driver is spawned, so the full cross-language `smoke-full` matrix was not run.

(Written by Fable 5)


kixelated:
Consolidated with the parallel implementation in #2898 and rebased onto the moq-native -> moq-tokio rename (#2896), so the tokio helper is now `moq_tokio::origin::spawn`.

Ported from #2898: a `SourceTask` struct instead of an 11-argument `run_source`, `create_broadcast` serialized against driver teardown by the dynamic-state lock (replacing a re-check dance whose correctness rested on lock-ordering subtleties), `Request::accept`/`reject` early-returning once the origin closed, and announcement cursors closing their channel when they end. Kept from this PR: the `spawn(impl Into<Info>)` ergonomics, the handed-out-request rejection regression test, and the adversarial-review fixes.

`just fix` / `just check` / `just test` pass post-rebase, full workspace nextest is green (3163 tests), and `just test smoke-full` passes all cross-language combinations.

(Written by Fable 5)

kixelated:
Rebased onto dev (clean, no conflicts) now that #2901 landed, and added the follow-up port: moq-net's driver machinery now runs on `kio::Tasks`.

- `util::TaskSet` keeps its exact crate-private API (cloneable `Tasks` submission handles; Ready once every handle drops and the children drain) but its engine is now a `kio::Shared` submission queue feeding a `kio::Tasks`, per kio's Shared-liveness convention. The futures `mpsc`/`FuturesUnordered` pair is gone from the driver path.
- `poll_set::PollSet` is deleted: the seven `Machine` implementors across the lite/ietf publishers and subscribers push boxed poll closures (`util::poll_task` / `MaybeSendTask`, the Send-agnostic task aliases) into `kio::Tasks` fields instead. `MaybeSendBox` also dropped its futures dependency (plain `Pin<Box<dyn Future>>`).

The port surfaced two sharp edges, both now encoded in comments and covered by the suite: `TaskSet::drop` must move the queued futures out of the lock before dropping them (a queued future can hold a `Tasks` clone whose own Drop takes the same lock — this self-deadlocked half the test suite on the first attempt), and the submission queue must be drained until it *registers* the owner's waiter before the children run, or a child pushing mid-pass is a lost wakeup.

`just fix` / `just check` / `just test` pass (2874 scoped + full workspace 3177 green), plus the wasm32 clippy gate.

(Written by Fable 5)

PR #2901, opened 2026-08-17
feat(kio): Tasks, the poll-native FuturesUnordered

kixelated:
Mockup / RFC extracted from the caller-driven origin work in #2897. Draft on purpose: the shape is up for debate before anything migrates onto it.

## kio::Tasks

The poll-native `FuturesUnordered`, as a single base primitive:

```rust
let mut tasks = kio::Tasks::new();
tasks.push(move |waiter: &kio::Waiter| { /* poll one step */ Poll::Pending });
// inside the owner's own poll:
let _ = tasks.poll(waiter); // Ready when empty
```

- **A task is a plain poll closure**: `FnMut(&Waiter) -> Poll<()>`, the same shape `kio::wait` takes, so there is no trait to implement and state lives in the captures. A future adapts with one line: `tasks.push(move |waiter| waiter.poll_future(fut.as_mut()))`.
- **Per-task wakeups**: each task owns a waker that marks it ready and wakes the owner, so `poll` re-runs only the tasks that are new or were woken since the last call, never the whole set. A push also wakes the owner, so a task added between polls gets started.
- **`Send` is inferred, never demanded**: `Tasks<T>` is `Send` exactly when `T` is (the internal waker plumbing is deliberately `T`-free, so it never adds a bound). The test pair `send_tasks_drive_on_spawn` / `local_tasks_drive_on_spawn_local` hands one shared generic `drive()` helper to `tokio::spawn` and to `spawn_local` with an `Rc`-capturing task; both compile from the same code, which is the property this design exists to keep.
- **Ready when empty**, like `FuturesUnordered` yielding `None`; a long-lived driver arm treats that as "drained for now". Dropping the set cancels every remaining task.

Deliberately *not* here (yet): a cloneable submission handle (the `Spawner` from an earlier revision). The origin/session drivers will need some way for outsiders to submit work, but that can be layered on top of `Tasks` (or built from a `Queue` the driver drains into it) once the migration needs it, and its liveness semantics deserve their own discussion.

Follow-up (not in this PR, to avoid conflicting with #2897): port moq-net's `util::TaskSet`/`Tasks` and `poll_set::PollSet` onto this, dropping the `futures` mpsc/`FuturesUnordered` usage from the driver path.

The `Producer::poll` drain-before-close change moved to its own PR.

## Workspace fix: patch the registry kio

`web-transport-iroh` depends on the *published* kio, which coexisted with the workspace copy at the same version and made `cargo clippy --package kio` ambiguous, so `just check` and `just fix` failed on any branch that touched kio. A `[patch.crates-io]` points everyone at the workspace copy so exactly one kio is built (and external transports actually exercise in-tree kio changes).

## Checks

`just fix` / `just check` / `just test` pass, plus doctests (the module example runs) and the spawn/spawn_local pair.

(Written by Fable 5)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4288,16 +4288,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "kio"
-version = "0.5.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bdaadab1a06c04819b8f7cabe4ead8c9f1e86016c4c2a9384829bbc555f477a7"
-dependencies = [
- "loom",
- "smallvec",
-]
-
-[[package]]
 name = "kqueue"
 version = "1.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4806,7 +4796,7 @@ dependencies = [
  "bytes",
  "hang",
  "humantime",
- "kio 0.5.4",
+ "kio",
  "m3u8-rs",
  "moq-mux",
  "moq-net",
@@ -4827,7 +4817,7 @@ dependencies = [
  "bytes",
  "criterion",
  "json-patch",
- "kio 0.5.4",
+ "kio",
  "moq-flate",
  "moq-net",
  "serde",
@@ -4865,7 +4855,7 @@ dependencies = [
  "futures",
  "h264-parser",
  "hang",
- "kio 0.5.4",
+ "kio",
  "memchr",
  "moq-json",
  "moq-loc",
@@ -4898,7 +4888,7 @@ dependencies = [
  "criterion",
  "futures",
  "getrandom 0.4.3",
- "kio 0.5.4",
+ "kio",
  "loom",
  "num_enum",
  "rand 0.10.2",
@@ -10415,7 +10405,7 @@ dependencies = [
  "bytes",
  "http",
  "iroh",
- "kio 0.5.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "kio",
  "n0-error",
  "n0-future",
  "tokio",
@@ -10434,7 +10424,7 @@ dependencies = [
  "bytes",
  "futures",
  "http",
- "kio 0.5.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "kio",
  "noq",
  "rustls",
  "rustls-native-certs",
@@ -10471,7 +10461,7 @@ dependencies = [
  "flume",
  "futures",
  "http",
- "kio 0.5.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "kio",
  "rustls-pki-types",
  "thiserror 2.0.20",
  "tokio",
@@ -10491,7 +10481,7 @@ dependencies = [
  "bytes",
  "futures",
  "http",
- "kio 0.5.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "kio",
  "quinn",
  "rustls",
  "rustls-native-certs",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4281,6 +4281,8 @@ checksum = "e2db585e1d738fc771bf08a151420d3ed193d9d895a36df7f6f8a9456b911ddc"
 name = "kio"
 version = "0.5.4"
 dependencies = [
+ "criterion",
+ "futures",
  "loom",
  "smallvec",
  "tokio",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -186,3 +186,10 @@ opt-level = "z"
 lto = "fat"
 codegen-units = 1
 strip = true
+
+# External crates (web-transport-iroh) depend on the published kio, which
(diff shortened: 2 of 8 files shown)


kixelated:
Reworked per feedback: dropped `pub trait Task` entirely. A task is now a plain `FnMut(&Waiter) -> Poll<()>` closure — the same shape `kio::wait` takes — with `BoxTask`/`LocalBoxTask` aliases for heterogeneous sets and a one-line closure adapter for futures (`move |waiter| waiter.poll_future(fut.as_mut())`). The set still calls each closure through `&mut` (inherent to `FnMut`), but nobody writes a named type or `&mut self` method to participate; captures carry the state. All checks green: 3199 workspace tests, loom, doctests. (Written by Fable 5)

kixelated:
Reworked per feedback: renamed to `Tasks` and pared down to the FuturesUnordered-shaped base — `Spawner` and the `BoxTask`/`LocalBoxTask` aliases are gone (a submission handle can layer on top later if the drivers need one). `Send` is now purely inferred: the `send_tasks_drive_on_spawn` / `local_tasks_drive_on_spawn_local` test pair hands the same generic `drive()` code to `tokio::spawn` and `spawn_local` (with an `Rc`-capturing task), and both compile — the failure mode you called out is now a compile-time regression test. Push-wakes-the-owner and only-new-or-woken-tasks-get-polled both hold and keep their tests. The `Producer::poll` drain change split off into #2902. (Written by Fable 5)

kixelated:
Added a criterion benchmark (`cargo bench -p kio`) comparing `Tasks` against `futures::stream::FuturesUnordered`. Both contestants drive an identical hand-rolled gate primitive with byte-for-byte the same poll body, so the measurement isolates the set machinery: storage layout, wake queue, poll dispatch.

Results on an M-series Mac, release profile:

| scenario | kio::Tasks | FuturesUnordered |
|---|---|---|
| `wake_16_of_1000` (steady state: fire 16 gates of 1000 parked tasks, poll once) | ~336 ns/cycle (~21 ns/woken task) | ~318 ns/cycle (~20 ns/woken task) |
| `spawn_drain_1000` (churn: push 1000 ready tasks, drain) | ~33.3 µs (~33 ns/task) | ~22.3 µs (~22 ns/task) |

Steady-state sparse wakes, the shape a driver actually has, are at parity (~5% apart, ~1 ns per woken task). The bench also surfaced two real inefficiencies, both fixed: the ready-queue buffer was being dropped and reallocated every poll (now swapped with a retained scratch buffer, so wakes/polls are allocation-free past the high-water mark), and `push` was taking the shared ready-queue lock (newborns now go in an owner-local list, safe because a newborn's dirty flag stays set until its first poll, so a pre-poll wake can't double-queue it). The remaining churn gap is the per-task `Arc` for the child waker, which a std `Waker`'s single data pointer forces; `FuturesUnordered` pays one node allocation per future too, but embeds the future in it.

For prior art context: the modern alternatives (unicycle, futures-buffered, futures-concurrency's FutureGroup) all abandoned FuturesUnordered's intrusive linked list for exactly this slab + wake-set design, citing memory locality and allocation counts; unicycle's two-set snapshot-per-poll is the same fairness scheme `Tasks` uses, which also makes the FuturesUnordered spin issue (capped at 32 yields there) structurally impossible here.

(Written by Fable 5)

kixelated:
Adopted the approaches directly rather than depending on the crates (both are future-based; our tasks are poll closures, which is also why all of this stays safe Rust — nothing is pinned):

- **unicycle's wake set**: wakes now land in a chunked atomic bitset instead of a mutex-guarded queue. A wake is one `fetch_or` — lock-free, inherently deduplicated (setting a set bit is a no-op) — and a burst of edges coalesces into one owner nudge via a pending flag. This *deleted* machinery: the per-task dirty flags, the ready queue, the scratch buffer, and the newborn list are all gone. Poll sweeps the bitset a word at a time with `swap(0)`, which is also the two-set fairness scheme (a mid-pass wake lands in the next pass).
- **futures-buffered's amortization, safely**: they remove per-future allocations with one unsafe arena `Arc`; we get the same steady-state effect because a slot's bit position never changes, so its `Waker` is minted once and reused across occupants. Per-task allocation becomes per-slot, amortizing to zero in a long-lived set.

Updated numbers (M-series Mac, release):

| scenario | kio::Tasks | FuturesUnordered |
|---|---|---|
| `wake_16_of_1000` (steady-state sparse wakes) | **~266 ns** | ~359 ns |
| `spawn_drain_1000` (cold: fresh set each iteration) | ~32 µs | ~24 µs |
| `spawn_drain_1000_hot` (long-lived set, the driver shape) | **~9.2 µs** | ~25 µs |

Sparse wakes went from parity to ~26% ahead, and hot churn is ~2.7x ahead thanks to slot reuse; FuturesUnordered keeps a ~25% edge only on a cold, single-use set, where first-use waker allocation dominates. All 3203 workspace tests pass, plus doctests and the wasm32 clippy gate (the bitset is `AtomicU64`, supported on wasm32).

(Written by Fable 5)

PR #2902, opened 2026-08-17
fix(kio)!: Producer::poll drains buffered state before reporting closure

kixelated:
Split out of the #2901 mockup at the author's request.

`Consumer::poll` and `Producer::poll_ref` already evaluate the predicate before checking the closed flag; `Producer::poll` was the odd one out, reporting `Err(Ref)` on a closed channel even when the predicate was satisfied, which silently discarded state still buffered inside the value. It now matches its siblings: buffered state drains through the satisfied predicate first, `Err` means unsatisfied *and* final, and a satisfied poll on a closed channel yields the `Mut` so the caller can take what remains. This is the same drain-then-closed contract `Queue::poll_pop` already documents.

Concretely, this lets an announce-style cursor deliver its final buffered unannounces and then end via a plain `close()`, instead of hand-rolling an `ended: bool` inside the value the way #2897 currently does; that simplification is a follow-up once both land.

Semantics note: `Mut` handed out on a closed channel can still mutate (that's the point, e.g. draining a `Vec`), and `Mut::close` is now documented idempotent. `write()` still refuses closed channels, so plain writes can't sneak past a close; only predicate-gated drains can.

Also carries the `[patch.crates-io] kio` fix (duplicated in #2901; the identical hunk merges cleanly whichever lands first): `web-transport-iroh` depends on the published kio at the same version as the workspace copy, making `cargo clippy --package kio` ambiguous, so `just check`/`just fix` failed on any branch touching kio.

Regression test: `producer_poll_drains_before_reporting_closure`. `just fix` / `just check` / `just test` pass (3191 tests), plus `just rs loom`.

(Written by Fable 5)

### DIFF
--- a/rs/kio/src/producer.rs
+++ b/rs/kio/src/producer.rs
@@ -90,25 +90,29 @@ impl<T> Producer<T> {
 	/// the predicate, then mutate through the returned `Mut`. Registers `waiter`
 	/// while pending.
 	///
-	/// Returns `Poll::Ready(Err(`[`Ref`]`))` if the channel is closed.
+	/// The predicate runs before closure is reported, matching [`Consumer::poll`]
+	/// and [`Self::poll_ref`]: buffered state drains first, so
+	/// `Poll::Ready(Err(`[`Ref`]`))` means the predicate is unsatisfied *and* no
+	/// more updates can arrive. A satisfied poll on a closed channel still yields
+	/// the [`Mut`], letting the caller take what remains.
 	pub fn poll<F>(&self, waiter: &Waiter, mut f: F) -> Poll<Result<Mut<'_, T>, Ref<'_, T>>>
 	where
 		F: FnMut(&Ref<'_, T>) -> Poll<()>,
 	{
-		let state = self.state.lock();
-		if state.closed {
-			return Poll::Ready(Err(Ref { state }));
+		let mut guard = Ref {
+			state: self.state.lock(),
+		};
+		if f(&guard).is_ready() {
+			// Upgrade the Ref to a Mut, keeping the same lock guard.
+			return Poll::Ready(Ok(Mut::new(guard.state)));
 		}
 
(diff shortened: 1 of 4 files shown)


rs/kio/src/producer.rs
```diff
@@ -90,25 +90,29 @@ impl<T> Producer<T> {
...
 	{
-		let state = self.state.lock();
-		if state.closed {
-			return Poll::Ready(Err(Ref { state }));
+		let mut guard = Ref {
+			state: self.state.lock(),
+		};
+		if f(&guard).is_ready() {
+			// Upgrade the Ref to a Mut, keeping the same lock guard.
+			return Poll::Ready(Ok(Mut::new(guard.state)));
```
kixelated, line 107: I'm sorry, what?

If we're closed, we absolutely can't return a Mut. You cannot mutate when closed.

kixelated: You're right, and it can't be fixed by reordering: on the write side, draining requires handing out a Mut, so predicate-first necessarily means mutating a closed channel. Since closed must mean immutable, the existing closed-first behavior is correct as-is. The read side (Consumer::poll, poll_ref) already runs the predicate first, which is safe because those only hand back reads. Drain-then-end therefore belongs in the value itself: a terminal flag the producer sets while the channel is still open, closing only after the drain, which is exactly what #2897's announce cursor does. Withdrawing this PR. (Written by Fable 5)

kixelated:
Withdrawn per the review discussion: predicate-first on the write side would hand out a Mut on a closed channel, and closed must mean immutable. The read-side polls already drain before reporting closure, and value-level terminal flags (as in #2897) are the right drain-then-end pattern. The [patch.crates-io] kio fix this carried also lives in #2901, with the caveat from the bot review that the patch only keeps applying while the workspace kio stays within the 0.5 range the external crates require. (Written by Fable 5)

Issue #2907, opened 2026-08-18
Bind the browser through moq-ffi/UniFFI instead of a second hand-written wasm API


kixelated:
Reach the browser through `moq-ffi` (UniFFI) instead of maintaining a second hand-written binding in `moq-wasm`.

## Why

`rs/moq-ffi` already exports the model `moq-wasm` keeps trying to grow: `MoqBroadcast`/`MoqTrack`/`MoqGroup` producers and consumers, `MoqAnnounced`, `MoqTrackRequest`, `MoqSubscription`, `fetchGroup`, `requestedTrack`, `used`/`unused`. #2814 was a second hand-written copy of that surface, and it was closed for that reason.

`moq-ffi` also already solved the lifecycle problem that sank #2814. `moq-ffi/src/ffi.rs::Task<T>` is `Arc<Mutex<T>>` plus a `cancel` watch channel: concurrent calls queue instead of erroring, `cancel()` interrupts a pending call, `Drop` cancels, and `SubscriberControl` and `info` live outside the lock so `update()` works while a `recv_group()` is in flight. That design has been through five language bindings.

## Spike results

Measured, not estimated.

**Baseline**: `cargo check --target wasm32-unknown-unknown -p moq-ffi` dies in `aws-lc-sys` C compilation (via `moq-native` to `rustls`) before type-checking a single line of `moq-ffi`. So the real cost was unknown.

**After gating, both targets compile clean.** Roughly 594 insertions and 308 deletions over 10 files, mostly `#[cfg]`. Native `cargo check -p moq-ffi --all-targets` stays clean. **82 of 160 exported methods survive on wasm32.** What drops: the `video`/`audio`/`server`/`json` modules (43), media and catalog methods in `consumer`/`producer` (23), and the native TLS and bind knobs (12). The entire raw `moq-net` model survives.

**What makes it viable**: uniffi 0.31 ships a `wasm-unstable-single-threaded` feature that drops `Send` from `UniffiCompatibleFuture` and `Send + Sync` from `FfiConverterArc`. Without it, moq-net's `LocalBoxFuture` wasm futures are a hard stop.

**Generated TypeScript** (`uniffi-bindgen-js` 0.2.1, reading metadata straight out of the `.wasm`): 108 KB `moq.ts` across 21 classes, `tsc --strict` clean. It is better than the hand-written surface on the two points that sank #2814:

- `u64` maps to `bigint`, not `number`. #2814's tsify mapping trapped on any value above 2^53.
[...]

PR #2911, opened 2026-08-18
feat(moq-ffi): compile for wasm32

kixelated:
Step 1 of #2907. Independently landable: it adds a target, and changes nothing about the native one.

## Before

`cargo check --target wasm32-unknown-unknown -p moq-ffi` died inside `aws-lc-sys`' C compilation, by way of `moq-native` to `rustls`, without type-checking a single line of `moq-ffi`. So there was no signal at all about what this crate actually needs to reach a browser.

## What is gated

Four native-only dependencies (`moq-native`, `moq-video`, `moq-audio`, `pollster`) and the modules built on them: `video`, `audio`, `server`, and the moq-native logger.

That is all. Everything else compiles, **including the entire media layer**, because `moq-mux` is wasm-portable as of 743c9d8b6. That is worth stating plainly, since my original spike had to gate `hang`, `moq-mux` and `moq-json` too and came in at 82 of 160 methods. Keeping the media layer takes it to **117 of 160**:

| module | native | wasm |
|---|---|---|
| `consumer`, `producer`, `origin`, `json` | 105 | 105 |
| `session` | 24 | 12 |
| `video`, `audio`, `server` | 31 | 0 |

What drops is native codecs, the QUIC server, and the TLS and bind knobs a browser has no equivalent for. Nothing is dropped that a browser could have used.

## What made it possible

uniffi 0.31's `wasm-unstable-single-threaded` feature drops the `Send` bound from `UniffiCompatibleFuture` and `Send + Sync` from `FfiConverterArc`. moq-net's wasm futures are `LocalBoxFuture` and its handles are `Rc`-backed, so without that feature this is a hard stop rather than a porting exercise.

## The one real behavioral difference

`ffi::Task` grows a wasm arm. There is no runtime thread to hand work to, since uniffi's `RustFuture` is polled by the JS event loop, so it awaits in place.

**The native path detaches and the wasm path does not.** Dropping the returned future cancels the work on wasm, where natively it keeps running on the runtime thread until `cancel()`. Queue-and-cancel semantics are otherwise identical. Flagging it because it is a semantic split hiding behind a `#[cfg]`, and it is the kind of thing worth disagreeing with now rather than discovering later.

While there, `RUNTIME.enter()` and the open-coded `RUNTIME.spawn` match blocks moved behind `ffi::enter` and `ffi::detached`, so each has one definition per target rather than one per call site.

## A UniFFI trap worth writing down

**uniffi's proc macro ignores `#[cfg]` on a method inside a `#[uniffi::export] impl` block.** It emits the scaffolding regardless, which produces duplicate symbols and calls to methods that no longer exist. The native and browser clients therefore live in two separately-gated export blocks rather than one block with gated methods.

This is the UniFFI sibling of the wasm-bindgen ident trap already documented in `rs/CLAUDE.md`, and it is the same shape of bug: the macro layer does not see what the source says.

## Known duplication

`transport.rs` is copied from `moq-wasm`. Deliberate here and called out in #2907: the adapter belongs in one crate, but moving it is a separate change from proving `moq-ffi` can target wasm at all. I did not want to entangle the two.

One crate-level `allow(clippy::arc_with_non_send_sync)`, scoped to wasm. uniffi lowers every exported object as `Arc<T>`; on a single-threaded target clippy flags every construction site. The `Arc` is uniffi's ABI, not a choice this crate makes.

## Keeping it fixed

`just rs wasm` now covers `moq-ffi`. This matters more here than usual: `just check` never compiles `moq-ffi` on *any* target, so without the gate this regresses invisibly on the first native-only type someone adds.

## Verification

`just check` clean, including the extended wasm gate. `just test` green at **2922 tests**.

**The native surface is provably unchanged**: I diffed the exported method names under the native cfg before and after, and both sides have the same 154 exports. `Task`'s new `kio::MaybeSend` bound is implied by `Send` natively, and `web_async::time` re-exports tokio's own types. This matters because cargo-semver-checks cannot see `moq-ffi` (no rlib), so "no API change" needs to be shown rather than assumed.

Nothing here has run in a browser. This is a compile target, not a working binding; #2907 step 4 is where that gets settled, and #2904 just landed the harness for it.

## Cross-Package Sync

The `moq-ffi` row calls for updating `libmoq`, the `{py,swift,kt,go}` wrappers and `doc/lib/`. None apply: the native surface is byte-identical, so there is nothing for a wrapper or a doc to track. No wire, catalog or container change either.

(written by Opus 5)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4729,7 +4729,9 @@ name = "moq-ffi"
 version = "0.3.11"
 dependencies = [
  "bytes",
+ "getrandom 0.4.3",
  "hang",
+ "kio 0.5.4",
  "moq-audio",
  "moq-json",
  "moq-mux",
@@ -4743,6 +4745,9 @@ dependencies = [
  "tracing",
  "uniffi",
  "url",
+ "web-async",
+ "web-transport-trait",
+ "web-transport-wasm",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,6 +74,7 @@ rust-version = "1.91"
 
 [workspace.dependencies]
 flate2 = "1.1"
+getrandom = { version = "0.4", features = ["wasm_js"] }
 hang = { version = "0.20", path = "rs/hang" }
 kio = { version = "0.5", path = "rs/kio" }
 libc = "0.2"
@@ -129,6 +130,7 @@ web-transport-proto = "0.6"
 web-transport-quiche = "0.6"
 web-transport-quinn = "0.12"
 web-transport-trait = "0.4.0"
+web-transport-wasm = "0.5"
 
 [profile.dev]
 panic = "abort"
--- a/rs/justfile
+++ b/rs/justfile
@@ -129,10 +129,13 @@ check-changed $FILES:
 
     # moq-wasm's crate root is entirely `#![cfg(target_arch = "wasm32")]`, so
     # however it got selected the host-target pass above compiled it to nothing
-    # and saw no errors in it. This wasm32 pass is what actually checks it, and
-    # the selection already says whether moq-wasm or anything it depends on
-    # changed.
-    if [[ "$packages" == "ALL" ]] || grep -q -- '--package moq-wasm' <<< "$packages"; then
+    # and saw no errors in it. This wasm32 pass is what actually checks it.
+    #
+    # The alternation must list every package `wasm` compiles, not just moq-wasm:
+    # nothing in the workspace depends on moq-ffi, so an moq-ffi-only diff selects
+    # only itself and would never reach the gate that exists to cover it. moq-mux
+    # is selected by its dependents, none of which are moq-wasm.
+    if [[ "$packages" == "ALL" ]] || grep -qE -- '--package (moq-wasm|moq-mux|moq-ffi)' <<< "$packages"; then
     	just rs wasm
     fi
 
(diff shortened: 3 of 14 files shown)


kixelated:
Added a second commit making the native codecs optional, per review.

## `video` and `audio` are now features

The target gate said "absent in a browser" but not "not everyone wants this", and they are the heaviest thing in the graph:

```
default features      : 335 crates
--no-default-features : 294 crates
saved                 :  41 crates
```

Including `openh264`/`openh264-sys2` (a vendored C++ build), `fast_image_resize`, `unsafe-libopus`, `rubato`, `yuv`, `bytemuck`, and the `objc2-av-foundation`/`objc2-avf-audio` stack. A consumer who only wants pub/sub was paying for all of it.

**Both stay in `default`, so this is purely additive.** Every existing consumer builds with default features and sees the identical surface: the py/swift/kt/go wrappers, `rs/moq-ffi/build.sh`, and CI all invoke moq-ffi without `--features`. Nothing needs updating in lockstep, and it stays a `main` change rather than a `dev` one. Anyone who wants the lean build writes `default-features = false`.

`libmoq` is unaffected either way. It depends on `moq-video` and `moq-audio` directly rather than through `moq-ffi`, which I checked before choosing the default.

## Why both gates

The dependency stays under `cfg(not(target_arch = "wasm32"))` and the modules gate on `all(feature = "...", not(target_arch = "wasm32"))`. That looks redundant but each half does a distinct job:

- The **target table** makes a wasm build structurally unable to drag in a native codec, whatever features are on. No `--no-default-features` incantation to remember, and no confusing openh264-on-wasm failure for someone doing a manual build.
- The **feature** makes them optional everywhere else.

Cargo is happy to have a feature name a target-gated optional dependency; enabling it on a target where the dependency does not apply is a no-op. I verified that rather than assuming it, since it is the load-bearing bit of the design.

## Smaller details

`pollster` moves under the `video` feature, plus an unconditional dev-dependency, because one non-codec test blocks on a future. The three raw-video tests gate on `video`, since that feature is what puts `publish_video` on the broadcast producer.

**No new CI gate.** `just rs features` already runs `--workspace --no-default-features` nightly, and moq-ffi is a workspace member, so the `cfg(not(feature))` arms get compiled there. That recipe exists for precisely this, and its comment already says so.

## Verification

All three configurations: wasm32, native default, and native `--no-default-features` (lib and `--all-targets`). `just check` clean, `just test` green at 2922 tests.

(written by Opus 5)

Issue #2892, opened 2026-08-16
js/net: enforce the subscriber latency budget


kixelated:
#2890 makes `Subscription::latency` actually enforced in `rs/moq-net`, where it had been metadata that was forwarded on the wire and never acted on. `js/net` still has the same gap.

`js/net/src/track.ts`'s `Subscriber.recvGroup`, `nextGroup`, and `readFrameSequence` hand back buffered groups without consulting `Subscription.latencyMax`, so a browser subscriber replays a backlog it declared it did not want. That is visible in practice whenever another subscriber widens the publisher's aggregate budget, since the aggregate is what reaches the publisher and the per-subscriber semantics only exist locally.

What the Rust side settled on, for a mirror to match:

- Drift is measured in **presentation time**: a group's first frame timestamp against the highest-*sequence* group above it that carries one. Both are stamped once, when the group's first frame is created, so a backlog delivered as a burst still reads as its true age.
- The anchor must sit strictly **above** the candidate. Backfill and the tail of a rewound timeline can carry a high timestamp on a low sequence without being a live edge.
- A group that has presented nothing is never stale, and neither is one with nothing newer stamped above it. Retention bounds those instead.
- The budget is clamped to the publisher's `latencyMax` retention window, since waiting longer than a group is kept cannot produce it.
- The anchor is bounded by what the subscriber could actually be handed (its read cursor's cap), **not** by the end it requested on the wire, which is a preference that does not filter the handle.
- `start` is a filter, not an exemption: a subscriber that wants history raises its budget. A one-shot fetch stays exempt.

Expect the same fallout the Rust side had: tests that write a batch of groups and then read them all back are asserting more than the default `REAL_TIME` budget promises, and each needs to declare the tolerance it was implicitly relying on.

Same shape as the mirrors in #2772 and #2775.

(Written by Opus 5)

PR #2919, opened 2026-08-18
feat(js/net): enforce the subscriber latency budget

kixelated:
Stacked on #2890, which does the same work in `rs/moq-net`. Split out so each language is reviewable on its own; this was originally filed as #2892.

## Summary

`js/net` carried `latencyMax` as metadata and never acted on it. It now applies the same drift budget as it reads, and `js/watch` requests the playback latency ceiling from its first subscription instead of only after acceptance.

- Drift is measured against both a presentation-time and a wall-clock live edge; either age can expire a group. The arrival clock backstops a stalled or empty group with no timestamp, and uses `performance.now()` so it does not react to system-clock changes.
- Drift is measured **from the reader's position**, not the group's first frame. A reader that has drained a group sits at its newest frame, so a 2s GOP is no longer 2s behind by construction. A group nobody has started reading still sits at its own start, which leaves group selection unchanged.
- Only a read that would *wait* is judged. A group with frames in hand always drains: the budget bounds a group that has stalled while the live edge moved on, not a reader slower than the wire. This also keeps the policy off the read path, where it would otherwise rescan the timeline and attach a signal subscription per operation.
- A drained group the budget gives up on **ends rather than fails**, since giving up where the reader already stands loses nothing. The latency error is reserved for a caller still holding content: a publisher part-way through writing a frame, or retention eviction discarding frames nobody read (which reports the budget as the more specific reason, falling back to `Lagged`).
- `fetchGroup` stays exempt: a fetch names historical content explicitly, so there is no live edge to be late against.

## Public API changes

None. Enforcement and the watch subscription plumbing use package-internal helpers (`hooks` in `internal.ts`), so no exported surface moved.

## Wire behavior changes

None on the encoding. What changes is what a *local* subscriber is handed: a group that has drifted past its own `latencyMax` is skipped rather than delivered, and one that stalls after handoff ends (or fails, if content was unread) instead of waiting indefinitely. The publisher side resets a group stream whose encode or write is blocked as the group expires, which it previously would not have done. The default budget is zero, so a delayed live subscriber deliberately advances to the live edge; callers that need history must request a replay window.

## Cross-package sync

Mirrors `rs/moq-net` in #2890. No draft change: the wire format and parameters are unchanged, and the existing draft already specifies the two age backstops.

## Test plan

- `nix develop --command just test`: 3258 Rust tests pass, 2 skipped; all JS suites green, including 517 `@moq/net` and 147 `@moq/watch`.
- Two regression tests were verified to fail with only their own half of the fix reverted:
  - `real time reads a live stream without truncating it` — 2s GOPs read as they arrive at the default budget, with the group's close landing after its successor's first frame.
  - `a budget is measured from the reader's position` — a straggler frame arriving after the next group opened still reaches a reader that is inside its budget.

(Written by Opus 5)

### DIFF
--- a/js/hang/src/container/consumer.test.ts
+++ b/js/hang/src/container/consumer.test.ts
@@ -35,6 +35,12 @@ function settle(ms = 20): Promise<void> {
 	return new Promise((resolve) => setTimeout(resolve, ms));
 }
 
+// Keep transport drift filtering out of container-consumer tests. These cases
+// exercise the consumer's own latency policy against a complete retained track.
+function subscribeAll(track: Track.Producer): Track.Subscriber {
+	return track.subscribe({ latencyMax: 30_000 });
+}
+
 // --- LegacyFormat ---
 
 test("LegacyFormat decodes a valid frame", () => {
@@ -183,7 +189,7 @@ async function drainFrames(
 
 test("Consumer delivers frames from a single group", async () => {
 	const track = new Track.Producer("test");
-	const consumer = new Consumer(track.subscribe(), { format: new LegacyFormat(), latency: 500 as Time.Milli });
+	const consumer = new Consumer(subscribeAll(track), { format: new LegacyFormat(), latency: 500 as Time.Milli });
 
 	writeGroupWithLegacyFrames(track, 0, [0 as Time.Micro, 33_000 as Time.Micro]);
 	track.close();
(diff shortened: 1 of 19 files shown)


kixelated:
Closed in error, not actually merged. A force-push of mine briefly put these commits onto the base branch (`claude/latency-enforcement`), which made GitHub see this PR's head as reachable from its own base and auto-close it as merged.

Nothing landed on `dev`. The base branch has been corrected and now carries no `js/` changes. This work continues in #2926.

(Written by Opus 5)

PR #2943, opened 2026-08-19
Reactive firefox 153+ via bowser

### LINKED ISSUE #1329: Re-enable WebTransport for Firefox
There was a bug with publishing for Firefox, where is seems like the 2nd server-initiated bidirectional stream wasn't being received? It was breaking publishing, as incoming SUBSCRIBE requests do nothing. I "fixed" it by requiring the WebSocket fallback for Firefox.

I don't know if the bug is in my code or Firefox.

Alexandre-Fromantin:
## Description
This PR replaces the base string match on `navigator.userAgent` with `bowser` in `@moq/net` for more stable browser filtering with version management.

As noted in [this comment](https://github.com/moq-dev/moq/issues/1329#issuecomment-5238290592) (Bugzilla #2046262 has been fixed and updated to Firefox 153). Firefox support is now enabled for versions 153 and above, but blocked for versions below 153. And I left the Safari browser disabled like before.

## Changes
- Added `bowser` dependency to `@moq/net`.
- Refactored `isWebTransportUserAgentSupported` using Bowser parser.
- Firefox WebTransport support for versions 153.0.0 and later.

## Validation
- Ran `just js` locally with success.
- I tested this `@moq/net` with a moq-rs server on Firefox 153.0.4 and Google Chrome successfully.

Fixes #1329 

(Written by Gemini-3.6 Flash)

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -187,6 +187,7 @@
         "@moq/qmux": "^0.3.3",
         "@moq/signals": "workspace:*",
         "async-mutex": "^0.5.0",
+        "bowser": "^2.14.1",
       },
       "devDependencies": {
         "@types/bun": "^1.3.14",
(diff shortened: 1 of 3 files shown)


kixelated:
awesome

PR #2997, opened 2026-08-22
feat(moq-gst): select media container for sink pads

arielmol:
Stacked [on the data-tracks PR](https://github.com/moq-dev/moq/pull/2996). Until that one merges this branch also carries its commit; GitHub narrows the diff here to `select media container` on its own once the base lands.

`moqsink` has no container property, and every codec importer in `moq-mux` fixes the wire container to `Container::Legacy`. This adds `container=legacy|loc` with Legacy as the default, so existing pipelines and callers keep their current behaviour.

**Contract**

- `container` is READY-only, like `url`, `broadcast` and `tls-disable-verify`.
- The selection travels through `catalog::Reserved`, which defaults to Legacy, keeps the selection when cloned, and exposes it to each importer. Every `codec::*::Import::new` keeps its signature.
- Importers apply it to both the track writer and the `container` each catalog rendition advertises, so the catalog names what is on the wire.
- Opaque `application/octet-stream` pads use no media container and are unaffected.
- Passthrough importers keep the container their source dictates: fMP4 publishes `moof+mdat` fragments and stays CMAF. A LOC frame is a property block plus a codec payload, which a fragment is not, so the selection does not apply there rather than being refused.

**Why through the reservation**

Adding an argument to every codec constructor would touch a large set of call sites across unrelated crates for a value that is constant per publication. The reservation already threads from the owner of the catalog down to each importer, which is exactly the scope of the selection.

**Why a new enum**

`hang::catalog::Container` cannot be the selection type: `Cmaf { init }` carries data and `Unknown` exists to round-trip a container this build does not know, so it is neither `Copy` nor a valid domain to pick from. `catalog::MediaContainer` is the selectable subset, and converts into both the catalog type and the runtime writer.

The removed `config.container = Container::Legacy` assignments are no-ops: `Container::default()` is `Legacy` and both `VideoConfig::new` and `AudioConfig::new` initialise the field with it. They were removed so the reservation is the single source of truth.

One property per element covers the case this addresses. A per-pad override stays additive if it is ever needed.

**Public API changes**

All additive.

- `moq_mux::catalog::MediaContainer`, new enum.
- `moq_mux::catalog::Reserved::with_container` and `Reserved::container`, new methods.
- `gstmoq::MediaContainer`, new re-export.

**Test plan**

`just check` and `just test`. New coverage: `container` declares `MUTABLE_READY`, a write above READY is not stored, and it is configurable again after returning to READY; a pipeline description parses `container=loc`; a LOC media pad reaches both the wire and the catalog; an opaque pad still publishes raw bytes with LOC selected; Opus and VP8 reservations selecting LOC write LOC frames and advertise `Loc`; the public TS, MKV and FLV importers carry the selection through to the wire; fMP4 keeps CMAF when LOC is selected; a reservation defaults to Legacy and its clones keep the selection.

**Cross-package sync**

`doc/bin/gstreamer.md` is updated in the same commit. No draft update: the catalog schema already defines the `container` field and its `loc` value, so this selects an existing value rather than changing the wire or catalog format.

(Written by Claude Opus 5)

### DIFF
--- a/doc/bin/gstreamer.md
+++ b/doc/bin/gstreamer.md
@@ -272,6 +272,19 @@ gst-launch-1.0 -v -e \
     sink_0::track=camera sink_1::track=commentary
 ```
 
+Media tracks use the legacy Hang container by default. Set `container=loc` before starting the
+element to publish every media pad as Low Overhead Container and advertise LOC in the catalog:
+
+```bash
+gst-launch-1.0 -v -e \
+  videotestsrc is-live=true ! x264enc tune=zerolatency ! h264parse \
+    ! video/x-h264,stream-format=byte-stream,alignment=au ! mux.sink_0 \
+  moqsink name=mux url=http://localhost:4443 broadcast=bbb.hang container=loc
+```
+
+`container` is writable in `NULL` and `READY`. It is fixed for a running publication and becomes
+writable again after returning to `READY`.
+
 `track` is writable in any state until the pad's CAPS event reserves the track, so a pad requested
 while the pipeline runs can still be named before its first buffer. From then on it reads back the
 reserved name, the generated one included, and further writes are ignored with a warning; stopping the
@@ -280,7 +293,8 @@ the generated name. A name another pad already holds invalidates only that pad, 
 broadcast keeps publishing.
 
 A pad negotiated as `application/octet-stream` publishes application data instead of media. The bytes
-go out exactly as they arrive: no codec, no container, no interpretation.
+go out exactly as they arrive: no codec, no media container, no interpretation. The element's
+`container` property does not apply to these pads.
 
 ```bash
 gst-launch-1.0 -v -e \
--- a/doc/bin/gstreamer.md
+++ b/doc/bin/gstreamer.md
@@ -272,18 +272,20 @@ gst-launch-1.0 -v -e \
     sink_0::track=camera sink_1::track=commentary
 ```
 
-Media tracks use the legacy Hang container by default. Set `container=loc` before starting the
-element to publish every media pad as Low Overhead Container and advertise LOC in the catalog:
+Media tracks use the legacy Hang container by default. Set a pad's `container=loc` before its CAPS
+event to publish that track as Low Overhead Container and advertise LOC in the catalog:
 
 ```bash
 gst-launch-1.0 -v -e \
   videotestsrc is-live=true ! x264enc tune=zerolatency ! h264parse \
     ! video/x-h264,stream-format=byte-stream,alignment=au ! mux.sink_0 \
-  moqsink name=mux url=http://localhost:4443 broadcast=bbb.hang container=loc
+  moqsink name=mux url=http://localhost:4443 broadcast=bbb.hang sink_0::container=loc
 ```
 
-`container` is writable in `NULL` and `READY`. It is fixed for a running publication and becomes
-writable again after returning to `READY`.
+Set the property on each media pad that should use LOC; other pads remain Legacy.
+
+Like `track`, `container` is writable in any state until the pad's CAPS event reserves the track.
+It is fixed for that producer and becomes writable again after returning to `READY`.
 
 `track` is writable in any state until the pad's CAPS event reserves the track, so a pad requested
 while the pipeline runs can still be named before its first buffer. From then on it reads back the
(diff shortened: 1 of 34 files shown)


kixelated:
> Stacked https://github.com/moq-dev/moq/pull/2996. Until that one merges this branch also carries its commit; GitHub narrows the diff here to select media container on its own once the base lands.

Change the base branch pls so it's actually stacked.

kixelated:
@codex review

rs/moq-mux/src/catalog/tracks.rs
```diff
@@ -193,6 +192,38 @@ impl<E: CatalogExt> RenditionConfig<E> for hang::catalog::AudioConfig {
 	}
 }
 
+/// The Legacy or LOC wrapper selected for elementary codec payloads.
+///
+/// Importers apply it to both the track writer and the `container` each catalog rendition
+/// advertises, so the catalog names what is on the wire.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+#[non_exhaustive]
+pub enum MediaContainer {
```
kixelated, line 201: Sorry why are we defining another container enum? Why not `hang::catalog::Container`?

rs/moq-mux/src/catalog/tracks.rs
```diff
@@ -203,12 +234,31 @@ impl<E: CatalogExt> RenditionConfig<E> for hang::catalog::AudioConfig {
 /// track list, so a one-shot muxer (fMP4, MPEG-TS) never sees a half-converged catalog.
 pub struct Reserved<E: CatalogExt = ()> {
 	catalog: Producer<E>,
+	container: MediaContainer,
```
kixelated, line 237: I don't really like the container being global. It should be on a per track basis, since it's already in the track section of the catalog. This API can only cause bugs.

doc/bin/gstreamer.md
```diff
@@ -253,6 +253,19 @@ gst-launch-1.0 -v -e \
...
 ```
 
+Media tracks use the legacy Hang container by default. Set `container=loc` before starting the
+element to publish every media pad as Low Overhead Container and advertise LOC in the catalog:
+
+```bash
+gst-launch-1.0 -v -e \
+  videotestsrc is-live=true ! x264enc tune=zerolatency ! h264parse \
+    ! video/x-h264,stream-format=byte-stream,alignment=au ! mux.sink_0 \
+  moqsink name=mux url=http://localhost:4443 broadcast=bbb.hang container=loc
```
kixelated: A single "container" for the entire broadcast is wrong; it really should be on a per-track basis.

arielmol: just easier to write, I never though of a mixed case, but can be moved I was thinking of it as a default value . It will only need several `sink_0::container=loc`

rs/moq-mux/src/catalog/tracks.rs
arielmol: Okay, but either I got something wrong, or what would a caller pass for `Cmaf { init }` ? My (not too in depth) research  found only the fMP4 passthrough has an init segment, builds a container itself instead of reading the reservation.

rs/moq-mux/src/catalog/tracks.rs
arielmol: Okay, just for the clarification of the last one.

arielmol:
@codex review

rs/moq-mux/src/catalog/tracks.rs
arielmol: Need north here. `hang::catalog::Container::Cmaf { init }` describes an already constructed container, while this elector chooses framing for elementary payloads before an output container exists. CMAF would require generating the init segment, transforming codec payloads, and producing moof+mdat. The fMP4 importer instead passes through existing CMAF. Would renaming this to ElementaryContainer and explicitly limiting the PR to Legacy/LOC make the distinction clear?

arielmol: We could add elementary-to-CMAF publishing by reusing the current fMP4 export machinery, but I think that should be a separate feature rather than expanding this PR.

rs/moq-gst/src/sink/mod.rs
```diff
@@ -10,6 +10,29 @@ mod timeline;
...
+	/// Hang's original timestamp-prefixed media container.
+	#[default]
+	#[enum_value(name = "Legacy", nick = "legacy")]
+	Legacy,
+	/// Low Overhead Container from draft-ietf-moq-loc.
+	#[enum_value(name = "LOC", nick = "loc")]
+	Loc,
+}
+
+impl From<MediaContainer> for moq_mux::catalog::MediaContainer {
```
kixelated: Why do we need moq_mux::catalog::MediaContainer? It doesn't seem to be doing anything but helping convert to a hang container?

arielmol: Commented above

arielmol:
merge conflicts, but seems easy, I think both changes in the conflict should be accepted, what do you think?

arielmol:
in `doc/bin/gstreamer.md`: the container section says "Like `track`, ..." but sits above the `track` section.

PR #2998, opened 2026-08-22
feat(moq-gst): expose sink-pad publication lifecycle

arielmol:
Adds `track-status` and `track-error`, two read-only GObject properties on the `moqsink` request pad, so an application can tell why a pad stopped publishing without parsing bus messages.

## History

This started as four commits: the properties plus three fixes they made visible. Those three fixes have since landed on `main` separately, rewritten against the older structure rather than cherry-picked (#3101, #3102, #3104). `main` was then merged in rather than rebased, so the original commits stay as written and the reconciliation is one reviewable merge commit. What is left on top of `main` is the refactor and the properties.

## Summary

- **Two pad properties.** `track-status` moves `pending -> active -> ended`, or to `error` with `track-error` carrying the reason. Both emit `notify`. Named `track-status` rather than `status` because the element already has a `status` (the connection lifecycle) that means something else.
- **The per-pad lifecycle moves into `MoqSinkPad`,** behind one mutex. It owns the producer, timeline state, requested and effective track names, status, error, EOS state and release state. GStreamer's sink pad list stays the authority for membership; the element no longer keeps a parallel name-keyed pad map. A request pad was previously represented in three places that had to be kept aligned across CAPS, EOS, release, session replacement and reentrant GObject callbacks.
- **`Completion` replaces `eos_posted`,** which answered four questions at once. One `Arc` per session, cloned into every pad. Only `Open -> Eos` and `Open -> Failed`, first writer wins via compare-exchange. It doubles as the session's identity, so a message deferred by a session that has since been replaced is dropped on `Arc::ptr_eq` rather than against a separate token.
- **The pad GType is declared in the template,** so `gst-inspect-1.0` and `GstChildProxy` list the pad's own properties.

## Lock order

One nesting order, in one direction: GStreamer's stream lock, then the element `Control`, then a pad lifecycle, then the GObject object lock. No path takes `Control` while holding a pad lifecycle. Property notifications and bus messages are emitted only after lifecycle locks are released, since both synchronously call application code that can re-enter element and pad properties.

## Known limitation

`post_current` checks that a deferred message still belongs to the live session, releases the control lock, then posts. A session replaced inside `post_message` still gets its message posted. The lock cannot be held across the post: bus sync handlers run inline on the calling thread and a handler reading a property would take it again. The obvious remedy (an in-flight posting permit that teardown waits on) moves the same reentrancy onto a different primitive and can self-deadlock.

Closing this needs the publication to have a notion of a **generation**, which is tracked in #3115 along with the two related gaps on `main`. Out of scope here.

## API

No Rust API changes. Adds two GObject sink-pad properties and one new GType, `GstMoqSinkPadStatus`.

## Test plan

`just check` and `just test`; 115 `moq-gst` tests pass. The merge preserved `main`'s regression tests for the three fixes that landed separately, and the review on top added `the_first_failure_reason_wins`.

One fix has no regression test: a failed `PLAYING -> PAUSED` no longer closes the EOS gate. Pads are deactivated on `PAUSED -> READY`, not on `PLAYING -> PAUSED`, so an activate-mode blocker never fires there and a plain `GstElement` parent has no other hermetic failure path.

## Cross-package sync

`doc/bin/gstreamer.md` documents both new properties.

(Written by Claude Opus 5)

### DIFF
--- a/doc/bin/gstreamer.md
+++ b/doc/bin/gstreamer.md
@@ -299,15 +299,15 @@ logs or asking a consumer:
 
 | Property      | Type   | Description                                       |
 | ------------- | ------ | ------------------------------------------------- |
-| `status`      | enum   | `pending` until CAPS builds a producer, `active` once the broadcast reserved the track, `ended` when it was finalized, `error` when the pad was invalidated |
+| `track-status` | enum   | `pending` until CAPS builds a producer, `active` once the broadcast reserved the track, `ended` when it was finalized, `error` when the pad was invalidated |
 | `track-error` | string | Why the pad was invalidated, null when it was not  |
 
-Both emit `notify`, so an application can connect to `notify::status` rather than poll. `active`
+Both emit `notify`, so an application can connect to `notify::track-status` rather than poll. `active`
 means the producer exists and the track is registered, not merely that the pad was requested, and a
 pad that sent EOS stays `active` until every pad has ended and the producers are finalized. `error`
 is terminal: it survives EOS, and clears when the pad is released or the element goes back to
 `READY`. A pad still waiting for CAPS is `pending` with no error, because nothing has failed yet.
-Connection loss is the element's own `status`, not the pad's.
+Connection loss is the element's own `status`.
 
 A pad negotiated as `application/octet-stream` publishes application data instead of media. The bytes
 go out exactly as they arrive: no codec, no media container, no interpretation. A data pad's
--- a/doc/bin/gstreamer.md
+++ b/doc/bin/gstreamer.md
@@ -294,6 +294,21 @@ element (back to `READY`) releases the reservation and makes it writable again. 
 the generated name. A name another pad already holds invalidates only that pad, so the rest of the
 broadcast keeps publishing.
 
+Each pad also reports what its track is doing, so a publication can be diagnosed without reading the
+logs or asking a consumer:
+
+| Property      | Type   | Description                                       |
+| ------------- | ------ | ------------------------------------------------- |
+| `status`      | enum   | `pending` until CAPS builds a producer, `active` once the broadcast reserved the track, `ended` when it was finalized, `error` when the pad was invalidated |
+| `track-error` | string | Why the pad was invalidated, null when it was not  |
+
+Both emit `notify`, so an application can connect to `notify::status` rather than poll. `active`
+means the producer exists and the track is registered, not merely that the pad was requested, and a
+pad that sent EOS stays `active` until every pad has ended and the producers are finalized. `error`
+is terminal: it survives EOS, and clears when the pad is released or the element goes back to
+`READY`. A pad still waiting for CAPS is `pending` with no error, because nothing has failed yet.
+Connection loss is the element's own `status`, not the pad's.
+
 A pad negotiated as `application/octet-stream` publishes application data instead of media. The bytes
 go out exactly as they arrive: no codec, no media container, no interpretation. A data pad's
 `container` property has no effect.
--- a/rs/moq-gst/src/sink/imp.rs
+++ b/rs/moq-gst/src/sink/imp.rs
@@ -22,7 +22,7 @@ use hang::moq_net;
 
 use super::pad::{CapsOutcome, ProducerOptions, PushOutcome, caps_supported};
 use super::request_pad::{MoqSinkPad, Notifications};
-use super::session::{CAT, ConnectionStatus, RUNTIME, ResolvedSettings, Session, SessionId};
+use super::session::{CAT, ConnectionStatus, RUNTIME, ResolvedSettings, Session, SessionId, SessionRegistration};
 
 #[derive(Debug, Clone, Default)]
 struct Settings {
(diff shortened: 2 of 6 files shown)


kixelated:
@codex review

arielmol:
@codex review

kixelated:
Heads-up on some changes to this branch, plus the review it was owed. Sorry for the delay in both.

## Three of your four commits already landed

The three fixes here were real and worth having, but they were bundled with a rewrite of the element's per-pad state, which made them hard to review or revert independently. They landed separately against `main`, rewritten against the old structure rather than cherry-picked (they conflicted, since they sit on top of the refactor here):

- #3101 — `READY -> PAUSED` rollback and the teardown ordering
- #3102 — session-scoped terminal errors
- #3104 — a flush restarting a pad that already sent EOS

So `5598b2bff`, `40bb1d17f` and `295e3f61d` are now largely redundant. What is left is the part that actually needed the discussion: the per-pad lifecycle refactor and the two pad properties.

## I merged `main` in

Every `moq-gst` hunk conflicted, since both sides now fix the same bugs in different shapes. A rebase would have meant inventing intermediate states that get overwritten two commits later, and emptying three commits while leaving their messages claiming to fix things already fixed. So it is one merge commit (`91001c614`) instead, and your four commits are untouched.

The resolution keeps this branch's architecture (`PadLifecycle`, `Control`, `Completion`, the pad properties) and folds in what `main` settled. `session.rs` was taken from here wholesale, since `SessionRegistration`, `mark_registered`, `wait_for_registration` and `post_session_error` exist on both sides and `main`'s `SessionState`/`errored` are superseded by `SessionId` + `CompletionState`.

Three behaviours came from `main`:

1. **`PAUSED -> READY` no longer tears the session down when the parent fails.** This is the one place the branch was wrong. `40bb1d17f` says the cleanup is unconditional, but GStreamer does not commit the transition when the parent fails, so the element stays in PAUSED. Tearing down left a PAUSED element publishing nothing with no transition left to build a replacement. `a_failed_paused_to_ready_still_releases_the_publication` asserted that behaviour, so it is replaced by `a_failed_paused_to_ready_keeps_the_publication`.
2. **A `FLUSH_START` arm**, expressed against `PadLifecycle`. Without it another pad's EOS finalizes the publication inside the flush window, which `FLUSH_STOP` cannot undo.
3. **`STREAM_START` re-anchors the media timeline**, not just the ended flag. Otherwise a stream restarting at zero reads as a rewind, invalidates the pad and silently drops its buffers.

Plus `main`'s regression tests that had no equivalent here. 114 `moq-gst` tests pass, clippy and rustfmt clean.

## Review of what remains

The integration tests are the strongest part of this PR, and the `eos_posted` diagnosis was correct: one flag answering four questions is a real problem.

Four things I would like to see changed before it lands:

**Two identity mechanisms.** `SessionId(Arc<()>)` and `CompletionHandle = Arc<CompletionState>` both exist to scope things to one session, and the latter is already a unique per-session `Arc`. `Arc::ptr_eq` on the completion handle gives identity for free. I left this alone deliberately, since it is your design call and not something to slip in through a conflict resolution, but `SessionId` looks like a parallel one-off that `Completion` can absorb.

**`MoqSinkPadStatus` should be `GstMoqSinkPadStatus`.** The other two are `GstMoqSinkConnectionStatus` and `GstMoqSinkMediaContainer`. GType names are ABI (gst-inspect, the Python/JS bindings), so this is cheap now and breaking later.

**The pad's `status` collides with the element's `status`.** They mean different things, and the docs have to say so explicitly ("Connection loss is the element's own `status`, not the pad's"). That sentence is the smell. `track-status` pairs with `track` and `track-error` and deletes the disclaimer.

**It adds more redundant state than it removes.** A pad now carries `Status` + `ended` + `releasing` + `completion` + `media.is_failed()`. `Status::Error` and `media.is_failed()` are two sources of truth for "did this pad fail", and `ended` duplicates `Status::Ended`. That is the same failure mode as `eos_posted`, just spread wider. Same question for `Control.playing` shadowing GStreamer's own state.

One smaller thing: the CAPS arm's two early returns call `gst::Pad::event_default` while still holding the `control` lock, where the Segment/FlushStop/StreamStart arms all `drop(control)` first. Probably benign for a sink pad, but it contradicts the lock discipline the module doc claims.

## Open question, filed as #3115

The residual gaps in all three merged PRs turned out to be one root cause: a `moqsink` publication has no notion of a **generation**, so "restart" can only mean a cycle through `READY`. That is what blocks a conformant flushing seek after EOS, and what leaves the check-to-post window in `post_session_error` unclosable without deadlocking on inline bus sync handlers.

`Completion` is most of the way to a generation. Worth deciding whether generations belong in this PR or in a follow-up on top of it — I would lean follow-up, to keep this one reviewable.

(Written by Claude Opus 5)

kixelated:
Went ahead and made the changes from the review, on top of the merge. Three commits, all with `just check` and `just test` green (114 `moq-gst` tests).

**`ee91f848` — the completion handle is the session identity.** `SessionId` is gone. `CompletionHandle` was already a unique `Arc` per session, so `Arc::ptr_eq` on it answers "is this still the live session?" without a second token. This also drops the `id` field from `Session` and the `id` parameter threaded through the reconnect task.

**`b23095db` — the pad property is `track-status`, the GType is `GstMoqSinkPadStatus`.** The rename removes the collision with the element's `status`, so the doc no longer has to disclaim it. The GType now matches `GstMoqSinkConnectionStatus` and `GstMoqSinkMediaContainer`; that one is ABI, so it was worth doing before release rather than after.

**`ae724b0f` — the CAPS arm drops the control lock before forwarding.** Its two early returns called `gst::Pad::event_default` while still holding `control`, where every other arm drops first. I checked the whole of `handle_event` and there are no remaining paths that forward under the lock.

**One review point I withdraw.** I said the pad carries redundant state, specifically `Status::Error` versus `media.is_failed()` and `ended` versus `Status::Ended`. Looking properly, that was wrong on both counts. `Pad::is_failed()` is the producer's own state and has its own coverage in `pad.rs`; `Status::Error` is the reported status, and `PadLifecycle::fail()` is already its single writer. Collapsing them would push producer state up into the outer type for no gain. And `ended` (this pad sent EOS) is genuinely not `Status::Ended` (the track was finalized) — they are different facts and the EOS aggregation needs the first. `Control.playing` is likewise justified by the comment already on it: the current state is not committed until `change_state` returns.

So the substantive design points are settled, and what is left is the refactor and the two properties on their own merits.

(Written by Claude Opus 5)

kixelated:
Final review pass before merging. Codex verdict: MERGE, no new findings.

**Fixed in `972122eb9`:**

- **A failed `PLAYING -> PAUSED` permanently stranded the EOS.** The arm closed the gate whatever the parent answered, calling itself fail-closed, while its own comment noted a failed parent leaves the element in PLAYING. `claim_eos` then refuses, and with no later entry to PLAYING nothing ever claims it, so an application waiting on EOS hangs. It now mirrors the `PAUSED -> READY` arm below it. This is the same mistake as the one corrected on `main` in #3101: treating a failed downward transition as though it committed.
- **`PadLifecycle::fail` overwrote the recorded reason.** `error` is terminal and `track-error` is meant to say why the pad stopped, but a pad that failed on its bitstream and then saw an unsupported caps event reported the caps instead, because the terminal check consulted only the session's completion and not the pad's own state. Fixed at the source rather than the call site, so the whole class goes away. `the_first_failure_reason_wins` covers it. Codex confirmed first-writer-wins cannot strand a stale error: `reset` clears it on session start, session stop, pad release and detached-pad reuse.

**Not fixed, deliberately:** the check-to-post window in `post_current`. Tracked in #3115 and described under "Known limitation" in the PR body. Codex agrees it is out of scope: the remedy it originally suggested (joining the old session task before installing a replacement) would block a GStreamer state change on network teardown and self-deadlock if the task is inside a bus sync handler.

**One gap worth stating plainly:** the EOS-gate fix has no regression test. Pads are deactivated on `PAUSED -> READY`, not on `PLAYING -> PAUSED`, so an activate-mode blocker never fires there. Codex checked this against the GStreamer source independently and confirmed base `GstElement` unconditionally succeeds for that transition, so there is no hermetic seam short of replacing a vfunc.

Two stale bot comments were verified as already fixed on the branch rather than acted on: "register the session before spawning its task" (the `SessionRegistration` gate is present) and "ignore unsupported CAPS after finalization" (the `terminal` check exists).

`just check` and `just test` clean, 115 `moq-gst` tests, CI green.

@arielmol — this ended up with a fair amount of my work on top of yours: the merge, the `Completion`/`SessionId` consolidation, the `track-status` rename, and the two fixes above. Happy to revert any of it if you disagree with a call, particularly the rename.

(Written by Claude Opus 5)

PR #3005, opened 2026-08-22
Specify dynamic (pattern) advertisements in moq-lite, plus a moqt extension draft

kixelated:
A publisher can advertise a **pattern** of paths it could serve on demand, a head and a tail with either half empty, instead of enumerating every broadcast.

## Why

An announcement says a broadcast exists and, by its route, where to reach it. Without one there is no route at all: an endpoint holds sessions rather than a directory, so a path nobody announced has no session to send a request down, however willing some peer would have been to serve it.

The set that would fix this cannot be enumerated, and not because it is large. A transcode worker cannot announce outputs before the inputs exist. A service cannot announce a customer who has not signed up yet. `origin::Dynamic` already solves this locally for `moq-mux`; this is the same idea given a wire representation and a pattern, so a relay can pick which session to route an unadvertised path down.

Three workloads motivate it (planned in [moq.pro's questline](https://github.com/moq-dev/moq.pro/pull/1111)): a transcode worker advertising `("", "transcode.pro")` once for a whole fleet, a chat backend advertising `("<pid>/chat", "")` for rooms that exist independently of any broadcast, and an archive advertising `("", "")`.

## moq-lite

- **A new Dynamic Stream (0x7)**: DYNAMIC_REQUEST/DYNAMIC_OK, then DYNAMIC_START, DYNAMIC_END, and DYNAMIC_UPDATE, mirroring the Announce Stream's mechanics with their own Dynamic ID space. Kept separate from Announce because the two answer different questions and are consumed differently: one enumerates content, the other describes capability, and a receiver that wants capability without inventory asks for exactly that.
- **Requests are patterns.** ANNOUNCE_REQUEST and DYNAMIC_REQUEST both carry a `Path Head` and a `Path Tail`, so one vocabulary covers asking and advertising. The tail filters on the wire; a peer whose version cannot carry it is filtered by the receiver instead, which is also how a relay serves several requests from one looser upstream subscription. Filtering a pattern against a pattern is a different test from filtering a path against one, so DYNAMIC_REQUEST states it as three explicit conditions rather than leaving "could match" to interpretation.
- **ANNOUNCE_START carries a `Path Middle`**, eliding the requested tail as well as the requested head. A matching path cannot overlap its own halves, so the middle is always well defined, and empty when the path is exactly head followed by tail.
- **Head/tail, not prefix/suffix.** Neither is a substring operation, and the tail is where that bites: a tail of `dash` does not match `room.dash`. The consequence is stated outright, that anything a path will be matched on has to be a segment of its own, with `customer/room.dash` named as the layout that cannot be selected by tail. `head/**/tail` is documented as an API spelling of the two fields; no glob reaches the wire.
- **A dynamic advertisement is a capability, not an inventory**: it never implies a matching path exists, and refusal denies one path. Pattern halves travel whole rather than rebased, since an advertisement is routinely broader than the request that surfaced it. No Epoch, and now the draft says why: two advertisers of one pattern are a pool rather than a contest, so there is no generation to arbitrate, and a fleet handing off prices its cost above the incumbent and drains instead of suppressing it.
- **Resolution**: only the most specific matching tier is consulted, and that tier is the whole answer rather than the head of a fallback list. Cost orders it and a pinned rendezvous FNV-1a of the normalized path against the advertiser identity spreads across the cost-equals. A standby seed SHOULD clear a RECOMMENDED constant floor of `2^32`.
- **New correctness rule**: a receiver MUST NOT resolve a path until the initial announcement set covering it has arrived. The two streams carry no ordering between them, so a handful of patterns arrives long before thousands of announcements, and resolving in that window starts a producer for a broadcast that already exists.
- **Resolved content is an ordinary broadcast**, announced concretely by whoever holds it so the next request is answered from the announcement. The advertisement is not consumed by resolution.
- **Typed refusal**: stream error code **0x30 DYNAMIC_CAPACITY** permits exactly one re-resolution, within the tier and excluding the refuser. Every other code is terminal. The reserved stream error range is split to make room without puncturing it: 32-47 stays reserved for the provisional placeholders implementations emit today, 48-63 becomes moq-lite's own assigned space. It is the one range that cannot forward across a bridge unchanged, so both documents name the mapping.
- **Pattern support is baseline in lite-06**, not something a peer may quietly not implement.

## moq-dynamic

The moq-transport half is its own extension draft, implementable standalone: cluster is a normative reference whose HOP_PATH/ROUTE_COST rules apply unchanged, and multi-hop resolution requires it because HOP_PATH is the only loop defense. A dynamic advertisement is a PUBLISH_NAMESPACE or NAMESPACE carrying the **DYNAMIC_SUFFIX parameter** (0x40B5F), with the DYNAMIC setup option (0x40B5D) declaring support and DYNAMIC_CAPACITY (0x40B5C) the typed refusal.

It keeps prefix/suffix vocabulary, deliberately: a namespace is a tuple of discrete fields, so the substring trap that motivates head/tail in lite does not exist there. It also documents that a moqt receiver takes both kinds and separates them itself, since moq-transport has no request that could carry the distinction.

## Not in this PR

- **`Ended` removal.** It was bundled here and is now restored; it stands alone and gets its own PR.
- **The request tail as a moqt SUBSCRIBE_NAMESPACE parameter.** Suffix filtering is useful with no patterns anywhere, so coupling it to negotiating pattern support is wrong. Its own small extension draft, in its own PR.

`just drafts check` passes on all ten drafts and the doc-site table test passes.

(written by Claude Opus 5)


kixelated:
Reworked in cdbafd7b per @kixelated's review: wildcards now live on their own Wildcard Stream (0x7) instead of riding the Announce Stream. Beyond the API-shape argument (routing capability vs content discovery), the shared design had two concrete problems the separation removes. Announce IDs are implicit — derived from message order — so a peer that merely tolerated unknown wildcard messages would still have had to count them to keep broadcast ids aligned; each stream now has its own counter. And a new message type on the shared stream would have killed an older peer's announce stream outright, forcing the accept-first fleet staging; the spec already defines unknown stream types as a non-fatal reset ("the fallback when an extension stream is opened against a peer that did not negotiate it"), so the new stream degrades gracefully and announcements are never at risk.

The cluster mirror is unchanged (parameter-based). The cross-kind rules the shared design needed (ANNOUNCE_UPDATE-on-a-wildcard-id violations, the Active Count contortion) are gone with it.

(written by Fable 5)

kixelated:
Two more review-driven changes in 307422ac:

**Renamed wildcard → dynamic** throughout, per @kixelated: the mechanism routes *dynamically produced* broadcasts/namespaces, while "wildcard" named the matching shape and implied glob syntax the design deliberately lacks (the message carries two literal halves, no wildcard character). The lite stream is now the Dynamic Stream (0x7) with DYNAMIC_REQUEST/OK/START/END/UPDATE and Dynamic IDs; prose says "dynamic advertisement".

**Split the moqt half into a standalone extension draft**, `draft-lcurley-moq-dynamic`, instead of a parameter buried in moq-cluster: a DYNAMIC setup option (0x40B5B, empty value) negotiates support — required, because an un-negotiated receiver would read the advertisement as a concrete namespace — and the DYNAMIC_SUFFIX parameter (0x40B59) carries the pattern's suffix on PUBLISH_NAMESPACE and, when cluster is also negotiated, the extended NAMESPACE. Selection, capacity-refusal retry, and authorization travel with it; ROUTE_COST ordering is an optional integration referencing moq-cluster informatively, so the extension is usable without the mesh machinery. moq-cluster is reverted to untouched.

`just drafts check` and `bun run --cwd doc check` pass; the new draft is picked up by the doc site automatically.

(written by Fable 5)

kixelated:
Ran a high-effort review over this PR (8 finder angles, adversarially verified); ten findings, nine confirmed, all fixed in eec69912:

- **Prefix rebasing was lossy** — a pattern forwarded onto a deeper-scoped stream lost its original prefix, inflating specificity per-relay and shrinking coverage on re-forwarding. The prefix now travels whole; the request prefix filters, never rebases.
- **The ended-broadcast lifecycle was undefined** after the Ended removal — keep-advertised and retract were both conformant and could not interop. Now: retraction claims nothing about content and never disturbs in-flight subscriptions; a recording retracts and serves stored groups over FETCH, discovered out of band.
- **CAPACITY sat at 0x20, which deployed moq-net already emits as its NotFound placeholder**, and the range narrowing left session-space 0x20 uncovered. Moved to 0x30 with an assigned-exception clause on the reserved range.
- **Selection**: strict cost-then-hash made the hash unreachable for unequal costs; the draft now states spreading among cost-equals is the design, pins the rendezvous FNV-1a hash (seed, input, advertiser identity = first hop entry), states per-receiver recursion, and defines the mixed dynamic-vs-concrete comparison (single cost as both halves, kind carries no rank).
- **CAPACITY semantics**: scoped to dynamically resolved requests, exclusion keys on the advertiser identity, and a receiver that spent its one re-resolution refuses downstream with a terminal code so the retry cannot compound per hop.
- **Authorization** now binds the receiver in both documents (the two drafts previously put the MUST on opposite actors) with discard as the action.
- **moq-dynamic is now implementable standalone**: cluster is a normative reference whose HOP_PATH/ROUTE_COST rules apply unchanged, multi-hop resolution requires it, DYNAMIC_CAPACITY is named and registered, and the codepoints moved to 0x40B5C/D/F — the old ones were moq-broadcast's two values swapped.
- **Path grammar and equality** are defined (normalized non-empty segments, byte equality).

Implementation of the Dynamic Stream in rs/moq-net is underway on a stacked branch.

(written by Fable 5)

PR #3011, opened 2026-08-23
Implement the Dynamic Stream: pattern advertisements in moq-net

kixelated:
The advertise half of [moq.pro's dynamic questline](https://github.com/moq-dev/moq.pro/pull/1111), implementing the Dynamic Stream that [#3005](https://github.com/moq-dev/moq/pull/3005) (this PR's base branch) specifies. Stacked on that branch; rebase onto `dev` once it merges.

## What's here

- **`Path::has_suffix` / `strip_suffix`** — the segment-aware suffix twins of the existing prefix pair, plus the **`Pattern`** type: `(prefix, suffix)` with `matches`, `overlaps`, and `specificity`.
- **Wire** (`lite/dynamic.rs`, stream type 0x7): `DYNAMIC_REQUEST`/`DYNAMIC_OK`, then `DYNAMIC_START`/`END`/`UPDATE` with implicit Dynamic IDs. The pattern prefix travels whole — the request prefix filters which patterns a stream carries, never rebases the encoding (the lossy-rebasing fix from the draft review). One route cost varint, no Epoch.
- **Model**: the pattern table lives inside the origin's existing shared dynamic state (beside the request queue, off the announce tree — a pattern is not a tree node). `Producer::advertise(pattern, cost)` checks the prefix against the publish scope — the catch-all and every suffix-only pattern demand an unscoped handle — and returns a handle that reprices in place (`set_cost`) and retracts on drop. `Consumer::patterns()` observes the table rebased by the cursor's root, filtered by its excluded origin, and deduplicated to the best route per pattern (cost, then chain length, then advertiser id), so duplicates present combined.
- **Subscriber**: one Dynamic Stream per allowed prefix on lite-06+, mirroring received advertisements into the table through the same hop-append / reflection-drop / empty-chain-fill / link-cost pipeline announcements use. Every failure retires the machine quietly — a peer without support resets the unknown stream type, and that must never cost the session (the announce machine's reset handling would have).
- **Publisher**: serves the table as an epoch-driven snapshot diff per subscriber (split-horizon exclusion via the cursor, scope-and-prefix overlap filtering). No demand watch, no cost discount, no linger — a pattern is never warm.
- **`StreamError::Capacity` = 0x30**, the draft's one assigned code in the reserved band; it round-trips, unlike the placeholders around it.

## Deliberately not here

Resolution — routing a subscribe/FETCH for an unadvertised path through the table (specificity tiers, the rendezvous hash, route install, capacity re-resolution) — is the questline's next quest and a follow-up PR; nothing consumes the table for dispatch yet. The JS client is a later quest too.

The last commit cherry-picks #3010 (the macOS bench-stub fix) so `just test` compiles workspace-wide here; it drops out on rebase.

`just check` and `just test` pass.

(written by Fable 5)


kixelated:
Closing as superseded by moq-dev/moq.pro#1226.

The `(prefix, suffix)` Pattern here is incompatible with the settled shared matcher: v1 is exact by default and supports whole-segment `*`, one in-segment `*`, and one whole-segment `**`, with exact containment, set-valued rebasing, and shared structural specificity.

The wildcard cost, pool, refusal, and routing work remains in the m2 wildcard questline. Implementation should resume after the core matcher lands rather than adapting this prototype into a second dialect.

(written by GPT-5.6)

PR #3012, opened 2026-08-23
feat(net): add insert_scope and remove_scope to AnnounceConsumer

### LINKED ISSUE #2714: Per-subscriber path predicate for OriginConsumer (0.1.x) / AnnounceConsumer (0.2.x)
Hi @kixelated 👋

We're building [WolvaneSfu](https://github.com/postanteGames/WolvaneSfu) — a MoQ-primary SFU backend (Rust media-engine + Elixir signaling) using `moq-net` for the relay side. We need a per-subscriber dynamic path exclusion filter to implement server-authoritative moderation (deafen enforcement — audio path subscribes filtered per-user based on server state) that a malicious/forked client cannot bypass.

Current `scope(&[Path])` is inclusion-only and immutable — it doesn't compose with a runtime-updated exclusion set. We'd like to add:

**Proposed API (0.1.x)** — `src/model/origin.rs::OriginConsumer`:

```rust
impl OriginConsumer {
    /// Returns a new consumer that drops paths from the announce stream when
    /// the predicate returns false. Unannounce events pass unconditionally
    /// (to preserve memory-cleanup semantics for previously announced paths).
    pub fn subscribe_filter<F>(self, predicate: F) -> OriginConsumer
    where F: FnMut(&Path) -> bool + Send + 'static;
}
```

Injection point: `OriginConsumerState::apply_announce` — predicate gate at the top.

**0.2.x equivalent** (`AnnounceConsumer`): a `with_filter(F)` chain builder after `Consumer::announced()`, applied inside `next() / poll_next()` drainage.

**Questions before I open a PR:**

1. Would you accept this API? Any alternative design you'd prefer (different predicate signature, or a fundamentally different approach)?
2. Should the PR target the `0.1.x` maintenance branch (we're pinned to 0.1.18) or `0.2.x` main? Happy to backport both directions if maintenance branches are still cut.
3. Is `0.1.x` still maintained, or should we be planning a `0.2.x` migration first?

If you'd rather I just open the PR and discuss inline, that works too — just wanted to check intent first since a "please migrate to 0.2.x" answer changes our plan meaningfully.

Test suite I'm planning (5 tests, matching the `#[tokio::test]` + `assert_next_wait` style in existing `origin.rs::tests`):

- `filter_blocks_matching_path` — paths where predicate returns false don't appear in `announced()`
- `filter_allows_unannounce_of_blocked` — unannounce passes unconditionally (memory cleanup)
- `filter_composable_with_scope` — `scope(&paths).subscribe_filter(pred)` chains correctly
- `filter_clone_independence` — filter is per-consumer, doesn't leak to siblings
- `filter_late_announce_gate` — announces arriving after subscribe still pass through the filter

Thanks — MoQ is finally starting to feel like it's ready for prod SFU work.

— Mustafa (WolvaneSfu)

retif:
## Summary

Closes #2714. Makes an `announce::Consumer`'s path scope mutable at runtime, using the
`insert_scope` / `remove_scope` shape suggested in the issue rather than a filter callback.

- A cursor now tracks two node sets: the roots it is currently registered on (its live scope) and
  the scope it was constructed with. Widening resolves against the latter, so a prefix that was
  removed can be restored, but a cursor can never out-grow the `origin::Consumer` that created
  it. Widening past construction would make `scope()` no restriction at all. A prefix that
  reaches past that scope is granted for the part inside it, matching how `scope()` already
  narrows to the intersection rather than refusing outright.
- `remove_scope` unannounces everything live under the prefix before unregistering, because once
  the cursor is off the node it can no longer be told what it lost. The existing coalescing in
  the consumer's pending map drops an announce that was never observed, so a broadcast the caller
  never saw produces no stray unannounce.
- Overlapping roots would double-deliver: an announce notifies every consumer registration on its
  way up the tree, so two roots where one is a prefix of the other deliver it twice. Inserting a
  prefix that spans existing roots therefore folds them in and suppresses replay for what they
  already covered, so widening `a/b` to `a` announces the rest of `a` and leaves `a/b` alone.

Both methods return whether the scope actually changed. This is the one deviation from the
signature in the issue, and it is deliberate: without it, the case below is a silent no-op.

## Public API changes

Two additions, both on `moq_net::announce::Consumer`:

- `pub fn insert_scope(&mut self, prefix: impl AsPath) -> bool`
- `pub fn remove_scope(&mut self, prefix: impl AsPath) -> bool`

`impl AsPath` rather than `Path` to match how the rest of `origin.rs` takes path inputs, including
`announce::Consumer::absolute` on this same type.

Additive only, hence `main` rather than `dev`: nothing is renamed, removed, or
signature-changed on the published surface. The remaining changes are private and not part of the
public API: `OriginNode::consume` and `consume_initial` gained a `skip` parameter, a private
`OriginNode::unannounce_all` and `AnnounceConsumer::notify` were added, and `AnnounceConsumer`
gained two private fields.

## Test plan

Twelve tests added in `rs/moq-net/src/model/origin.rs`:

- `test_scope_mutation_is_idempotent`
- `test_remove_scope_unannounces`
- `test_insert_scope_announces_live_broadcast`
- `test_insert_scope_cannot_widen_past_construction`
- `test_insert_scope_grants_the_part_within_the_construction_scope`
- `test_remove_scope_of_descendant_is_unsupported`
- `test_remove_scope_of_never_announced_is_silent`
- `test_scope_mutation_is_per_consumer`
- `test_scope_mutation_composes_with_scope`
- `test_late_announce_respects_current_scope`
- `test_remove_scope_keeps_active_subscription`
- `test_scope_mutation_under_a_root`

Ran `just check` and `just test` on Linux, against this branch merged with `main`, and both
pass: the full unit suite is green and clippy is clean with `-D warnings` on the host and
`wasm32` targets. Not run: the macOS and Windows recipes, and `just rs features`, none of which
this change touches.

## Two things worth flagging

**Removing a scope does not cancel an active subscription.** You flagged this as unknown, so I
tested it. It does not. To be exact about what was measured, since there was no `remove_scope` to
call at the time: a cursor scoped to `room` took the `room/alice` subscription off the announced
stream, then the whole `announce::Consumer` was dropped, which is the maximal version of the
scope going away, and the held `broadcast::Consumer` was still open afterwards. The step to
`remove_scope` is a fortiori rather than a direct measurement. That matches what you said about
only being able to filter broadcasts and not tracks: scope filters discovery, not delivery. It is
called out in the doc comment, because "remove them from my scope" reads like it should stop
delivery and someone will reach for it as a kick mechanism.

**The inclusion-set model cannot express the exclusion the issue asks for.** `OriginNode` has no
link to its parent (the nearby `parent` belongs to `NotifyNode`, for propagating announces
upward), and `OriginNodes::select` only ever narrows. So `remove_scope` can drop a root, but not
a strict descendant of one: dropping `room/alice` from a cursor scoped to `room` would mean
restating the scope as every other child of `room`, which goes stale the moment a new participant
publishes. `remove_scope` returns `false` in that case rather than silently doing nothing.

That leaves the original per-subscriber exclusion use case unserved. This PR works for an SFU
that scopes each subscriber to the participants it may see and toggles them, which is a real
shape (`lite/publisher.rs` already builds one `announce::Consumer` per peer, so mutating it
reaches that peer over its existing stream). It does not work for one that scopes to the room and
subtracts a member. If you want that, it needs an exclusion set alongside the inclusion set, with
a precedence rule, and I am happy to do that instead.

## Cross-package rows skipped

- `drafts/`: no wire change. The prefix set is a local filter, never serialized as a set; each
  prefix already corresponds to its own `ANNOUNCE_REQUEST` stream, and the draft already allows
  multiple Announce Streams with overlapping prefixes.
- `rs/moq-ffi` and `rs/libmoq`: neither exposes `scope` or a prefix-set concept today, so there is
  nothing to mirror unless you want the new capability surfaced through the bindings.
- `js/net`: its `announce.Consumer` holds a single fixed prefix, so matching this would be a
  redesign of that type rather than a mirror. Happy to follow up separately if you want parity.

`doc/concept/layer/moq-lite.md` is updated, since it described the announce prefix as fixed for
the life of the session.

(Written by Claude Opus 5)


kixelated:
afaik the scope needs to be on `origin::Consumer` so we actually disallow/allow specific broadcasts, not just hide announcements?  This feels a little bit difficult.

kixelated:
I think this needs a rethink. Scope should be an authorization primitive, but this implementation only filters out announcements, but doesn't stop accessing broadcasts that should be outside of scope. 

IMO we might just need to rethink auth before doing this.

PR #3030, opened 2026-08-24
feat(cli)!: migrate command-line parsing to usage-rs

kixelated:
## Summary

- Replace direct Clap usage across the Rust binaries and examples with `usage-rs`, including native completion endpoints, strict unknown and duplicate argument handling, and typed adapters for custom values.
- Preserve `moq-cli`'s repeated stage syntax while delegating each stage's parsing, validation, help, specification, and completions to Usage.
- Merge relay and benchmark configuration in the order CLI > TOML > environment > built-in defaults, including canonical handling of legacy TOML aliases.
- Materialize effective defaults instead of representing known defaults with `Option`, backed by a reusable human-readable `moq_tokio::Duration` type.
- Accept AAC as a valid playback selection while continuing to reject unsupported VP8 and VP9 video selections.

## Public API changes

- Breaking: `moq_tokio::connect::Config::{race,resolution_delay,timeout}` now use concrete `moq_tokio::Duration` values instead of optional standard durations.
- Breaking: `moq_tokio::connection::Backoff::{initial,multiplier,max,timeout}` and `GoawayConfig::{redirect,handover}` now expose concrete defaulted values.
- Breaking: `moq_tokio::quic::Config::{idle_timeout,keep_alive}`, `unix::Config::allow`, and `websocket::Config::{enabled,delay}` now expose concrete defaulted values.
- Additive: `moq_tokio::Duration` provides human-readable parsing, display, serde, conversion, and comparison support.
- `moq-ffi`, `libmoq`, and `moq-gst` only adapt their internal construction to these changes. Their published ABI and wrapper surfaces are unchanged, so bindings and binding documentation were not regenerated.

## Wire behavior changes

- None. Protocol negotiation, framing, and peer-visible transport behavior are unchanged.

## Documentation

- Document the relay's CLI, TOML, environment, and default precedence.
- Existing `moq-cli` examples remain valid because canonical command and flag spellings are unchanged.

## Test plan

- `MOQ_STRICT=1 nix develop --command just fix`
- `MOQ_STRICT=1 nix develop --command just check`
- `MOQ_STRICT=1 nix develop --command just test` (3,472 Rust tests passed, 2 skipped; 52 Python tests passed)
- `git diff --check`
- Audit for remaining direct Clap dependencies and attributes

(Written by GPT-5)


kixelated (commented):
I think Usage is a positive direction for these binaries, but I found three regressions that should be fixed before landing, in addition to the existing `moq-relay` / `moq-bench` help-exit finding.

1. **[P1] `moq` prints no help or version output.** In `rs/moq-cli/src/args.rs::parse_error`, `Help`, `HelpAll`, and `Version` are classified as successful control flow, but their message is still built with `usage::render_failure`. Usage 6.3 explicitly renders those variants as an empty string because callers must handle them first. `ParseError::exit` therefore exits 0 after printing nothing for `moq --help`, stage help, and `moq --version`. These variants need their actual help page or version response rendered before conversion, with subprocess tests pinning stdout and status.

2. **[P2] Completion endpoints are bypassed by the custom entry points.** `#[usage(completion)]` installs the `__complete_word__` interception in the generated `Cli::parse()` method. `Invocation::parse` calls `try_parse_from` instead, while relay and bench call their own `parse_and_merge` paths, so all three feed completion requests into the ordinary grammar and reject them. Call `completion_request` before custom parsing, or expose an upstream process-entry helper that custom loaders can reuse. For `moq`, the completion route also needs to account for the repeated stage grammar after `--`, since the root Usage spec models only the first stage.

3. **[P2] The old command aliases become visible.** Clap's `#[command(alias = "publish")]` and `subscribe` aliases were hidden. For Usage subcommands, `alias` is advertised in help and completions; `alias_hidden` is the hidden spelling. The migrated attributes on `Command::{Import, Export}` therefore reintroduce the old names into the published surface. Use `alias_hidden` to preserve the existing canonical `import` / `export` interface.

Two design concerns are worth resolving while this is still a spike:

- Relay and bench now implement layer provenance indirectly through serialize, TOML overlay, deserialize, `update_from`, alias normalization, and manual `has_long_flag` exceptions. Usage's settings/config layer records explicit CLI presence directly and is a better owner for `CLI > TOML > env > defaults`; otherwise each new default-true boolean needs another parser-specific exception.
- The accepted MoQ versions moved from `Version::names()` to four literal `choices(...)` lists across connect/listen and legacy/current flags. A new protocol version can now parse via `FromStr` but remain unavailable through these CLIs until every copy is updated. Please restore a single source of truth, or at least add a spec regression test comparing every version flag's choices with `Version::names()`.

(Written by GPT-5)

PR #3041, opened 2026-08-25
feat(relay): make the auth API the source of truth for live sessions

kixelated:
Supersedes #2974, whose commit this builds on so @wrangelvid's authorship stays in the history. GitHub won't take a fork branch as a base, so this targets `main`.

Split out of what this PR used to be: the `--auth-api-mode proxy` half now lives in #3042, stacked on this.

## The predicate

#2974 decided "still vouched for" by asking whether the endpoint still returns a `key` for the session's `kid`. That can't see:

- **Anonymous sessions.** No `kid`, and no `exp` either, so they got no re-check at all. Gating a project stopped new admissions while its existing tokenless viewers streamed until the peer hung up - the case a project-wide stop exists for.
- **A key replaced under that same `kid`.** Every session the compromised key admitted stays authorized until `exp`, while only new connections see the new key.

So the re-check **replays the admission request** and keeps the session only while the reply still grants what it holds. One predicate, every credential shape. A narrowed grant closes the session and the client reconnects into the narrower one; a widened grant disturbs nobody. mTLS peers are exempt, so a customer-facing decision can't tear down the relay mesh.

## Timings belong to the endpoint

No flag - revalidation is what `--auth-api` means, since an endpoint that can refuse a connection should be able to stop one. `--auth-api-revalidate` and `--auth-api-revalidate-stale` from #2974 are gone, and both timings come off the response instead, so they can be decided per project and per request without a fleet roll:

| directive | meaning here | default | guardrail |
|---|---|---|---|
| `max-age` | re-check cadence, so also the bound on how long a revoked grant keeps serving | 60s | clamped to 1s..24h |
| `stale-while-revalidate` | how long to keep serving on the last good answer while re-checks keep failing | 3x cadence | capped at 24h |

A network error, 5xx, or garbage body is evidence of nothing, so the session keeps serving through jittered retries until that window passes. A brief auth outage doesn't mass-disconnect; a sustained one still fails closed.

## Cost

Two things hold the request rate to one per grant rather than one per viewer: identical requests share one in-flight re-check, and re-checks ride the same cached client as admission, which merges the ones that aren't concurrent (sessions start at staggered times, so without it a staggered audience each dials on its own schedule). The price is a reply up to one `max-age` old, so worst-case revocation is `2 x max-age` - documented rather than traded away for a per-viewer request rate.

`MokaManager::default()` is a 42-entry cache, a library sample value that was fine when it was only read at admission and isn't once every live session reads it on a cadence. Raised to 10k.

## Public API

- `Expired` and `Auth::expired` are `pub`. They were `pub(crate)`, unusable from a crate embedding `moq_relay::Auth` - exactly the consumers that need to close a session on revocation, and the reason to have one seam rather than one revocation path per gateway.
- `AuthParams` derives `Clone`; `AuthError` gains `NotFound`.
- `AuthConfig::revalidate` / `revalidate_stale` removed (added in #2974, never released).

## Test plan

`cargo test -p moq-relay` (204 lib + all integration suites), `cargo clippy --locked --all-targets -- -D warnings`, `cargo fmt --all --check`, `RUSTDOCFLAGS="-D warnings" cargo doc`, `cargo check --workspace --all-targets`: clean.

Two end-to-end tests encode the root cause, and both fail on #2974 as written:

- `revoked_public_grant_closes_anonymous_sessions` - an anonymous publisher and subscriber round-trip a frame, the API withdraws `public`, both close. On #2974 the session carries no re-check at all.
- `rotated_key_closes_live_sessions` - same `kid`, different material. On #2974 `key.is_some()` is still true.

Unit coverage: scope coverage (unchanged / widened / narrowed / re-rooted), withdrawn `public`, rotated key, cadence for absent / zero / tiny / huge `max-age`, `stale-while-revalidate` governing the outage window end to end, a garbage body classified unavailable rather than revoked, and coalescing across one kid's audience (two distinct tokens, one request) but not across roots.

Cross-Package Sync: `doc/bin/relay/auth.md` gains the revalidation contract.

(Written by Claude Opus 5)

### DIFF
--- a/doc/bin/relay/auth.md
+++ b/doc/bin/relay/auth.md
@@ -241,6 +241,36 @@ auth_api = "https://api.moq.dev/cluster/auth"
 
 Unlike the standalone flags, the unified call **fails closed**: any network error, non-2xx status, or unparseable response rejects the connection. The verifying key itself comes from this call, so there is no safe fallback; the endpoint's `Cache-Control` softens transient failures. This applies to mTLS peers as well, including root (`/`) connections such as cluster peers: when an auth API is configured it is the source of truth for every connection (so it can alias and tier the root too), and a failed lookup rejects the connection so the peer reconnects and self-heals once the API recovers. The only fail-open case is when **no** auth API is configured, where the client certificate is the sole credential and the path is used unchanged.
 
+### Revalidating live sessions
+
+A grant that is only checked at connect time can only stop NEW connections. Revoking a key, gating a project, or exhausting a quota would leave every session it already admitted running until its token's `exp` - and an anonymous session has no `exp` at all, so it would run until the peer hung up.
+
+So `--auth-api` keeps asking. The relay re-issues each live session's admission request on the endpoint's own `Cache-Control: max-age` cadence, and closes the session once the reply no longer grants what the session holds. There is no flag: an endpoint that can refuse a connection can stop one, or the two would disagree.
+
+- **Still granted**: the reply still authorizes at least the scope the session already has - a `key` that verifies its credential, or the `public` prefixes it was admitted under. The next re-check waits out the new `max-age`.
+- **Refused** (404, or a reply that no longer grants the session's scope): the session closes immediately.
+- **Unavailable** (network error, 5xx, unparseable body, or a 401/403 rejecting the *relay's own* credential): evidence of nothing about this session. The session keeps serving and the re-check retries with jittered backoff until the outage window passes without a success, then closes. A brief auth outage does not mass-disconnect; a sustained one still fails closed.
+
+The re-check REPLAYS the admission request rather than asking a narrower question, which is what makes one mechanism correct for every credential: a key replaced under an existing `kid` no longer verifies the retained JWT, and a withdrawn `public` block revokes anonymous sessions. `exp` still applies as the outer bound wherever the credential has one. mTLS peers are never revalidated, so a customer-facing decision cannot tear down the relay mesh.
+
+**`max-age` is the opt-in.** Revalidation is switched on by the endpoint, not by relay config: a reply that names a `max-age` is telling the relay how long its answer is good for, and that is the cadence. A reply with no usable `max-age` - none at all, `no-store`, `no-cache`, or `max-age=0` - has not asked to be re-consulted, so the session is never re-checked and its credential's own `exp` remains the only bound, exactly as before revalidation existed. Nothing is invented on the endpoint's behalf, and an existing deployment that sends no `Cache-Control` is unaffected until it opts in.
+
+| directive | meaning here | when absent | relay guardrail |
+|---|---|---|---|
+| `max-age` | re-check cadence, so also the bound on how long a revoked grant keeps serving | no revalidation at all | floor of 1s |
+| `stale-if-error` | how long to keep serving while re-checks FAIL, measured from where `max-age` ends | falls back to `stale-while-revalidate` | - |
+| `stale-while-revalidate` | the same window, for endpoints that express it that way | 1 hour | - |
+
+Either stale directive grants the outage window, and `stale-if-error` wins when both are present - it is the precise license for the case the relay is actually in: revalidation is erroring and the session keeps serving on the last good answer. The window runs from the END of freshness, so `max-age=300, stale-if-error=60` means a session survives 60s of a failing auth API, not that it closes 4 minutes before the first re-check. Zero is allowed and means "close on the first failed re-check" - the re-check itself still gets a full request budget, so a healthy endpoint renews the grant rather than the session closing without ever being asked.
+
+**The relay does not impose a ceiling.** A long `max-age` is a long revocation window, by the endpoint's explicit choice; the only bound is far beyond any real value and exists so the relay's clock arithmetic cannot overflow. Pick the number you are willing to wait for a ban to land.
+
+**The outage default is an hour, on purpose.** A short cadence is a request for a tight *revocation* window, not permission to sever every live session over a brief auth outage. If the default were proportional to the cadence, a 60s cadence would drop the fleet after three minutes - which an ordinary Worker deploy or transient incident can exceed. An hour of the endpoint being unreachable is worth surviving; past that, failing closed is the right call. Send `stale-if-error` to ask for less.
+
+**Budget for twice the `max-age`.** Two things hold the request rate to one per grant rather than one per viewer, and the second costs window. Sessions that would issue an identical request share one in-flight re-check, which merges an audience whose timers fire together. Re-checks also ride the same cached HTTP client as admission, which merges the ones that do not - sessions connect at staggered times, so without it a staggered audience would each dial on its own schedule. The cost is that a re-check may be answered from a cache entry up to one `max-age` old, so worst case is one `max-age` of staleness plus one until the session asks again: `2 x max-age`. Cap what you emit at half the window you are willing to promise.
+
+Note the asymmetry when choosing a long `max-age`: the cadence is set by the reply the relay is already holding, so shortening `max-age` later cannot pull in a re-check that is already scheduled. Whatever TTL you hand a healthy connection is how long an unannounced revocation takes to reach it.
+
 ### Authenticating the relay to the auth API
 
 The outbound HTTP the relay makes for auth (`--auth-api` requests and JWK fetches) reuses the cluster client's TLS configuration. The same `--client-tls-cert` / `--client-tls-key` the relay presents when dialing cluster peers also identifies it to the auth API, and `--client-tls-root` trusts a private CA on the endpoint (env `MOQ_CLIENT_TLS_*`, or `[client.tls]` in TOML). So an auth API can require mTLS and recognize the relay by the same certificate it uses for clustering.
--- a/rs/moq-relay/src/auth.rs
+++ b/rs/moq-relay/src/auth.rs
@@ -1,22 +1,27 @@
 use anyhow::Context;
 use axum::http;
+use futures::FutureExt;
 use moq_native::Transport;
 #[cfg(test)]
 use moq_net::AsPath;
 use moq_net::{Path, PathOwned, PathPrefixes, stats::Tier};
 use moq_token::{Key, KeyId};
+use rand::RngExt;
 use reqwest_middleware::ClientWithMiddleware;
 use serde::{Deserialize, Serialize};
 use serde_with::{OneOrMany, formats::PreferMany, serde_as};
+use std::collections::HashMap;
 use std::path::PathBuf;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
+use std::time::Duration;
+use tokio::time::Instant;
 use url::Url;
 
 /// Parameters extracted from an incoming connection for authentication: the
 /// request-derived path + JWT, plus metadata about the connection itself that the
 /// auth API can bucket on (e.g. the transport). Connection metadata is set by the
 /// relay after parsing the request (the URL/SETUP parsers don't know it).
-#[derive(Default, Debug)]
+#[derive(Default, Debug, Clone)]
 pub struct AuthParams {
 	/// The URL path identifying the broadcast root.
 	pub path: String,
@@ -153,6 +158,12 @@ pub enum AuthError {
 	#[error("key not found")]
 	KeyNotFound,
 
+	#[error("the auth API has no grant for this connection")]
+	NotFound,
+
+	#[error("the auth API could not be revalidated; served a stale cached response")]
+	ApiStale,
+
 	#[error("missing key ID in token")]
 	MissingKeyId,
 
@@ -169,6 +180,29 @@ pub enum AuthError {
 	InvalidKeyId(#[from] moq_token::KeyIdError),
 }
 
+impl AuthError {
+	/// True when the auth API answered and the answer was "no".
+	///
+	/// The distinction is what lets [`Auth::revalidate`] tell a withdrawn grant
+	/// from an unreachable API: a refusal closes the session immediately, while
+	/// everything else is evidence of nothing and only closes once the staleness
+	/// window passes. An unrecognized error is NOT a refusal, so a new failure
+	/// mode keeps sessions serving rather than mass-disconnecting them.
+	fn is_refusal(&self) -> bool {
+		matches!(
+			self,
+			Self::UnexpectedToken
+				| Self::ExpectedToken
+				| Self::DecodeFailed
+				| Self::IncorrectRoot
+				| Self::KeyNotFound
+				| Self::MissingKeyId
+				| Self::NotFound
+				| Self::InvalidKeyId(_)
+		)
+	}
+}
+
 // reqwest::Error → AuthError flows through reqwest_middleware::Error so callers can use `?`
 // on both .send() (returns reqwest_middleware::Error) and .error_for_status() / .text()
 // (return reqwest::Error) without a manual map_err.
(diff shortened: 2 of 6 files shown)


kixelated:
self-reviewed by Claude Code (built-in `code-review` skill, `high` effort, `--fix`). Five findings, all fixed in fa19490. Three were introduced by the first commit.

**Revalidation cost one request per viewer, not per grant.** The flight key was the raw credential, but every viewer of a broadcast holds a *distinct* JWT signed by the same key - so re-checks for one `kid` stopped sharing a flight. Bypassing the HTTP cache compounded it: coalescing only merges re-checks in flight at the same moment, and sessions start at staggered times, so the cache was what merged the rest. Together they'd have turned a 10k-viewer broadcast into 10k auth-API requests per cadence against `/cluster/auth` and the ProjectAuth DO - the exact inverse of the property the design rests on.

Fixed by restoring the cache and giving both the request and the flight key one `Lookup` to derive from, since that drift is what let them disagree. The honest cost, now documented rather than traded away: a re-check can be answered from an entry up to one `max-age` old, so worst-case revocation is `2 x max-age`. Regression test asserts two sessions holding different tokens under one `kid` make exactly one request.

**A `grant` was honored on a `kid` lookup.** There the relay never forwards the token, so it cannot check the signature - and the reply is cached per `kid` and shared across the audience, so a forged token with a known `kid` would have inherited the grant. A grant is now honored only when the credential was actually forwarded. Test presents an unsigned token with a valid `kid` header against an endpoint answering with a grant.

**A verdict grant with a past `exp`** was admitted and then closed on the next tick, which reads as a flap and drives a reconnect loop. Refused at admission, matching what `Key::verify` does with an expired JWT.

One pre-existing issue fixed in passing: `MokaManager::default()` is `Cache::new(42)`, a library sample value. That was tolerable when the auth cache was only read at admission; now every live session reads it on a cadence, so a relay with more than 42 distinct grants would thrash it and re-dial on every miss. Raised to 10k.

`cargo test -p moq-relay` (212 lib + integration), `cargo clippy --all-targets`, `cargo fmt --check`, `cargo check --workspace --all-targets`: clean.

(written by Claude Opus 5)

kixelated:
Reworked in 85cd019 after @kixelated pushed back that this had gotten more complicated than the problem warranted. He was right, and the cause was a decision I inherited rather than questioned: making verdict "a response shape, not a mode" so one endpoint could answer either way per connection.

That put both paths inside a single request, and everything awkward in the previous revision followed from it - the kid-vs-credential split, the never-send-the-credential-with-a-kid rule, the cache-key partitioning, and the coalescing bug the self-review caught. None of it was the cost of revocation. It was the cost of dual-mode.

`--auth-api-mode token|proxy` deletes all of it. `token` is untouched, so existing operators (including #2974's use case) are unaffected. `proxy` forwards the connection verbatim and enforces the grant - the relay verifies nothing and holds no keys.

The flight key is now taken from the request that actually gets sent, rather than rebuilt beside it, which is what made the earlier bug possible in the first place.

(written by Claude Opus 5)

kixelated:
Pushed e612197: the endpoint now owns **both** timings, per response, with no relay config for either.

| directive | meaning here | default | guardrail |
|---|---|---|---|
| `max-age` | re-check cadence, so also the bound on how long a revoked grant keeps serving | 60s | clamped to 1s..24h |
| `stale-if-error` | how long to keep serving while re-checks FAIL | 3x cadence | capped at 24h |

The cadence already came from `max-age`; the outage window was a hardcoded `3 x` that, which is the part that didn't belong in the relay. Both being per-response means an endpoint can hand a project nowhere near its limits a long window and a project close to them a short one, decided per request, with no fleet roll.

`stale-if-error` rather than `stale-while-revalidate` on purpose. RFC 5861 splits them: `stale-if-error` is "keep using what you have when revalidation ERRORS", which is exactly this; `stale-while-revalidate` licenses serving a stale answer during a routine background refresh, which the relay never does. Not pedantic in practice either - moq.pro's `/cluster/auth` already emits `stale-while-revalidate=300`, so honoring that one would have silently moved the outage window from 180s to 300s with nobody deciding it. There's a unit test asserting SWR is *not* read as a session bound.

The relay keeps only guardrails, since an endpoint shouldn't be able to make it poll hot or serve forever.

One asymmetry worth knowing before picking a long `max-age`, now documented: the cadence is set by the reply the relay is already holding, so shortening `max-age` later can't pull in a re-check that's already scheduled. Whatever TTL you hand a healthy connection is how long an unannounced revocation takes to reach it.

(written by Claude Opus 5)

kixelated:
Split into two, since both are landing:

- **#3041** (this one) — revalidation. The auth API becomes the source of truth for a live session, not just for admitting one.
- **#3044** — `--auth-api-mode proxy`, stacked on this. The endpoint decides and the relay just enforces.

They're independently useful and independently reviewable: revocation fixes a hole that exists today, and the mode is a new capability.

Also switched the outage window from `stale-if-error` to `stale-while-revalidate` (d831a91). @kixelated's reading is the better one: past `max-age` the loop is literally serving the session on the last good answer while it keeps retrying the re-check, which is what the directive describes. `stale-if-error`'s trigger matches too, but it's a request-satisfaction rule for a response cache, which this loop isn't — and SWR is the far more widely implemented directive.

Worth flagging for the moq.pro side: `/cluster/auth` already emits `stale-while-revalidate=300`, which is inert today. On the pin bump it becomes live meaning, moving the outage window from the 180s default to 300s. Benign, but it should be a deliberate value rather than one inherited from when it did nothing.

(written by Claude Opus 5)

kixelated:
Adversarial review by Codex (a different model), triaged and applied in 32692b2. Verdict was "do not ship" and it was right — three high-severity findings, all of which let a revoked grant keep serving or a healthy one get dropped.

**The 24h ceiling was decorative.** The loop clamped how often it polls, but every poll went through a cache whose freshness still followed the endpoint's raw `max-age`. With `max-age=31536000` the relay would wake every 24h, be handed the same cached grant, reset its timer, and never reach the origin at all — a revoked session surviving indefinitely, contradicting both the guardrail and the documented `2 x max-age` bound. The cache now gets the same ceiling as its `max_ttl`.

**The singleflight shared a verdict, not a fetch.** The flight captured whichever `Revalidate` created it, verified *that* session's credential, compared *that* session's scope, and returned the answer to every waiter. Two anonymous sessions on one path share a flight key but can have been admitted either side of a narrowed `public` block; two JWTs under one `kid` can carry different claims or expiries. So a session could keep authority the reply no longer granted, or be closed because a different viewer's token had expired. The flight now returns the shared reply and each waiter authorizes it against its own credential and scope — the HTTP request is still coalesced, which was the whole point. CodeRabbit independently flagged this one too.

**The outage deadline expired before revalidation began.** It was anchored to the last success rather than to the end of freshness, so `max-age=300, stale-while-revalidate=60` put the deadline four minutes *before* the first re-check: a single transient 500 would have disconnected every affected session at once — precisely the mass-disconnect the window exists to prevent. It now runs from where freshness ends, and bounds the request in flight rather than only being checked after it, so a stalled endpoint can't carry a session past the deadline on the client's 10s timeout.

Each has a regression test that fails without its fix.

Also applied: `Expired` is `#[non_exhaustive]`, repeated `Cache-Control` field lines are read rather than only the first, and two doc links left dangling by the PR split.

**Rejected — 401/403 should be treated as revocation.** A 401/403 from the auth API rejects the *relay's own* credential (its client cert), not the user's session. Treating that as a per-session refusal would disconnect every live session on the fleet the moment the relay's cert expires — the exact failure the staleness window is for. Per-session refusal is expressed as 404, or a 200 that no longer grants the scope. Documented in `auth.md` rather than changed.

**Not fixed here** — a re-check that returns a changed `tier` keeps the admission tier. Real, but the tier is bound to the session's stats handle at admission and changing it mid-session needs that handle rebuilt; it takes effect on reconnect. Worth its own change if it matters.

`cargo test -p moq-relay` (207 lib + integration), clippy, fmt, rustdoc, and workspace check all clean. #3044 has been rebased onto this and carries its own fix for a `grant.exp` overflow panic the same review surfaced.

(written by Claude Opus 5)

kixelated:
Force-pushed 436ec2e. Two changes.

**No revalidate flags anywhere, including in history.** The final tree already had none - revalidation is unconditional - but the history added `--auth-api-revalidate` / `--auth-api-revalidate-stale` in the first commit and removed them in the second, so a reader or a bisect still met them. Squashed to one commit; `auth-api-revalidate` now appears nowhere in the diff. @wrangelvid keeps credit via `Co-Authored-By`.

**Refactor: `auth_api` and the revalidator were two `Option`s that could never disagree.** `Auth` held `auth_api: Option<(Url, Client)>` and `revalidate: Option<Arc<Revalidator>>`, built as `auth_api.is_some().then(...)` - so they were always both-Some or both-None, and the code carried an `.expect("revalidation requires an auth API")` plus an `Option`-returning `flight_key` to handle a state that cannot occur. They are now one `AuthApi { base, client, revalidator }`, which deletes the expect, the impossible-state plumbing, and the `if self.revalidate.is_some()` guard on arming a re-check. The 3-argument `(client, base, request)` calls became `api.fetch(request)`.

That the two were inseparable is the same fact the flag removal asserts: revalidation is what `--auth-api` MEANS. Worth having the type say it rather than a constructor.

`cargo test -p moq-relay` (207 lib + integration), clippy, fmt, rustdoc, workspace check: clean. #3044 rebased onto this and squashed likewise.

(written by Claude Opus 5)

kixelated:
Two changes in 5c800fa.

## Revalidation is now opt-in *by the endpoint*

`max-age` is the opt-in. A reply naming one is telling the relay how long its answer is good for, which is exactly a cadence. A reply with no usable `max-age` - none at all, `no-store`, `no-cache`, `max-age=0` - has not asked to be re-consulted, so the session gets **no re-check at all** and its credential's `exp` stays the only bound, exactly as before this feature existed.

That removes the last arbitrary constant. The old 60s default was the relay inventing a polling interval for an endpoint that had declined to give one, and it meant this change silently added request load to any existing deployment that wasn't sending `Cache-Control`. Now such a deployment is untouched until it opts in.

It also keeps the property the flag removal was about: the *relay operator* still cannot configure revocation away. The opt-in moved to the endpoint, which is per-project, per-response, and the side that actually knows.

`stale-if-error` is now read too. Either stale directive grants the outage window and `stale-if-error` wins when both are present, being the precise license for the case the relay is in - revalidation is erroring, session keeps serving on the last good answer. Absent both it stays 3x the cadence; there is deliberately no way to set it to zero, since one transient 500 would then disconnect every affected session.

## Refactor

Six loose `REVALIDATE_*` constants and three free functions (`cadence`, `staleness`, `cache_directive`), with `CacheHints` defined 1700 lines away from the constants governing it, are now one type:

- `CacheHints` - the raw directives, with the guardrails as associated consts (`MIN_CADENCE`, `MAX_CADENCE`, `STALE_CADENCES`, `MAX_STALE`) and `directive()` as a private helper.
- `CacheHints::schedule() -> Option<Schedule>` - the whole policy in one function, where `None` *is* "the endpoint didn't opt in". The opt-in change fell out of this rather than being bolted on.
- `Schedule { cadence, staleness }` - what a live session actually runs on, so the loop reads `schedule.cadence` instead of recomputing from raw hints each iteration.

`REVALIDATE_BACKOFF` became `Revalidator::BACKOFF`, since it belongs to the retry loop rather than to anything the endpoint says. `http_client::build` now takes `CacheHints::MAX_CADENCE`, which makes the "cache ceiling must equal cadence ceiling" coupling visible at the call site instead of living in a comment.

Also restored the `2 x max-age` explanation, which I'd dropped while rewriting that section.

`cargo test -p moq-relay` (209 lib + integration), clippy, fmt, rustdoc, workspace check: clean. New tests cover no-opt-in (`no-store`/`no-cache`/`max-age=0`/absent, and a stale directive alone), cadence clamping, and `stale-if-error` winning over `stale-while-revalidate`. #3044 rebased on top.

(written by Claude Opus 5)

kixelated:
**Not merging.** Codex's re-review returned HOLD and it was right on both counts. Pushed 73dd7be with what is unambiguous; one open question needs a decision.

## Open: the 24h cadence ceiling does not bound a revoked grant

The earlier round found that clamping the poll to 24h is meaningless if the cache keeps serving a `max-age=31536000` entry, and I "fixed" it by passing `max_ttl` to the cache. **`max_ttl` is inert in the version we actually resolve.** Proven, not argued: fetch twice through the real client with a 1ms ceiling against a `max-age=3600` origin, plus an uncapped control - the control is served from cache (1 request) and so is the capped one (1 request, should have been 2).

The cause was mine: I read `http-cache` **1.0.0-alpha.6** source for every conclusion, while the lockfile resolves **alpha.7** (`http-cache-reqwest` alpha.8). The caret range moved under me and I never checked. Every claim I made about `max_ttl`, `Warning: 111`, the moka default, and `stale-while-revalidate` support came from the wrong file.

So the hole is open: an endpoint sending a `max-age` above the ceiling keeps its cached grant fresh for that whole duration, and the clamped poll is answered from cache. The `MAX_CADENCE` doc no longer claims otherwise, and the behaviour is pinned by a test so a dependency bump turns it red rather than passing unnoticed.

Three ways out, and the choice is a product call rather than a mechanical fix:

1. **Bypass the cache for a re-check whose raw `max-age` exceeded the ceiling.** Precise - only the pathological endpoint pays, everyone else keeps the one-request-per-grant property.
2. **Drop the ceiling and honour whatever `max-age` says**, keeping only an arithmetic-safety cap. Most consistent with "the endpoint owns the timings", and makes a year-long window the endpoint's explicit choice.
3. **Find the right knob in alpha.7** (`cache_options`, `response_cache_mode_fn`) - unknown cost, and pins us harder to an alpha.

## Fixed in 73dd7be

`ApiStale` fell through the status mapping to 401. The endpoint was unreachable and the cache answered for it, so reporting a rejected credential is both wrong and misdiagnoses the outage. It joins the 502 arm.

## Verified against Codex

- **Stale-on-error reset (its earlier critical): confirmed fixed.** `Warning: 111` is rejected before the body is parsed, `ApiStale` is not a refusal, and only `Recheck::Valid` touches the deadline.
- **The synthetic-`max-age` concern: my first test proved nothing** - it used an uncacheable reply, so the second fetch hit the origin and would have passed whatever the cache did. Codex asked directly whether I was fooling myself, and I was. Rewritten with request-count assertions covering both halves.
- **Zero stale window: correct**, and Codex flagged a comment left stale by the fix, now rewritten.

Still deferred, by request: the `tier`/`alias`-change-on-recheck issue, and `Auth::expired` resolving the endpoint from `self` rather than the token.

(written by Claude Opus 5)

kixelated:
Resolved the HOLD, squashed to one commit (f714d6c). Two decisions.

## No relay-side ceiling on `max-age`

The 24h cadence ceiling is gone rather than reimplemented. It never actually bounded anything - `max_ttl` is inert in the resolved version, so the cache kept serving the entry regardless - and the honest fix was to stop imposing it. A long `max-age` is a long revocation window by the endpoint's explicit choice, which is the same principle that makes `max-age` the opt-in in the first place. The only remaining bound is far beyond any real value and exists so `Instant` arithmetic cannot overflow on a `max-age` near `u64::MAX`.

That deletes the inert `max_ttl` plumbing too, so cluster peer-list polling is back to untouched caching and `cluster.rs` is no longer in the diff at all.

## The outage default is an hour, not three cadences

An auth API going down must not sever the fleet. The default was proportional to the cadence, so a 60s cadence dropped every live session three minutes into an outage - which an ordinary Worker deploy can exceed. It is now a flat hour: long enough to ride out a real incident, still failing closed past that. A short cadence is a request for a tight *revocation* window, not permission to disconnect everyone over a blip.

`stale-if-error` is read and wins over `stale-while-revalidate`, so an endpoint that wants less says so explicitly.

Covered by a paused-time test that admits a session, takes the endpoint down entirely, and proves it is still serving after 30 minutes and closed by two hours.

## Also fixed

`ApiStale` fell through the status mapping to 401 - telling a client its credential was rejected when the endpoint was simply unreachable and the cache answered for it. It joins the 502 arm.

## Method note

The `max_ttl` reasoning was wrong because I read `http-cache` **alpha.6** source while the lockfile resolves **alpha.7** (`http-cache-reqwest` alpha.8). The caret range had moved. Everything I concluded about `max_ttl`, `Warning: 111`, the moka default and stale-directive support came from the wrong file, and it took an empirical probe rather than more reading to establish what actually happens.

213 lib tests plus the integration suites, clippy, fmt, rustdoc and a workspace check all clean.

(written by Claude Opus 5)

PR #3061, opened 2026-08-26
refactor(watch)!: narrow Sync's latency input to Paced

### LINKED ISSUE #3052: watch: narrow SyncInput.latency from Latency to Paced
Follow-up to #3048, which added `latency="instant"` to `<moq-watch>`.

## The problem

`Sync` accepts the full `Latency` union, including `"instant"`, but has no way to honor it. Not pacing video and disabling audio both happen in `MoqWatch`, so a consumer wiring one up directly:

```ts
const sync = new Sync({ latency: "instant" });
new Video.Decoder(source, sync, { enabled });
```

compiles and silently does nothing of the sort. The clock still gates video and an independently wired audio decoder stays enabled. `latencyBounds` resolves `"instant"` to zero bounds, so it degrades to a zero jitter buffer rather than breaking, but that is not the mode that was asked for.

## The fix

```diff
-	latency: Getter<Latency>;
+	latency: Getter<Paced>;
```

plus the matching `as Paced` cast in the constructor. `Paced` is already exported by `js/watch/src/sync.ts`, so this is a three-line diff. A value the component can't implement then stops typechecking, rather than being ruled out by a doc comment.

## Call sites

Narrowing an input rejects callers passing a `Signal<Latency>`. Two in-tree:

- `js/moq-boy/src/game.ts` — its public `latency` knob is typed `Watch.Latency`, so today it advertises `"instant"` and does nothing with it. Same bug, one layer out.
- `js/watch/src/sync.test.ts` — same over-wide annotation.

Both become `Watch.Paced` / `Paced`.

## Why it isn't in #3048

Narrowing an input is a semver break in a published package, so it belongs on `dev` per [CONTRIBUTING.md](https://github.com/moq-dev/moq/blob/main/CONTRIBUTING.md#branch-targeting). #3048 targets `main` and is additive, so this can't be based on anything until it lands and `main` merges into `dev`.

(Written by claude-opus-5)

kixelated:
Closes #3052. Follow-up to #3048, which added `latency="instant"` and deliberately left this out because it targeted `main`.

## Why

`Sync` accepted the full `Latency` union, including `"instant"`, but has no way to honor it. Not pacing video and disabling audio both happen in `MoqWatch`, so this compiled and did nothing of the sort:

```ts
const sync = new Sync({ latency: "instant" });
new Video.Decoder(source, sync, { enabled });
```

The clock still paced video, and an independently wired audio decoder stayed enabled. `latencyBounds` resolved `"instant"` to zero bounds, so it degraded to a zero jitter buffer rather than breaking, but that isn't the mode that was asked for. The restriction was documented in a comment; now it's in the type.

## What changed

```diff
-	latency: Getter<Latency>;
+	latency: Getter<Paced>;
```

plus the matching constructor default cast, and the `Latency` doc comment restored to the accurate "`Sync` takes the narrower `Paced` and never sees it".

`Paced` was already exported by `js/watch/src/sync.ts`, so the type side is a three-line change.

## Call sites

Narrowing an input rejects callers passing a `Signal<Latency>`. Both in-tree ones were over-wide in the same way:

- `js/moq-boy/src/game.ts` — its public `latency` knob was typed `Watch.Latency`, so a caller could set `"instant"` and get nothing. Same bug, one layer out.
- `js/watch/src/sync.test.ts` — same annotation.

## Reviewer notes

- **Targets `dev`**: narrowing an input is a semver break in a published package.
- `moq-boy`'s knob narrows too. It's an app rather than a library, and the alternative is leaving it advertising a mode it can't perform.
- `just fix`, `just check`, and `just test` all pass.

(Written by claude-opus-5)

### DIFF
--- a/js/moq-boy/src/game.ts
+++ b/js/moq-boy/src/game.ts
@@ -74,7 +74,7 @@ export class Game {
 	// Reactive state exposed to UI.
 	readonly hovered = new Moq.Signals.Signal(false);
 	readonly active = new Moq.Signals.Signal(false);
-	readonly latency = new Moq.Signals.Signal<Watch.Latency>("real-time");
+	readonly latency = new Moq.Signals.Signal<Watch.Paced>("real-time");
 	readonly userMuted = new Moq.Signals.Signal(false);
 	readonly volume = new Moq.Signals.Signal(0.25);
 	readonly status = new Moq.Signals.Signal<GameStatus | undefined>(undefined);
--- a/js/watch/src/sync.test.ts
+++ b/js/watch/src/sync.test.ts
@@ -1,7 +1,7 @@
 import { describe, expect, it } from "bun:test";
 import type { Time } from "@moq/net";
 import { Signal } from "@moq/signals";
-import { type Latency, Sync } from "./sync";
+import { type Paced, Sync } from "./sync";
 
 // Effects in @moq/signals flush on a microtask, so let pending updates drain before asserting.
 const flush = () => new Promise((resolve) => setTimeout(resolve, 0));
@@ -39,7 +39,7 @@ describe("latency range", () => {
 	});
 
 	it("reacts to a ceiling set after construction", async () => {
-		const latency = new Signal<Latency>("real-time");
+		const latency = new Signal<Paced>("real-time");
 		const sync = new Sync({ latency });
 		await flush();
 		expect(sync.out.buffered.peek()).toBe(false);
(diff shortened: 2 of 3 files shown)


kixelated:
Superseded by #3068, which takes the simpler route: `Sync` now honors `"instant"` itself rather than being kept away from it by the type system. That deletes `Paced`, `DecoderInput.paced`, and the element's mapping signal, so there is nothing left for this PR to narrow.

(written by claude-opus-5)

PR #3070, opened 2026-08-26
moq-native: instrument spawned QUIC driver with the caller's span

gquintard:
Hi!

I've been looking at observability and found I couldn't easily link log messages to an client address, so I fired up an AI agent to get this started. This PR doesn't immediately solve my issue, but it spawns the session with a span which can be used further down the line.

The change itself in `rs/moq-native/src/lib.rs` is quite simple, but testing got a bit hairy and while the code makes sense to me, I'm not experience enough with the codebase to guarantee it odes the right thing™.
The test was initially gnarlier because it tried to work around the log being filtered, which the change in `Cargo.toml` fixes, but I'm not clear if there are caveats to this approach.

I have tested this branch (with some extra local commits to propagate the client address) and it works, but it's my first contribution and it was mostly AI-driven, so beware. Please let me know if I can do anything to make this code change more palatable.

### DIFF
--- a/rs/moq-native/Cargo.toml
+++ b/rs/moq-native/Cargo.toml
@@ -104,4 +104,7 @@ tempfile = "3"
 # test-util adds tokio::time::pause/auto-advance for the time-dependent tests.
 tokio = { workspace = true, features = ["test-util"] }
 toml = "1.1"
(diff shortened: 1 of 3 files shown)


kixelated:
thx

PR #3081, opened 2026-08-26
feat(uring): speak WebTransport for browser peers

kixelated:
## Summary

M4 PR 3 on #2875, stacked on #3078: the **WebTransport-over-quiche adapter** the epic's M4 requires ("implement the WebTransport-over-quiche adapter from M3c as `web_transport_trait::poll` session and stream types"). A browser cannot speak raw QUIC, so without this the relay integration could only serve native peers; with it, `h3` connections get the HTTP/3 CONNECT handshake on the same adapter #3033 built.

- **`quic::web::Request`** runs the server handshake using `web-transport-proto`'s sans-IO state machines over the existing poll streams (its async helpers want tokio I/O, so decoding is incremental over buffered bytes): SETTINGS exchange, the CONNECT request with its URL (what the relay routes and authenticates on), and subprotocol selection, the WebTransport equivalent of ALPN that carries the moq version, mirroring the moq-tokio dispatch.
- **`quic::web::Session`** is now the runtime's one transport type (`Handle::Transport` changed from `quic::Connection`), following web-transport-quinn's `Session::raw` pattern rather than a delegation enum: raw QUIC and WebTransport run the same machinery, with the layering compiled down to an `Option`. Web mode frames opened streams (`0x54`/`0x41` + session id, written lazily ahead of the first payload byte), classifies accepted streams by reading header varints exactly (never past them), prefixes and filters datagrams, holds the peer's control and QPACK streams open for the session's life, and maps stream/close codes through the HTTP/3 error space in both directions.
- **Close semantics for browsers**: `close(code, reason)` writes a `CloseWebTransportSession` capsule in a DATA frame on the CONNECT stream (what `WebTransport.closed` reads), then closes the connection after a one-second grace or the peer acting, whichever first. Inbound, a capsule reader task surfaces the peer's code and reason as the session error, so moq's close codes survive the H3 mapping round trip; the echo test asserts code 42 and its reason arrive intact.
- **Pipelined peers**: a client may open WebTransport streams before its CONNECT is answered. The handshake classifies every early arrival by its header (settings, QPACK, WT-with-claimed-session-id, noise) and seeds matching early streams into the session's accept queue, rather than dropping them. Bidirectional order is safe by construction: the accept queue is in stream-id order, so the client's first bidi stream (CONNECT) always arrives first.

## Validation

`tests/web.rs` runs interop against **web-transport-quinn**, the reference stack browsers interoperate with, on a tokio thread dialing the uring server:

- Hand-driven echo through the framing: bidirectional and unidirectional streams both ways, datagrams, URL path/query and subprotocol assertions, and the close code/reason through the capsule and back out of the H3 error mapping.
- A **full moq-lite session over WebTransport**: subprotocol negotiation, SETUP on the CONNECT-adjacent bidi stream, announce, subscribe, and a group on a unidirectional stream, with the tokio stack as subscriber. This is the browser-to-relay path end to end.

All 27 moq-uring tests pass; `just check` and `just test` are green. Existing raw-QUIC call sites wrap with `Session::raw` (deliberate API change to the unreleased runtime surface).

## Notes

- `moq-tokio` and `web-transport-quinn` join as Linux dev-dependencies only (the client side of the interop tests); the library gains `web-transport-proto`, `http`, and `url`.
- Client-side WebTransport (dialing out over H3) is not implemented: browsers are the client on this path, and cluster dials are raw QUIC by design.
- Capsules split across DATA frames are treated as unparseable (logged); nothing we interoperate with splits them.

(Written by Claude Fable 5)

https://claude.ai/code/session_01XLJU7rEfPe7hn7K5jX77mb


kixelated:
## Codex finding triage

Walked every unresolved Codex thread against the current head. Most were already fixed by 82e4d5203 and c1184ffff (the 64 KiB `MESSAGE_LIMIT`, close-on-handshake-failure, the owned `Response`/`Rejected` types, capsule reassembly) and the threads were just never resolved.

**Fixed here:** `Request::url` returned a bare `url::Url` with no re-export. moq-uring is published as of #3082, so that made `url` part of the public API without saying so. Re-exported it and documented that a major bump is breaking.

**Filed, not fixed:**

- #3103 - an abandoned `Request` never closes the connection (the `respond()` subprotocol early return escapes the `accept()` guard), and `reject()` resets the H3 critical streams instead of shutting down deliberately.
- #3105 - web-mode stream wrappers skip the WebTransport code mapping on `Drop`, and `finish()` turns transient flow-control backpressure into a terminal error.
- #3106 - one stalled stream head-of-line blocks handshake classification.

These want a Linux test to prove the behavior rather than a blind edit, and none of them reach a default build: the whole crate is `#![cfg(target_os = "linux")]` and the relay integration is behind the off-by-default `io-uring` feature.

(written by Claude Opus 5)

PR #3082, opened 2026-08-26
feat(relay): serve QUIC from io_uring workers

kixelated:
## Summary

M4 PR 4 on #2875, stacked on #3081: **the relay integration**. `--runtime-io-uring` (with the existing `--runtime-workers N`) swaps the worker mode's engine from tokio to io_uring: each pinned worker owns a `SINGLE_ISSUER` ring with batched UDP, its own steered `SO_REUSEPORT` socket, and a `quic::Endpoint` serving browsers (WebTransport over `h3`) and native peers (raw QUIC, moq-lite ALPNs) on one socket. moq-transport ALPNs are deliberately not offered (the ietf de-future is still open), which was option 2 on the earlier scope ledger.

**The topology** follows the plan's shape exactly: workers own everything transport-shaped (the QUIC and WebTransport handshakes, the SETUP, the session's protocol machine), while authentication and session supervision run on the shared runtime, which owns the auth API's HTTP client, the timers, and the origins. The handoff is free because #3027 made the session handle `Send + Sync` whatever drives its transport: the worker completes `Request::ok()` and hands the session to a supervision task on tokio (`tokio::runtime::Handle::spawn` from the worker thread; awaiting the JoinHandle needs no reactor).

## moq-net: `accept_request_lite`

The relay authenticates raw-QUIC lite peers from the path + `?jwt=` in their SETUP, which needs the `Request` pause, and `accept_request` demands the `Boxable` (Send) bundle a pinned transport cannot satisfy. Rather than a parallel lite type, the ietf and legacy handshake continuations are now **boxed at construction**, where the caller proved the bounds; `Request` itself, its accessors, and its lite path shed the thread-affinity bounds entirely (they were previously required even to call `Request::path()`, which was an artifact, not a design). `accept_lite` becomes a thin wrapper over the new pause. All 1007 moq-net tests pass unchanged.

## Relay policy shared, not copied

`Connection`'s authorization block (role gating, origin pruning, stats context) and its supervision block (credential expiry, GOAWAY drain) extract into `connection::{authorize, supervise}`, used by both engines. The uring path adds only what is genuinely transport-specific (~100 lines of handshake dispatch).

## A latent `Relay::run` bug, fixed

The integration exposed a pre-existing bug: with QUIC owned by a worker group (tokio workers too!) and no `tcp`/`unix` listeners configured, the stream-only server's accept loop returns `None` immediately, `serve()` bails "stopped accepting connections", and `run()` takes the healthy relay down milliseconds after start. The existing worker test never hit it because it drives the pieces manually rather than through `run()`. `run()` now pends the shared accept loop when QUIC lives on workers and the server has nothing it could ever accept (`accept_health()` is empty).

## Scope notes (deliberate, documented in `doc/bin/relay/config.md`)

- **Cluster/upstream dials stay on tokio** for now; moving them onto the workers' endpoints (the dial side already works, per #3078's test) is the natural follow-up.
- One `listen.tls` certificate, read once at startup: no hot reload, no multi-cert SNI yet.
- mTLS client roots are not yet honored on the uring listener (JWT/anonymous auth only); cluster peers should use token auth against it.
- `moq-uring` drops `publish = false`: moq-relay is published, so its dependency must be publishable. release-plz will pick it up at 0.0.1. Flag this if unwanted.

## Validation

`tests/runtime_uring.rs` runs the whole thing through a real relay via the stock `Relay::run` loop: a WebTransport publisher and four subscribers (two WebTransport, two raw QUIC) against a two-worker steered group, with the frame crossing worker threads through the shared origin, then a clean teardown. Passes in ~160ms on this kernel; kernel-gated below the 6.12 floor. The config-merge regression test covers the new flag. `just check` and `just test` pass across moq-net, moq-tokio, moq-uring, and moq-relay (233 relay tests, 1007 moq-net tests).

(Written by Claude Fable 5)

https://claude.ai/code/session_01XLJU7rEfPe7hn7K5jX77mb


kixelated:
## Codex finding triage

Walked every unresolved Codex thread against the current head. The config-refusal findings (mTLS roots, QUIC-LB, an explicit `listen.backend`, legacy lite versions), the `--quic-*` plumbing, and starting the workers before readiness are all already fixed by 20e252ea2 and 1135856ab; those threads were just never resolved.

**Accepted deliberately:** `Server::accept` / `accept_request` gained `SendStream: MaybeSync` + `RecvStream: MaybeSync`. That is strictly more than `transport::poll::Boxable` promises (it constrains the associated streams to `MaybeSend` only), so it narrows a published method's accepted transport set. It falls out of boxing the IETF continuation at construction, which is what frees `Request` and the lite path from the thread-affinity bounds a pinned transport cannot satisfy; the `MaybeSync` on `Paused` is needed because the relay holds a `Request` across an await behind `&self`. Confirmed as intentional, and `dev` is where a narrowing like this belongs. It is a no-op on wasm, and every native transport in the tree already has `Sync` stream handles.

**Filed, not fixed:**

- #3107 - `/certificate.sha256` is empty when io_uring owns QUIC, so a browser cannot pin a file-provided self-signed certificate.
- #3108 - each worker rereads the certificate files in `Endpoint::new`, and `--quic-mtu-discovery` is inert while both payload sizes stay pinned to the 1350-byte `SEGMENT`.

Worth noting for anyone reading the green checkmark: `rs/moq-relay/src/uring.rs` sits behind the off-by-default `io-uring` feature, so PR CI never compiles it. The nightly `just rs features` pass is what covers it.

(written by Claude Opus 5)

PR #3095, opened 2026-08-27
test(net): assert a stalled write releases the group it was serving

kixelated:
## Summary

A group being actively served can expire mid-stream when a single transport write stalls. CodeRabbit raised this as a truncation bug on #3090 ([thread](https://github.com/moq-dev/moq/pull/3090#discussion_r3867669769)); it is pre-existing on `main` and it is **working as intended**.

Mechanism: `GroupState::poll_frame_source` stamps the group's cache access on every frame the publisher takes, immediately before writing it. Nothing re-stamps inside `write_chunk`. A peer whose flow control window stays shut for longer than the whole retention window therefore stops accessing the group, `evict_expired` ages it out, and the stream resets with `Old`.

That absence of a refresh *is* the DoS bound. The obvious fix (an RAII retention lease held across the delivery) would let any subscriber pin cache indefinitely by simply refusing to read, which is precisely what the retention window exists to prevent. A bounded lease only moves the cliff, and a pool-budgeted one is a no-op in the default config where the pool is unbounded.

Three details worth recording, all verified by the test:

- The grace is a full `latency_max` **per frame handed out**, not per group delivery. The failure needs one write to move zero bytes for the entire window, not merely a slow subscriber.
- The abort only truncates frames the publisher has **not yet taken**. A handed-out `frame::Consumer` owns a cloned `Bytes` that `release()` cannot reclaim, and `poll_terminal` gives a reader that consumed everything a clean end rather than an error.
- The live edge is protected on both eviction paths and is never enqueued into `evict` at all (`insert_group` pushes only the *previous* latest, once superseded). Since `evict_expired` runs solely from `commit_group`, an idle publisher's latest group is never scanned. Already covered by `evict_keeps_max_sequence` and `latest_group_never_evicted`.

The policy was emergent and unasserted, which is how it got re-filed as a bug. This locks it in so the next reader finds the rationale instead of reaching for a lease.

## Changes

- `rs/moq-net/src/lite/publisher.rs`: `stalled_write_releases_the_group`. Stalls a uni-stream write past the retention window while the source keeps publishing (which is what runs the expiry scan), then asserts the serve fails with `Old`, the stream carries that reset code so the peer can tell truncation from a routine cancel, and the in-flight frame still reached the wire while the untaken one did not.
- `rs/moq-net/src/lite/test_transport.rs`: `SinkSession::gated_uni`, mirroring the existing `gated_bi`. Groups travel on uni streams, so this is how a test stalls delivery the way a shut flow control window does.

## Public API changes

None. Test-only; `gated_uni` is on a `#[cfg(test)]`-gated transport double.

## Test plan

- `just test` — 2795 passed, 0 failed, 1 skipped.
- `just rs check-changed` (clippy `-D warnings`, workspace + wasm target) and `just _check-common` — clean. `just check` additionally runs `_flake` (`nix flake check`), which needs network unavailable locally; it is unaffected by this diff.
- Confirmed the new test fails as a genuine repro before being inverted: asserting the group *survives* the stall produced `Old`, which is what pinned the diagnosis.

## Cross-Package Sync

No rows apply: no wire format, `moq-ffi`, config, or CLI surface changed.

(Written by Claude Opus 5)

### DIFF
--- a/rs/moq-net/src/lite/publisher.rs
+++ b/rs/moq-net/src/lite/publisher.rs
@@ -2392,6 +2392,98 @@ mod serve_group_test {
 			"rank 0 must reach the transport as send order 255: {priorities:?}",
 		);
 	}
+
+	/// A subscriber that stops reading must not pin the group it was being served.
+	///
+	/// The publisher stamps the group's cache access once per frame, immediately
+	/// before writing it, so a delivery in progress gets a full `latency_max` of
+	/// grace per frame handed out. Nothing re-stamps inside the write itself: a peer
+	/// whose flow control window stays shut for longer than the whole retention
+	/// window lets the group expire mid-stream and the stream resets with `Old`.
+	/// That is the point. Holding the group for as long as a wedged peer refuses to
+	/// read would let any subscriber pin cache indefinitely.
+	#[tokio::test(start_paused = true)]
+	async fn stalled_write_releases_the_group() {
+		let gate = kio::Producer::new(true);
+		let session = SinkSession::gated_uni(gate.consume());
+		let log = session.log.clone();
+
+		let track_priority = kio::Producer::new(0u8);
+		let subscription = Subscription {
+			session,
+			id: 0,
+			track_name: "test".into(),
+			priority: PriorityQueue::default(),
+			track_priority: track_priority.consume(),
+			track_priority_seen: 0,
+			version: Version::Lite06Wip,
+			timescale: Some(crate::Timescale::default()),
+		};
+
+		let mut track = track::Producer::new(Arc::new(broadcast::Info::default()), "test", None);
+		let mut group = track.create_group(group::Info { sequence: 0 }).unwrap();
+		group
+			.write_frame(Timestamp::from_millis(0).unwrap(), b"first".as_slice())
+			.unwrap();
+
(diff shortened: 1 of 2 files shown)


kixelated:
Not taking the `start_paused = true` → `#[tokio::test]` + `tokio::time::pause()` nitpick.

The guideline it cites ([`rs/CLAUDE.md`](https://github.com/moq-dev/moq/blob/main/rs/CLAUDE.md), "Async tests that depend on time call `tokio::time::pause()` first") is about *having* a paused clock, which this test does. `start_paused = true` is the attribute spelling of the same thing, and it is what the codebase actually uses: 306 occurrences against 122 for the explicit call, including all 16 other tests in `rs/moq-net/src/lite/publisher.rs`. Switching just this one would make it the odd test out in its own file.

It is also the stronger form. `start_paused = true` pauses before the runtime starts the test body, whereas `pause()` as the first statement leaves the setup preceding it on the wall clock.

On the docstring-coverage warning: both symbols this PR adds (`stalled_write_releases_the_group` and `SinkSession::gated_uni`) carry doc comments. The rest of the 8 analyzed functions are pre-existing `web_transport_trait` impl methods in a `#[cfg(test)]` transport double, which the repo's "document every exported symbol" rule does not reach.

(Written by Claude Opus 5)

PR #3102, opened 2026-08-27
fix(moq-gst): scope a terminal session error to the session that earned it

kixelated:
Split out of #2998. That PR needs all three of its fixes, but bundles them with a rewrite of the element's per-pad state; this is the second of the three against `main` on its own.

## Summary

- **Root cause:** the reconnect task posted its terminal error straight onto the element. `Session`'s `Drop` aborts that task rather than joining it, and an abort is not synchronous: a task already inside its terminal arm runs to completion. An application that cycles `PAUSED -> READY -> PAUSED` could therefore see the run it stopped report a fatal error onto the bus of the run that replaced it.
- The task now hands its error to the element, which drops it when the session has been replaced.
- **No new identity token.** The session already allocates an `Arc` for its terminal flag, and every clone of it belongs to exactly one session, so `Arc::ptr_eq` answers "is this still the live session?" for free. `errored` and the identity are the same fact, so they are now one type, `SessionState`.
- The control lock is released before posting, because `post_message` runs the bus sync handlers inline and a handler reading a property would take it again.

## Public API changes

None public. `SessionState` is `pub(crate)`, and `Session::errored()` keeps its signature and meaning.

## Test plan

`just check` and `just test`, plus `a_stopped_session_does_not_report_into_its_replacement`: it starts a session, keeps its identity, cycles the element to READY and back to PAUSED, then posts an error as each session. The stopped session's error must not reach the bus and the live one's must. Verified failing without the guard.

It is a unit test in `imp.rs` rather than in `tests/element.rs` because reaching this path for real needs a terminal reconnect error (an auth rejection), which the hermetic suite cannot produce.

## Cross-package sync

No row applies: no wire format, no CLI surface, and no change to what `doc/bin/gstreamer.md` documents. The bus error's text and type are unchanged; only the session it is attributed to.

(Written by Claude Opus 5)

### DIFF
--- a/rs/moq-gst/src/sink/imp.rs
+++ b/rs/moq-gst/src/sink/imp.rs
@@ -19,7 +19,7 @@ use hang::moq_net;
 
 use super::pad::{Pad, ProducerOptions, caps_supported};
 use super::request_pad::MoqSinkPad;
-use super::session::{CAT, ConnectionStatus, RUNTIME, ResolvedSettings, Session};
+use super::session::{CAT, ConnectionStatus, RUNTIME, ResolvedSettings, Session, SessionState};
 
 #[derive(Debug, Clone, Default)]
 struct Settings {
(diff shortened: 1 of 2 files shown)


kixelated:
Merging with a known residual, tracked in #3115: the identity check and the bus post are not atomic, so a `PAUSED -> READY -> PAUSED` completing inside `post_message` can still land a stale error on the replacement's bus. The lock cannot be held across the post (bus sync handlers run inline and would deadlock on it), and the natural remedy has the same problem in a different place. This PR still narrows the window from unbounded to the duration of one `element_error!`.

The adversarial review also caught a regression this PR introduced: a terminal error arriving before the element installed its session was discarded as stale, which `main` surfaces unconditionally. Fixed in 761a8426 by parking the task until registration.

(Written by Claude Opus 5)

PR #3104, opened 2026-08-27
fix(moq-gst): let a flush restart a pad that already sent EOS

kixelated:
Split out of #2998. That PR needs all three of its fixes, but bundles them with a rewrite of the element's per-pad state; this is the last of the three against `main` on its own.

## Summary

- **Root cause:** `FLUSH_STOP` re-anchors the pad's timeline and it flows again, but the element never removed it from the `ended` set it aggregates EOS over. The pad stayed ended, so the *next* pad's EOS completed the element while the flushed one was still publishing: the pipeline saw EOS with a live track. `STREAM_START` had the same problem and was not handled at all.
- Both now clear the pad's ended flag, but only while the publication is still open.
- Once the publication is finalized the producers are gone and a flush cannot bring them back, so a buffer arriving after that reports `Eos` to its streaming thread instead of being written into nothing. Recovery is a cycle through READY, which builds another session, broadcast and catalog.

## Scope

#2998 goes further here and separates "the producers were finalized" from "the EOS message was posted", which `eos_posted` conflates today. That separation is what its per-pad lifecycle rewrite buys, and it is not reproducible standalone. This change keeps the conflation and fixes only the reset, which is the user-visible half.

## Public API changes

None. `handle_event` and `forward_buffer` are internal.

## Test plan

`just check` and `just test`, plus three regression tests, all verified failing on `main` and passing with the fix:

- `a_flush_after_eos_makes_the_pad_flow_again` — two pads; `sink_0` sends EOS then flushes, `sink_1` sends EOS. The element must not post EOS.
- `a_new_stream_after_eos_makes_the_pad_flow_again` — the same via `STREAM_START`.
- `a_buffer_after_the_publication_ended_reports_eos` — a buffer pushed after the publication finalized answers `Eos` rather than being accepted.

## Cross-package sync

No row applies: no wire format and no CLI surface. `doc/bin/gstreamer.md` documents `moqsink`'s properties and pad naming, none of which change; it does not document flush or EOS aggregation semantics.

(Written by Claude Opus 5)

### DIFF
--- a/rs/moq-gst/src/sink/imp.rs
+++ b/rs/moq-gst/src/sink/imp.rs
@@ -490,6 +490,12 @@ impl MoqSink {
 		if state.session.errored() {
 			return Err(gst::FlowError::Error);
 		}
+		// The publication was finalized: the producers this buffer would be written into are gone, and a
+		// flush cannot bring them back. Answer the streaming thread rather than dropping it as `Ok`
+		// against nothing. Recovery is a cycle through READY, which builds another session.
+		if state.eos_posted {
+			return Err(gst::FlowError::Eos);
+		}
 
 		// The pad almost always exists already (caps arrive before buffers), so look it up without
 		// allocating an owned name; only the rare first-buffer insert pays for the key.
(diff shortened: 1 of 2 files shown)


kixelated:
Merging with a known residual, tracked in #3115: a flushing seek after the element has already completed EOS stays broken. Once the last pad ends, the producers are finalized and the catalog is taken, and `FLUSH_STOP` cannot reopen them, so this returns `FlowError::Eos` where GStreamer specifies that flow resumes. `main` is already broken here (it pushes into finalized producers); this makes the failure explicit rather than silent. The real fix is a publication generation, which is the lifecycle work in #3115.

The adversarial review also found two gaps in the original commit, both fixed in abdb0df7: FLUSH_START was unhandled, so another pad's EOS could finalize the publication mid-flush; and STREAM_START cleared the ended flag without re-anchoring the timeline, so a stream restarting at zero read as a rewind and had its buffers silently dropped.

(Written by Claude Opus 5)

Issue #3108, opened 2026-08-27
moq-relay: io_uring listener re-reads the certificate per worker, and honors --quic-mtu-discovery in name only


kixelated:
Two fidelity gaps in the io_uring listener, both found by Codex review on #3082 and both behind the non-default `io-uring` feature that PR CI does not compile.

### The certificate is loaded once per worker, not once per listener

`Workers::serve` calls `Endpoint::new` on each worker thread, and each call invokes `server.quiche()`, which rereads the certificate and key paths. If the files are replaced while the threads are starting, earlier and later workers end up serving different identities, and a client pinning either one fails depending on which worker accepts it. This contradicts the documented single startup read in `doc/bin/relay/config.md`.

Parse the identity once before spawning the workers and share the resulting TLS configuration.

### `--quic-mtu-discovery` cannot discover anything

`rs/moq-uring/src/quic/mod.rs` enables quiche's discovery while pinning both payload sizes to the 1350-byte `SEGMENT`:

```rust
config.set_max_recv_udp_payload_size(SEGMENT);
config.set_max_send_udp_payload_size(SEGMENT);
...
config.discover_pmtu(transport.mtu_discovery);
```

and `Driver::flush` only ever hands `conn.send` 1350-byte slices, so the ceiling never moves. The relay accepts and documents the setting as honored while nothing changes.

Either raise the configured and staging-buffer ceiling when discovery is on, or reject the option in this backend the way #3082 already rejects mTLS roots, QUIC-LB, and an explicit `listen.backend`.

(written by Claude Opus 5)

kixelated:
Fixed in #3116, merged to `dev` as 1935549bd.

The certificate is now read once, in `Workers::bind`, before any worker thread exists: `quic::Identity` holds PEM rather than paths. `an_identity_is_read_once` deletes the files from disk and then builds two endpoints from one `Identity`.

`--quic-mtu-discovery` went the second way you offered: the relay already refused it here (that landed with #3082), so what was left was `moq_uring::quic::Transport::mtu_discovery` being a public knob that did nothing for every other caller. It is removed. Making it work means variable-size sends, which the pooled GSO path can't express without a rewrite.

(Written by Claude Opus 5)

Issue #3097, opened 2026-08-27
moq-relay: io_uring listener cannot publish its certificate or authenticate mTLS peers


kixelated:
Two gaps in `--runtime-io-uring` (#3082) that need a new API rather than a local fix. Both are refused or documented there rather than silently ignored, but neither is actually implemented.

### 1. `/certificate.sha256` serves nothing

`Relay::load` takes the web-visible fingerprints from either the tokio worker group or the shared server:

```rust
let certificates = match &workers {
    Some(workers) => workers.certificates(),
    None => server.certificates(),
};
```

Under io_uring, `workers` is `None` and the shared server is stream-only, so the handle is empty and the endpoint 404s — even though the workers are serving a certificate. A client that pins a self-signed relay through that endpoint stops being able to connect the moment the flag is flipped.

The fix needs a way to build a `moq_tokio::tls::Certificates` from outside moq-tokio: `Certificates::new` is `pub(crate)` and `Info` is private, so `moq_relay::uring::Workers` cannot expose what it is serving. Either a public constructor that reads and fingerprints a cert/key pair, or a way for the uring group to register with the shared handle.

Documented in `doc/bin/relay/config.md` in the meantime.

### 2. mTLS client authentication is not wired through

`moq_uring::quic::server::Config` supports `ClientAuth::{Optional, Required}` with roots, but the relay never sets it, and `serve_connection` has no peer-identity path — it always calls the JWT/public authenticator. `listen.tls.root` is refused at startup in #3082 so the relay cannot start while claiming mTLS it will not enforce, but the capability is simply missing.

Wiring it up means passing `listen.tls.root` into `ClientAuth`, surfacing the verified peer identity out of `moq_uring::quic::Connection`, and taking the `verify_mtls` branch in `serve_connection` the way `Connection::authenticate` does on the shared path.

---

Both were raised by Codex reviewing #3082 and verified against the code. The rest of that round is fixed in the PR: the Iroh accept-loop regression, auth status mapping, worker-exit supervision, readiness ordering, and the startup refusals.

(Written by Claude Opus 5)

kixelated:
Split so each half stands alone:

- Section 1 (`/certificate.sha256` serves nothing under io_uring) → #3107
- Section 2 (mTLS client authentication is not wired through) → #3112

Closing this umbrella; nothing is dropped.

(written by Claude Opus 5)

PR #3110, opened 2026-08-27
fix(uring): close abandoned handshakes and stop conflating H3 codes

### LINKED ISSUE #3096: moq-uring: HTTP/3 layer edge cases left from the #3081 review
Five findings from the review of #3081 that are real but each need a decision, so they were left out of that PR rather than growing it further. Roughly in the order I'd care about them.

### 1. `unmap_err` reports unmappable HTTP/3 codes as MoQ application errors

`unmap_err` falls back to the original code when `error_from_http3` returns `None`, but keeps the `App`/`Reset`/`Stop` variant:

```rust
let unmap = |code: u64| proto::error_from_http3(code).map(u64::from).unwrap_or(code);
```

Those variants are what `session_error()` / `stream_error()` advertise, so an HTTP/3 *connection* error such as `H3_NO_ERROR` (0x100) reaches moq-net looking like an application code the peer chose, and an invalid RESET_STREAM code looks like a valid MoQ stream error. It conflates "HTTP/3 failed" with "the peer sent you this WebTransport code", which is the distinction the mapping exists to preserve.

The fix is a contract change — an unmapped code should probably become a transport-flavored variant whose trait accessors return `None` — which is why it is here rather than in #3081. This is the one I would want settled before the layer ships.

### 2. `reject` tears down the critical H3 streams

`Request::reject` writes the response and returns, dropping the consumed `Request`. That drops the local control stream and the peer's held control/QPACK receive streams, whose `Drop` sends `RESET_STREAM` / `STOP_SENDING`. HTTP/3 treats closing a critical stream as a connection error, so a client can see an H3 connection failure instead of the rejection it was sent. The streams need to outlive the response, with the connection shut down deliberately afterwards.

### 3. The handshake head-of-line blocks on one stalled stream

`Request::accept` awaits each unidirectional stream's classification in turn. A peer that sends a stream type and then stalls before the session-id varint keeps the loop from ever reaching a control stream that has already fully arrived. The established-session path already polls partially classified streams concurrently (`classify_uni`); the handshake should do the same. The arrival cap added in #3081 does not help, since only one stream has been adopted.

### 4. `finish()` turns transient backpressure into a terminal error

`SendStream::finish` on a web-mode stream that has not written its header yet calls `try_write`, and reports `Error::Web` when connection-level flow control returns zero. That is ordinary backpressure that would clear on its own, and callers generally drop the stream on the error, resetting a stream that should have finished cleanly. Either queue the prefix before handing the stream out, or keep enough state to finish it asynchronously.

### 5. Dropping a web stream sends an unmapped code

A web-mode `SendStream` / `RecvStream` dropped unfinished lets the inner stream's `Drop` send raw code 0 rather than `error_to_http3(0)`, so routine cancellation reads to a browser as an HTTP/3 stream error instead of a WebTransport one.

---

All five were raised by Codex reviewing #3081 and verified against the code. #3081 fixed the resource-bound and API-shape findings from the same round (connection leak on handshake failure, the arrival-count bypass, `held_recv` growth, the abandoned close capsule, the unmapped datagram error, and `respond`/`reject`).

(Written by Claude Opus 5)

### LINKED ISSUE #3103: moq-uring: an abandoned WebTransport handshake leaks its connection, and reject() resets critical H3 streams
Two teardown holes in `rs/moq-uring/src/quic/web.rs`, both on the unauthenticated handshake path. Found by Codex review on #3081; deliberately deferred rather than fixed blind, since the fix wants a Linux test to prove it.

### An abandoned `Request` never closes the connection

`Request::accept` closes `conn` when `handshake` fails, but that guard covers only `Self::handshake`. `respond()` has its own early return:

```rust
if !self.request.protocols.iter().any(|offered| offered == protocol) {
    return Err(Error::Web(format!("subprotocol {protocol:?} was not offered")));
}
```

which consumes and drops `self` without closing `conn`. The peer controls its subprotocol offer, so it can trigger this at will. Dropping the public `Connection` does not close QUIC: the endpoint keeps the connection, its routes, and its driver task until the driver sees a terminal state, and the backlog stopped counting it once accepted. A peer that ignores the resulting stream resets and keeps sending can therefore hold each abandoned handshake open indefinitely.

The clean fix is a close-on-drop guard on `Request`, disarmed only after a completed response or rejection, so every path out is covered rather than each one remembering.

### `reject()` tears down the H3 critical streams

`reject()` writes the response, finishes `self.send`, and returns. Dropping `self` then drops `held` (the peer's control and QPACK receive streams) and `control` (ours, unfinished), and those `Drop` impls issue `RESET_STREAM`/`STOP_SENDING`. RFC 9114 treats closing a critical stream as a connection error, so the client can surface an H3 connection failure instead of the 4xx it was sent. The streams should stay alive until the response is delivered, with the connection then shut down deliberately.

(written by Claude Opus 5)

### LINKED ISSUE #3105: moq-uring: web-mode streams skip the WebTransport code mapping on Drop, and turn backpressure into a terminal error
Two issues in the web-mode stream wrappers in `rs/moq-uring/src/quic/web.rs`. Found by Codex review on #3081.

### `Drop` bypasses the error mapping

`SendStream::reset` and `RecvStream::stop` map the application code through `proto::error_to_http3` when the stream is web-framed. Neither wrapper implements `Drop`, so an unfinished stream falls through to the inner `Drop` in `rs/moq-uring/src/quic/stream.rs`, which issues `stream_shutdown(.., 0)` with a raw 0. Browsers therefore see routine application cancellation as an HTTP/3 stream error rather than a WebTransport cancellation with code 0. moq cancels subscriptions by dropping streams, so this is the common path, not an edge case.

The wrapper needs its own `Drop` that maps 0 before the inner fallback runs. It needs to know whether the inner stream was already finished or reset, which the inner's private `fin`/`reset` flags currently do not expose.

### `finish()` converts transient backpressure into a terminal error

```rust
fn finish(&mut self) -> Result<(), Self::Error> {
    if !self.prefix.is_empty() {
        let n = self.inner.try_write(&self.prefix);
        self.prefix.advance(n);
        if !self.prefix.is_empty() {
            return Err(Error::Web("no capacity to frame the stream before finishing".into()));
        }
    }
    ...
```

Finishing a newly opened, empty WebTransport stream while connection-level flow-control credit is exhausted makes `try_write` return zero and reports a terminal `Error::Web`, even though capacity normally arrives once the peer advances flow control. Callers drop the stream after that unexpected failure, resetting a stream that should have finished cleanly.

Either queue the prefix before handing the opened stream back, or retain enough state to complete it asynchronously instead of failing.

(written by Claude Opus 5)

### LINKED ISSUE #3106: moq-uring: one stalled stream can stall the whole WebTransport handshake
`Request::handshake` in `rs/moq-uring/src/quic/web.rs` classifies each arriving unidirectional stream one at a time:

```rust
let recv = accept_uni(&mut conn).await?;
let mut pending = PendingUni::new(recv);
let class = std::future::poll_fn(|cx| pending.poll_classify(cx)).await;
```

An early WebTransport stream that sends its type byte and then stalls before completing the session-id varint blocks the loop, so a fully available control stream carrying SETTINGS is never accepted. The `HANDSHAKE_STREAMS` cap does not apply either, because only that one stream has been adopted. A peer can keep the handshake stuck for as long as it keeps the QUIC connection alive.

The established-session path already polls partially classified streams concurrently; the handshake should do the same rather than awaiting each classification in turn.

Bounded in practice by the connection idle timeout, and the cost is one connection plus its driver task, so this is lower severity than #3103. Found by Codex review on #3081.

(written by Claude Opus 5)

kixelated:
Six findings from the Codex review of #3081, all in `rs/moq-uring/src/quic/web.rs`. They were filed rather than fixed in that PR because each needed a decision or a Linux test; this is that pass.

Closes #3096, closes #3103, closes #3105, closes #3106.

## `unmap_err` reported unmappable HTTP/3 codes as application errors

Only a sparse range of HTTP/3 codes names a WebTransport error. The fallback kept the original `App`/`Reset`/`Stop` variant for anything else, and those variants are exactly what `session_error()` / `stream_error()` advertise, so `H3_NO_ERROR` (0x100) reached moq-net looking like an application code the peer chose.

Unmappable codes now become a new `Error::Http3 { code, reason }`, whose trait accessors report neither a session nor a stream error. Additive on a `#[non_exhaustive]` enum, and it keeps the code for logs.

## An abandoned `Request` never closed the connection

`Request::accept` closed `conn` when `handshake` failed, but `respond` has its own early return when the peer names a subprotocol it did not offer, and that consumed and dropped `self`. Dropping the public `Connection` does not close QUIC: the endpoint keeps the connection, its routes, and its driver task until the driver sees a terminal state, and the backlog stopped counting it at accept. The peer picks its own subprotocol offer, so it could hold each abandoned handshake open indefinitely.

A close-on-drop `Guard`, disarmed only by a completed response or rejection, now covers every path out rather than each one remembering.

## `reject()` tore down the H3 critical streams

`reject` wrote the response and returned; dropping the consumed `Request` then dropped the peer's control and QPACK receive streams and our unfinished control stream, whose `Drop` impls send `RESET_STREAM` / `STOP_SENDING`. RFC 9114 treats closing a critical stream as a connection error, so a client could surface an H3 connection failure instead of the 4xx it was sent.

It now waits for the response to be acknowledged (bounded by `CLOSE_GRACE`, the same one the close capsule uses), then closes with `H3_NO_ERROR`. Everything stays open until that returns.

## One stalled stream head-of-line blocked the whole handshake

The handshake awaited each unidirectional stream's classification in turn. The accept queue is strictly id-ordered, so a stream that sent its type byte and then stalled before the session-id varint was always adopted first and parked the loop there, while a fully arrived control stream carrying SETTINGS sat behind it. `HANDSHAKE_STREAMS` did not help, since only one stream had been adopted.

It now polls every adopted stream each pass, the way the established session already does, and hands the still-unclassified ones to the session's `pending_uni` instead of dropping them, so a pipelined WebTransport stream is no longer lost to the handshake ending.

## `finish()` turned transient backpressure into a terminal error

Finishing a web-mode stream that had not written its header yet called `try_write` and reported `Error::Web` when connection-level flow control returned zero. That is ordinary backpressure that clears once the peer reads, and callers drop the stream on an unexpected error, resetting one that should have finished cleanly.

`finish` now records that the header is still owed and returns `Ok`; `poll_closed` writes the rest and finishes then. moq-net's writers already pair `finish()` with `poll_closed`, so nothing else changes.

## Dropping a web-mode stream sent an unmapped code

Neither wrapper implemented `Drop`, so an unfinished stream fell through to the inner `Drop` in `stream.rs`, which issues `stream_shutdown(.., 0)` with a raw 0. Browsers therefore saw routine application cancellation as an HTTP/3 stream error. moq cancels subscriptions by dropping streams, so this was the common path.

Both wrappers now map 0 through `error_to_http3` before the inner fallback runs, using a new crate-private `ended()` on the inner streams so a finished stream is not reset instead.

## Testing

Five new tests in `rs/moq-uring/tests/web.rs`, plus a unit test for `unmap_err`. Four of them cover a stall or a leak, so they are bounded by a `within(..)` helper that fails instead of hanging.

- `a_stalled_stream_cannot_block_the_handshake` and `finishing_under_backpressure_is_not_an_error` drive a raw quiche peer (a real client will not produce either case). The second advertises 64 KiB of connection-level flow control and stops reading, so the server genuinely runs out of credit mid-stream.
- `an_abandoned_handshake_closes_the_connection`, `a_rejection_reaches_the_peer`, and `a_dropped_stream_carries_a_webtransport_code` go through `web-transport-quinn`, which is what browsers interop with. The last one asserts `ReadError::Reset(0)` / `WriteError::Stopped(0)` rather than the `InvalidReset` / `InvalidStopped` the unmapped code produced.

Verified each fails without its fix: reverting all five behaviors at once fails exactly the four tests that cover them, on the right assertion each time.

`a_rejection_reaches_the_peer` is honest about its limit: it proves the 404 reaches the client, but `web-transport-quinn` does not police critical-stream teardown, so it would not have caught the reset on its own. The stream lifetimes are the reviewable part there.

Run on real io_uring (kernel 6.19 container): `cargo test -p moq-uring` is 40 tests green, `cargo clippy -p moq-uring -p moq-relay --features moq-relay/io-uring --all-targets -- -D warnings` and `RUSTDOCFLAGS=-D warnings cargo doc -p moq-uring` are clean. `just check` and `just test` pass locally (they compile no moq-uring on macOS, since the crate is `cfg(linux)`).

Not in scope, still open: #3097 and #3108, the relay-side gaps. Those need a new `moq_tokio::tls::Certificates` constructor and the mTLS wiring, and are a separate PR.

(Written by Claude Opus 5)

### DIFF
--- a/rs/moq-uring/src/quic/mod.rs
+++ b/rs/moq-uring/src/quic/mod.rs
@@ -318,6 +318,16 @@ pub enum Error {
 	/// stream that could not be framed.
 	#[error("webtransport error: {0}")]
 	Web(String),
+	/// HTTP/3 failed with a code of its own, one that names no WebTransport
+	/// error (`H3_NO_ERROR`, say). Neither trait accessor reports it, because
+	/// it is not a code the peer's application chose.
+	#[error("http/3 closed: code={code} reason={reason:?}")]
+	Http3 {
+		/// The HTTP/3 error code.
+		code: u64,
+		/// The UTF-8 lossy close reason, empty for a stream-level code.
+		reason: String,
+	},
 }
 
 impl web_transport_trait::Error for Error {
--- a/rs/moq-uring/src/quic/stream.rs
+++ b/rs/moq-uring/src/quic/stream.rs
@@ -39,6 +39,13 @@ impl SendStream {
 		self.id
 	}
 
+	/// Whether the send side is already terminated, so [`Drop`] would do
+	/// nothing. The WebTransport wrapper asks before mapping a reset of its
+	/// own, since a finished stream must not be reset instead.
+	pub(crate) fn ended(&self) -> bool {
+		self.fin || self.reset
+	}
+
 	/// Queue as much of `buf` as quiche will take right now, without parking.
 	/// Best-effort, for the close path where nobody is left to poll.
 	pub(crate) fn try_write(&mut self, buf: &[u8]) -> usize {
@@ -218,6 +225,12 @@ impl RecvStream {
 			stopped: false,
 			backlog: BytesMut::new(),
 		}
+	}
+
+	/// Whether the read side is already terminated, so [`Drop`] would do
+	/// nothing.
+	pub(crate) fn ended(&self) -> bool {
+		self.finished || self.stopped
 	}
 
 	/// [`stop`](web_transport_trait::poll::RecvStream::stop) with a
--- a/rs/moq-uring/src/quic/web.rs
+++ b/rs/moq-uring/src/quic/web.rs
@@ -177,16 +177,16 @@ impl Request {
 				// Stop adopting past the cap, but do not give up on what is
 				// already here: the control stream may be sitting at the head
 				// of a queue a pipelining peer filled behind it, and refusing
-				// that peer is the bug the cap is not for.
+				// that peer is the bug the cap is not for. The one that tips
+				// it over is kept rather than dropped, since dropping it would
+				// cancel a legitimate stream on a handshake that then succeeds.
 				let mut over = false;
 				while !over {
 					match web_transport_trait::poll::Session::poll_accept_uni(&mut conn, cx) {
 						Poll::Ready(Ok(recv)) => {
 							arrivals += 1;
+							pending.push(PendingUni::new(recv));
 							over = arrivals > HANDSHAKE_STREAMS;
-							if !over {
-								pending.push(PendingUni::new(recv));
-							}
 						}
 						Poll::Ready(Err(err)) => return Poll::Ready(Err(err)),
 						Poll::Pending => break,
--- a/rs/moq-uring/src/quic/web.rs
+++ b/rs/moq-uring/src/quic/web.rs
@@ -63,6 +63,9 @@ const HELD_STREAMS: usize = 8;
 /// `H3_GENERAL_PROTOCOL_ERROR`, for closing a connection whose HTTP/3
 /// handshake never produced a session.
 const H3_GENERAL_PROTOCOL_ERROR: u64 = 0x0101;
+/// `H3_NO_ERROR`, for closing a connection that did what it came to do: a
+/// rejection the peer has been told about.
+const H3_NO_ERROR: u64 = 0x0100;
 
 /// An incoming WebTransport handshake: the CONNECT request, ready to answer.
 ///
@@ -72,6 +75,8 @@ const H3_GENERAL_PROTOCOL_ERROR: u64 = 0x0101;
 pub struct Request {
 	handle: Handle,
 	conn: Connection,
+	/// Closes the connection on every way out of here that is not an answer.
+	guard: Guard,
 	request: proto::ConnectRequest,
 	send: super::SendStream,
 	recv: super::RecvStream,
@@ -83,6 +88,42 @@ pub struct Request {
 	/// WebTransport streams a pipelining peer opened before its CONNECT was
 	/// answered, headers consumed, keyed by the session id they claimed.
 	early: Vec<(u64, super::RecvStream)>,
+	/// Streams that arrived during the handshake and are still mid-header;
+	/// the session goes on classifying them.
+	pending: Vec<PendingUni>,
+}
+
+/// Closes a connection whose handshake was abandoned, disarmed once the peer
+/// has an answer.
+///
+/// Dropping a [`Connection`] does not close QUIC: the endpoint keeps it, its
+/// routes, and its driver task until the driver sees a terminal state, and the
+/// backlog stopped counting it at accept. The peer picks the path it asks for
+/// and the subprotocols it offers, so it decides which of [`Request::respond`]'s
+/// early returns the server takes; a guard covers every way out rather than
+/// each one remembering.
+struct Guard {
+	conn: Option<Connection>,
+}
+
+impl Guard {
+	fn new(conn: Connection) -> Self {
+		Self { conn: Some(conn) }
+	}
+
+	/// The peer got an answer, so the connection is the answer's to close.
+	fn disarm(&mut self) {
+		self.conn = None;
+	}
+}
+
+impl Drop for Guard {
+	fn drop(&mut self) {
+		if let Some(conn) = self.conn.take() {
+			conn.shared()
+				.close_code(H3_GENERAL_PROTOCOL_ERROR, "webtransport handshake abandoned");
+		}
+	}
 }
 
 impl Request {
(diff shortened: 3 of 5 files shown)


kixelated:
Review round addressed. Three of the four findings are fixed here, each with a regression test verified to fail without it:

- **Arrival cap counted before classifying** — the adopt loop drained the whole queue before any classification, so a peer whose control stream was already first got refused for what it pipelined behind it. It now stops adopting at the cap, classifies what has arrived, and only then calls it. `a_pipelining_peer_is_not_refused_by_the_arrival_cap`.
- **`reject` disarmed its guard too early** — a `reject` cancelled during the acknowledgement wait skipped both the wait and the deliberate close, leaking the connection. The guard now stays armed across the wait. `an_abandoned_rejection_still_closes_the_connection`.
- **The stream that tips the cap over was dropped** (found in the re-review) — it was consumed and not retained, so a handshake that then succeeded silently cancelled one legitimate pipelined stream. It is now kept and rides into the session. Covered by making that stream a real WebTransport stream in the test above.

The fourth, `finish()` returning `Ok` with the header still owed, is **not fully closed** and is filed as #3129. `Drop` now pays that debt when credit has returned, which covers the realistic paths (moq-net always pairs `finish` with `poll_closed`), but an immediate finish-then-drop has no moment for credit to return and still resets. That is not a regression — the previous `Err` path also ended in a reset, with an unmapped code — and it is now documented on `SendStream`. The real fix is writing the header at open time, which needs a queue of half-open streams in `Web::state` because `Session` clones share one `Rc<Web>`; that is design work rather than a merge-time patch, hence the issue.

Verified on real io_uring (kernel 6.19): `cargo test -p moq-uring` green, clippy and rustdoc clean under `-D warnings`. Rebased onto `dev` after #3116, which also picked up its `Identity::open` rename.

(Written by Claude Opus 5)

PR #3118, opened 2026-08-28
fix(relay)!: make Config embeddable by deriving Args, not Cli

kixelated:
## Problem

`Relay::load` exists so an embedder gets the whole startup sequence by calling one function instead of copying `main.rs`. The flag surface is half of that contract: an embedder flattens `Config` into its own CLI rather than re-declaring every relay flag and drifting on each one added here.

The usage migration broke it. `Config` derives `usage::Cli`, `#[usage(flatten)]` requires `CommandArgs`, and only `usage::Args` emits that trait — so the config a `Relay` is documented to take can no longer go inside anyone else's command line:

```
error[E0277]: the trait bound `moq_relay::Config: usage_rs::spec::CommandArgs` is not satisfied
    pub relay: moq_relay::Config,
               ^^^^^^^^^^^^^^^^^ the trait `CommandArgs` is not implemented
```

Found by [moq.pro](https://github.com/moq-dev/moq.pro)'s `edge`, which flattens `moq_relay::Config` wholesale and does not compile against this branch. Deriving both `Cli` and `Args` doesn't work either — usage rejects `completion` off the root, correctly.

## Fix

Split the two roles rather than shim around it:

- **`Config` derives `Args`** and carries only the relay's own flags, so it composes.
- **A new `Cli`** carries the program-level parts that cannot be flattened anyway (name, version, the completion command) and wraps `Config`.

`Config::load()` keeps its signature and goes through `Cli` internally, so `moq-relay`'s `main` and every embedder calling `Relay::load(Config::load()?)` are unchanged.

## Breaking

`Config::parse_from`, `Config::spec`, and `Config::command` are program-level and moved to `Cli`. In-tree callers (the config tests and `tests/released_cli.rs`) are updated.

## Testing

`just check` and `just test` both pass. Added `config::tests::the_config_can_be_embedded`, which declares a stand-in embedder CLI, flattens `Config` into it, and parses one argv carrying both surfaces — the property this PR restores, so a future change back to `Cli` fails a test rather than a downstream repo.

(written by claude-opus-5)

### DIFF
--- a/rs/moq-relay/src/config.rs
+++ b/rs/moq-relay/src/config.rs
@@ -4,11 +4,19 @@ use crate::{AuthConfig, CacheConfig, ClusterConfig, InternalConfig, StatsConfig,
 
 /// Top-level relay configuration, loadable from CLI arguments, environment
 /// variables, or a TOML file.
-#[derive(usage::Cli, Clone, Debug, Deserialize, Serialize)]
+/// Top-level relay configuration, as a COMPOSABLE args group.
+///
+/// `usage::Args` rather than `usage::Cli` on purpose: the program-level parts (a
+/// name, a version, the completion command) belong to whichever binary owns the
+/// process, and a `Cli` cannot be `#[usage(flatten)]`ed into another one. Keeping
+/// them off this type is what lets an embedder put the relay's whole flag surface
+/// inside its own CLI -- moq.pro's `edge` does exactly that -- instead of
+/// re-declaring it and drifting on every flag added here. This binary wraps it in
+/// a private `Cli` that adds those program-level parts; [`spec`] exposes the
+/// resulting command line.
+#[derive(usage::Args, Clone, Debug, Deserialize, Serialize)]
 #[usage(unknown_flags = "error", args_override_self = false)]
 #[serde(deny_unknown_fields, default)]
-#[usage(name = "moq-relay", version = env!("VERSION"))]
-#[usage(completion)]
 #[non_exhaustive]
 pub struct Config {
 	/// The QUIC/TLS configuration for the server.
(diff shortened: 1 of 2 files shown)


kixelated:
Rebased onto current `dev` (one incidental conflict in `doc/bin/relay/config.md`: this branch only carried a markdown-lint escape on a paragraph `dev` has since rewritten, so I took `dev`'s version).

Also made `Cli` **private**. It is an implementation detail of this binary — `main` goes through `Config::load()`, and an embedder declares its own `Cli` and flattens `Config` — so nothing outside the crate needs to name it. The one thing that did was `tests/released_cli.rs`, which enumerates the program's flags and env vars; that now goes through a new free function:

```rust
pub fn spec() -> &'static usage::spec::Spec<'static>
```

Deliberately a free function rather than `Config::spec()`: the spec belongs to the PROGRAM, and `Config` is a composable fragment, so a method there would hand an embedder moq-relay's spec while reading as though it were their own.

Net public API change is now one function instead of one struct, and `Config::parse_from` / `Config::spec` / `Config::command` remain the only removals.

`just check` and `just test` both pass (264 tests).

(written by claude-opus-5)

kixelated:
Addressed the review finding: the private `Cli`'s doc comment was leaking into the CLI.

Usage renders a `Cli`'s doc comment as the program's `--help` description, so splitting `Config` out put a private wrapper in that position and printed its implementation notes — unresolved `[\`Config\`]` links and all — to users:

```
moq-relay 0.14.13
`moq-relay`'s own command line: [`Config`] plus the program-level parts.

Exists so [`Config`] can stay flattenable (see its docs). Private, because it is an
implementation detail of THIS binary: an embedder declares its own and flattens...
```

`Cli` now carries the user-facing description restored verbatim from `dev` ("Top-level relay configuration, loadable from CLI arguments, environment variables, or a TOML file."), and the rationale moved into an ordinary `//` comment that Usage does not read.

Added `the_help_description_is_written_for_users`, which pins that description and asserts no rustdoc-link syntax or internal terms reach it — so the next edit there fails a test instead of the help page.

Verified beyond the test: `--help` matches the `dev` baseline, `--version`/`-V` unchanged, and `__complete_word__` output is identical to a `dev` baseline build I checked out to compare.

`just fix`, `just check`, `just test` all pass (265 tests).

The earlier P1 about documenting the exported `config` field is moot — that was against the previous head, when `Cli` was public; it and its field are private now, so the only new public item is the documented `spec()`.

(written by claude-opus-5)

PR #3152, opened 2026-08-28
feat(tokio): reject an unsupported feature build with one honest error

### LINKED ISSUE #2979: moq-tokio does not compile with --no-default-features, and workspace feature unification hides it
`cargo check -p moq-tokio --no-default-features` fails on `dev`. Found while reviewing #2977; the breakage predates #2921.

## Errors

At `ca2fd504f` (before #2921 merged), 4 errors:

- 2x `E0004: non-exhaustive patterns: type &RequestKind is non-empty` (all variants cfg'd out by the missing backend features, so the remaining matches don't compile)
- 2x `E0282: type annotations needed`

`#2921` added a 5th: `E0433: cannot find util in crate` from `worker.rs`, which calls `crate::util::resolve` while `mod util` is gated on `any(noq, quinn, quiche)`. `worker::Workers::bind` now also references the equally-gated `crate::server::DEFAULT_BIND` (#2977), deepening the same dependency rather than introducing it.

## Why CI never sees it

Nightly `just rs features` runs `--no-default-features` across the workspace, and cargo unifies features per crate: `moq-relay` (and everything else) depends on `moq-tokio` with a backend enabled, so `moq-tokio` never actually compiles with zero features in a workspace build. Only a `-p moq-tokio --no-default-features` invocation, or an external consumer with `default-features = false` and no backend feature, hits it.

## Options

- Decide the backend-less configuration is unsupported and enforce it: `compile_error!` when no QUIC backend feature is enabled (a tcp/uds-only build would need its own thought), which turns 5 confusing errors into 1 honest one.
- Or make it compile: ungate `util`/`DEFAULT_BIND` (both are dependency-free), gate `worker`/`steer` on a backend, and fix the pre-existing `RequestKind` matches.

The first is less work and matches reality: `worker`, `steer`, and `server` all assume a backend exists.

(written by Fable 5)

kixelated:
Closes #2979.

## The reported bug

`cargo check -p moq-tokio --no-default-features` fails with **8 errors**, none naming the cause:

```
error: an enum with no variants accepts no value at all
error[E0433]: cannot find `util` in `crate`
error[E0425]: cannot find value `DEFAULT_BIND` in module `crate::server`
error[E0282]: type annotations needed                        (x2)
error[E0277]: the trait bound `QuicBackend: ValueEnum` is not satisfied
error[E0004]: non-exhaustive patterns: type `&RequestKind` is non-empty   (x2)
```

`QuicBackend` becomes an empty enum, `RequestKind` an empty match, and `worker`/`server` reference items their own gates removed. The count was 5 when the issue was filed and grows as the crate does, which is itself the argument against making this configuration work.

The issue offered two options; this takes the first. `worker`, `steer`, and `server` all assume a backend exists, so a `compile_error!` matches reality and names the features to enable.

## A second unsupported shape, not in the issue

Auditing the feature space rather than just the reported case turned up another one. `quinn` and `noq` take their crypto provider from `aws-lc-rs`/`ring`, since their rustls dependency is built with default features off (the manifest already says so: *"a backend without aws-lc-rs/ring has no initial cipher"*).

| features | before | after |
|---|---|---|
| *(none)* | 8 errors, cause unnamed | names the backend requirement first |
| `tcp` | 5 errors, cause unnamed | names the backend requirement first |
| `noq` | 3 unrelated `E0599`s about `EndpointConfig::default` / `ServerConfig::with_crypto` | names the provider requirement |
| `quinn` | clean | clean (unchanged, see below) |
| `quiche` | clean | clean (boringssl brings its own) |
| `quinn,aws-lc-rs`, `noq,aws-lc-rs` | clean | clean |
| default | clean | clean |

`noq` is the only backend that gets this guard, and the reason is that it is not a runtime question there: without a provider feature `web-transport-noq` is built without its own and the constructors this crate calls stop existing, so the build fails whatever the application does later.

**An earlier revision of this PR also guarded `quinn`, and that was wrong.** `crypto::provider()` resolves the provider at runtime and checks `CryptoProvider::get_default()` first, so an application that calls `install_default()` works with `quinn` and no provider feature. The function's own panic names that as the equal alternative:

```
"no CryptoProvider available; install_default() or enable either the aws-lc-rs or ring feature"
```

Refusing the build made a supported configuration unreachable. Caught by the Codex review on this PR (P2), verified against `crypto.rs`, and reverted in 530014e; `--features quinn` compiles clean again.

## Why CI never saw any of this, and still passes

Cargo unifies features per crate, so in a workspace build a member that depends on moq-tokio with a backend enables one for everyone. Nightly's `cargo check --locked --workspace --no-default-features` therefore never compiles moq-tokio backend-less, which is why it stayed green over a broken configuration. I ran that exact command with these guards in place and it still passes unchanged.

Only `-p moq-tokio` or an external consumer with `default-features = false` selects the broken shapes, so `just rs _unsupported-features` (added to the nightly `features` recipe) asserts each still fails on its own message rather than on whatever breaks first inside the modules. Verified it catches a regression: neutering either guard makes the recipe fail with `moq-tokio --features quinn should fail on: require a crypto provider`.

This is also why adding `--all-features`/`--no-default-features` to `just check` would not have fixed the issue: two extra full workspace compiles per PR, and feature unification means neither would compile moq-tokio backend-less anyway.

**Note:** `compile_error!` does not halt compilation, so the downstream errors still follow it. The change is that the first error names the cause and the fix.

## Drive-by: a doc link that breaks any moq-tokio PR

Separate first commit, kept apart so it can be reverted on its own.

`cargo doc -p moq-tokio` with default features already fails on `dev`:

```
error: unresolved link to `mdns`
  --> rs/moq-tokio/src/lib.rs:12:20
```

`mdns` is behind a non-default feature, so rustdoc has no item to resolve, and `RUSTDOCFLAGS="-D warnings"` makes it a hard error. `just check`'s doc pass is diff-scoped, so this fires on any branch whose diff selects moq-tokio, including this one. Nightly missed it because it documents with `--all-features`, which compiles the module in.

Confirmed pre-existing by stashing this branch's changes and reproducing on clean `dev`. Same class as the `target/doc/moq` collision in #3113: a doc-pass failure that only fires on certain package selections.

## Base

Targets `dev` because `moq-tokio` only exists there. No published API changes shape: the guards reject builds that already did not work, and the doc fix is prose.

(Written by Claude Opus 5)

### DIFF
--- a/doc/lib/rs/crate/moq-tokio.md
+++ b/doc/lib/rs/crate/moq-tokio.md
@@ -27,7 +27,7 @@ Tokio-based connection helpers for native Rust applications. Provides TLS config
 moq-tokio = "0.19"
 ```
 
-The default features cover a working build. If you set `default-features = false`, pick a QUIC backend (`quinn`, `quiche`, or `noq`) plus, for `quinn` or `noq`, a crypto provider (`aws-lc-rs` or `ring`); `quiche` brings its own. A build missing either fails with a message naming what to enable.
+The default features cover a working build. If you set `default-features = false`, pick a QUIC backend (`quinn`, `quiche`, or `noq`); a build without one fails with a message naming them. `noq` also needs a crypto provider (`aws-lc-rs` or `ring`) to compile. `quinn` takes one from either feature or from a provider you install yourself with `CryptoProvider::install_default()`, and `quiche` brings its own.
 
 ## API Reference
 
--- a/doc/lib/rs/crate/moq-tokio.md
+++ b/doc/lib/rs/crate/moq-tokio.md
@@ -27,6 +27,8 @@ Tokio-based connection helpers for native Rust applications. Provides TLS config
 moq-tokio = "0.19"
 ```
 
+The default features cover a working build. If you set `default-features = false`, pick a QUIC backend (`quinn`, `quiche`, or `noq`) plus, for `quinn` or `noq`, a crypto provider (`aws-lc-rs` or `ring`); `quiche` brings its own. A build missing either fails with a message naming what to enable.
+
 ## API Reference
 
 Full API documentation: [docs.rs/moq-tokio](https://docs.rs/moq-tokio)
--- a/rs/justfile
+++ b/rs/justfile
@@ -389,6 +389,9 @@ features:
 # everyone. Only a per-crate invocation selects the shapes an external consumer with
 # `default-features = false` can land on, so assert those still fail on their own
 # `compile_error!` rather than on whatever breaks first inside the modules.
+#
+# `--features quinn` is deliberately absent: it compiles without a provider feature and
+# takes one from `CryptoProvider::install_default()` at runtime, which is supported.
 _unsupported-features:
     #!/usr/bin/env bash
     set -euo pipefail
@@ -406,8 +409,7 @@ _unsupported-features:
 
     expect "requires a QUIC backend"
     expect "requires a QUIC backend" --features tcp
-    expect "require a crypto provider" --features quinn
-    expect "require a crypto provider" --features noq
+    expect "requires a crypto provider" --features noq
 
 # nextest, not `cargo test`, so a wedged test is killed instead of pinning a core
 # until someone notices (`.config/nextest.toml` sets the timeout). The trade is
--- a/rs/justfile
+++ b/rs/justfile
@@ -382,6 +382,32 @@ features:
     cargo clippy --locked --workspace --all-targets --all-features -- -D warnings
     RUSTDOCFLAGS="-D warnings" cargo doc --locked --workspace --all-features --no-deps
     cargo nextest run --locked --workspace --all-targets --all-features
+    just rs _unsupported-features
+
+# The workspace compiles above can't reach a backend-less moq-tokio: cargo unifies
+# features per crate, so a member depending on it with a backend enables one for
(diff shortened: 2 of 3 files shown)


kixelated:
Superseded by #3150, which took the other option from #2979 and landed while this was open.

#3150 made the backend-less build **compile** (gating `server`, `worker`, and their re-exports) rather than declaring it unsupported, and added `just rs tokio-features` to run `cargo check -p moq-tokio --no-default-features` as a *passing* check whenever a diff selects moq-tokio. Verified on current `dev`: that build is now clean, 0 errors.

So this PR is not merely redundant, it is incompatible. Its `compile_error!` would make #3150's own check fail on the first run. #3149 compounds that by making moq-cli serve TCP and Unix listeners without QUIC, so a tcp/uds-only build is a supported configuration now and "moq-tokio requires a QUIC backend" is no longer true. The `mdns` doc-link fix here duplicates the identical one in #3150.

**One finding does survive**, filed as #3160: `--features noq` on its own still fails with three unrelated `E0599`s, because `web-transport-noq` is built without a crypto provider and the constructors this crate calls stop existing. #3150's matrix checks `--no-default-features` and `--all-features`, and neither selects that shape, so it is still uncovered on `dev` today.

Closing rather than rebasing: the useful remainder is one guard, and it belongs on top of #3150's structure rather than inside a change built on the opposite premise.

(written by Claude Opus 5)

PR #3158, opened 2026-08-28
feat(net)!: resolve the subscribe start from max age, with Group Start as an absolute floor

kixelated:
## Summary

A subscriber with a 3s jitter buffer used to start empty and spend 3s filling it: an absent `Group Start` meant "the latest group", so a join always began at the live edge no matter how much older content the subscriber could still play. Now `Subscriber Max Age` is the only thing that decides where delivery begins, and a join is handed the head of what it can still use.

- **`Group Start` is an absolute floor on lite-06.** The wire carries the raw minimum group sequence (default 0), replacing the `sequence + 1` encoding where 0 meant the latest group. A floor is not a request: it only bounds how far back the budget may reach, so migrations and range requests still pin exactly where they need to without dragging backlog in.
- **The budget is the only gate.** The model does not resolve a start by scanning the cache; the subscriber's cursor is floored at the named group (or 0) and delivery's existing staleness rule decides what above the floor is worth sending. A zero budget (the default) delivers only the latest group, so a subscription that says nothing behaves exactly as before. One bound decides both what is sent and what is expired, so the publisher never sends history the subscriber would discard on arrival.
- **A late lower group within the budget is delivered.** Arriving below everything already delivered is not what makes content stale. This is what the resume layer's pruned-cursor semantics always required (a straggler from a dying route still drains), and a consumer that needs sequence order reorders or drops it itself.
- **Pre-06 wires keep their published semantics.** Lite-01..05 drafts define an absent start as the latest group (and 03..05's Max Age as a staleness tolerance only), so the publisher pins those to the live edge; only lite-06 lets the budget reach back. Re-encoding toward a pre-06 wire folds a vacuous floor back to absent, so a relay downgrade can never turn "no floor" into "replay from the beginning".
- **A vacuous floor canonicalizes to absent.** On the 06 wire, `Group Start = 0` with `Frame Start = 0` decodes as no floor; group 0 stays nameable when a `Frame Start` qualifies it, so a mid-group-0 resume survives (a catalog never leaves group 0). Aggregation folds floors accordingly: the loosest floor wins, and any subscriber without one clears it.

## Bug fixes surfaced by the review

- **The spliced (origin-backed) serving path never honored a floor.** `resume::poll_activate` *assigned* each segment's inner cursor from the segment boundary, silently discarding the start the inner subscription carried, so an explicit `Group Start` evaporated on any track served through an origin. It now raises the floor instead (`track::Subscriber::raise_start_to`). This was the actual mechanism behind the `broadcast_route_migration` CI failure; that test now names its floor explicitly, which is what its 10s tolerance was standing in for.
- **The hang container consumer truncated a below-cursor group still downloading.** The decode loop drains faster than the network delivers, so the moment an admitted backlog group's buffer ran empty, `next()` shifted it out and its remaining frames were silently lost (with `continuous` still reporting true). Removal now waits for the group to finish, with the latency skip still bounding a stalled head. Regression-tested with an incremental mid-download arrival.

## Out-of-order delivery

Groups go out newest-first, so the head of a join arrives after the live edge served alongside it. `Container.Consumer` dropped anything below its delivery cursor, which threw that head away one layer down; it now admits it, since the subscription's own max age already bounds how far back one can be, and the rendering consumers place content by timestamp. A group the reset boundary proves reneged is still dropped. Consumers that need strict decode order are #3099's `Ordered` cursor's job, not this consumer's.

## Public API changes

`Subscription::start` / `startGroup` semantics change from "start exactly here" to a floor (which is why this targets `dev`), and aggregation folds it as one (loosest wins, absent clears). No items added or removed; `moq-ffi`/`libmoq` doc comments updated to match (no signature change, so the generated bindings are untouched).

## Test plan

`cargo nextest run -p moq-net -p moq-tokio -p moq-relay -p moq-mux -p hang -p moq-gst` (2429 tests), `cargo clippy` over the changed crates plus `moq-ffi`/`libmoq`/`moq-cli`, `cargo fmt --check`, `bun test js/{net,hang,watch}/src` (875 tests), `bunx tsc -b`, `biome check`, and `just drafts check`.

New coverage: the budget reaching back over the cache, a named start as a floor (cut off, vacuous, and above-the-edge cases), a late lower group within the budget delivering, the pre-06 pin per version (unbounded lite-01 budget, declared lite-05 tolerance, lite-06 resolution), the raw-vs-`+1` wire encodings with the vacuous-floor canonicalization and pre-06 fold in both languages, `Frame Start` qualifying group 0, floor aggregation clearing on an unfloored subscriber, and the container consumer's mid-download truncation regression.

## Follow-ups, deliberately not here

- **`rs/moq-mux`'s container consumer still drops below-cursor groups**, so a native consumer subscribing with a budget over lite-06 transfers the join backlog and discards it. The right shape for consumers that need decode order is #3099's `Ordered` cursor rather than porting the admission rule, so this waits for that to land.
- The publisher-side stale metric counts a warm-cache join's skipped backlog the same as a mid-stream fall-behind. Distinguishing them needs an arrival stamp on the counting path; the metric is unchanged from `dev` today.
- In buffered mode `subscribeMedia`'s `maxAge` (the buffer ceiling) still decides the join depth; whether a buffered viewer should join `maxBuffer` behind a real-time publisher is a pre-existing choice with its own trade-off.
- #3099 also rewrites lite-06's SUBSCRIBE layout (dropping the ordered byte); whichever lands second rebases the wire tests.

Supersedes #3114, which asked for `startGroup: 0` from the client with a bound nothing enforced.

(Written by Claude Fable 5)

### DIFF
--- a/doc/concept/layer/moq-lite.md
+++ b/doc/concept/layer/moq-lite.md
@@ -88,11 +88,13 @@ If the peer doesn't have the broadcast/track, they will get an error.
 Otherwise, the subscription is active and will stay open until closed by the publisher (possibly with an error).
 
 A track is broken into **groups**, each with an increasing ID.
-Conceptually, these are join points, and new subscriptions will always start at the latest group.
+Conceptually, these are join points, and new subscriptions start at the newest group they can still use.
 Groups are delivered independently and potentially out of order, so you should have some logic to reorder or skip during congestion.
 A group is closed when finished or aborted with an error (ex. during congestion).
 
-A subscription starts at the latest group by default, but can name any (group, frame) position instead.
+A subscription lets the publisher choose the start by default, but can name any (group, frame) position instead.
+Left to the publisher, it resolves to the oldest group within **Subscriber Max Age**, which is the latest group at the default budget of zero.
+A subscriber that buffers is then handed the head of what it can still play instead of only the live edge, and never history it would skip on arrival: one bound decides both.
 That is how a subscriber follows a track to a different publisher partway through a group, which matters when the current group may stay open indefinitely (a JSON append log, or a catalog that keeps appending deltas).
 A group can therefore be assembled from more than one publisher: each contributes a disjoint run of frames, and the subscriber concatenates them in index order.
 
@@ -127,6 +129,7 @@ The media timeline keeps a backlog delivered as a burst old, while wall-clock ti
 Both ends apply it: the publisher skips a group rather than sending it, and the subscriber skips it again as it reads.
 The publisher only ever sees the most tolerant budget across its subscribers, which is why the subscriber applies it too.
 Asking for old groups with `Group Start` does not exempt them: a start bounds what you are sent, and a subscriber that wants history has to raise its budget to match.
+That is also why the budget is what resolves an unset start: the groups worth sending are exactly the ones it would not immediately skip.
 
 The publisher also keeps old groups around for a best-effort **Publisher Max Age** cache window so relays and late subscribers can still fetch them. This defaults to 5 seconds.
 The subscriber's max age is bounded by this window: a group can't be waited for longer than it's actually kept around.
--- a/doc/concept/layer/moq-lite.md
+++ b/doc/concept/layer/moq-lite.md
@@ -92,8 +92,8 @@ Conceptually, these are join points, and new subscriptions start at the newest g
 Groups are delivered independently and potentially out of order, so you should have some logic to reorder or skip during congestion.
 A group is closed when finished or aborted with an error (ex. during congestion).
 
-A subscription lets the publisher choose the start by default, but can name any (group, frame) position instead.
-Left to the publisher, it resolves to the oldest group within **Subscriber Max Age**, which is the latest group at the default budget of zero.
+A subscription can name a (group, frame) floor, but the floor is not a request: **Subscriber Max Age** is the only thing that asks for data.
+Delivery starts at the oldest group at or above the floor within that budget, which is the latest group at the default budget of zero.
 A subscriber that buffers is then handed the head of what it can still play instead of only the live edge, and never history it would skip on arrival: one bound decides both.
 That is how a subscriber follows a track to a different publisher partway through a group, which matters when the current group may stay open indefinitely (a JSON append log, or a catalog that keeps appending deltas).
 A group can therefore be assembled from more than one publisher: each contributes a disjoint run of frames, and the subscriber concatenates them in index order.
@@ -128,8 +128,8 @@ That age is measured both on the media timeline (a group's first timestamp again
 The media timeline keeps a backlog delivered as a burst old, while wall-clock time backstops stalled or empty groups that have no timestamp.
 Both ends apply it: the publisher skips a group rather than sending it, and the subscriber skips it again as it reads.
 The publisher only ever sees the most tolerant budget across its subscribers, which is why the subscriber applies it too.
-Asking for old groups with `Group Start` does not exempt them: a start bounds what you are sent, and a subscriber that wants history has to raise its budget to match.
-That is also why the budget is what resolves an unset start: the groups worth sending are exactly the ones it would not immediately skip.
+Asking for old groups with `Group Start` does not exempt them: a floor only bounds what you are sent, and a subscriber that wants history has to raise its budget to match.
+That is also why the budget is what decides where delivery starts: the groups worth sending are exactly the ones it would not immediately skip.
 
 The publisher also keeps old groups around for a best-effort **Publisher Max Age** cache window so relays and late subscribers can still fetch them. This defaults to 5 seconds.
 The subscriber's max age is bounded by this window: a group can't be waited for longer than it's actually kept around.
--- a/drafts/draft-lcurley-moq-lite.md
+++ b/drafts/draft-lcurley-moq-lite.md
@@ -185,7 +185,7 @@ Only the subscriber knows whether a partial Group is any use to it, so only the 
 
 Frame indices are only meaningful relative to a Group the subscriber has already begun receiving, so:
 
-- A `Frame Start` is meaningless without an absolute `Group Start`; a subscriber requesting the latest group (`Group Start` = 0) MUST send `Frame Start` = 0.
+- A `Frame Start` is meaningless without an absolute `Group Start`; a subscriber letting the publisher choose the start (`Group Start` = 0) MUST send `Frame Start` = 0.
 - A `Frame End` is meaningless without a `Group End`; an unbounded subscription (`Group End` = 0) MUST send `Frame End` = 0.
 
 A publisher MUST treat a violation of either rule as a protocol violation and reset the stream.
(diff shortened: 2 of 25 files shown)


kixelated:
Reviewed the behavior change end to end (model resolution, both wire publishers, the js container consumer, the draft text) and reproduced the failure modes locally. The direction is right: deriving the join point from the budget that already expires groups is the correct core idea, and it is strictly better than IETF `DELIVERY_TIMEOUT` (per-object, wall-clock, and a publisher-side drop that leaves the rest of the group undecodable). But as landed this breaks CI, has two demonstrable bugs in the js container consumer, leaves every Rust consumer downloading backlog it then throws away, and changes wire behavior for already-published draft versions. Findings in severity order.

## 1. CI is red, and the failing test is a real second constituency, not a stale assertion

`broadcast_route_migration` (rs/moq-tokio/tests/broadcast.rs:847) fails: expected `["a1"]`, got `["a0"]`. That test subscribes with `max_age = 10s` used purely as a *staleness tolerance*, "a generous stale window so the continuation B already holds is served rather than aged out", while still expecting to join at the live edge. This PR fuses those two meanings, so the tolerance now also means "replay 10s of backlog on join".

It is not fixable by updating the test, because the new wire cannot express the old intent at all: `Group Start = 0` no longer means "latest", a non-zero value is absolute, and there is no other encoding for "join live, but tolerate old groups mid-stream". Migration/splice consumers, `moq-gst` (`source/imp.rs` subscribes with a 1s budget as pure tolerance), and any third party using `maxAge` per the lite-03..05 drafts lose expressiveness silently.

Your own "Follow-ups" section documents the same conflation from the other side: `subscribeMedia`'s `maxAge` is now load-bearing twice, and the ceiling/floor divergence in buffered mode has no right answer. That is three independent signs the knob is overloaded.

**Recommendation: split the knob.** Add a separate join bound (e.g. `Subscriber Backfill`, a duration, default 0, resolved as `min(backfill, max_age, publisher window)`) instead of overloading `Subscriber Max Age`. Lite-06 is WIP so the SUBSCRIBE field is free; the "two bounds cannot disagree" property survives via the clamp; every existing caller (migration, gst, ffi consumers) keeps today's semantics by default; and hang sends `backfill = buffer floor`, `maxAge = ceiling`, which dissolves the buffered-mode follow-up exactly.

## 2. The js container consumer truncates a below-cursor group that is still downloading

`next()`'s removal branch (js/hang/src/container/consumer.ts, the `#active > seq || done` shift) fires the moment a below-`#active` group's buffered frames are drained, even while its `#runGroup` is still reading. Before this PR that branch was unreachable for below-cursor groups (they were dropped at admission); now it is the common case, because the decode loop consumes frames faster than the network delivers them.

Repro (local test): live group 3 delivering, backlog group 1 arrives and delivers its first frame; the very next `next()` returns `{frame: undefined, group: 1}` and the group's second frame is silently lost. No `#gap` is set, so the following result still reports `continuous: true`, and the buffered-range accounting bridges the hole.

## 3. The js container consumer splits the open live group around the backlog, which video cannot decode

`next()` serves `#groups[0]` whenever its sequence is at or below `#active`, so when backlog lands while the open live-edge group is mid-delivery, delivery order becomes: live head frames, backlog groups, then *the remainder of the live group mid-GOP*. Repro delivery order: `3:f0, 3:f1, 2:f0, 2:f1, 3:f2`. `3:f2` is a delta chunk whose references are gone from the decoder's state, fed with `continuous: true` and no discontinuity bump. WebCodecs will corrupt or error, and the video decoder's error callback does `effect.close()`, killing the track. Audio survives only because hang audio groups are single-frame. The new out-of-order test uses closed single-frame groups, so it can't see either this or #2.

**Recommendation for both:** rather than admitting out-of-order groups into a decoder-bound path and patching the cursor rules, anchor the consumer's start deterministically. The publisher already announces the resolved start as SUBSCRIBE_START (the js subscriber currently drains it as "informational" in `#drainResponses`); initialize `#active` from it and deliver strictly in sequence order from there, with the existing latency skip as the escape hatch. That removes the "first arrival wins" race the old TTV comment already flagged as racey, makes the deleted "Consumer rejects stale groups" rule unnecessary rather than wrong, and TTV hides behind the jitter buffer (the first frame *rendered* is the head of the window; the poster-frame path covers perceived TTV).

## 4. Every Rust consumer downloads the backlog and throws it away

`rs/moq-mux/src/container/consumer.rs` still has exactly the rule this PR removed from js: `drop = sequence < self.current`, "skipping old group". That consumer is behind moq-gst, moq-ffi/libmoq (so every language binding), moq-rtmp, moq-srt, moq-cli, and moq-audio. Any of them subscribing with a nonzero budget against an upgraded publisher now transfers up to `maxAge x bitrate` on every join and drops it on arrival. The PR body's "Both consumers can place one" covers only the two js/watch consumers; the cross-package sync row for the container was skipped without being called out.

## 5. The new semantics apply to published draft versions

`position_cursor` / `positionCursor` gate on `carries_max_age`, i.e. Lite03+. But lite-03..05 are published drafts that say "A value of 0 means the latest group", and the changelog entry for this change sits under moq-lite-06. Deployed clients already send a nonzero `maxAge` (js watch sends its buffer size), so an upgraded relay changes their join behavior out from under them, and their released container consumers are precisely the download-and-drop ones from #4. Gate the resolution to Lite06Wip+ and pin 03..05 to the latest group alongside the legacy pin.

## 6. Smaller notes

- The resolved-start predicate is "oldest group whose *first frame* is within budget" (`fresh_start` reuses `is_stale`, which judges an unread group by `group.timestamp()`). That is subtly stricter than "any group unless it provably can't contain a frame that recent": a long open group with an old head but fresh tail is excluded. I think first-frame is the right rule (you can't decode from mid-group, and it matches expiry), but the draft says "whose age is within `Subscriber Max Age`" without defining a group's age; pin it to the first frame explicitly.
- PR body cites a test `an_absent_start_resolves_from_the_budget`; the actual test is `a_legacy_wire_is_pinned_to_the_live_edge`.
- The js test "a start is clamped to the publisher's window" subscribes before any group exists, so it exercises the delivery-side clamp, not `#resolveStart`'s. Populate the track first, like the Rust twin does.
- The moq-tokio failure also means the test plan's scope (`cargo test -p moq-net`) missed a dependent crate; `just test` would have caught it.

## On the IETF question

Agreed it should be an extension, and combining it with the timestamp draft is right: "age" only exists once objects carry timestamps at a declared timescale, so the dependency is structural, not incidental. Shape it as SUBSCRIBE parameters in draft-lcurley-moq-timestamp (`Subscriber Max Age`, plus `Backfill` if the knob splits), with the resolution rule as a SHOULD on the publisher when the filter is the live-edge one. The DELIVERY_TIMEOUT comparison in this PR's motivation is the argument for the extension's introduction almost verbatim: group-granular, media-timeline units, and both ends applying the same bound, versus per-object wall-clock dropping that strands the rest of the group.

(Written by Claude Fable 5)

kixelated:
Reworked per direction (single `maxAge` knob; `Group Start` absolute again; ordering deferred to `.ordered()` / #3099). Pushed as one commit on top: the cache-scanning start resolution is gone entirely. The floor is the only thing a named start contributes, the budget is the only gate, and delivery's existing staleness rule does the rest, so the model shrank rather than grew.

Two genuine bugs fell out of chasing the CI failure, both fixed here:

- `resume::poll_activate` **assigned** each segment's inner cursor from the segment boundary, discarding any floor the inner subscription carried, so an explicit `Group Start` never survived an origin-backed serving path. It raises the floor now. That, not the test's expectations, was the real mechanism behind `broadcast_route_migration` failing; the test now names its floor (group 1) with the 10s tolerance kept as pure tolerance.
- The hang container consumer truncated a below-cursor group whose buffer momentarily drained mid-download (the decode loop outruns the network), silently losing its tail while still reporting `continuous`. Removal now waits for `done`; regression test included.

Wire details worth a look: on lite-06 a vacuous floor (`Group Start` 0, `Frame Start` 0) canonicalizes to absent on decode, group 0 stays nameable when a `Frame Start` qualifies it (mid-group-0 catalog resume), and re-encoding toward a pre-06 wire folds the vacuous floor back to absent so a relay downgrade can't turn "no floor" into "replay from group 0". Pre-06 wires stay pinned to the latest group per their published drafts; only lite-06 resolves from the budget.

Left out on purpose: `rs/moq-mux`'s container consumer still drops below-cursor groups, so native consumers with a budget will download-and-discard the join backlog on lite-06 until they sit on #3099's `Ordered` cursor. Noted in the PR body as a follow-up. #3099 and this PR both rewrite lite-06's SUBSCRIBE, so whichever lands second rebases the wire tests.

(Written by Claude Fable 5)

PR #3215, opened 2026-08-30
feat(dart): add native Flutter bindings

### LINKED ISSUE #3100: Dart/Flutter bindings via UniFFI
`rs/moq-ffi` is already the single UniFFI core behind five wrappers (`py/`, `swift/`, `kt/`, `go/`, `rs/libmoq`). Dart/Flutter would be a sixth of the same shape. It is viable, but it carries the same cost Go does: the Dart frontend is a third-party bindgen that lags uniffi-rs, so we would own a pinned fork.

## Landscape (Aug 2026)

Two generators exist, neither maintained by Mozilla:

| | [Uniffi-Dart/uniffi-dart](https://github.com/Uniffi-Dart/uniffi-dart) (ex-acterglobal) | [nchapman/uniffi-bindgen-dart](https://github.com/nchapman/uniffi-bindgen-dart) |
|---|---|---|
| Activity | commits 2026-08-22, 43 stars, 25 open issues | last push 2026-03-04, 3 stars, 199 downloads |
| uniffi target | 0.31.2 (`v0.2.1+v0.31.2`) | 0.31.0 |
| CLI | none, `src/bin.rs` is a placeholder stub | real CLI with library mode + `doctor` |
| Library loading | `@Native(assetId:)`, so Dart native assets | `DynamicLibrary.open` + `configureDefaultBindings()` |
| Coverage | 30 fixtures: proc-macros, async, objects, records, enums, errors, maps, bytes, traits, callbacks | similar claims, no fixture suite |

Proposal: **uniffi-dart**, on maintenance grounds. Its README's "5 critical blockers" list is stale. `MapCodeType` is implemented in `src/gen/compounds.rs` and `fixtures/proc-macro/` exists, so `HashMap` and proc-macros both work.

## Fit against the actual `moq-ffi` surface

33 `uniffi::Object`, 25 `Record`, 6 `Enum`, one `flat_error` enum, 45 `pub async fn`, `HashMap<String, V>` in `MoqCatalog`, `Vec<u8>` frames. All covered. We use **zero** callback interfaces and zero trait interfaces, which avoids uniffi-dart's weakest area entirely.

Async works across threads: `src/gen/types.rs` emits `NativeCallable<UniffiRustFutureContinuationCallback>.listener(...)`, so completions from the dedicated `moq-ffi` tokio thread land correctly on the Dart isolate. That was the main technical risk and it is already handled upstream.

Flutter packaging is no longer the obstacle it once was. Build hooks (`hook/build.dart`) and code assets are stable as of Dart 3.10 / Flutter 3.38; link hooks and tree-shaking landed in 3.13. Cargokit was archived 2026-03-26 in favor of exactly this, and `native_toolchain_rust` drives cargo from a build hook.

## The blocker

`moq-ffi` is on uniffi **0.32**; uniffi-dart targets **0.31.2**. Per #2946 / #2949, `UNIFFI_CONTRACT_VERSION` stayed at 30 while the metadata encoding changed, so a 0.31 bindgen dies on the first symbol. Upstream [#149](https://github.com/Uniffi-Dart/uniffi-dart/pull/149) (uniffi 0.32) and [#152](https://github.com/Uniffi-Dart/uniffi-dart/pull/152) (library-mode CLI) are open, both by DenisovAV, both `mergeable_state: unstable`, last updated 2026-08-23.

The library API `gen::generate_dart_bindings(..., library_mode: true)` already works on `main`; only the *binary* is missing. Worst case we write a 30-line bin like `rs/moq-ffi/uniffi-bindgen.rs`.

### Phase 0: spike, gate everything else on it (~1 day)

Cherry-pick #149 + #152 + #151 onto a branch, build `moq-ffi` with `--no-default-features`, generate against `libmoq_ffi.dylib`, run `dart analyze` on the output, and drive one connect + announce + subscribe round-trip against a local relay from a plain Dart CLI. If the generated Dart does not analyze clean or async deadlocks, stop and reconsider.

### Phase 1: the generator

Publish `kixelated/uniffi-dart` tag `v0.3.0+v0.32.0`, mirroring the `kixelated/uniffi-bindgen-go` arrangement. Add a `buildRustPackage` derivation next to `uniffi-bindgen-go` in `flake.nix`. The repo + tag will need naming in the same handful of places: `flake.nix`, `dart/scripts/check.sh`, the release workflow env, `dart/README.md`, `doc/lib/dart/`.

Test the *unlocked* `cargo install --git` resolve explicitly. That is where the `toml = ">=0.9, <2"` range bit us on #2949, and no local path (cargo build, nix vendoring, the fixture suite) exercises it.

### Phase 2: `dart/` layout

Mirror the `kt/` two-package split:

```
dart/
  moq_ffi/    # generated bindings + hook/build.dart, version tracks the crate
  moq/        # ergonomic wrapper, versioned independently
  scripts/{check,package,publish}.sh
  justfile    # `mod dart` in the root justfile
```

Open decision: does `hook/build.dart` **compile** Rust on the consumer's machine (requires their Rust toolchain, ~5 min first build) or **download** a prebuilt from the existing `moq-ffi-v*` release assets? Leaning download-with-cargo-fallback, matching how Swift ships an XCFramework and Kotlin a Maven `.so`.

Ship `--no-default-features` first. `audio` / `video` drag in ~40 crates plus openh264's vendored C++.

### Phase 3: ergonomic `moq` package

Port `kt/moq/src/jvmAndAndroidMain/kotlin/dev/moq/Moq.kt`: `Moq.connect()`, announcements as a Dart `Stream` rather than a poll loop, `Future`-based everything, `close()`. Plus a Flutter example app.

### Phase 4: CI and release

`just dart check` skipping unless the diff touches `dart/` or `rs/moq-ffi/`, a `_tools` entry for `MOQ_STRICT`, `dart` + `flutter` in the devShell, and `release-dart-ffi.yml` / `release-dart.yml` modeled on the `release-kt-*` pair. pub.dev accepts GitHub Actions OIDC publishing, so there is no token to manage.

### Phase 5: docs

`doc/lib/dart/`, a `dart/` entry in the CLAUDE.md Cross-Package Sync row for `rs/moq-ffi`, and `doc/lib/index.md`.

## Explicitly out of scope for phase 1

- **Flutter Web.** uniffi-dart has no wasm support ([PR #126](https://github.com/Uniffi-Dart/uniffi-dart/pull/126) is dirty and stalled since May). Point web users at the JS packages.
- **Video rendering.** `moq-ffi` hands back decoded frames, not a Flutter `Texture`. A texture bridge is per-platform plugin work and a separate project. Phase 1 is data-plane only: connect, announce, publish/subscribe, catalog, raw frames.

## Costs

We would own a second bindgen fork, on a repo badged "experimental" with 43 stars and effectively one maintainer plus two drive-by contributors. Every `moq-ffi` change would then ripple to six wrappers instead of five.

The alternative generator, flutter_rust_bridge, is worse for us specifically: it wants its own Rust-side API definition, so it duplicates `moq-ffi` rather than reusing it, which cuts against the single-core rule.

## References

- [Uniffi-Dart/uniffi-dart](https://github.com/Uniffi-Dart/uniffi-dart)
- [nchapman/uniffi-bindgen-dart](https://github.com/nchapman/uniffi-bindgen-dart)
- [Dart hooks](https://dart.dev/tools/hooks)
- [code_assets](https://pub.dev/packages/code_assets)
- [native_toolchain_rust](https://pub.dev/packages/native_toolchain_rust)
- [flutter_rust_bridge: migrate from Cargokit to native assets](https://cjycode.com/flutter_rust_bridge/manual/integrate/migrate-cargokit-to-native-assets)

(written by Claude Opus 5)

kixelated:
Closes #3100.

## Summary

- Add generated Dart bindings for `moq-ffi` plus an ergonomic `moq` package, published to pub.dev.
- Load checksum-verified native assets on Android, iOS, Linux, macOS, and Windows via Dart Native Assets.
- Add local generation and validation recipes, documentation, and release workflows.
- Pin a public UniFFI 0.32 compatible Dart generator fork because upstream has not released that support.

## Known blocker

**The generated bindings leak native memory on every call.** `dart/moq_ffi/lib/src/uniffi_runtime.dart` is generated (`dart/scripts/check.sh` diffs it against a fresh `generate.sh` run), and it leaks two ways:

- `toRustBuffer` `calloc`s the scratch buffer and `ForeignBytes` that Rust copies from, and never frees them, so every string or struct **argument** leaks. Passing the `RustBuffer` itself is fine, since uniffi's callee takes ownership.
- `rustCallWithLifter` frees the `RustCallStatus` but never the returned `RustBuffer`, so every returned `String`, struct, and list leaks. `RustBuffer.free()` is called in exactly one place in the whole binding: the panic path.

`announcement.path()` in a loop therefore leaks without bound. The fix belongs in `kixelated/uniffi-dart` (tag `v0.3.0+v0.32.0`) followed by a regeneration here; patching the file in-tree fails the staleness diff. This should land before the packages go to pub.dev.

## No codecs

The Dart libraries are built `--no-default-features`, so unlike Swift and Kotlin they carry no `audio`/`video`: no `MoqAudioProducer`, `MoqVideoProducer`, or any encoder or decoder. Catalog and container types are present, so an application can carry already-encoded frames. This is now documented in `/doc/lib/dart`, but it is worth a deliberate decision rather than an accident: `release-kt-ffi.yml` already builds the same Android and desktop targets with defaults, and `release-swift-ffi.yml` the same iOS targets, so there is no cross-compilation reason for the difference.

## Changes since the first revision

The Flutter example is gone. It was 116 of the original 161 files, per-platform `flutter create` scaffolding for an app with a single Connect button. No workflow built it, so the Xcode, Gradle, CMake, and win32 runner sources were review surface with nothing behind them, and every open review finding on this PR sat inside that scaffolding. `dart test` already drives the Native Assets hook, so the round-trip tests in `moq/test/` exercise the same load path an application uses. Dropping it also drops Flutter from the toolchain, and with it the extra `nixpkgs-dart` flake input that existed only to get a newer Flutter package set: the pinned nixpkgs already carries Dart 3.12.

Also:

- **Android from a checkout was broken.** The build hook ran plain `cargo build --target aarch64-linux-android`, which has no NDK linker or sysroot. It now dispatches through `cargo ndk`, matching `rs/moq-ffi/build.sh`.
- **The hook packaged stale libraries.** It returned whatever already sat in `target/` without letting Cargo run, so an artifact predating a Rust edit, or built by another recipe with different features, was shipped silently. It now always builds, and declares the Rust sources as hook inputs so Dart re-runs the hook at all when they change.
- **Version pinning follows `kt/gradle.properties`.** `pubspec.yaml`, `hook/build.dart`, and `CHANGELOG.md` carry a `0.0.0-dev` sentinel that `package.sh` replaces from the tag. A hardcoded version goes stale the moment release-plz bumps the crate, and the hook names the assets it downloads after it.
- **`release.sh` was not safe for two callers.** `release-swift-ffi.yml` and `release-dart-ffi.yml` both publish into the release for a `moq-ffi-v*` tag, and both wrote one shared `SHA256SUMS` covering only their own artifacts, uploaded `--clobber`. Whoever finished last erased the other's manifest. The manifest name now comes from `$RELEASE_CHECKSUMS` (Dart uses `SHA256SUMS-dart`), and a lost `gh release create` race falls back to upload.
- Release-path `dart pub get` resolves from committed lock data wherever a lock survives packaging. `moq` loses its lock along with the path override, so `package.sh` drops it rather than ship one describing a different graph.
- `check.sh` matches its scope anywhere in the changed-file list. A `dart/` path listed after any other silently skipped the entire check.
- Documented the codec omission and the Android API 24 / iOS 16 deployment floors, neither of which any `/doc` page mentioned.

## Testing

- `nix develop --command just dart check` (regenerates the bindings and diffs them, analyzes, tests, and validates the publish layout for both packages)
- `nix develop --command just _check-common` (shellcheck, shfmt, actionlint, remark, taplo, nixfmt)
- `dart/scripts/package.sh` exercised for both packages, confirming the sentinel is replaced in the pubspec, the hook, and the changelog, and that `moq` loses its override and stale lock

## Release setup

The first `moq` and `moq_ffi` versions still need a manual pub.dev upload to claim the package names and configure trusted publishing. Later tagged releases use the included OIDC workflows.

---

Original implementation by @kixelated (written by GPT-5); taken over and revised (written by Claude Opus 5).


kixelated:
I've taken this over. Disposition of every review finding below.

### Fixed

| Finding | Reviewer | Resolution |
| --- | --- | --- |
| Centralize writes to the shared moq-ffi release | Codex | `write_checksums` now names the manifest from `$RELEASE_CHECKSUMS`, so swift-ffi and dart-ffi no longer clobber one shared `SHA256SUMS`; a lost `gh release create` race falls back to upload. Confirmed swift-ffi was the only other `release.sh create` caller, so this PR did introduce the race. |
| Build checkout Android assets with cargo-ndk | Codex | The hook dispatches Android through `cargo ndk --platform 24`, matching `rs/moq-ffi/build.sh`. |
| Rebuild native assets after Rust sources change | Codex | The hook always runs Cargo instead of returning a pre-existing artifact, and declares `Cargo.lock` plus every `.rs`/`Cargo.toml` under `rs/` as inputs, so Dart re-invokes it at all. The stale-artifact risk was worse than described: another recipe's build with different features was also eligible. |
| Enforce committed Dart lockfiles in release jobs | Codex, CodeRabbit | `--enforce-lockfile` wherever a lock survives packaging. `moq` is the exception: `package.sh` strips its path override, which invalidates a lock that resolved `moq_ffi` by path, so it now drops the lock rather than ship one describing a different graph. |
| Exclude Cargo version bumps from the PR matrix | CodeRabbit | `!rs/moq-ffi/Cargo.toml` added. |
| Match relevant paths anywhere in `files` | CodeRabbit | Switched to a whitespace boundary. This was a live bug: any `dart/` change listed after another path skipped the whole check. |
| iOS deployment target vs. supported apps | Codex | The conflicting 15.0 declaration was in the example, now deleted. The floors are documented instead: Android API 24 and iOS 16. |

### Moot — the Flutter example is deleted

The `INTERNET` permission, the macOS `network.client` entitlement, `DEVELOPMENT_TEAM = D7D5SDDB5Z`, the `flutter_window.cpp` null dereference, and the concurrent-`connect()` race were all in `dart/example/flutter/`, which is gone. Rationale is in the PR description; briefly, it was 116 of 161 files, no workflow built it, and `dart test` already covers the Native Assets load path.

### Acknowledged, cannot be fixed here

**Free temporary native buffers after string conversion** (CodeRabbit) is correct, and it is the one blocker on this PR. Both leaks are real and unbounded:

- `toRustBuffer` leaks the `calloc`'d scratch buffer and `ForeignBytes` on every string or struct argument.
- `rustCallWithLifter` never frees the returned `RustBuffer`, so every returned `String`, struct, and list leaks. `RustBuffer.free()` appears exactly once in the binding, on the panic path.

`uniffi_runtime.dart` is generated and `check.sh` diffs it against a fresh `generate.sh` run, so an in-tree patch fails the check. This needs a fix in `kixelated/uniffi-dart` and a new tag, then a regeneration here.

### Rejected

**Use one platform-specific library path** (CodeRabbit, `rs/moq-ffi/README.md`). The `libmoq_ffi.{dylib,so}` brace matches the three `cargo run --bin uniffi-bindgen` lines directly above it, and the line before those already says the extension depends on your platform. It reads as a documentation convention, not a runnable command. Making only the Dart line concrete would be less consistent, not more.

### Out of scope, worth a follow-up

`swift/Package.swift.template` declares `.iOS(.v15)` while `release-swift-ffi.yml` builds with `IPHONEOS_DEPLOYMENT_TARGET=16.0`. That is the same mismatch Codex flagged here, but it predates this PR and lives in the Swift package. Happy to file it separately.

(written by Claude Opus 5)

kixelated:
Rebased onto current `main` and added one commit.

`main` moved twice while this was in flight. The second move, #3227, added `#[uniffi(default = None)]` to the optional record fields, which changes what the generator emits: optional record parameters now carry an explicit `= null`. The committed bindings predated that, so `check.sh`'s staleness diff failed on the merge commit even though the branch passed in isolation. Regenerated in `chore(dart): regenerate bindings for the record field defaults`; the diff is exactly those 6 lines.

Worth flagging for whoever merges: that coupling only becomes real once this lands. #3227 was right not to touch `dart/`, because `dart/` did not exist on `main` yet. From here on, any `rs/moq-ffi` record or signature change needs a `just dart generate` in the same PR, which is the `rs/moq-ffi` -> `dart/` row of the Cross-Package Sync table and is what `check.sh` enforces. It also means this branch re-stales whenever an `moq-ffi` change merges ahead of it.

CI is green: 32 passing, the 5 skipped jobs being the tag-only publish steps.

(written by Claude Opus 5)

kixelated:
Ran an adversarial Codex review ahead of merging. **Holding** — one finding is a blocker I should have caught earlier, and it needs a scope decision rather than a patch.

### iOS is advertised but cannot work

[`dart/moq_ffi/hook/build.dart`](https://github.com/moq-dev/moq/blob/codex/dart-bindings/dart/moq_ffi/hook/build.dart) emits iOS assets with `StaticLinking()`. The pinned `code_assets` 2.0.0 says of that class, in its own source:

```
/// Not yet supported in the Dart and Flutter SDK.
// TODO(https://github.com/dart-lang/sdk/issues/49418): Support static linking.
final class StaticLinking extends LinkMode {
```

So no iOS application can consume the `libmoq_ffi.a` assets. Meanwhile `release-dart-ffi.yml` builds and uploads three iOS targets on every release, and `doc/lib/dart/index.md` tells readers the iOS floor is 16. I added that documentation line during the takeover without checking that the link mode was implemented, so the docs currently promise a platform that cannot load.

Two ways out, and it is a product call: drop iOS from the matrix and the docs until dart-lang/sdk#49418 lands, or establish that Flutter's iOS embedder consumes `.a` assets through some path the `code_assets` doc does not describe.

### `MOQ_DART_FFI_LIB` is dead

Build hooks run with a filtered environment, so the override never reaches the hook. Verified directly: pointing it at a nonexistent path should raise `StateError`, but `dart test` passes and silently builds through Cargo instead. `check.sh` exports it believing it pins the library; it works only because the checkout path happens to do the right thing. For a published package there is no override at all, so an offline build still tries the network.

### Two more, both real

- The hook declares `Cargo.lock` and everything under `rs/` as inputs, but Cargo also reads the root `Cargo.toml`, `.cargo/config.toml`, and `rust-toolchain.toml`. Editing `[profile.release]` leaves the cache valid and bundles a stale library.
- `_download` returns an existing cache file without re-verifying its checksum, and creates then writes it non-atomically. A build interrupted mid-write leaves a truncated file that later builds accept.

### Cleared

Codex confirmed the `release.sh` concurrency fix is race-safe for both callers, that Swift's behavior is unchanged, and that the packaging scripts are portable across macOS and Linux. CI is green at 32 passing.

Not merging. The last three are straightforward to fix; the iOS question should be settled first, since it decides whether the iOS matrix entries and docs stay at all.

(written by Claude Opus 5)

PR #3225, opened 2026-08-31
feat(net)!: announcements are prefix routes

kixelated:
## What

IETF moq-transport's `*_NAMESPACE` messages are specified as prefix routes, but we treated
them as per-broadcast announcements. This embraces the route model everywhere: an
announcement is now a **route**, a claim that paths under a prefix can be served, and it is
up to the application to decide that a path is a broadcast and request it.

Core model (`moq-net`):

- `origin::Producer::announce(Route) -> Announcement`: advertise a prefix explicitly. The
  `Announcement` guard keeps it advertised; `update()` re-prices it (hops/cost); dropping it
  retracts. `origin::Route { prefix, hops, cost }` is the new announced unit.
- `origin::Consumer::announced()` yields `announce::Update { route, active }`, never a
  `broadcast::Consumer`. Resolution is explicit: `request_broadcast(path)` walks local
  broadcasts, then the most specific covering route (served on demand by the session that
  announced it), then a `dynamic()` fallback.
- `create_broadcast(path)` no longer takes a `Route` and no longer announces. The
  convenience `publish_broadcast(path) -> (broadcast::Producer, Announcement)` does both,
  and is what the FFI's `create_broadcast` maps to (so wrappers keep announcing exact paths
  by convention). `Consumer::routed(path)` awaits a covering route (replaces
  `announced_broadcast` in the Rust API; the FFI keeps `announced_broadcast` as
  routed-then-request).
- `broadcast::Broadcast` goes back to being a simple path: `Route`, `Cost`, epoch, and all
  the `set_route`/`route_changed`/`routes()` machinery are gone from it. `Cost`/`MAX_COST`/
  `DRAIN_COST` move to `origin`.
- `origin::Producer::dynamic()` now works with Rust consumers: announce a short prefix (for
  example `.dash`) and serve whatever is requested beneath it.

Wire: **no byte-format change** to the lite-06 announce messages. The suffix is
reinterpreted as a route prefix, and the drafts are updated to say so (Epoch, which was
spec-only and never implemented, is removed; `draft-lcurley-moq-broadcast` is retired).
IETF `PUBLISH_NAMESPACE`/`SUBSCRIBE_NAMESPACE` now map naturally: each namespace is a route.

Applied to every language: Rust workspace, JS (`@moq/net` route table with lazy
materialization), Python, Swift, Kotlin, Go, C (`libmoq`), and wasm, plus the docs and the
IETF drafts.

## Deliberate semantic changes

- **Failover is abort-and-resubscribe, not a transparent splice.** Cross-session splicing
  relied on content identity (epoch / first hop) that per-broadcast announces carried; with
  epoch gone the splice is unsound, so a subscription through a dead route ends and the
  consumer resubscribes through the best remaining route (matching "refusals are
  authoritative; abort so the decoder resubscribes"). Local fronts (two `create_broadcast`
  at one path in the same origin) still splice, newest source wins.
- **The per-broadcast warm-cost discount is dropped** (COST_LINGER, the handover hold, the
  reselect wait). The wire keeps the warm/cold pair and the draft keeps the discount as a
  MAY, but the relay forwards accumulated costs only for now. Back to the drawing board
  with route-level identity.
- **Takeover is per-request**: new requests resolve through the best current route (newest
  wins ties, and identical-meta reannounces are invisible), but a new announce no longer
  ends the incumbent's existing subscriptions; those end when their session does.
- Known follow-ups: moq-hls loses its warm-track linger (it re-resolves per fetch); JS has
  no producer-side `announce(prefix)` yet (it consumes prefix routes; publishing still
  announces exact paths).

## API consistency round (post-review)

- **`Prefix` newtype**: what a route covers is its own opaque type with `covers()` and
  per-segment matching, so wildcard patterns can land later without a breaking change.
- **`Route { hops, cost }`**: the prefix is no longer part of the route; it is what the
  route covers. `announce(prefix, route)`, `Update { prefix, route, active }`, and
  `update(route)` can no longer express a prefix change at all (the earlier runtime
  NotFound check and its test are deleted as unrepresentable).
- **`announce::Producer` is the advertisement guard** (was `origin::Announcement`),
  completing the split-handle symmetry with `announce::Consumer`/`announce::Update`. The
  old `announce::Producer`, a consumer factory with one method, is deleted.
- **`publish_broadcast` removed**: announcing before a broadcast has tracks is usually a
  bug, so the blessed order is create, populate, announce. `moq-cli`'s `Publish` attaches
  its announcement after constructing the catalog for the same reason.
- **JS parity**: `insertRoute` is now `Producer.announce(prefix, provider)` (the fused
  announce+dynamic: serve a subtree from one provider), and the announce `Event` carries
  `prefix` rather than `path`. FFI + all wrappers take `announce(prefix, route)` with
  `MoqRoute` reduced to hops and cost.

## Wildcards (planned ahead)

`Prefix` is opaque with `new`/`covers`/`as_path`, and matching is segment-wise
intersection in one place, so a future pattern type (prefix + suffix, per the moq.pro
wildcard quest) extends `Prefix` internally without touching `Route` or any signature.

## Cross-Package Sync

- `rs/moq-net` wire/API -> `js/net`, `doc/concept`, `drafts/draft-lcurley-moq-lite.md`: done.
- `rs/moq-ffi` -> `rs/libmoq`, `py/`, `swift/`, `kt/`, `go/wrapper`, `doc/lib/*`: done.
- `rs/moq-relay` -> `doc/bin/relay/` (cluster costs, 1+1): done.
- drafts: `moq-lite` rewritten (announce = routes, Epoch removed), `moq-cluster` updated
  (metadata-only updates, no content identity), `moq-broadcast` retired.

## Bugs found while validating

- **Lost wakeup in the lite announce loop** (`AnnouncePrefix`): the `Run` arm served
  route requests before parking on the announce decode, so a request that arrived
  after a route attached in the same pass had no registered waiter and nothing else
  re-polled the machine. Deterministic on the internal unix listener (fast in-process
  connects), hidden by latency elsewhere. Fixed by draining buffered announces first
  and registering on every route's request queue before parking.
- **Fetch-gate gap in `ServeLoop::new`**: `track::Request::accept` releases the
  request's built-in fetch handler before the relay's own `dynamic()` registered, so a
  cache-miss `fetch_group` issued during the TRACK_INFO round trip drained as
  `NotFound`. Fixed by grabbing the dynamic from the request before accepting
  (regression covered by `broadcast_moq_lite_05_fetch_webtransport`).
- **Dropped announcement guards**: `publish_broadcast` returns `(Producer,
  Announcement)`, and three callers held the producer but let the guard die early,
  silently unannouncing: moq-cli's `Publish` (stdin/capture publishing was
  unannounced the moment setup returned), moq-bench's per-connection publish loop,
  and several tests. `Publish::new`/`capture` now take the pair.

## Testing

- Full Rust workspace: `cargo nextest run --workspace` (all pass) + clippy `-D warnings`,
  fmt, rustdoc.
- The goaway cluster tests were reworked to the new failover contract: the draining
  session serves through the handover window, the subscription ends when it closes,
  and a resubscribe carries on; route re-pricing surfaces as metadata updates, never a
  retraction. The diamond test also stops expecting live-edge catch-up: on lite-05 Max
  Age is a staleness tolerance, so a fresh subscription starts at the live edge (dev's
  #3158 semantics, unchanged here).
- JS: full typecheck + package tests.
- `just drafts check`.
- Smoke matrix: `just test smoke-full` (wire + ffi change).

(Written by Claude Fable 5)


kixelated:
All three addressed in the force-pushed branch (now also rebased onto the latest `dev`):

- **Fuzz corpus**: the 24k generated files under `fuzz/seeds`/`fuzz/corpus` and the crash artifact were untracked local state swept in by the first commit; the branch history was rewritten to exclude them entirely (they were also what broke CI: the changed-file list blew past `ARG_MAX` in the diff-aware `just` recipes, exit 126).
- **Two-step routed/request race**: both `moq-ffi`'s `announced_broadcast` and `libmoq`'s `consume_announced` now loop, waiting for the path's coverage to change before retrying an `Unroutable` resolution, so retraction churn keeps the wait alive instead of surfacing a spurious terminal failure. The announce-cursor dedupe now also keys on servability so a served route replacing an advertise-only one under identical metadata wakes such a waiter. Regression test: `announced_broadcast_survives_an_unservable_route`. One known corner remains: a local `announce(P)` cover plus a later local `create_broadcast(P)` with identical metadata is invisible to announce cursors (the broadcast lands in the trie, not the route table), so a waiter parked on exactly that path under its own cover stays parked; noted as a follow-up rather than special-cased here.
- **Mismatched update prefix**: the prefix is now pinned on the `Announcement` at announce time and `update` returns `NotFound` for any other prefix instead of silently applying the metadata under the old one. Regression test: `update_rejects_a_mismatched_prefix`.

(Written by Claude Fable 5)

PR #3265, opened 2026-09-01
fix(watch): guard read() against a re-anchor rebasing the ring under it

### LINKED ISSUE #3253: SharedRingBuffer: an in-flight read() can commit an old-timeline READ across a re-anchor
Found by Codex while reviewing [#3197](https://github.com/moq-dev/moq/pull/3197). Pre-existing: `read()` is byte-identical on `main` and on that branch, and the interleaving does not involve anything the PR adds.

## Mechanism

`read()` checks STALLED once at entry, snapshots READ and WRITE, copies samples, and only then advances READ:

```ts
read(output: Float32Array[]): number {
    if (Atomics.load(this.#control, STALLED) === 1) return 0;

    let read = Atomics.load(this.#control, READ);
    const write = Atomics.load(this.#control, WRITE);
    ...
    // copy samples
    Atomics.casAdvance(this.#control, READ, (read + count) | 0);
}
```

A read that starts just before `reset()` passes the stall check. `reset()` and the next `insert()` then rebase READ and WRITE to the new anchor's origin. The in-flight read resumes and advances READ using its old-timeline snapshot, which lands far beyond the new WRITE on a long-running stream.

The ring then reads as empty and `insert()` discards new samples as too old until WRITE catches up, which is the same failure shape [#3197](https://github.com/moq-dev/moq/pull/3197) fixes for the resize handoff. Codex reproduced it: after re-anchoring with 64 new samples, READ was 64 with WRITE at 64, and the next read returned nothing.

## Scope note

[#3197](https://github.com/moq-dev/moq/pull/3197) adds a GENERATION slot to the control array and a protocol for detecting exactly this class of re-anchor race, but only the resize reconcile honours it. `read()` never consults it. So after that PR the invariant exists but is only half enforced, which is worth closing.

## Possible fix

Make `read()` generation-aware: snapshot GENERATION alongside its cursors, re-validate before committing the advance, and discard the quantum when it changed. Discarding costs one quantum of silence, which is the tradeoff `truncate()` already documents.

Wants a deterministic regression test that pauses `read()` after its cursor snapshot, performs `reset()` plus a re-anchoring `insert()`, then resumes and verifies the new utterance is still readable. The parameterised test added in #3197 does this for the reconcile path and is a reasonable template.

(written by Claude Opus 5)

kixelated:
Closes #3253, found by Codex while reviewing [#3197](https://github.com/moq-dev/moq/pull/3197) and confirmed pre-existing (`read()` was byte-identical on `main` and that branch).

## Root cause

`read()` checks STALLED only on entry, then snapshots READ and WRITE, copies samples, and only afterwards advances READ. A `reset()` plus a re-anchoring `insert()` landing in that window rebases both cursors, so the advance commits a playhead measured against an anchor that no longer exists.

On a long-running stream READ lands far past the new WRITE. The ring then reads as empty and `insert()` discards new samples as too old until WRITE catches up, which is the same failure shape [#3197](https://github.com/moq-dev/moq/pull/3197) fixed for the resize handoff. The regression test reproduces it at 2000 samples past WRITE.

## Fix: one guarded advance, not three

[#3197](https://github.com/moq-dev/moq/pull/3197) already had to solve this for the reconcile, and got it wrong twice before it was right. So rather than duplicate that reasoning into `read()`, this folds it into a single `#advanceRead(candidate, generation)` that returns the resulting cursor, or `undefined` when the timeline moved under it. It keeps the three properties that took two rounds to get right:

- clamp to a freshly loaded WRITE, so a racing advance cannot park READ beyond the new timeline
- compare-exchange from an exact observed READ, so the re-anchor's `store(READ, 0)` makes it fail and retry
- re-read GENERATION after a successful exchange and undo, catching the ABA where the observed READ was already 0

All of it rests on `insert()` bumping GENERATION *before* rebasing the cursors, which is what [#3197](https://github.com/moq-dev/moq/pull/3197) established.

Both of `read()`'s advances now go through it: the latency skip and the final commit. Either one seeing a moved timeline drops the quantum, since the copied samples belong to a playhead that no longer exists. A quantum of silence is what `truncate()` already trades for the same reason.

`#reconcile` uses the same helper, so the concurrency argument lives in exactly one place.

## Test plan

- `nix develop --command just js test` (all packages, 0 fail)

Two tests force the interleaving by firing the re-anchor from inside `Atomics.compareExchange`, which is precisely the window: the read has already snapshotted its cursors by then. `buffered` mode picks which advance is under test, since the latency skip only runs in non-buffered mode, leaving the commit as the only advance in buffered mode.

Both guards are independently load-bearing: removing either one fails exactly its own test.

Cross-package sync is not applicable: no wire format, package API, or watch UI change.

(written by Claude Opus 5)

### DIFF
--- a/js/watch/src/audio/shared-ring-buffer.test.ts
+++ b/js/watch/src/audio/shared-ring-buffer.test.ts
@@ -893,7 +893,12 @@ describe("read() racing a re-anchor", () => {
 			};
 
 			try {
-				worklet.read(out);
+				// The raced read must discard its quantum: not just report zero, but leave no
+				// old-timeline samples in the buffer it was handed, since the worklet renders
+				// whatever is there regardless of the count it gets back.
+				out[0].fill(-1);
+				expect(worklet.read(out)).toBe(0);
+				expect(out[0].some((sample) => sample !== 0 && sample !== -1)).toBe(false);
 			} finally {
 				atomics.compareExchange = realExchange;
 			}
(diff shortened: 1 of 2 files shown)


kixelated:
Superseded by [#3286](https://github.com/moq-dev/moq/pull/3286), which is merged.

This branch guarded `read()`'s advances with the generation counter from [#3197](https://github.com/moq-dev/moq/pull/3197). Review found the guards themselves unsound: returning 0 did not clear the buffer the worklet renders, the no-op path could report an advance nobody made, and the rollback compare-exchange could rewind the new timeline when the rebased cursor happened to equal the value it was restoring.

[#3286](https://github.com/moq-dev/moq/pull/3286) reverts that protocol and packs the epoch and cursor into one atomic word, which removes the need for these guards entirely. The `read()` race this PR was filed against (#3253) is fixed there and covered by a sweep that fires a re-anchor at each atomic load inside `read()`.

(written by Claude Opus 5)

PR #3268, opened 2026-09-01
chore(quest): replace the Dart epic with its two open follow-ups

### LINKED ISSUE #3100: Dart/Flutter bindings via UniFFI

kixelated:
Scoping pass over `quest/m3/plan-dart-bindings.md`, which wanted to "run the spike as its own quest and let its verdict decide whether the rest gets planned at all".

## The spike is answered by an implementation

[#3215](https://github.com/moq-dev/moq/pull/3215) is open against `main`, carries `Closes #3100`, and delivers every phase the epic planned: the pinned UniFFI 0.32 generator fork, the two-package `dart/` layout, the ergonomic `moq` wrapper, checksum-verified native assets across Android/iOS/Linux/macOS/Windows, `just dart check`, the OIDC release workflows, and `/doc/lib/dart`.

Phase 0 asked whether the generated Dart analyzes clean and whether async deadlocks across the `moq-ffi` tokio thread. A working implementation with round-trip tests answers both. So the placeholder has nothing left to gate, and both quests are deleted without a closing keyword — #3215 owns that.

## Two quests, from what that PR names as unfinished

**`m0/dart-leak.md` (M)** — the generated bindings leak native memory on every call, so `announcement.path()` in a loop grows without bound:

- `toRustBuffer` allocates the scratch buffer and the `ForeignBytes` Rust copies from and frees neither, leaking every string or struct **argument**. (Passing the `RustBuffer` itself is fine — uniffi's callee takes ownership of that one.)
- `rustCallWithLifter` frees the `RustCallStatus` but not the returned `RustBuffer`, leaking every returned `String`, struct and list.

`RustBuffer.free()` is called in exactly one place in the whole binding: the panic path.

The fix is upstream **by construction**, not by preference: `uniffi_runtime.dart` is generated and `dart/scripts/check.sh` diffs it against a fresh `generate.sh` run, so an in-tree patch fails the staleness check by design. Held as a plain-text `Required` on a `kixelated/uniffi-dart` release. **m0**, because once #3215 lands this is a defect in what `main` ships, and it gates the first pub.dev upload — which is manual anyway (package names claimed, trusted publishing configured), so there is a natural place to hold it.

**`m2/dart-codecs.md` (M)** — Dart is the only binding that cannot originate media. Built `--no-default-features`, it has no `MoqAudioProducer`, no `MoqVideoProducer`, no encoder or decoder. Catalog and container types are present, so an application can carry already-encoded frames and nothing is broken; it is simply the odd one out. The PR flags this as "worth a deliberate decision rather than an accident", and the deciding fact is that there is **no cross-compilation reason** for it: `release-kt-ffi.yml` already builds the same Android and desktop targets with defaults, and `release-swift-ffi.yml` the same iOS targets. The cost is known (~40 crates plus openh264's vendored C++), which is why Dart shipped lean first.

Neither stays in m3. Dart stopped being a prototype the moment the bindings worked.

`just check` passes (`quest: 234 documents ok`).

(written by claude-opus-5)


kixelated:
Codex's P2 finding was correct and is fixed in the follow-up commit: the Dart epic is restored (with a note that #3215 is executing it and should delete it on merge), and both follow-up quests now carry a `Required` on it so neither reads as ready. An open PR is not grounds for deleting the only quest tracking the work.

(written by claude-opus-5)

Issue #3193, opened 2026-08-30
Expose a cancellable route watch API in Python


kixelated:
## Problem

Python handles route changes differently from the other bindings. `BroadcastConsumer.route_changed()` lazily creates a route watch, caches it in a private field, and awaits its next value:

- [Python `BroadcastConsumer` route handling](https://github.com/moq-dev/moq/blob/7494084aaf7e2fa6abe553ac83101ac4ef19f33a/py/moq-rs/moq/subscribe.py#L376-L405)

The watch handle is not public and has no deterministic cancellation path. A caller that abandons route observation must rely on garbage collection or broadcast shutdown to release the native watch.

Other wrappers expose explicit lifetime ownership:

- [Go `RouteWatch.Next(ctx)` and `Cancel`](https://github.com/moq-dev/moq/blob/7494084aaf7e2fa6abe553ac83101ac4ef19f33a/go/wrapper/subscribe.go#L15-L40)
- Swift exposes a cancellable asynchronous sequence.
- Kotlin exposes a `Flow` that cancels the watch in `finally`.

## Proposed direction

Expose a public Python route watch abstraction with deterministic cleanup, following the existing wrapper conventions. Suitable shapes include:

- an async iterator returned by `routes()`
- an async context-managed `RouteWatch`
- both, with `route_changed()` retained only as a convenience over the owned watch

The owner should be able to cancel a pending wait and release the native resource without waiting for garbage collection.

## Acceptance criteria

- Python callers can explicitly own and close a route watch.
- Abandoning iteration deterministically cancels the native watch.
- A pending route wait is cancellable by normal asyncio task cancellation.
- Tests cover early iterator exit and cancellation while waiting.
- Lifecycle documentation matches the Go, Swift, and Kotlin ownership contract.

(Written by GPT-5.6)

kixelated:
Finalized plan:

- Add a public `RouteWatch` wrapper around the native route-update consumer.
- Add `BroadcastConsumer.route_updates() -> RouteWatch` as the canonical API.
- Make `RouteWatch` an async iterator and async context manager, with a terminal `cancel()` for explicit cleanup.
- Keep `route_changed()` compatible by lazily caching a `RouteWatch`.
- Export `RouteWatch` from the `moq` package and document canonical `async with ...` / `async for ...` usage.
- Cover the initial route, ordered updates, clean termination, context-manager cleanup, pending-read cancellation, reuse after a timed-out read, and existing `route_changed()` behavior.

[...]

PR #3274, opened 2026-09-01
chore(quest): rescope routing quests after prefix routes

kixelated:
## What

[#3225](https://github.com/moq-dev/moq/pull/3225) made announcements prefix routes and deleted machinery three questlines were built on. This rescopes them against what `dev` actually does, rather than leaving quests that send someone at APIs and specs that no longer exist.

## New m0 quests

Both are regressions on `dev`, not plans, so m0's goal widens to cover defects in what main **or dev** ships.

- **[Route resume identity](quest/m0/route-resume.md)** `[L]` - a subscription whose route dies now ends instead of resuming through the next one. #3225 called this unavoidable once `Epoch` was gone, but the first hop is not gone: hop chains still travel on the wire, `RouteEntry` still stores them, `route_order` still ranks on them. What was deleted is `FrontState.publisher` and `same_identity`, the structure that read the first hop. Restoring that restores the failover behavior and supplies the identity two other questlines need.
- **[Lite draft routing text](quest/m0/lite-draft-routing.md)** `[S]` - `draft-lcurley-moq-lite` still normatively specifies the warm-cost discount and the `(cold, hash)` adoption rank that the relay no longer implements, which misleads anyone interoping against it.

## pop-skipping

`warm-lifecycle` is replaced by **[warm-advertise](quest/m2/pop-skipping/warm-advertise.md)**. The core tension: warmth is per-broadcast, but a route covers a prefix, and carrying `pid/foo.hang` says nothing about the rest of `pid/`. So a carrying relay advertises the **exact broadcast path as its own warm route** instead of discounting a broader one. No wire change; `best_server` already prefers the more specific route.

Two rulings recorded:

- **Selection keeps specificity above cost.** That is what `best_server` and the draft do, and it matches longest-prefix-match everywhere it appears (IP forwarding, BGP, DNS, URL routers) for the same reason: routes of different prefix length describe different destination sets, so comparing their costs is a category error. Cost decides between routes covering the same thing, which is where warm/cold was designed to work.
- **A draining carrier retracts rather than reprices.** Under specificity-first, a ceiling-priced exact-path route still wins and keeps attracting the subscribers a drain is trying to move. This replaces the draft's ceiling-exemption paragraph.

`rank` now requires warm-advertise and route-resume. `peer-reconfigure` is untouched.

## wildcard

`epoch` is deleted: #3225 removed `Epoch` from the drafts and retired `draft-lcurley-moq-broadcast`, so the quest implements a spec that no longer exists, and route-resume covers colliding publishers (more weakly - it can end the loser, not declare two workers interchangeable, which is called out).

The rest is corrected against dev, and mostly gets **easier**:

- `origin::Prefix` is an opaque newtype built as the pattern extension point, so a pattern extends it internally instead of needing a parallel table.
- The routing table exists. `request_broadcast` resolves local, then `best_server` (longest covering prefix, exclusion filter, `route_order`), served on demand and cached in `ServeState.served`. `resolve` extends it rather than standing one up.
- `announced()` yields `Update { prefix, route, active }` and never a broadcast, so a pattern needs no distinct event kind.
- `js/watch`'s `#isPathAnnounced` is already prefix-aware; only pattern matching is left, so `demand` drops `[M]` to `[S]`.
- `draft` gains an open decision: pattern in the existing prefix field (smaller change, but an older peer mis-covers silently) vs a new message type (fails loudly).

## relay-memory

**standby-routes is closed - #3225 delivered it.** `lite::subscriber` now calls `origin::Producer::announce_served`, so an announcement costs a `RouteEntry` plus one `ServeState` whose `served` cache materializes a broadcast only on request. No `broadcast::Producer`, no front, no lifecycle task per announced path. Its accessor rider survives as [route-gauge](quest/m2/relay-memory/route-gauge.md) `[XS]`. The questline's measurement table predates prefix routes and is flagged as needing a rebuild.

## Fallout elsewhere

| Quest | Correction |
|---|---|
| [2933](quest/m1/2933-moq-ffi-moqroute-round-trip-erases-a-routes-cold-cost.md) | `set_route`/`route_updates` are gone, but the erasure survives via `announced()` → `announce()`; the severity argument rested on the deleted `handover_allowed` |
| [2985](quest/m1/2985-js-net-path-keyed-publisher-state-goes-stale-when-a.md) | its remedy was the broadcast epoch; the three JS bugs remain with no proposed fix |
| [2895](quest/m1/2895-add-an-atomic-readiness-gate-for-origin-broadcasts.md) | `create_broadcast` no longer announces, so only the exact-path half remains |
| [3190](quest/m1/3190-align-origin-broadcast-creation-naming-across-language.md) | Rust's `create_broadcast` no longer announces, the FFI's still does: a semantics divergence, not a naming one |
| [2714](quest/m1/2714-per-subscriber-path-predicate-for-originconsumer-0-1-x.md) | a predicate over prefix routes decides about a *set*, and cannot be the enforcement point the reporter needs |

**3193 is deleted.** `BroadcastConsumer.route_changed()` no longer exists in any binding, and a broadcast carries no route to watch. Issue [#3193](https://github.com/moq-dev/moq/issues/3193) should be closed as obsolete - I have not touched it.

## Testing

`just check` passes: `quest: 235 documents ok`.

(Written by Claude Opus 5)


kixelated:
Taking this over. The two blocking questions are settled as follows (per Luke's direction on the first):

1. **route-resume stands, as a deliberate spec reversal rather than a regression fix.** Routes keep carrying the hops/cost of what they serve, and a relay resumes/stitches a broadcast across routes when the first hop is the same. Epoch stays dead (it was per-broadcast; nothing per-broadcast survives prefix routes). The quest now states this as a decision, requires the implementing PR to amend the draft's "route carries no content identity" / MUST NOT splice paragraph in the same change, and is resized to XL: ServeState is per-announcer and create_broadcast sources are not inserted into the tree, so the per-path front spanning routes has to be rebuilt, not just given back a field.

2. **Specificity stays above cost, and the pricing bullet now says so honestly.** A concrete claim is maximally specific and shadows every pattern behind it at any cost, so "there is no concrete-beats-wildcard rule" was wrong and is gone. Cost orders within a tier. The seed floor keeps its job where equal-specificity claims actually meet (standby vs running concrete claims, and warm-advertise's exact-path warm routes), and the stale-concrete-claim shadow is documented as an accepted consequence, bounded by the claiming session's lifetime.

Also dropped lite-draft-routing.md from this PR: the work is already open as #3278 against dev, so the quest would land pre-completed. rank.md and warm-advertise.md now reference the PR instead.

(Written by Fable 5)

PR #3280, opened 2026-09-01
refactor(watch): give every shared-ring control slot a single writer

### LINKED ISSUE #3253: SharedRingBuffer: an in-flight read() can commit an old-timeline READ across a re-anchor

kixelated:
Closes #3253, and supersedes [#3265](https://github.com/moq-dev/moq/pull/3265).

## Why not another guard

Every bug in this ring has been the same one. The playhead was a single slot both threads compare-exchanged, and a cursor *value* does not identify which timeline it belongs to. So a re-anchor landing between any two of the reader's atomics let it publish a position measured against an anchor that no longer existed, and the ring then read as empty while `insert` discarded everything as too old.

Four guards were added for that across [#3197](https://github.com/moq-dev/moq/pull/3197) and [#3265](https://github.com/moq-dev/moq/pull/3265): clamp to a freshly loaded WRITE, compare-exchange from an exact observed cursor, re-validate the generation after the exchange, re-validate it on the no-op path too. Each was individually correct. None of them fixed the shape, and review kept finding the next hole — the last being that the rollback could itself rewind the new timeline when the rebased cursor happened to equal the value it was restoring, producing a read spanning both timelines.

## The change

Stop storing the playhead. It becomes a difference of two slots, each with exactly one writer:

| slot | writer | meaning |
|---|---|---|
| `WRITE` | main | end of buffered audio |
| `CONSUMED` | **worklet** | total samples it has ever played, monotonic, never reset |
| `ANCHOR` | **main** | the `CONSUMED` value the current timeline calls position 0 |
| `LATENCY`, `STALLED` | main | unchanged |

`read = CONSUMED - ANCHOR`. Nothing compare-exchanges a cursor, because nothing shares one.

The failure mode changes in kind, not degree. A race can misplace the playhead by at most what the reader played in that window, since both terms sit on one monotonic scale. It can no longer be off by the length of the timeline just left, which was the root of every previous bug.

This deletes `GENERATION`, `#reconcile`, `#advanceRead` and its rollback, and `casAdvance`. The resize handoff is now the reader carrying one integer across the swap, so there is nothing to reconcile: the samples it played from the old ring are already in its own tally.

## Residual window

A read that is already past its `STALLED` check when the writer re-anchors can play up to one quantum of stale samples. That is inherent to the reader checking a flag and then acting, it is bounded by the reader's own quantum, and it is the same trade `truncate()` already documents. Closing it would mean reintroducing a generation, which is what this removes.

## Test plan

- `nix develop --command just js test` (all packages, 0 fail)

56 ring tests. Two sweeps force the interleavings a single-threaded test cannot otherwise reach: one replays `insert`'s re-anchor store by store with a handoff dropped between each pair, the other fires a re-anchor at each atomic load inside `read()`.

Run against `main`'s protocol, the new tests fail in 6 places, three of them the `read()` race from #3253.

## Review note

Two things worth pushing on, since they are where I would expect to be wrong: `resize()` has the main thread seed `CONSUMED`, a worklet-owned slot, before posting the message (the post should give happens-before, but it is the kind of assumption that has been wrong here before); and whether the one-quantum stale-audio window is genuinely bounded in every interleaving.

Cross-package sync is not applicable: no wire format, package API, or watch UI change.

(written by Claude Opus 5)

### DIFF
--- a/js/watch/src/audio/shared-ring-buffer.test.ts
+++ b/js/watch/src/audio/shared-ring-buffer.test.ts
@@ -52,7 +52,7 @@ describe("initialization", () => {
 		expect(init.capacity).toBe(128);
 		expect(init.rate).toBe(1000);
 		expect(init.samples.byteLength).toBe(2 * 128 * 4); // 2 channels * 128 samples * Float32
-		expect(init.control.byteLength).toBe(5 * 4); // 5 control slots * Int32
+		expect(init.control.byteLength).toBe(6 * 4); // 6 control slots * Int32
 	});
 
 	it("rounds capacity up to a power of two", () => {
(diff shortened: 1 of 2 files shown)


kixelated:
Superseded by [#3286](https://github.com/moq-dev/moq/pull/3286), which is merged.

This branch made every control slot single-writer and derived the playhead as `CONSUMED - ANCHOR`. That removed the ABA on the shared cursor, but it could not make a two-step sequence atomic, and review found three races it left open: `resize` double-counting between its two `CONSUMED` snapshots, the handoff carrying a tally across a re-anchor, and `read` rendering samples from a discarded timeline. Fixing the last two needed an epoch counter back, and `#seek` still had an unbounded overshoot I could not close without one either.

[#3286](https://github.com/moq-dev/moq/pull/3286) packs the epoch and the cursor into one 64-bit word instead, so the reader's publish is a single compare-exchange over both and the epoch check *is* the write. That is what actually closed the class of bug this branch was chasing.

(written by Claude Opus 5)

PR #3287, opened 2026-09-01
fix(rs): repair the test call sites the main merge missed

kixelated:
## Summary

`origin/dev` at 3227c195e ("Merge main into dev") does not compile: the merge carried three tests over from `main` without updating them for dev's APIs. The first two are hard errors, and both crates are dependents of widely-used ones, so `just check` and `just test` fail for essentially every branch based on `dev` (that is how I ran into this, from #3282).

- **`moq-video`** (`encode/producer.rs`): a test called `catalog.track_info()`, but dev's `moq_mux::catalog::Producer::track_info` takes a `priority: u8` (#2854). The production call site in the same file was updated by the merge; this one was missed.
- **`moq-rtc`** (`lib.rs`): the WHIP/WHEP round-trip test built origins with `moq_net::Origin::random().produce()` and announced by passing a `Route` to `create_broadcast`. On dev an origin is `origin::Info` plus a driver (`moq_tokio::origin::spawn`, already a dev-dependency), and announcing is separate from creating: `create_broadcast(path)` then `announce(prefix, Route)`, in create-populate-announce order. The announcement assertions move from `path`/`broadcast.is_some()` to `prefix`/`active`.
- **`moq-video`** again: with the first fix in place, `idle_capture_publishes_a_discontinuity_before_resume` compiled and **failed** (`[1, 1]`, expected `[1, 0, 1]`) — a real behavior difference the compile error had been hiding.

  Root cause: the marker is still published (`container::Producer::discontinuity` appends an empty group and finishes it). What changed is delivery. A group's staleness is measured by its *reach*, which is bounded by its successor's start, so the empty group's reach is the resumed group's timestamp — equal to the live edge, and `>=` a `Duration::ZERO` budget. The default subscription therefore sheds it. That is correct for a live consumer, which takes the marker as the live edge when it arrives with no successor; it is wrong only for a test that walks the finished track afterwards. Fixed by subscribing with a replay window, the same thing the `moq-binary` / `moq-json` stream tests do for the same reason.

Test-only changes; no production code touched.

## Public API changes

None.

## Test plan

- `just check` and `just test` (diff-scoped to `moq-rtc`, `moq-video`, and their dependents): 376/376 pass. Both failed on `origin/dev` before this.
- `cargo test -p moq-rtc`: 44 pass, including `whip_and_whep_round_trip_opus`, which had not been able to run.
- A full `just check-all` / `just test all` is running; I will note anything else it turns up.

## Not fixed here

`just drafts check` fails on `draft-lcurley-moq-hang.md:425`, a `{{field-timeline}}` cross-reference with no anchor (from #3109). The `timeline` field exists on the `json` and `binary` rendition schemas, but the draft never got a `### timeline {#field-timeline}` section, and the sentence claims the field "carries the same meaning here as for a media rendition" when media renditions have no such field. Writing that section is spec authoring rather than a build repair, so it wants an author. Nothing in CI runs the drafts check, so it blocks no workflow.

(Written by claude-opus-5)

### DIFF
--- a/rs/moq-rtc/src/lib.rs
+++ b/rs/moq-rtc/src/lib.rs
@@ -85,11 +85,11 @@ mod tests {
 
 	#[tokio::test]
 	async fn whip_and_whep_round_trip_opus() {
-		let source_origin = moq_net::Origin::random().produce();
+		let source_origin = moq_tokio::origin::spawn(moq_net::Hop::random());
 		let source_consumer = source_origin.consume();
 		let mut announcements = source_consumer.announced();
 		let mut source = source_origin
-			.create_broadcast("source", moq_net::broadcast::Route::new().with_announce(true))
+			.create_broadcast("source")
 			.expect("create source broadcast");
 		let catalog = moq_mux::catalog::Producer::new(&mut source).expect("create source catalog");
 		let mut opus = crate::codec::opus::Bridge::new(source, catalog, 48_000, 2).expect("create Opus bridge");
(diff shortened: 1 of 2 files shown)


kixelated:
Closing: every breakage this PR repaired landed first, so nothing here is still needed.

- moq-rtc's origin/announce migration: #3284
- moq-video's `track_info(priority)` and the idle-capture test's replay window: #3285 (a superset of what I had, since it widens the publisher-side window on the track info as well as the subscription)
- `MoqVideoHint`'s `#[uniffi(default = None)]` attributes, the `AudioSampleFormat` / `max_age_ms` names in `py/moq-rs/tests/test_local.py`, and the Go `ErrUnresolvableBroadcast` sentinel: all present on `dev` as of d680b54b1

The one thing still outstanding is unrelated to the build: `just drafts check` fails on `draft-lcurley-moq-hang.md:425`, a `{{field-timeline}}` cross-reference with no anchor (from #3109). The `timeline` field is real on the `json` and `binary` rendition schemas, but the draft has no `### timeline {#field-timeline}` section, and the sentence claims the field "carries the same meaning here as for a media rendition" when media renditions do not have it. That wants an author rather than a build fix, and nothing in CI runs the drafts check.

(Written by claude-opus-5)

PR #3329, opened 2026-09-02
feat(moq-video): add the VAAPI H.264 decoder

Frando:
Adds `decode/backend/vaapi.rs`, 130 lines including its test, adapting `moq_vaapi::decode` to the `Backend` trait. It is the mirror of the 115-line encode backend, registered next to NVDEC and gated on Linux and `vaapi`.

*This PR is part of a series to update iroh-live to latest moq, see n0-computer/iroh-live#45. The code and below description was written by Claude Code*

Output is CPU I420 via `vaDeriveImage`, which works on this hardware, so the only copy is the strided row copy that strips pitch padding. There is a `vaCreateImage` and `vaGetImage` fallback for drivers that cannot derive; instrumented, it never fires here.

## Position in the series

Depends on moq-dev/vaapi#2, which adds the decoder this calls. That is not in moq-vaapi 0.0.3, so this PR points `moq-vaapi` at the branch to give CI something to build. **That dependency line has to go back to a released version before merge.**

#3331 stacks on this one. It needs `I420::from_nv12`, added here, and the same `moq-vaapi` dependency. This PR stands alone and can merge first.

## Verified

Tested on Intel Meteor Lake with iHD 26.1.5. `decode::backend::vaapi::tests::vaapi_h264_round_trip` encodes a gradient with openh264, decodes it with VA-API, and compares planes by mean absolute error with exact timestamps. It self-skips rather than being `#[ignore]`d, so it runs wherever the hardware exists and stays silent elsewhere. `Kind::Auto` selects `vaapi` on this machine with NVDEC compiled in and refusing.

The underlying decoder is verified byte-for-byte against ffmpeg in moq-dev/vaapi#2.

## Decoder delay

Output trails input by `num_ref_frames`, not by the reorder depth the VUI declares, because the DPB bumps only when a new picture needs the slot. With x264's default `ref=3`, a short stream yields nothing until the fourth access unit. NVDEC gets zero delay from an explicit cuvid knob and VA-API has no equivalent. This is documented in the backend's module doc.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5133,8 +5133,7 @@ dependencies = [
 [[package]]
 name = "moq-vaapi"
 version = "0.0.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e4209e5f9401f2ed76eb5a82c807966be235c483a6932777623cec30950ba7d4"
+source = "git+https://github.com/Frando/vaapi?branch=pr%2Fdecode#b2550c8f63a7a982e607dc482a5131457f9dd5d2"
 dependencies = [
  "anyhow",
  "bindgen 0.70.1",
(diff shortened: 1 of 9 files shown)


Frando:
Replaced by #3331, which carries these commits plus the zero-copy import that was stacked on them. One PR instead of two stacked ones.

Issue #3334, opened 2026-09-02
`moq export ts`: PCR byte position and release instant are functions of frame arrival, so #2967's grid cannot be recovered from the wire


t0ms:
### Summary

[#2967](https://github.com/moq-dev/moq/pull/2967) made the PCR **values** a uniform 25 ms grid, and they are now exact — 1,811 consecutive intervals with median and maximum both 25.000 ms, one histogram bin. [#3006](https://github.com/moq-dev/moq/pull/3006) made the stdout writer pace on each frame's timestamp, and it does.

What is left is that neither the **byte position** nor the **release instant** of a PCR packet is constrained by the PCR's own elapsed media time:

- **Position.** A PCR packet can only ever be placed *between* media frames, never among the media bytes of the slot it labels. Backfilled slots emit as consecutive single-packet frames, so they land back-to-back: **56.5% of PCR packets sit one packet from the previous one**, with the media bytes they label heaped between the clusters (worst gap 2,611 packets).
- **Release.** A slot cannot be emitted until a media frame arrives whose timestamp proves the slot has elapsed. By then the instant the slot asserts is already in the past, `Pacer::pace` returns a `send_at` behind `now`, and the sleep is a no-op — so the run is written as a burst. **626 of 1,811 intervals are released more than 10 ms *early*** against the interval their own PCR values assert.

These are not two independent observations. Of the 626 early releases, **615 are byte-adjacent**; of the 136 late ones, **none** are. The clusters and the bursts are the same packets.

I am filing this separately rather than reopening anything. [#2937](https://github.com/moq-dev/moq/issues/2937) is correctly closed — #2967 delivered the contract it asked for. [#2984](https://github.com/moq-dev/moq/issues/2984) is correctly closed by #3006 — it asked for the stdout writer to honour `frame.timestamp`, and it does. This is the part #3006 could not reach, with a root cause I did not have when I filed #2984.

### Reproduction

Ubuntu 26.04 on EC2 (`c6in.2xlarge`, 8 vCPU, eu-west-1b), loopback relay, 45 s window. Source is a 1080i25 H.264 contribution capture, 9,945,951 b/s CBR, single video PID.

```bash
# publisher
tsp --realtime -I file capture.ts --infinite \
    -P regulate --pcr-synchronous --wait-min 5 -O file - \
  | moq --broadcast live.hang import ts

# grade the export in all three domains, in one pass
moq --broadcast live.hang export ts --latency-max 3s \
[...]

PR #3335, opened 2026-09-02
test(ts): add pcr-timing.py, grading PCR release timing and byte position

t0ms:
Test tooling only. No behaviour changes.

`compliance.py` grades a captured file on the stream's own PCR clock. That is the right basis for the IRD model math it does, and this directory's README records the consequence in its own caveats:

> Wall-clock delivery jitter/burstiness is intentionally out of scope: all timing is derived from the stream's PCR, not from socket arrival times.

Which means nothing here can observe a change to *when* `export ts` hands bytes over — a file has no release timing left in it. [#3006](https://github.com/moq-dev/moq/pull/3006) landed without a test for exactly that reason, and I could not report [#3334](https://github.com/moq-dev/moq/issues/3334) without first building an instrument that could see it.

### What it adds

`test/ts/pcr-timing.py` grades the three claims a PCR makes, in one pass:

| Domain | What it claims | Graded from |
|---|---|---|
| value | consecutive PCR values are spaced within the repetition limit | the values |
| release | the bytes carrying a PCR were handed over when that PCR asserts | arrival stamps |
| position | a PCR packet sits among the media bytes it describes | packet offsets |

```bash
# live: all three domains, reading the exporter directly
moq --client-connect http://localhost:4443 --broadcast live.hang export ts \
  | ./pcr-timing.py --live --seconds 45

# offline: value and position only
./pcr-timing.py capture.ts
```

It needs only `python3` — no TSDuck, no source file, no declared mux rate — because every check is graded against the stream's **own** PCR values rather than a nominal you supply. If two consecutive PCRs are 25 ms apart in value then they must be ~25 ms apart in arrival, whatever rate the stream runs at. It pays the same price as `compliance.py` for that basis, and for the same reason: a PCR emitted at the wrong rate stays internally consistent, so absolute rate is not what either grades.

Reads stdin one packet at a time. The granularity is deliberate — a coarser read cannot distinguish a run of packets released together from a run released on a cadence, which is the distinction `release` exists to measure.

### Why the checks are severities rather than all-hard

`pcr-position` is a **shape** check (WARN unless `--strict`), because `export ts` is VBR by design. It is reported anyway because a consumer holding only the byte stream — every MPEG-TS tool, and the thing stdout is — recovers the clock from where the PCR packets sit, so it distinguishes a stream whose values are exact from one whose values are also *recoverable*. `pcr-value-interval`, `pcr-release-timing`, `continuity`, `sync` and `pcr-single-pid` are hard.

When `release` and `position` both flag, the report cross-tabulates them, because two invariants failing on the *same* PCRs is one cause rather than two and two aggregate percentages cannot show that. Report-only: it explains a failure, it does not define one.

### Calibration

The property that matters for a gate, and the one that is easy to lose: it **passes a real broadcast source unmodified**. Offline over a 1080i25 contribution capture that never went through moq, every check passes — 1,318 PCRs, median interval 24.648 ms, worst 24.951 ms, 0% adjacent at a median gap of 163 packets. Over a pre-#2967 export capture the same command fails `pcr-value-interval` alone, at the characteristic 0.011 ms median, and *passes* `pcr-position` — pre-#2967 the PCR rode payload packets and so was naturally spread. So neither the value nor the position check is calibrated to moq's output, in either direction.

Against the three builds I have to hand:

| | value | release | position |
|---|---|---|---|
| pre-#2967 | FAIL | FAIL | PASS |
| #2967+#3006 | PASS | FAIL | FAIL |

`just fix` and `just check` are clean on this branch. Happy to wire it in under `just test ts --live` rather than as a sibling script if you would prefer one entry point — I left it standalone because the live mode needs a pipe rather than a capture, which does not fit `run.sh`'s round-trip-then-analyze shape.

### DIFF
--- a/test/ts/README.md
+++ b/test/ts/README.md
@@ -116,8 +116,21 @@ internally consistent, so absolute rate is not what this grades.
 | `continuity` | hard | no discontinuities, and a payload-less packet must not advance the counter (ISO 13818-1 2.4.3.3) |
 | `pcr-single-pid` | hard | every PCR rides one PID |
 | `pcr-value-interval` | hard | no interval above `--repetition-ms` (default 40, TR 101 290) |
-| `pcr-release-timing` | hard | each interval's arrival is within `--release-ms` of the interval its own values assert, and accumulated drift over the whole sample stays within `--drift-ms` (`--live` only) |
+| `pcr-release-timing` | hard | no more than `--release-pct-max` of intervals arrive further than `--release-ms` from the interval their own values assert, and accumulated drift stays within `--drift-ms`, being the standing lag the sender is allowed to hold (`--live` only) |
 | `pcr-position` | shape | share of PCR packets within `--adjacent-packets` of the previous one |
+
+Accumulated drift has two shapes and only one is a defect, so the check bounds
+the total and reports the rate over the tail of the sample beside it. A sender
+that buffers builds a standing lag once and then runs at the media rate: the lag
+is a constant offset no receiver can see, and it cannot grow past the latency
+budget the sender is allowed to hold, so set `--drift-ms` to that budget
+(`export ts --latency-max`, 500 ms by default). A pipe that is not running at the
+media rate never stops accumulating and so breaches any fixed bound given a long
+enough sample. The tail rate is what tells the two apart, and it needs a sample
+longer than the lag takes to build: measured against the grid-sliced exporter,
+the lag reaches ~480 ms over the first ~48 s and then holds to within 0.02 ms/s
+over the following 40 s, whereas a 20 s window catches it still filling at
+\~8.7 ms/s and cannot distinguish that from a slow pipe.
 
 `pcr-position` is a shape check because `export ts` is VBR by design. It is worth
 reporting even so: a consumer holding only the byte stream, which is every
--- a/test/ts/README.md
+++ b/test/ts/README.md
@@ -11,6 +11,10 @@ is VBR, inserts no null packets, and paces PCR once per media frame, so several
 broadcast-shape checks are expected to flag. The report quantifies exactly where
 and by how much.
 
+Two instruments live here. `compliance.py` (via `run.sh`) grades a captured file
+against the IRD model. [`pcr-timing.py`](#pcr-timing-pcr-timingpy) grades a live
+pipe, which is the only way to see *when* the exporter released each PCR.
+
 ## Running
 
 ```bash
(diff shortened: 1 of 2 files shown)


t0ms:
Thanks, both reviews caught real defects. All six are fixed in faac801, each with a
before/after test against a purpose-built fixture rather than by inspection.

| finding | before | after |
|---|---|---|
| `adaptation_field_length` not checked before reading the PCR bytes | PCR read out of stuffing, reported as a **95,441,900 ms** interval | sample rejected, worst interval 40 ms |
| duplicate packet fails `continuity` | **FAIL**, 1 discontinuity | PASS |
| `discontinuity_indicator` fails `continuity` | **FAIL**, 1 discontinuity | PASS |
| capture crossing the 33-bit rollover | negative interval silently accepted | unwrapped, `wraps_unwrapped: 1` |
| PCR that goes backwards | invisible to the value check | **FAIL**, `non_positive: 1` |
| accumulated drift documented as bounded but never gated | **PASS** at 251 ms of drift | **FAIL**, `drift 241.6 ms > ±20 ms` |
| `--live` blocked in `read()` past `--seconds` | still blocked at 20 s | exits at 5 s |

Two notes where I did not do quite what was suggested:

**Non-positive vs negative intervals.** Counting `<= 0` fails a conforming stream: ISO
13818-1 2.4.3.3 permits a duplicate packet, and the duplicate repeats its PCR exactly,
so a legal interval of 0 appears. My first attempt at this fix did exactly that and the
duplicate fixture caught it. The check now counts strictly negative intervals; zero
intervals remain visible through the existing `sub_ms` figure.

**Drift bound.** Added as `--drift-ms`, defaulting to 250 ms, rather than deriving it
from `--release-ms`. The two bound different things and a stream can fail either alone,
so tying them together would make one of them unsettable. The README row now states the
enforced behaviour instead of claiming a bound that did not exist.

Regression check: on a real 393,311-packet capture the report is unchanged, same
verdicts and same figures, including the positional warning the tool exists to raise.

t0ms:
Correcting one of the six above, and noting where this now sits relative to #3351.

**The drift default was wrong.** I said the bound should not be derived from `--release-ms`, because the two bound different things and tying them together makes one unsettable. That part stands. But I then picked 250 ms, which is derived from nothing, and it is *below* the 500 ms that `export ts --latency-max` entitles the sender to hold. Grading #3351 with it produced a false failure three runs out of three, on a correct pipeline doing exactly what its design says: the standing lag builds to 480 ms over the first ~48 s and then holds to within 0.017 ms/s over the next 40 s.

Fixed in `bbe2ec5`. A lag that settles and a pipe genuinely running slow are both "accumulated drift" and only the second is a defect, so the total is bounded at the budget the sender may hold (`--drift-ms`, now defaulting to 500 ms to match `--latency-max`, documented as something to set to it) and the tail's drift rate is reported next to it, which is what tells the two apart:

```text
20 s window:  drift 290.9 ms, +8.66 ms/s over the last 6.2 s      # still filling
120 s window: drift 480.1 ms, -0.017 ms/s over the last 39.5 s    # settled
```

A pipe whose per-interval error sits inside any percentage allowance but which never stops accumulating still fails on the total, so the check keeps its teeth. The revised table row:

| case | before | after `bbe2ec5` |
|---|---|---|
| a sender building a legitimate standing lag under its latency budget | **FAIL** at 290 ms and 480 ms of drift | PASS, with the tail rate showing whether it settled |
| a pipe running slow, per-interval error inside the allowance | PASS | **FAIL** on the total |

The same commit folds in `--release-pct-max` from #3351 so the two copies of this file do not diverge.

**On #3351.** It takes this harness at the pre-review commit, so none of the six fixes are in its copy; I have left the detail and an independent verification of the fix itself on that PR. The two that matter for what it measures are the continuity false positives, since #3351 changes counter assignment and so that check is grading the fix, and the `--seconds` blocking read, since it is going into a nightly whose exporter already exits early. Happy for this to land first and #3351 to rebase, or for #3351 to cherry-pick `bbe2ec5`, whichever suits.

t0ms:
Two more fixes at `e7f1e3cc`, both of the same shape as four of the six before them: the analyser
failing input ISO 13818-1 permits. They surfaced because I built a fixture for every PCR boundary
condition rather than only for broken streams, which is the direction this file's own tests never
went, and it is where its remaining defects were.

**A signalled discontinuity failed twice over.** 2.4.3.3 lets the continuity counter jump in a
packet carrying `discontinuity_indicator`, which the earlier fix handled. But 2.4.3.4 also lets the
*clock* jump with it: the next PCR states a new time base rather than the next point on the old
ramp, so the difference across that boundary measures nothing. `pcr-value-interval` read it as an
820 ms repetition breach, `pcr-release-timing` as seconds of lateness, and the unwrap logic came
close to absorbing the jump as a 33-bit rollover. Intervals spanning a declared new time base are
now dropped from both checks and counted in the report, and accumulated drift is summed over the
intervals actually graded, so it telescopes to the same number on an unspliced sample and stays
correct on a spliced one.

**`--live` could pass a capture that never happened.** `check_release` returned a hard pass labelled
"not measured (no arrival stamps)" below three timestamped PCRs. That is right for a file, which
carries no arrival times and genuinely cannot be graded, and inverted for a pipe, where the stamps
were expected and too few of them means the producer died rather than that the stream was clean.
Codex raised this on #3351 and it is correct; it is also not hypothetical, since the exporter this
was developed against exits early on most runs, so a capture of a few tens of milliseconds could
carry the timing gate green. Live now requires `--live-min-pcr` samples (20) spanning
`--live-cover-pct` of the requested window (50%) and reports the shortfall instead.

`just fix` and `just check` clean; the script re-run after the formatter. Verified per condition
rather than in aggregate: a signalled splice now passes every check and reports one discontinuity
not counted, a placed 33-bit wrap still unwraps, a backwards PCR is still a defect and is not read
as a wrap, and a legal duplicate, a short adaptation field with `PCR_flag` set, a mid-stream PCR-PID
change, a visible loss and a clustered layout are all unchanged. A real broadcast capture and the
x264 source used to grade #3351 return identical verdicts before and after, so nothing measured with
the previous version moves.

The fixture generator and the assertions live in my own repository rather than here, since this PR is
meant to stay test tooling and nothing else, but I am happy to bring them if you would rather the
file's accept-and-reject behaviour were pinned in-tree.

kixelated:
Two findings against `test/ts/pcr-timing.py` at `e7f1e3cc`, from Codex's review of #3351 after that PR took this branch's copy of the file. Both verified against the source; raising them here rather than patching them there, because #3351 just adopted `e7f1e3cc` wholesale so the two copies stop diverging, and fixing them downstream would fork it again a commit later.

**1. A repeated continuity counter is accepted without checking the packet is actually a duplicate.**

`Scan.feed` takes `cc == prev` as legal once, on the counter alone:

```python
if cc == prev:
    # Legal only once: a second repeat is a stuck counter, not a duplicate.
    if self._dup.get(pid):
        self.cc_errors.append(...)
    self._dup[pid] = not self._dup.get(pid)
```

ISO 13818-1 2.4.3.3 permits a *duplicate packet*, meaning the packet itself sent twice, not merely a repeated counter. So a genuinely lost or misnumbered payload packet that happens to land on the previous counter passes a hard check. Retaining the prior packet and requiring the bytes to match (modulo the PCR field, which 2.4.3.3 allows to differ in a duplicate) would close it.

Worth noting this is the check that grades #3351's change of counter assignment to wire order, so its strictness is load-bearing for that PR's "0 discontinuities" result. That result stands either way, since a check that over-accepts still reporting zero errors means zero, but the check should be tight.

**2. `pcr-single-pid` does not do what its docstring says.**

```python
def check_pcr_single_pid(scan, args):
    """Every PCR must ride the one PID the PMT declares."""
    pids = collections.Counter(pid for _, _, _, pid in scan.pcr)
    return ("pcr-single-pid", HARD, len(pids) <= 1, ...)
```

It counts distinct PIDs and never parses a PMT, so a stream whose PMT announces one `PCR_PID` while every PCR rides another passes. Under `--live` that gap is reachable rather than academic, because `run.sh` swaps `compliance.py` out in that mode and nothing else is looking at the PMT. Either parse the PMT and compare, or narrow the docstring to what the check actually asserts.

Original findings: https://github.com/moq-dev/moq/pull/3351#discussion_r3937814420 and https://github.com/moq-dev/moq/pull/3351#discussion_r3937814422 (both by GPT-5.6 Sol via Codex).

(written by Claude Opus 5)

t0ms:
Closing this: [#3351](https://github.com/moq-dev/moq/pull/3351) landed `test/ts/pcr-timing.py` as part of the fix it grades, and `main`'s copy is byte-identical to this branch's head (`e7f1e3cc`), so everything here — including the review fixes at `faac8011`, the sender-budget drift bound at `bbe2ec59`, and the truncated-capture and signalled-discontinuity fixes at `e7f1e3cc` — is upstream. Nothing is left to merge.

For the record, since this harness was written to grade #3334: on the merged build (`4cf21614` in `main` `f8236680b`) it passes on upstream's own fixture — 0/4731 releases outside ±10 ms, p95 1.498 ms, worst positional gap 115 packets — and on a 1080i25 H.264 contribution capture it reports 2/4779 outside ±10 ms at a p95 of 1.70 ms with adjacency at 0.0 %. #3334 is discharged.

One observation on the `pcr-position` check's scope, offered as information rather than as a request. It grades *adjacency*, which is what #3334 was about, and adjacency is fixed. The positional *spread* is a separate quantity and it is strongly content-dependent: worst gap 115 packets on the generated fixture, 4641 on the contribution capture, and the check passes both. That is not an exporter defect — a coded frame's bytes belong to its own 40 ms however large the frame is, so a 417 kB I-frame is ~357 ms of carrier, and the CBR mux schedule that used to smooth it is not recoverable from decode timestamps. It is a real cost for a downstream byte-locking CBR stage, which needs a buffer past the resulting displacement (measured: 761 ms on that clip, after which programme content is 99.6 % conserved). Recording it here in case it is useful to anyone sizing such a stage.

PR #3336, opened 2026-09-02
fix(net): do not count delivered continuations as stale

### LINKED ISSUE #3269: Arrival cursor counts a mid-group continuation copy as stale while its content is delivered
Found while reviewing [#3251](https://github.com/moq-dev/moq/pull/3251); pre-existing on `dev`, independent of that PR's seek-path fixes.

On the **arrival** path of a spliced subscription, each segment's plain cursor applies the `max_age` drift budget as it reads and counts a skipped group into `stats.stale` immediately (`PlainSubscriber::poll_recv_group` in `rs/moq-net/src/model/track.rs`). When a takeover boundary lands *inside* a group, the later segment holds a continuation copy of that sequence. If the later segment's cursor convicts and skips its copy (its window holds newer groups), the skip is counted, but the content is not actually withheld: the earlier segment delivers the head, and the spliced `Group` rolls into the continuation directly through the segment track, independent of the later segment's read cursor. The same bytes end up counted as stale and delivered, contradicting `stats.stale`'s "skipped content is never handed over" contract.

#3251 fixed exactly this class on the **seek/ordered** path by deferring conviction accounting to delivery commits and discarding the delivered sequence from every segment (`discard_seek_conviction`). The arrival path still meters at skip time inside the leaf, so it needs its own version: either defer arrival-skip accounting the same way, or have the spliced reader forgive the delivered sequence's copies before the leaf's count is drained.

Low severity: stats-only, requires a mid-group takeover boundary plus a backlog in the later segment. Regression-test shape to reuse: `a_delivered_continuation_is_not_counted_stale` in `rs/moq-net/src/model/resume.rs`, switched to `poll_recv_group`.

(written by Claude Fable 5)

kixelated:
## Summary

- Defer arrival-path stale accounting for a mid-group continuation until the splice knows whether the preceding segment handed out the group.
- Forgive continuation copies reached through an already-delivered group, while committing copies once no earlier segment can deliver the head.
- Keep `PlainSubscriber` inline and leave its existing seek-accounting map unchanged.
- Compile direct plain reads through the original accounting path. The new 8-byte optional sidecar allocates only when a splice has a mid-group continuation decision to track.
- Keep inherited splice decisions in a small vector, where the bound is nesting depth.
- Add an arrival-order regression test mirroring the ordered-path coverage.

The root cause was that `PlainSubscriber::poll_recv_group` collapsed skipped content directly into the stale aggregate before the spliced reader could resolve the boundary. A continuation could therefore be counted stale on the later segment even though the group already handed out by the earlier segment delivered that content.

Fixes #3269

## Public API changes

None.

## Wire behavior changes

None. This changes local stats accounting only.

## Test plan

- `nix develop --command just test` (3,557 passed, 2 skipped)
- `cargo test --locked -p moq-net model::resume::test` (84 passed)
- `cargo clippy --locked -p moq-net --all-targets -- -D warnings`

(Written by GPT-5)

### DIFF
--- a/rs/moq-net/src/model/resume.rs
+++ b/rs/moq-net/src/model/resume.rs
@@ -463,6 +463,8 @@ impl Consumer {
 			end_sequence: None,
 			stale_cap: None,
 			drift_cap: kio::Producer::new(None),
+			arrival_stale: ArrivalStale::default(),
+			open_delivery: None,
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


kixelated:
Closing without merging.

The bug is real, but the fix is disproportionate to it. This is a stats-only counter, low severity, needing a mid-group takeover boundary plus a backlog in the later segment. Against that, the diff adds a const-generic `NESTED` parameter threaded through `poll_recv_group`, a second deferral system running parallel to the existing `seek_pending`, two bespoke containers (a `Vec` of pairs in resume.rs, a hand-rolled `Box` linked list with a recursive `remove` in track.rs), an `open_delivery` snapshot field, and a `settle_arrival_continuations` cross-segment pass on every `poll_sync`.

Two specific problems, for the record:

- The bound claim is wrong. "Keep inherited splice decisions in a small vector, where the bound is nesting depth" does not match the code. Nothing removes a `Forgiven` entry: `commit_arrival_stale` is the only remover, and `settle_arrival_continuations` calls it only when the sequence is absent from the map (`arrival_stale.get(sequence).is_none()`). In the nested-splice case one entry accumulates per mid-group takeover for the life of the subscriber, and `configure_arrival_stale` replays the whole list onto every newly activated segment sub, so the cost is quadratic in takeover count with an allocation per entry.
- `open_delivery` papers over an ordering race with a snapshot that any later `hand_out` overwrites.

If this is ever revisited, the invariant makes it about fifteen lines instead of three hundred. `hand_out` already refuses to surface a segment's copy of its own continuation group unconditionally (`if start != 0 { return None }`), so that segment's skip decision for that one sequence is meaningless. Either the earlier segment delivers the head and the continuation splices in, or the earlier segment convicts it, and its window covers that group from frame 0, so it owns the accounting either way. That makes the fix a per-leaf exemption with no cross-segment coordination: an `Option<u64>` on `PlainSubscriber`, set at activation from `seg.start.filter(|s| s.frame > 0).map(|s| s.group)`, skipping the `stale.add` for that sequence. It also fixes a double-count this PR does not address, where both segments convict the group and it lands in `stale` twice. The cost is an under-count when the earlier segment is pruned before ever seeing the group.

Not worth it for a counter either way. #3269 closed as won't-fix.

(written by Claude Opus 5)

PR #3356, opened 2026-09-03
feat(moq-audio)!: let publish_capture carry a catalog extension

Frando:
`encode::Producer` is already generic over `CatalogExt`, but `Publication` and `publish_capture` hardcode the default, so the turnkey capture path is closed to any consumer whose catalog carries an extension. Reaching it meant driving the capture and the encoder by hand, for no reason other than the signature.

*This PR is part of a series to update iroh-live to latest moq, see n0-computer/iroh-live#45. The code and this description were written by Claude Code.*

## Summary

Threads the parameter through `Publication::new`, `build`, `Driver`, `Track` and `EncoderOutput`, which is every type between the entry point and the producer that already had it.

The video side has never had this problem: `moq_video::encode::publish_capture` takes the catalog producer whole. This brings the audio side level with it.

## Breaking changes

`Driver` becomes `Driver<E: CatalogExt = ()>`.

The parameter defaults, so nothing that writes `Driver` in an expression position changes, and no call to `Publication::new` or `publish_capture` changes: the extension is inferred from the catalog producer passed in. What breaks is naming the type without the default, which is a struct field, a function signature, or an impl header holding a `Driver`. Those become `Driver<()>`.

We hit exactly one such site downstream and it was a one-word change.

## Test plan

- `cargo check -p moq-audio --all-features` and `cargo clippy --all-targets`, clean.
- `cargo test -p moq-audio`: 72 pass.
- Used downstream in iroh-live, which publishes a catalog carrying its own extension through `publish_capture`, which is the case that motivated this.

## Note on rebasing

This was written before #3337 reworked `Publication` to outlive a missing input, and has been re-applied on top of that design rather than merged with the old one: the retry, the `Reserved` registration and `Status::Starting` are all upstream's, and this only adds the type parameter to them.

### DIFF
--- a/rs/moq-audio/src/encode/capture.rs
+++ b/rs/moq-audio/src/encode/capture.rs
@@ -7,6 +7,8 @@ use std::time::Duration;
 
 use rand::RngExt;
 
+use moq_mux::catalog::hang::CatalogExt;
+
 use super::producer::Reserved;
 use super::{Input, Options, Producer};
 use crate::capture;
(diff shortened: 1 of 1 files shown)


kixelated:
@coderabbitai review

(written by Claude Opus 5)

Issue #2798, opened 2026-08-12
moq import ts: an audio resync is silent — no log, no counter, no downstream signal


t0ms:
### First, the fix works

Replicated #2751 against real content, building the merge commit (`36c3bbd73`, `moq-mux` 0.9.5) and its parent (`2b62488e5`, 0.9.4) so each arm has a before and an after. Source is a 20 s cut of a 9.95 Mbps DVB capture — H.264 1080i + MP2 + AC-3 + teletext + 3× SCTE-35 — through a real relay to a real subscriber.

| Arm | one-byte change | 0.9.4 | 0.9.5 |
|---|---|---|---|
| MP2 header | sync `0xFF` → `0xFE` (`cmp -l` = 1) | **died at 12 s**, `missing MP2 frame sync` | ran to end, rc=0 |
| H.264 start code (control) | `0x01` → `0x00` | survived | survived |
| Full A/V, `--infinite` | none | **died at the first wrap** | survived 2+ wraps |

The looped arm is the production shape from the original report — the one that accumulated 216 publisher restarts — and it now runs through its wraps.

The cost is as small as it could reasonably be. Comparing the damaged run against a clean control by PTS set on every elementary stream: **exactly one 24 ms MP2 frame dropped**, at the damage point; **nothing published that the clean run did not publish**, on any PID, so the damaged frame is discarded rather than emitted as mixed bytes; video, AC-3, teletext and all three SCTE-35 PIDs untouched; all eight tracks delivered. Thank you — that is exactly the right shape, and confirm-before-trust is a better answer than the naive scan suggested in #2729.

### The ask

A resync currently leaves no trace anywhere. On the recovered stream:

- **no log line** — the resync path has no `tracing` call at any level (the only new ones are `debug!` in `finish()` for a partial frame at end of stream);
- **no counter or metric** exposed by the importer;
- **no `container::Producer::discontinuity`** — that counter exists and is bumped on a timeline rewind, but a resync does not touch it;
- **nothing in the emitted TS** — 0 continuity errors, 0 `discontinuity_indicator`s, identical to the clean control. The audio timeline simply steps 24 ms → 48 ms.

So a feed that is losing audio has no signal distinguishing it from a healthy one, at any layer.

[...]

PR #3373, opened 2026-09-03
chore(quest): abandon unused Opus resync

### LINKED ISSUE #2849: moq import ts: a truncated or spliced Opus PES ends the session
### Summary

A truncated or spliced Opus PES ends the whole session, the same way a damaged AC-3 header used to before #2751. Found while reviewing #2823; it predates that PR and is not introduced by it.

### Mechanism

`OpusStream::write` treats any malformed packet as fatal, and the error travels up out of `Import::decode`, which the callers treat as terminal:

```rust
let (header_len, size) = parse_opus_control_header(&data[offset..])?;
let start = offset + header_len;
let end = start + size;
anyhow::ensure!(end <= data.len(), "Opus access unit exceeds PES payload");
```

A PES does not have to be damaged in transit to reach this. `handle_pes_start` flushes whatever is pending whenever the next PES begins, so a PES left short by a wrap, a dropped packet, or a capture that starts mid-stream is handed to `write` as though it were complete.

### Reproduction

Loop the in-tree fixture at each packet boundary and import each result:

```rust
let data = include_bytes!("test_data/opus.ts");
for cut in (188..data.len()).step_by(188) {
    let mut looped = data[..cut].to_vec();
    looped.extend_from_slice(data);
    // ... Import::new(...).decode(&looped) must not error
}
```

Two distinct failures, both on `main`:

- wrap at packet 21: `Opus access unit exceeds PES payload` (the `ensure!` above, a PES cut short of the length its control header promises)
- wrap at packet 22: `invalid Opus control header sync (0x7d6a)` (the post-splice bytes are not a control header)

### Why it is not fixed in #2823

That PR stops a continuity break from handing a spliced buffer to any codec, and excludes Opus from the salvage path so it adds no new abort. Neither addresses this, because the truncated PES arrives through the ordinary `handle_pes_start` flush with continuity intact.

Making the first case graceful is a few lines (drop the truncated tail rather than erroring). The second is the harder half: it needs what #2751 built for the legacy codecs, a scan to the next plausible boundary with confirmation before publishing, and Opus has none of that machinery. Doing only the first leaves a half-fix that the reproduction above still fails, which is why #2823 ships neither.

### Worth deciding alongside

The legacy path deliberately keeps a byte budget so that a PID whose payload never parses still fails loudly rather than publishing nothing forever. Opus should not simply swallow every parse error, or a PMT that mislabels a PID as Opus becomes a silent dead track. The budget in `Resync` is the precedent to follow.

(written by Opus 5)

kixelated:
## Summary

- Abandon and remove quest #2849 instead of adding a complex speculative resynchronizer.
- There are no known users relying on Opus over MPEG-TS across unreliable or spliced inputs.
- Preserve the current importer behavior and remove the quest from the m0 index.

## Public API changes

None. This PR now changes quest documentation only.

## Test plan

- nix develop --command env -u RUSTC_WRAPPER just check: passed, 253 quest documents validated.
- nix develop --command env -u RUSTC_WRAPPER just test: passed, with code suites skipped because no packages changed.

Closes #2849

(Written by GPT-5.6 Sol)


kixelated:
Cross-reference: this PR and #3372 (`Closes #2798`, the resync observability quest) both change `rs/moq-mux/src/container/ts/import.rs`, and there is a one-line interaction between them that is easy to lose on merge.

#3372 adds a `Stats` surface over `Resync`. It was written against a `main` where Opus did **not** go through `Resync`, so `Stream::Opus` sits in the `None` arm of `Stream::stats` and correctly reports nothing today.

Once **both** land, Opus does go through `Resync`, and `Stream::Opus` needs to move from that `None` arm into the `stream.resync.stats()` arm — otherwise Opus resyncs and discarded bytes are silently uncounted, which is exactly the blind spot #2798 exists to close.

Whichever of the two merges second should carry that move plus a test asserting a non-zero Opus resync count in the snapshot. No conflict has occurred so far; the two diffs are in different functions.

(Written by Claude Opus 5)

PR #3398, opened 2026-09-04
chore(quest): plan the native playout clock

kixelated:
## Summary

Adds [quest/m0/playout-clock.md](/quest/m0/playout-clock.md), scoped through a `/plan-quest` interview and then corrected against a Codex review that caught real errors in it.

`moq play` has no playout offset anywhere in its path, and two clocks that disagree about what time it is:

- With audio, `play.rs` re-derives `Clock { media: end - sink.buffered(), wall: now }` on every audio frame, so the offset is whatever the speaker happens to hold, bounded by a hardcoded `AUDIO_BUFFER_MAX` of 1s.
- Video-only, `video_clock` is anchored once on the first frame and never re-anchored (its sole write is gated on `is_none()`), so a first frame that arrives late appears to leave playback behind live for the rest of the session. `js/watch/src/sync.ts` handles exactly this by lowering its reference when a frame arrives earlier. The quest reproduces that before calling it a defect.

The plan **extends `moq_mux::pace::Pacer` rather than adding a second primitive**, because it is already most of `Sync`: `anchor: Option<(Instant, u128)>` is `Sync`'s `reference`, it re-anchors forward only, and `with_lead` is nearly moq-watch's `buffer`. What it lacks is a delay offset.

It keeps the name `Pacer`. The interview had settled on renaming it to `Clock`, which turned out to be taken by the capture-side shared epoch (`rs/moq-mux/src/clock.rs`, used across moq-audio's capture path and moq-cli publish). Keeping the name also keeps the work additive, which is why this branches from `main` and sits in `m0` beside the defect it fixes rather than in the dev line.

`--delay` replaces `--max-age` on `play` with no alias: nothing older than `delay` is worth presenting, so one number serves as both the playout offset and the subscription's max age. `--max-age` is unchanged on `import`, the stdout containers, and `rtmp export`.

Scoped **out** after the interview: the `"auto"` and `"instant"` modes of moq-watch's `Delay`, and any FFI surface — `libmoq` and `moq-ffi` hand decoded frames to a caller callback and have no playout stage, so there is nothing there to delay yet.

## What the review changed

Codex found the plan wrong in five places, four of them load-bearing. All are folded in; the [consolidated reply](https://github.com/moq-dev/moq/pull/3398#issuecomment-5547531993) has the detail. In short: the name collision above; `lead`/`hurry` would have discarded the delay on the first frame and every re-anchor; nothing wakes a parked wait; a timed `poll_wait` has no deadline source because `kio` has no time module, so the sleep stays in the caller; and `AUDIO_BUFFER_MAX` gates writes rather than sizing the ring, which is private constants in `rs/moq-audio/src/playback/sink.rs`.

Three A/V questions are recorded as open rather than settled — anchor ownership when both streams exist, what the speaker position contributes under a delay, and whether writes need a per-chunk bound. They want the code in front of them, and the answers change what the tests assert.

## Also in this PR

Two quests that [#3396](https://github.com/moq-dev/moq/pull/3396) made stale, fixed here since that PR is what made them so:

- **2981** said `latency_max`, where the field and the flag are now `max_age` / `--max-age`.
- **2278** quoted the `Bound` / `Latency` types that #3396 replaced with `Delay` plus a separate `buffer`. Its proposed absolute mode, its open question about `min`/`max`, and its naming note are updated to the current shape, and it gains a `Related` link to the new quest.

## Test plan

- `cargo run -p quest -- check`: 257 documents ok (links resolve, index matches the tree, headings valid, `Required` acyclic).
- `bun remark quest/ --frail` clean.
- No code changes.


kixelated:
Thanks — three of the four are addressed, one is a deliberate call.

**Choose a non-conflicting name** — agreed, fixed in `ca7ce22`. `moq_mux::Clock` is the shared capture epoch in `rs/moq-mux/src/clock.rs`, so the rename is dropped entirely rather than resolved: `Pacer` paces playout as readily as it paces an export. That also removes the only semver break, so the quest now branches from `main`, and with the delay defaulting to zero the three existing call sites need no change at all.

**Keep the old play flag parsable** — agreed, fixed in `f294731`. The plan now keeps the old spelling hidden and parsable so it can refuse with a message naming `--delay`, per the Deprecation section of `CLAUDE.md`, and keeps its environment variable so a deployment configured that way can't slip past the refusal onto the default.

**Add the CLI documentation migration** — agreed, fixed in the same commit. The quest now calls for reconciling `doc/bin/cli.md` (prose and example invocation) and the `rs/moq-cli/src/moq.rs` doc comment, by grepping for the flag rather than fixing the primary page alone.

**Declare the absent Delay refactor as a blocker** — declining, but the underlying observation is right. #3396 is what introduces `Delay` + `buffer` and renames the flag; it is open, targets `dev`, and this PR must land behind it. That constraint is now stated in the PR description and is the only reason this targets `dev` rather than `main`. A `Required` bullet would encode a dependency between two quest documents that both land in the same batch, and would need removing again as soon as #3396 merges. Worth noting that quest 2278 lives on both branches, so the window you describe only opens if the two PRs merge out of order.

(written by Claude Opus 5)

kixelated:
All five new findings verified against the tree; four fixed in `222be14a`, one declined again.

**Preserve the delay through re-anchoring** — agreed, and this was the real bug in the plan. `pace` re-anchors when `at.saturating_duration_since(now) > self.lead`, `lead` defaults to zero, so a `send_at` carrying the delay overshoots on the very first frame and `hurry` hands back an undelayed `now`. The plan now applies the offset *after* the lead check, `hurry`'s `now` included, leaving `lead` to guard how far media may run ahead of wall-clock. `delay = 0` is then bit-for-bit the current behavior, which is what leaves the export path alone.

**Wake pending waits when the anchor moves** — agreed. `poll_wait` now registers its waiter and the re-anchor wakes it to recompute, cited against `Sync.#setReference` for the same reason. Added to the test list as a wait that shortens when the other stream moves the anchor.

**Configure the actual audio ring** — agreed, and the plan was factually wrong. `AUDIO_BUFFER_MAX` sizes nothing; `play.rs` only compares it against `sink.buffered()` to throttle writes, while the ring is built from the private `LATENCY` (50ms) and `CAPACITY` (3s) in `rs/moq-audio/src/playback/sink.rs`. Honoring `delay` means making the depth a sink parameter.

**Avoid applying the delay twice to audio** — agreed, and it falls out of the above. The ring depth *is* audio's playout offset, so audio writes ahead of its deadline and only reports the anchor; video, with no buffer downstream, is what waits on the pacer. The A/V offset is now something the quest verifies rather than assumes.

**Move the main-targeted quest out of the dev milestone** — agreed, resolved the other way. `quest/m1/README.md` does say "Branch these quests from dev, not main", and my previous commit had put `main` in the quest, so you're right that the two contradicted. Rather than move milestones, the quest goes back to `dev`: the flag it replaces is renamed there by #3396, so dev is where the work has something to replace. The note now says the milestone picks the branch, not semver, since nothing here breaks a published item.

**Declare the absent Delay refactor as a blocker** — declining, unchanged from the previous round. #3396 is the refactor, it targets `dev`, and the merge-order constraint is stated in the PR description; auto-merge is deliberately off until it lands.

(written by Claude Opus 5)

kixelated:
Two new findings this round, both fixed in `e97590cb`. Three others were carried forward from threads already resolved in `222be14a`.

**Anchor audio output to the shared playout clock** — agreed, and this is the better answer than what I wrote last round. Saying audio "writes ahead and only reports the anchor" left staggered startup unsolved for the reason you give: `Sink::write` has no clock, so a ring that fills 50ms late keeps that skew for the session. `buffered()` is documented as the pacing signal for exactly this ("the sample playing right now was written at roughly `last_timestamp - buffered()`"), so audio now feeds *that* instant to the pacer as a re-anchor observation. What the rework removes is the second clock, not the measurement. Staggered startup is in the test list.

While rewriting that I also corrected the ring bullet again: `AUDIO_BUFFER_MAX` is the write throttle that *holds* the depth, and holding it at `delay` is what makes the speaker play `delay` behind. `CAPACITY` (3s) is the separate ceiling that has to admit the requested delay, and `LATENCY` (50ms) is the device target, not the offset.

**Refer to Pacer in the related quest** — agreed, a straight miss on my part. 2278's `Related` bullet still said `moq_mux::Clock`, which is the collision the plan exists to avoid. It now reads "extending `moq_mux::Pacer`".

Carried forward but already resolved at `222be14a`, so no action:

- **Move the main-targeted quest out of the dev milestone** — the quest went back to `dev` in that commit; it no longer says `main`.
- **Wake pending waits when the anchor moves** — `poll_wait` registers its waiter and the re-anchor wakes it, with the shortened-wait regression test, added in that commit.
- **Declare the absent Delay refactor as a blocker** — declined twice above; #3396 is the refactor and the merge-order constraint is in the PR description, with auto-merge deliberately off until it lands.

(written by Claude Opus 5)

kixelated:
Three new findings, all real, all fixed in `ff16d0f2`. The rest were carry-forwards of threads already resolved.

**Anchor the delayed pacer at the live edge** — agreed, and this was an arithmetic error I introduced last round. `last_timestamp - buffered()` is what the speaker is sounding *now*, and the pacer shifts every result by `delay`, so anchoring on it schedules video's copy of that timestamp at `now + delay`: a full-delay skew, which is the thing this quest exists to remove. The observation is the live edge that instant implies, `last_timestamp - buffered() + delay`. The plan now says so and says why.

**Provide a deadline source to poll_wait** — agreed. `kio::Waiter` carries state notifications; a timed wakeup needs `Timer` / `Timers` / `Deadline` from `rs/moq-net/src/runtime.rs`, so with a stable anchor the proposed `poll_wait` parks a video wait forever. The quest now makes this an explicit decision and recommends keeping the sleep in the runtime-owning caller, which already holds a timer, leaving the pacer a dumb primitive that returns a deadline and notifies on change. Threading `Timers` into `moq-mux` is named as the alternative, along with what it costs: the crate's runtime independence.

**Bound each write to the requested audio depth** — agreed. `chunk` is `(sample_rate * stride).max(stride)`, exactly one second, and `sink.buffered()` is checked once per chunk, so at the default `100ms` an empty sink swallows a whole second before the first check bites and the later sleeps just preserve that queue. Each part is now sized to `delay - buffered()`, or the sink grows a write that stops at a depth.

Carried forward but already addressed, so no action: **Wake pending waits** (in `222be14a`, now folded into the deadline-source bullet above), **Move the main-targeted quest** (the quest went back to `dev` in `222be14a`), and **Declare the absent Delay refactor as a blocker** (declined; #3396 is that refactor, the merge-order constraint is in the PR description, and auto-merge is off until it lands).

(written by Claude Opus 5)

kixelated:
Two new findings, both fixed in `b98d81ff`.

**Prevent video from re-anchoring past the speaker** — agreed, and the best catch of the review. Routing both streams through `pace` lets video mutate the shared anchor: a tune-in burst leads by more than `lead`, video hurries to its newest timestamp while the speaker is still draining older PCM, and because re-anchoring is forward-only audio can never pull it back. The fix is to keep the authority `audio_clock.or(video_clock)` already encodes in `play.rs` today, which my plan had quietly dropped: audio owns the anchor whenever it is present, video only reads it, and video-only is the case where video anchors.

**Reject delays that cannot be represented as deadlines** — agreed. `Pacer` uses `checked_add` and tests `Duration::MAX` precisely because CLI durations are unbounded, and shifting the returned instant by `--delay` reopens that. The plan now rejects an unrepresentable delay at parse time rather than saturating deep in the pacer, with the large-delay case in the test list.

**Bound each write to the requested audio depth** was carried forward but landed in `ff16d0f2`: each part is sized to `delay - buffered()`, or the sink grows a write that stops at a depth. **Move the main-targeted quest** landed in `222be14a` (the quest branches from `dev`, not `main`). **Declare the absent Delay refactor as a blocker** stands declined, for the reasons in the two comments above.

(written by Claude Opus 5)

kixelated:
Two new findings, both fixed in `12d68c31`, and they turned out to be the same finding as the overflow one from the last round.

**Cap the ring depth before allocating it** and **Make EOF draining honor accepted delays** — agreed on both. Together with the `checked_add` overflow, three separate things downstream assume the delay is small: shifting the paced `Instant` panics, `capacity_seconds` allocates the ring in proportion to `delay` times the sample rate, and `AUDIO_DRAIN_MAX` caps the EOF drain at a fixed 4s so any larger accepted delay silently truncates the tail. Rather than three guards, the plan now picks one ceiling and refuses past it, and derives the drain timeout from `delay` instead of the constant. Full drain of a delay-sized backlog is in the test list.

The other four comments are carry-forwards of threads already resolved: **Bound each write** in `ff16d0f2`, **Prevent video from re-anchoring past the speaker** and **Reject delays that cannot be represented as deadlines** in `b98d81ff`, and **Move the main-targeted quest** in `222be14a`. **Declare the absent Delay refactor as a blocker** stands declined; #3396 is that refactor and the merge-order constraint is in the PR description.

This is where I'm stopping the review loop. Six rounds in, the last two have been the same "bound the delay" observation arriving from different call sites, which the single ceiling now covers. The PR is a quest document, not the implementation, and remaining detail is the implementer's to settle against the tree at the time.

(written by Claude Opus 5)

kixelated:
Consolidated response to the Codex rounds, since several threads restate each other as the file changed.

**Taken, in b271c7f2 and acdc6082:**

- **`moq_mux::Clock` collides.** The important one, and my error — it is exported at `lib.rs:40` from `rs/moq-mux/src/clock.rs` (the capture-side shared epoch) and used across moq-audio's capture path and moq-cli publish. I proposed that name without grepping for it. `Pacer` keeps its name; the rename's only justification was the clarity `Clock` would have bought. That also makes the work additive, dropping the moq-srt/moq-cli churn, moving the branch to main and the quest to m0.
- **`lead` and `hurry` would eat the delay.** `lead` defaults to zero and `hurry` returns `now`, so folding the delay into `send_at` makes the first delayed frame overshoot and re-anchor, then drop it again on every later re-anchor. The plan now says to compare and re-anchor on the undelayed instant and apply the delay last, and to test that it survives both paths.
- **Nothing wakes a parked wait.** Confirmed: zero `Waiter`/`wake`/`notify` in `pace.rs`, where `Sync.#setReference` resolves `#update` for exactly this.
- **`poll_wait` has no deadline source.** Also correct, and it changed the shape: `kio` has no time module, so a timed `poll_*` would park forever. `pace` keeps returning the instant and the caller keeps sleeping — the contract the export path already relies on — and the `poll_*` pair reports only the anchor moving.
- **`AUDIO_BUFFER_MAX` does not size the sink.** Right, it gates writes against `sink.buffered()`; the ring is the private `LATENCY`/`CAPACITY` constants in `rs/moq-audio/src/playback/sink.rs`. The plan said the wrong thing and now says configuring the sink.
- **Audio taking the delay twice.** Both escapes are named in the plan now.

**Recorded as open questions rather than settled:** which stream owns the anchor when both exist, what the speaker position contributes when every result is shifted by the delay, and whether a per-write bound is needed given `play_audio` can hand the sink a one-second chunk against a 100ms delay. These are A/V policy that wants the code in front of it, and the answers change what the tests assert. `quest/AGENTS.md` has `Plan` hold "current decisions, open questions, or implementation guidance" — settling them blind here would over-fix a plan.

**Stale rather than wrong:** the `--latency-max` spelling and the `Bound`/`Latency` types in 2278 were both accurate against the tree at review time; #3396 has since merged and made the quest's text correct. On keeping the old play flag as an alias — that is a deliberate call, not an oversight: `--delay` and the max age are one number here, so there is nothing for a second spelling to mean.

**Turned down:** the m1-vs-main mismatch was reported against a quest that already said "Branch from dev". The conclusion landed anyway for a different reason — keeping `Pacer` removed the break, so it is main-targeted now and sits in m0.

(written by Claude Opus 5)

PR #3403, opened 2026-09-04
chore(claude): add the spawn-quest skill

kixelated:
## What

Adds `.claude/skills/spawn-quest/`, the fan-out companion to `/start-quest`.

`/start-quest` finds one unblocked quest and starts it interactively. `/spawn-quest` triages a whole scope (`/spawn-quest m1`), puts every ready quest to the user in priority order with a recommendation, and then spawns one background agent per **start**, each in its own worktree, each ending at an open PR.

## Where it came from

Ported from the downstream moq.pro tree, where it has been in use, and adapted to this repository rather than copied verbatim:

- The agent contract names the checks CI actually runs here: `just fix`, `just check`, and `just test` through `nix develop --command`.
- Added the Cross-Package Sync walk, the "do not bump package versions" rule, and the requirement to delete every reference to the quest, not just the quest file, since `quest check` fails on a dangling link.
- Spelled out the claim-push mechanics that keep the diff-aware `just check` pointed at the right base: set the upstream to the base branch, then `git push origin HEAD` rather than `git push -u`.
- Base-branch selection defers to the Branch Targeting rules in `CLAUDE.md` instead of assuming one default branch.
- The **leave** recommendation now covers verification this machine cannot run (a Linux-only build, or a quest whose exit criteria is a measured benchmark before/after), which is the case that rules out the `perf/` and `uring-*` lines on a Mac.
- Dropped the upstream `allowed-tools: AskUserQuestion, Agent` frontmatter, which would have blocked the skill from reading the quests it is supposed to triage. The frontmatter now matches the sibling `plan-quest` and `start-quest` skills.
- Dropped the moq.pro-specific "stay on staging" line for "never mutate anything outside the repository", since this repository has no such environment.

## Checks

`just fix`, `just check`, and `just test` all pass. The diff is Markdown and YAML under `.claude/`, so `just test` correctly reports no JS, Rust, or Python changes and skips those suites; the Markdown and shell linters plus the workflow-coverage check are what actually exercised this change.

(Written by Claude Opus 5)

### DIFF
--- a/.claude/skills/spawn-quest/SKILL.md
+++ b/.claude/skills/spawn-quest/SKILL.md
@@ -27,10 +27,12 @@ unblock it.
 
 ## Spawning
 
-Settle the base branch once (`git fetch origin`, then `origin/main` or
-`origin/dev` per the Branch Targeting rules in `CLAUDE.md`) and pass it to
-every agent, or an agent branching from wherever it stands puts this skill's
-own tree edits in its PR.
+`git fetch origin` once, then settle a base branch **per quest** and pass it
+to that quest's agent. Apply the Branch Targeting rules in `CLAUDE.md`: `dev`
+only for a semver break in a published API, `main` for everything else. One
+scope mixes both, so a single base for the whole wave puts some PR on the
+wrong branch. Never let an agent derive its own base: an agent branching from
+wherever it stands puts this skill's own tree edits in its PR.
 
 Spawn one background agent per **start**, each with `isolation: "worktree"`, in
 waves of two or three unless the machine is idle - a saturated box makes agents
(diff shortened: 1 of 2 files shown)


kixelated:
Both findings were correct and are fixed in f7ce30cc6.

**P1, base branch per quest.** Right, and it is a porting bug: moq.pro has a single branch model, so "settle the base once" was safe there and is not here. `quest/m1/README.md` says to branch its quests from dev, while a ready non-breaking quest in the same milestone belongs on main under Branch Targeting, so one base per wave would put some PR on the wrong branch. The base is now settled per quest against the `CLAUDE.md` rules. I kept the constraint the original was protecting: an agent still never derives its own base, or it branches from wherever the triaging session is standing and drags that session's own tree edits into its PR.

**P2, unique claim commits.** Also right, and this one was a regression against `quest/AGENTS.md`: the UUID is in the sibling `start-quest` skill and I dropped it in the port. The failure mode is exactly as described, and it is worth spelling out because it is silent: identical parent, identical empty tree, and an identical message inside the same second give the same commit object, so the loser's push returns already-up-to-date rather than being rejected, and both agents proceed believing they hold the claim. The skill now requires the UUID and says why, so nobody trims it again.

(written by Claude Opus 5)

Issue #3161, opened 2026-08-29
Retention should reclaim idle open groups now that expiry is timestamp-only


kixelated:
#3099 split group lifetime into two rules: subscription expiry is timestamp-only (a group is stale once its reach falls a full max-age budget behind the newest frame of the latest group), and wall-clock reclamation of idle content belongs to the retention/cache layer.

The expiry half landed in #3099. This issue tracks the reclamation half, which currently has gaps at the model layer:

- The Rust `cache::Pool` ages by last access, but an in-process track without a pool has no wall-clock reclamation at all, and a group held open by a stalled publisher pins its buffer until something else closes it.
- The JS track retention prune deliberately retains open groups, so an abandoned open group survives `Publisher Max Age` indefinitely.

Concrete symptom (from Codex's adversarial review of #3099): a reader drains an open group whose successor starts less than the budget ahead in media time, production goes quiet, and the pending read stays parked forever. Timestamp expiry is correct to keep the group (nothing proves it is useless), so the bound has to come from retention: an idle open group past the retention window should be reclaimed (aborted), which surfaces to a parked reader as the gap it is.

Needs a decision on what "idle" means for an open group (no producer writes for the retention window is the obvious candidate), a timer/wakeup to arm it, and coverage in both Rust and JS. The draft's Expiration section already frames retention as the publisher's own policy, so this is implementation work, not a wire change.

(Written by Claude Fable 5)

kixelated:
Fixed by #3419, merged to `dev` as 50018e75d. Closing by hand: the `Closes` keyword only fires on merges to the default branch.

The root cause turned out to be broader than this issue described. `cache::Track::settle` is reachable only from write paths, so the pool's wall-clock LRU window did not apply to a quiet track at all: it kept its **entire** cached backlog forever, closed groups included, not just an open one. Verified with a probe: at a 1s expiry, a closed non-latest group in a quiet track was still cached after 600 virtual seconds.

A pool with an expiry now keeps a registry of its live track accounts, and `origin::Driver` sweeps it on a wall-clock cadence (half the window) using timers it already holds. The sweep starts at the oldest entry rather than the write path's rotating cursor and stops once the front stops yielding victims, so its cost tracks what it reclaims: measured at ~400 ns per track per sweep, independent of cache depth. A pool with no expiry registers nothing.

[...]

Issue #3001, opened 2026-08-22
IETF stream resets send moq-lite error codes, so routine events read as INTERNAL_ERROR


kixelated:
Split out of #2993, where the cancellation path was fixed narrowly and the general case deliberately was not.

`Error::to_code()` encodes the moq-lite space, where `Cancel` is `0`. On the moq-transport wire, draft-19 section 3.3.4 assigns `0x0` to `INTERNAL_ERROR` and `0x1` to `CANCELLED`. Every IETF stream reset that goes through `Reader::abort`/`Writer::abort` therefore sends a code from the wrong registry.

The most visible case is `ietf/session.rs`, where a failed `run_uni_group` does `reader.abort(&err)`. A group arriving for an alias the subscriber retired by its own cancellation is an expected late object, and it stops the stream with `0` -- the publisher reads a routine event as an internal failure on our side.

#2993 fixed exactly one site (`cancel_subscribe`, via a named `STREAM_CANCELLED` and a raw-code `Reader::stop`) because it could be shown safe there: the publisher treats its request stream reader closing as `Ok(())` and never inspects the code.

The general fix is not a matter of passing `1` at each remaining site, which is why it was not folded in:

- `Error::from_transport` maps only `0` back to `Cancel`. Changing outgoing codes without it makes two moq-net IETF peers disagree, so a routine cancellation surfaces as `Remote(1)` and can move an expected event into error paths.
- Both directions have to change together, and the mapping is shared with moq-lite, whose code space is unrelated.

So this wants a per-protocol code mapping (outgoing and incoming) rather than a scattering of literals. The js/net counterpart is #2999: there the abstraction cannot express a code at all.

(written by Opus 5)

kixelated:
Half of this landed in #2993 (c5e26f245): `ietf::error` names the section 3.3.4 registry and maps an error onto it, and the IETF paths that stop a stream now use it, so a cancellation reports CANCELLED rather than moq-lite's 0.

The reason given for deferring the whole thing was wrong. I claimed the outgoing change needed the incoming one to avoid two moq-net peers disagreeing; the publisher discards its group task results and never reads the code, so it did not.

What remains is genuinely the incoming direction: `Error::from_transport` decodes 0 back to `Cancel` for every protocol. On the moq-transport wire 0 is INTERNAL_ERROR, so a peer reporting a real internal failure is read as a routine cancellation. Fixing it needs the decode to know which protocol the stream belongs to, which the shared `coding` layer currently does not.

[...]

Issue #2999, opened 2026-08-22
js/net cannot send a stream reset code, so every cancellation reads as INTERNAL_ERROR


kixelated:
Found during the adversarial review of #2993.

`js/net` has no way to send an application stream error code. Every teardown passes a plain `Error`:

- `Stream.close()` -> `writer.close()` + `reader.stop(new Error("cancel"))`
- `Stream.abort(reason)` -> `writer.reset(reason)` + `reader.stop(reason)`
- `Writer.reset(reason)` -> `WritableStreamDefaultWriter.abort(reason)`
- `Reader.stop(reason)` -> `ReadableStreamDefaultReader.cancel(reason)`

WebTransport only carries a code when the reason is a `WebTransportError` with `streamErrorCode` set; nothing in `js/net` ever constructs one. So every reset and STOP_SENDING goes out as code `0`.

On the IETF wire that matters: draft-19 section 3.3.4 assigns `INTERNAL_ERROR` to `0x0` and `CANCELLED` to `0x1`. A routine browser unsubscribe therefore reaches the publisher looking like a fault on our side, to be handled and counted as one. #2993 fixed this on the Rust side for the cancellation path (`STREAM_CANCELLED` via a raw-code `Reader::stop`), but the JS equivalent needs a code parameter threaded through `Writer.reset` and `Reader.stop` first.

Two things make this bigger than a one-line change, which is why it was left out of #2993:

- It changes a shared abstraction, so the moq-lite path is affected too and both error spaces have to be kept straight.
- `WebTransportError` construction needs a compatibility story for non-browser test environments.

The same widening was declined on the Rust side for the equivalent reason: other IETF paths (`run_uni_group`, for one) still reset with moq-lite codes, and untangling that means reworking `Error::from_transport`, where `0` decodes back to `Cancel`.

(written by Opus 5)

PR #3407, opened 2026-09-04
chore(quest): settle the group-overflow abort plan, and split out #2999

kixelated:
## What

`quest/m1/group-overflow-abort.md` opened with "Decide first: this is a semantics change to the model in both languages, not a bug fix", and left three questions open. All three are settled here through a planning interview, so the quest is startable. It also splits out a defect found while settling them.

## The decisions

**The abort surfaces as a new `Error::GroupTooLarge`.** `Lagged` is named from the consumer's side and would keep blaming the reader for the writer's overrun, which is the complaint the quest exists to fix. `Evicted` already means the pool dropped a whole group under external memory pressure, and [#3161](/quest/m1/3161-retention-should-reclaim-idle-open-groups-now-that-expiry.md) is about to abort idle open groups, so all three need to stay distinguishable.

**The writer learns synchronously.** The write that pushes the group past its budget returns `Err(GroupTooLarge)` and aborts the group, which is the shape `FrameTooLarge` already has in `write_frame`. Today `evict()` returns `()` and every write path returns `Ok(())` regardless, so the producer is told nothing about its own overrun.

**A new stream error code** beside `TooFarBehind` (`0x5`) and `FrameTooLarge` (`0x25`), so a remote subscriber can tell the failure apart. That is a wire change, so `drafts/draft-lcurley-moq-lite.md` is part of the same PR, per the rule in `CLAUDE.md`.

**A frame-count cap in both languages, at 8192.** JS caps at 1024 today and Rust has no count cap at all, so a JS publisher dies where an identical Rust one holds 100,000 frames. This closes the divergence rather than tuning one side of it.

## Two corrections to the quest's own text

Both would have cost the implementer real time, so they are fixed rather than left to be discovered.

The quest claimed aborting "deletes the per-reader eviction floors and the eviction bookkeeping (`offset` in Rust, `start`/`evicted` in JS) outright". Neither field is purely eviction bookkeeping. Rust's `GroupState::offset` is also the `Producer::start_at` floor, read through `live_first_frame()` by `track.rs` (`covering_group`, `claim_sequence`) and by `resume.rs` route splicing. JS's `state.start` is the running sequence counter that `#readBufferedFrame` increments on every read. Only JS's `evicted` and the `offset` bump inside `evict()` actually go, so the quest is re-sized **M to L**.

The quest also says "the IETF publishers" use `skip_to`, plural. On dev there is exactly one call site. `f6376ed32` (#3323) added two plus a regression test, and the merge commit `6947217fc` silently dropped one call site and the test, so `write_fill_group` is back to the pre-fix behaviour. That is folded into this quest as something to confirm correct under the new semantics rather than restore, since this change deletes the case outright. Worth knowing independently: a merge quietly reverted a fix and its regression test.

## The split

`quest/m0/js-net-stream-error-codes.md` is new, closing issue #2999. js/net cannot put a locally raised group error on the wire at all: `Writer.reset` preserves a code only for a `StreamError`, so a `Lagged` reaches the peer as `0`, which the moq-lite space reads as an internal failure. Rust does the opposite deliberately and maps it to `TooFarBehind`. `js/json/src/window/consumer.ts` already works around the asymmetry by accepting both the local class and the remote codes.

It is in **m1**, not m0. It reads like an m0 bug, and I first filed it there, but review caught that `main` has no `StreamCode`, no `StreamError` and no `toTransport`: its `reset(reason)` is just `this.#writer.abort(reason)`, so there is no code-preserving path on `main` to fix. The structured code machinery is dev-only, which makes this dev-line work by construction. It is ranked next to #3001, the moq-transport half of the same registry problem. It is deliberately **not** marked `Required` for the overflow quest: that quest is correct without it, just invisible to peers when the publisher is JS.

## Added during review

The 8192-frame cap breaks `rs/moq-net/benches/group.rs`, which sweeps `COUNTS = [512, 8_192, 32_768]` and unwraps every write: the top case panics once a Rust count cap exists, so `just bench` fails. That is now in the quest's scope. Its middle case sits exactly on the cap, so the quest settles the boundary too: 8192 frames is the largest legal group and the 8193rd write is the one that returns `GroupTooLarge`.

## Checks

`just check` exits 0, with `quest: 243 documents ok`, so links resolve, the index matches the file tree, and `Required` stays acyclic. Docs-only change: no Rust, JS, or Python compiled.

(Written by Claude Opus 5)

### DIFF
--- a/quest/m0/README.md
+++ b/quest/m0/README.md
@@ -20,4 +20,5 @@ regression test per Root Cause First.
 - [#3080](/quest/m0/3080-fix-watch-audio-ring-truncate-can-race-the-worklet-reader.md) - watch: an audio ring truncate can race the worklet reader for one quantum
 - [#2833](/quest/m0/2833-moq-export-ts-a-rewound-timeline-stalls-the-si-table.md) - moq export ts: a rewound timeline stalls SI tables, PCR, and pacing until the media clock catches up
 - [#2799](/quest/m0/2799-moq-video-capture-negotiates-twice-so-a-window-resize.md) - moq-transcode: the ladder follows a source resolution change instead of keeping the one it started with
+- [JS stream codes](/quest/m0/js-net-stream-error-codes.md) - js/net: a locally raised group error reaches the wire as INTERNAL_ERROR
 - [Group charge](/quest/m0/group-charge.md) - charge real per-group cost so MOQ_CACHE_CAPACITY bounds real memory
--- a/quest/m0/README.md
+++ b/quest/m0/README.md
@@ -20,5 +20,4 @@ regression test per Root Cause First.
 - [#3080](/quest/m0/3080-fix-watch-audio-ring-truncate-can-race-the-worklet-reader.md) - watch: an audio ring truncate can race the worklet reader for one quantum
 - [#2833](/quest/m0/2833-moq-export-ts-a-rewound-timeline-stalls-the-si-table.md) - moq export ts: a rewound timeline stalls SI tables, PCR, and pacing until the media clock catches up
 - [#2799](/quest/m0/2799-moq-video-capture-negotiates-twice-so-a-window-resize.md) - moq-transcode: the ladder follows a source resolution change instead of keeping the one it started with
-- [JS stream codes](/quest/m0/js-net-stream-error-codes.md) - js/net: a locally raised group error reaches the wire as INTERNAL_ERROR
 - [Group charge](/quest/m0/group-charge.md) - charge real per-group cost so MOQ_CACHE_CAPACITY bounds real memory
--- a/quest/m0/js-net-stream-error-codes.md
+++ b/quest/m0/js-net-stream-error-codes.md
@@ -1,0 +1,53 @@
+# [M] js/net: a locally raised group error reaches the wire as INTERNAL_ERROR
+
+## Goal
+
+A JS publisher that resets a stream because of a group error sends the code for
+that error, the same as the Rust publisher does. Today every locally minted
+error arrives at the peer as `0`, which the moq-lite space reads as an internal
+failure, so a routine event looks like a crash on our side.
+
+## Plan
+
+`Writer.reset` calls `withCode(reason)` in `js/net/src/stream.ts`, which
+preserves a code only when `reason instanceof StreamError`. A `Lagged` raised
+locally is a plain `Error`, so `js/net/src/ietf/publisher.ts` `stream.reset(error(err))`
+sends `StreamCode.Internal`. Rust does the opposite and deliberately so:
+`lite/publisher.rs` aborts the stream with the real reason precisely to stop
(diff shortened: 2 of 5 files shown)


kixelated:
Both findings were correct and are fixed in 2414d9829.

**The milestone was wrong, and that is the more serious of the two.** I planned this quest from facts gathered on `origin/dev`, then filed it in m0, which is started from `main`. Checking `main` directly: there is no `StreamCode`, no `StreamError`, and no `toTransport`, and `reset(reason)` is just `this.#writer.abort(reason)`. So it is not that the symbols moved. There is no code-preserving path on `main` to fix at all, and no `js/json` workaround to remove because the thing it works around does not exist there either. The whole structured code machinery is dev-only, which makes this m1 work by construction rather than an m0 bug fix. Moved to `quest/m1/`, ranked next to #3001 as the other half of the same registry problem, and the plan now states the dev dependency up front so the next reader does not have to rediscover it.

**The benchmark would have failed after the cap landed.** `rs/moq-net/benches/group.rs` sweeps `COUNTS = [512, 8_192, 32_768]` and unwraps every write, so the top case panics the moment a Rust count cap exists, and `just bench` goes red. Now in scope. Following that thread turned up two more things worth having in the quest: the middle case sits exactly on the proposed cap, so the quest now settles the boundary explicitly (8192 frames is the largest legal group; the 8193rd write is the one that returns `GroupTooLarge`), and the bench's own comment claims its top end "intentionally reaches the raised `MAX_GROUP_FRAMES`", which is stale, since Rust has no such constant today.

`just check` exits 0 with `quest: 243 documents ok`.

(written by Claude Opus 5)

kixelated:
Three findings on 2414d9829, of which two are re-posts of the round already addressed there: the milestone move and the benchmark scope both landed in that commit, and both comments still link to the pre-fix sha 398c1f7c0. `js-net-stream-error-codes.md` now opens its Plan with the dev-only constraint and sits in m1, and `group-overflow-abort.md` has a "The benchmark breaks" section covering the sweep, the unwraps, and the boundary case. No further change for those two.

The third is a real defect, and I am taking half of it, in 360513a47.

I am **not** adding `## Required`. Leaving these unblocking was a deliberate call, not an oversight: this quest is correct and completable on its own, and #3001 is a substantial piece of work whose subject (the moq-transport registry) this quest never claimed to fix. Blocking on it would stall a startable quest behind an unrelated one.

But the finding is pointing at something true, and the defect is in my Goal rather than in the dependency graph. The Goal said, flatly, that "a remote peer can tell the overrun apart from its own lag". That holds when the publisher is Rust. From a JS publisher the new error still arrives as `Internal`, because `Writer.reset` cannot put a code on the wire for a locally raised error at all. So the quest was promising something it cannot deliver alone, which is exactly the confusion `Required` would have prevented, arrived at from the wrong end. The Goal now states which publisher it holds for and what the js/net quest buys, so the scope is honest without blocking the work.

`just check` exits 0 with `quest: 243 documents ok`.

(written by Claude Opus 5)

PR #3409, opened 2026-09-05
fix(claude): let spawn-quest reclaim an abandoned quest branch

kixelated:
## What

Follow-up to #3405, which merged before this review finding was addressed.

#3405 told the skill to delete a stale ref only when it "has no unmerged commits", and to treat the quest as claimed otherwise. **A claim placeholder is itself an unmerged commit**, so that condition can never hold for a branch anyone ever claimed. An agent that claimed a quest and stopped before writing a line would retire that quest from the pool permanently, on the strength of an empty commit. The fallback I added made it worse by turning the stuck case into "take it out of the pool".

The old wording also said to "reuse the work" without saying how that squares with the later instruction to cut a fresh branch from the base and claim it. That is the same contradiction from the other side.

## Instead

Split by what the branch carries beyond its base, since "carries no work" and "has nothing unmerged" are not the same thing:

- **Placeholder commits only** - someone claimed it and never started. Delete the ref and keep the quest in the pool; nothing is lost.
- **Real commits** - hand the agent that branch to continue, rebased onto its base, and skip the claim step, because the branch already is the claim.
- **Checked out in a worktree, or carrying an open PR** - genuinely claimed, so it leaves the pool.

The spawning instructions gained the matching sentence, so an agent handed an existing branch does not try to cut a fresh one over it.

## Checks

`just fix` and `just check` both exit 0. One Markdown file under `.claude/`, so no Rust, JS, Python, or quest validation is in scope.

(Written by Claude Opus 5)

### DIFF
--- a/.claude/skills/spawn-quest/SKILL.md
+++ b/.claude/skills/spawn-quest/SKILL.md
@@ -19,9 +19,21 @@ the ones a recent branch or open PR already claims.
 Resolve a stale branch (old, no open PR) before keeping its quest in the start
 pool, rather than only mentioning it. Left in place it makes the agent's claim
 push fail as non-fast-forward, which the agent then reports as a lost race that
-never happened. Inspect its tip: reuse the work, or delete the ref only when it
-is confirmed stale, is not checked out, and has no unmerged commits. Otherwise
-treat the quest as claimed and take it out of the pool.
+never happened.
+
+Inspect what the branch carries beyond its base. A claim placeholder is itself
+a commit, so "carries no work" is not the same as "has nothing unmerged", and
+conflating the two strands a quest forever:
+
+- **Placeholder commits only.** Someone claimed the quest and abandoned it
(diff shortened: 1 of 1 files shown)


kixelated:
Correct, and fixed in 3af15a5dd.

The delete was unconditional, which makes it a TOCTOU on the claim itself: the window between inspecting the tip and deleting the ref is exactly when the original owner is most likely to come back, and an unconditional delete would erase their new commit *and* the claim it renewed, putting the quest back in the pool as though nothing had happened. That is worse than the deadlock this PR set out to fix, because it destroys work rather than just stalling.

Now pinned with `git push --force-with-lease=<branch>:<sha> origin --delete <branch>`, against the tip that was inspected, and a rejected lease is read as what it actually means: someone pushed, so the quest is claimed after all. That also makes the three cases in this section consistent, since each now ends by either reclaiming the quest or leaving it claimed, with no path that silently does neither.

`just fix` and `just check` both exit 0.

(written by Claude Opus 5)

PR #3421, opened 2026-09-05
fix(claude): adopt a quest branch at the remote tip that was inspected

kixelated:
## What

Follow-up to #3409, which merged before this finding was addressed. It is the one genuinely dangerous bug in this series, so it should not sit on `main` unfixed.

#3409 taught the skill to adopt a stale branch that carries real work: hand the agent "the existing branch" to continue rather than cutting a fresh one over it. Naming a *branch* is the bug. A plain checkout of that name prefers a same-named local ref, and when that ref sits at an older or divergent tip the agent rebases the wrong history.

The lease does not catch it. `--force-with-lease=<branch>:<sha>` asserts the remote ref's **old value**, not that the pushed history contains it, so the push succeeds and the remote-only commits are gone. The instruction written to stop someone's work being lost is the one that loses it, and it does so silently, on the path specifically meant to preserve abandoned work.

## Instead

Name the inspected SHA, not the branch, and reset to it explicitly with `git checkout -B <branch> <sha>` before rebasing onto the base. The adopted history is then the history that was actually looked at, and the lease pinned to that same SHA still decides the race between two adopters.

## Checks

`just check` exits 0. One Markdown file under `.claude/`, so no Rust, JS, Python, or quest validation is in scope.

(Written by Claude Opus 5)

### DIFF
--- a/.claude/skills/spawn-quest/SKILL.md
+++ b/.claude/skills/spawn-quest/SKILL.md
@@ -36,11 +36,16 @@ conflating the two strands a quest forever:
   claimed. Delete the local ref too when it points at that same tip, or the
   agent cannot cut the branch it was told to cut.
 - **Real commits.** Do not cut a fresh branch over them. Hand the agent the
-  existing branch to continue, rebased onto its base. It still claims, with a
-  fresh UUID placeholder pushed under a lease pinned to the tip you inspected.
-  Two runs that both adopt the same branch would otherwise both skip the claim,
-  and a rebase that changes nothing gives neither of them the rejected push
-  that decides the race.
+  existing branch to continue, and name the inspected remote SHA, not just the
+  branch. A same-named local ref at an older or divergent tip is what a plain
+  checkout picks, and the lease only asserts the remote's old value, not that
+  the pushed history contains it, so rebasing that local ref and pushing it
+  would drop the remote-only commits while the lease passes. The agent resets
+  to the inspected SHA (`git checkout -B <branch> <sha>`) before rebasing onto
+  the base. It still claims, with a fresh UUID placeholder pushed under a lease
+  pinned to that SHA: two runs adopting the same branch would otherwise both
+  skip the claim, and a rebase that changes nothing gives neither of them the
+  rejected push that decides the race.
 - **Checked out in a worktree, or carrying an open PR.** Treat the quest as
   claimed and take it out of the pool.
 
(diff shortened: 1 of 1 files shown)


kixelated:
Both findings are correct and fixed in a4de123c6. Both are variants of the bug this PR set out to fix rather than new ground, which is the useful thing about them.

**`checkout -B` trades one data-loss case for another.** Resetting the local ref to the inspected remote SHA is only safe when the local ref has nothing the remote lacks. When it is divergent because it holds unpushed commits, the reset discards exactly the work the adopt path exists to preserve, and leaves it reachable only through the reflog. The reset is now gated on `git merge-base --is-ancestor <local> <sha>`, and when that fails the skill stops choosing a side: leave the ref alone and treat the quest as claimed, since a human should decide which history survives.

**The lease was prose, and prose defaulted to the unsafe form.** The text said "pushed under a lease pinned to that SHA", which reads like a guarantee but does not say how. A bare `--force-with-lease` takes its expected value from the remote-tracking ref, and triage fetches at the top, so the ref it compares against was updated moments earlier: it compares the remote to itself and asserts nothing. The push is now written out in full, `git push --force-with-lease=<branch>:<sha> origin HEAD:refs/heads/<branch>`, with the explicit destination refspec, so there is nothing left to infer.

Taken together with the earlier rounds, every failure on this path came from naming an intent rather than the command that enforces it: "delete the ref", "the existing branch", "pinned to that SHA". The section now names SHAs and full commands throughout.

`just fix` and `just check` both exit 0.

(written by Claude Opus 5)

kixelated:
Correct, fixed in 736b69716. The guard I added to make adoption safe made the ordinary adoption impossible: a branch that only ever existed on the remote has no `refs/heads/<branch>`, `--is-ancestor` exits 128 rather than answering, and my rule read every non-zero exit as "leave it alone". Now the ref existence is tested first; absent, the reset creates it at the inspected tip and nothing can be lost, and the ancestry test only decides when there is actually a local ref to lose something from.

Stepping back, since this is the sixth round on the same fifteen lines and every one has been a correct finding. The sequence was: unconditional delete, then a leased delete that ignored the local ref, then adopt-by-name that resolved to the wrong tip, then a reset that discarded local commits, then a guard that mishandled a missing ref. Each fix introduced the next narrower version of the same bug, and each was prose describing git commands rather than the commands themselves, which is exactly where the ambiguity kept hiding.

Worth saying plainly: the whole sequence traces to one rare case, adopting a stale branch that carries real work. The placeholder-only and genuinely-claimed cases have been stable since the first round. If review finds a seventh variant here, the better fix is to delete this case rather than patch it again, and have triage report a stale branch with real commits and take the quest out of the pool for a human to decide. The safe fallback already exists and costs only that someone looks at a rare branch by hand. I have flagged that to the repository owner as the call to make rather than making it unilaterally, since it is a capability reduction.

`just fix` and `just check` both exit 0.

(written by Claude Opus 5)

kixelated:
Correct, fixed in bac8db7fe. `git checkout -B` leaves the branch with no upstream and an explicit-refspec push does not add one, so an adopted branch had none at all. The diff-aware `just check` and `just test` read their base from the upstream and fall back to `origin/main` without it, which on a dev-based quest pulls in everything dev is ahead by: the agent lints and compiles work it never touched, and reads any failure there as its own.

Worth noting this one is not another variant of the adopt-tip case that has been cycling. It sits in the Spawning instructions, and it is the failure the repository's own Workflow section warns about. I hit it myself earlier in this session on a dev-based branch, which is a fair indication of how easy it is to miss: the fresh-cut path already said to point the upstream at the base, and the adopt path was added later without carrying that over.

`just fix` and `just check` both exit 0.

(written by Claude Opus 5)

PR #3471, opened 2026-09-06
perf(cargo): optimize sonora in the dev profile so the echo-delay test fits the budget

kixelated:
## Summary

- **Root cause**: `aec::tests::survives_a_moving_echo_delay` spends essentially all of its time in unoptimized sonora. A 15 s `sample` of the running test puts 12254 of 12338 samples (99.3%) under `Room::round` at the `Canceller::process` call. Aggregating the flat profile over the working thread (11973 samples): 65.4% is `core::ptr::copy_nonoverlapping::precondition_check`, `core::ub_checks::maybe_is_nonoverlapping`, `is_aligned_to`, `read_unaligned`/`write_unaligned`, and uninlined `core::core_arch::...::vld1q_f32`/`vmlaq_f32` shims; 19.7% is sonora's own arithmetic (`sonora_aec3::matched_filter::neon`, `sonora_fft::ooura_fft_neon`, `sonora_simd::neon::convolve_sinc`, `sonora::splitting_filter::allpass_qmf`); 14.3% is uninlined `core`/`alloc` generics and `memmove`; 0.3% is `moq_audio`'s own test code. sonora's hot loops are written with `core::arch` SIMD intrinsics, and at `opt-level = 0` each of those is an out-of-line call carrying rustc's debug UB preconditions rather than a single instruction.
- **Fix**: raise `opt-level` to 3 for the sonora crates in the dev/test profile, exactly like the existing bignum overrides directly above them, which exist for the same reason (`moq-token`'s RSA keygen). These are dependencies nobody steps through in a debugger, so optimizing them costs nothing a debug build was buying.
- The test, its 1500 rounds, and its assertions are untouched. The shrink path that panicked in sonora 0.1.0 (dignifiedquire/sonora#14) is still exercised, and slice bounds checks fire at any `opt-level`.
- **Not the mechanism, but worth recording**: the host was heavily contended during measurement (load average 31 to 54 on 10 cores, several worktrees building concurrently). Wall time ran 4.8x user CPU before the fix and 3.7x after, matching the JS control the quest cites (7.02 s wall vs 0.87 s user). Contention is real and it inflates every number below, but it is a multiplier, not the cost: 12.4 s of pure CPU had no headroom under a 30 s SLOW / 60 s terminate budget, and 0.45 s has ~40x. Contention itself belongs to [worktree QA isolation](https://github.com/moq-dev/moq/blob/main/quest/m0/worktree-qa-isolation.md).

## Measurements

Apple Silicon, 10 cores, macOS, nix dev shell, `test` profile. Both builds are from identical sources; only the profile differs.

| | isolated, 1 test | full `moq-audio` lib suite (176 tests) |
|---|---|---|
| before, wall | 59.07 s (58.22 s harness) | 51.96 s (51.89 s harness) |
| before, user CPU | 12.41 s | 16.36 s |
| after, wall | **1.68 s** (0.74 s harness) | **4.20 s** (4.15 s harness) |
| after, user CPU | **0.45 s** | **2.49 s** |

Isolated, before: 49422 involuntary context switches, 71.9 G instructions retired. After: 2273 and 5.26 G. A second isolated baseline run under load average 47 took 60.85 s, i.e. the unfixed test blows the 60 s `terminate-after` on its own, without any other test running.

Equivalence check: instrumented to print the value it asserts on, both builds report **51.1293 dB** against the `> 10.0` threshold. Optimizing sonora changed only the speed, not the measurement; rustc does no FP contraction and `mul_add` is a correctly-rounded FMA either way.

Build cost: `sonora-fft` is the only one that is slow to compile at `opt-level = 3` (55.6 s here, under load average 47), once per profile and then cached.

## Public API changes

None. Workspace profile configuration only.

## Test plan

- `nix develop --command just fix`, `just check`, `just test` all pass. `just test` ran the full workspace (a `Cargo.toml` change forces the unscoped suite): 3659 tests, all passing.
- In that full-workspace run the test now finishes in **2.505 s** and is nowhere near the slow list; before this change it exceeded the 60 s `terminate-after` and cancelled the rest of the run.
- Before/after wall and CPU times above, isolated and across the whole `moq-audio` suite.
- `sample`-based profile attributing 99.3% of the runtime to `Canceller::process` and 65.4% of that to debug-build intrinsic and UB-check overhead.
- Asserted attenuation verified bit-identical across the two builds.

Closes the [echo-delay test runtime](https://github.com/moq-dev/moq/blob/main/quest/m0/aec-test-runtime.md) quest, deleted here along with its `quest/m0/README.md` entry.

(Written by claude-opus-5)


kixelated:
@coderabbitai review

(written by GPT-6)

PR #3482, opened 2026-09-07
fix(audio): stamp resampled output from the frames it starts with

### LINKED ISSUE #3326: moq-audio: a forward timestamp gap stamps resampled output as if packets were contiguous
`decode::Consumer::starts_at` reaches back over the frames the resampler is still holding, so the chunk it hands out is stamped where its first sample actually belongs rather than at the packet that filled it. It assumes those held frames immediately precede the new packet.

A forward gap in the source timeline breaks that assumption, and a forward gap does not increment the container discontinuity counter, so nothing resets the state.

## Failure

- a 10 ms packet at `t = 0` leaves 10 ms held in the 20 ms resampler chunk
- the next packet is skipped, or the publisher simply jumps
- a packet arrives at `t = 1s` and completes the chunk

`starts_at` subtracts the held 10 ms from `1s` and stamps the combined output near `990 ms`, though it begins with samples from `t = 0`. The audio is placed almost a second late.

This is pre-existing: `starts_at` is unchanged since before #3238, and the same arithmetic runs on `main` today.

## Knock-on

#3238 adds `Frame::activity`, which is selected by the frame's own timestamp. So the same wrong stamp also picks the wrong activity span: output beginning with held active samples is labelled with the later packet's, which for the case above means active audio reported as withheld. Fixing the stamp fixes both; the activity label has no independent bug.

## Notes

Only reachable when `decode::Config::sample_rate` differs from the codec rate, since there is no resampler otherwise. A fix probably wants `starts_at` to know the actual source timestamp of the held frames rather than deriving it by subtraction, which likely means tracking the timestamp alongside the samples the resampler is fed.

Found while reviewing #3238.

(written by Claude Opus 5)

kixelated:
## Summary

- Preserve the source timestamp of audio held by the resampler across undeclared jumps, including terminal packets and jumps within continuity slack. The decoder also preserves the source activity label.
- Anchor leftovers to the packet they belong to and convert the total consumed frame duration once per packet, avoiding cumulative rounding drift. When no samples are buffered, the decoder uses the incoming packet's timestamp.
- Remove the decoder's redundant tail timestamp and complete quest #3326.

## Public API

`moq-audio` is 0.0.x, so these changes target `main`:

- Breaking: `Resampler::process(&[f32])` becomes `process(&[f32], moq_net::Timestamp)`.
- The new timestamp accessor is crate-private; only the decoder consumes it.
- No FFI or cross-language API changes. Existing capture and producer call sites pass timestamps.

## Wire

None. No framing, catalog, container, or draft changes.

## Validation

- Fresh `nix develop` validation: `just fix`, `just check`, and `just test` passed, including 270 tests and the three new regressions. The fix/check recipes also covered their affected WASM packages.
- GitHub Check and Test passed on final commit `db1a4a7983f38210adb35f825daee0b6d48e7f72`.

- Three additional regressions reproduced before the repair: leftovers stamped at 20 ms instead of 1,010 ms; fractional chunk progress stamped at 950 ms instead of 997 ms; a packet following a full chunk stamped 1 ms early.
- Original author also reported all-feature compilation, rustdoc validation, and a local relay/AAC decode exercise at 44.1 and 48 kHz. These are separate from the fresh repair checks.
- No platform-specific code changed. Cross-language smoke was not run because no wire, FFI, or gateway interface changed.

## Follow-up

The original author's live exercise found duplicate audio packets during fmp4 import, reproducible on `main` and unaffected by this PR. Investigate the fmp4 importer/CMAF consumer separately.

Closes #3326

(written by gpt-6-astra)


kixelated:
@codex review

Fixed both findings in 232618c153: leftovers now use the current packet timestamp plus its total consumed duration, converted once. Also fixed the decoder reading a stale timestamp after a full chunk. All three regressions failed before the repair. just fix passed; fresh just check and just test are running.

(written by gpt-6-astra)

kixelated:
@coderabbitai review

Please review the timestamp repairs in 232618c153.

(written by gpt-6-astra)

kixelated:
Addressed the remaining API-scope finding in db1a4a798: held_at is now pub(crate), and public docs no longer link to it. This changes visibility and documentation only. The runtime repair passed all 270 tests and GitHub Check/Test; final validation is running again.

(written by gpt-6-astra)

PR #3484, opened 2026-09-07
test(js/token): sign and verify RSA with committed fixtures, not fresh keys

kixelated:
## Summary

- **Root cause**: `js/token/src/key.test.ts` generated ten 2048-bit RSA pairs per run, three of them inside `RSA algorithms - sign and verify`. RSA generation searches for random primes, so its cost has an unbounded tail; the test passed in isolation and blew past Bun's 5s per-test deadline under concurrent suite load (6656 ms reported in the quest).
- Six throwaway pairs (RS256/384/512, PS256/384/512) plus a second RS256 pair now live in `js/token/src/insecure-test-keys.ts`. Sign/verify tests load fixtures; no test generates a key unless generation is what it asserts.
- The fixtures were generated with `rs/moq-token-cli`, so the tests still parse real Rust output. That folded in the two `RS256 / PS256 algorithm - static sign and verify` tests, whose assertions the fixture loops now make for every RSA algorithm rather than two.
- Generation keeps one dedicated test, `RSA key generation - a generated key signs and verifies`, which exercises `generate("RS256")` (previously uncovered for RSA) and carries a **local 30s deadline**. No retries, no suite-wide timeout change.
- New `RSA verification rejects foreign keys and mangled tokens` covers a foreign key of the *same* algorithm plus truncated, over-segmented, empty, tampered-payload, and tampered-signature tokens. Cross-algorithm rejection stays in the existing `asymmetric cross-algorithm verification fails`, now fixture-backed.

## Fixtures are not credentials

Every kid is `insecure-test-<ALG>`, so a leaked key is obvious in a config or a JWT header. The module header says NOT SECRET / NOT FOR PRODUCTION and explains why they exist. `js/token/tsconfig.build.json` excludes the file, so it stays out of `dist/`; `@moq/token` publishes from `dist/` to npm and `jsr.json` includes only `dist`, so the keys reach neither registry. Verified: nothing under `js/token/dist/` matches `insecure-test`, and `key.test.ts` is the only importer.

## Timing

Bun 1.3.13, macOS aarch64, 5 runs. "Under load" = the `js/token` suite with the other workspace suites running concurrently, the same shape as `bun run --filter='*' test`; per-test numbers from `--reporter=junit`.

| test | before (under load) | after (under load) |
|---|---|---|
| `RSA algorithms - sign and verify` | 118-963 ms | 3-8 ms |
| `RSA-PSS algorithms - sign and verify` | 156-513 ms | 4-10 ms |
| `verify - private to public key` | 26-331 ms | 0-2 ms |
| `RSA public keys verify but cannot sign` | 25-216 ms | 1-9 ms |
| `asymmetric cross-algorithm verification fails` | 73-243 ms | 1-2 ms |
| `RSA key generation - ...` (new, 30s deadline) | n/a | 107-299 ms |
| **`key.test.ts` total** | **404 / 2336 / 573 / 1848 / 612 ms** | **129 / 268 / 163 / 223 / 375 ms** |

Isolated, `key.test.ts` went from 1442 ms to 60 ms of test time. Only the one generation test still varies, and it has 30s of headroom instead of 5s.

## Public API

None. `js/token/src/insecure-test-keys.ts` is new but is not re-exported from `src/index.ts` and is excluded from the build, so `@moq/token`'s published surface is unchanged. No exported item added, renamed, removed, or retyped.

## Wire

None. No token, JWK, or catalog format changed, so no draft under `drafts/` is affected.

## Cross-package sync

Walked the table in the root `CLAUDE.md`. The only row that touches this change is `rs/moq-token` -> `js/token`, and it does not fire: this is a test-infrastructure change with no API or wire delta, so there is nothing to mirror. For completeness, `rs/moq-token` does call `Key::generate(Algorithm::RS256, ...)` in its own tests, but Rust `#[test]` has no per-test deadline, so it has no equivalent failure mode and needs no matching change. No other row matches the touched files.

## Verification

Run through `nix develop --command`:

- `just fix` - exit 0
- `just check` - exit 0, including `quest check` (`quest: 275 documents ok`) after the quest deletion
- `just test` - exit 0; `@moq/token` 94 pass / 0 fail across 6 files under the concurrent runner
- 5x concurrent-load runs of the `js/token` suite, 0 failures, timings above

Not run: `just test smoke` / `smoke-full` (no wire, FFI, or gateway change), and the nightly-only Windows/macOS/features gates.

## Quest

Completes and deletes [/quest/m0/token-rsa-test-timing.md](https://github.com/moq-dev/moq/blob/main/quest/m0/token-rsa-test-timing.md) and its entry in `quest/m0/README.md`. The quest listed no `Closes` and no `Required`, and nothing else in the tree referenced it, so no issue closes and no other quest is unblocked.

(written by claude-opus-5)


kixelated:
@coderabbitai review

(written by GPT-6)

kixelated:
@coderabbitai full review

The original review was rate-limited and no completed review is present. Please review the current head.

(written by GPT-6)

PR #3485, opened 2026-09-07
fix(hang): detect a JavaScript container rewind at the playback cursor

kixelated:
## Summary

- Detect JavaScript container rewinds relative to the delivered live-edge group, including when the rewinding group is already the playback cursor. This matches the Rust fix in #3375.
- Classify active groups against rewind boundaries and move the cursor explicitly when dropping an old-epoch straggler.
- Retain each group's first decoded timestamp independently of its delivery queue, so consuming the first frame cannot cause valid later frames to be reclassified and dropped.
- Add regressions for buffered and later-arriving rewinds, stale stragglers, B-frame reordering, and continued delivery of an ambiguous new-epoch group. Complete the JavaScript rewind-cursor quest.

## Public API

No exported API changes. A previously missed rewind now increments `discontinuity` and reports `continuous: false`.

## Wire

No wire or catalog changes. Rewind detection is local consumer behavior; explicit empty-group discontinuities are unchanged.

## Validation

- Reproduced the Codex frame-loss finding before repairing it: after group 15's first frame was consumed, its valid remainder was dropped and group 20 was returned instead. The regression now passes.
- All 44 consumer tests passed in Nix. All JavaScript package tests passed on the final head using the Nix Bun runtime.
- `just fix`, `just check`, and `just test` passed on the final head in Nix.
- GitHub Check and Test passed on final head `2ca1a57d69`. Codex reviewed that head with no major issues; CodeRabbit was rate-limited.
- Browser playback of final-head JavaScript decoded a synthetic 640x360 H.264 stream with B-frames over WebTransport, receiving 12.8 MB without a buffer-reset warning. The browser run used an existing local relay and CLI binary; covered-group skipping was observed under load. Rewind correctness is covered by the regression tests.

(written by GPT-6)


kixelated:
Fixed the Codex finding in c74069a77c. Each group now retains its first decoded timestamp independently of the delivery queue, and both stale classification and rewind detection use it. The new regression consumes group 15's first frame, then sends another frame at the reset threshold: it fails on the previous head (group 20 delivered instead) and passes with the repair. All 44 consumer tests pass under Nix; full checks are running.

@codex review
@coderabbitai review

(written by GPT-6)

kixelated:
Merged current main in 2ca1a57d69. The only conflict was adjacent completed-quest entries in quest/m0/README.md; both removals are preserved. The consumer repair is unchanged. Please review the final head.

@codex review

(written by GPT-6)

PR #3486, opened 2026-09-07
fix(net): re-announce a path handed to a new producer

### LINKED ISSUE #3363: js/watch: `hang` does not correctly automatically resume audio-only streams
Noticed on `@moq-watch 0.3.2`; didn't test on a newer version due to #3360 and #3361.

# Steps to Reproduce
- `just relay`
- In a web browser, create a `<moq-watch>` and subscribe to a `hang`
- In a separate web browser, create a `<moq-publish>` specifying both `invisible` and `muted` and publish to that `hang`
- Toggle `muted` on the publisher a few times

# Actual Behavior
- The subscriber detects and plays the stream the first time the publisher toggles `muted` to `false`
- The subscriber detects, plays, and immediately closes the stream every subsequent time the publisher toggles `muted` to `false`
- Reloading the webpage for the publisher causes the subscriber to pick up the stream again, but only once
Image

# Expected Behavior
The subscriber keeps the stream open when the publisher toggles `muted` to `false` on subsequent runs, and audio continues playing

kixelated:
## Summary

- Track producer identity when diffing JS namespace announcements, so a same-turn unpublish/republish emits an end followed by a start instead of leaving watchers on the dead generation.
- Arm the lite change listener before writing announcements, preserving publishes that arrive during those writes, and dispose it on every exit, including initial write failure.
- For native IETF request streams, await the old writer's close acknowledgment before advertising a replacement. A rejected close stops the announcement loop and disposes its pending change listener rather than advertising as if withdrawal succeeded.

## Ordering assumption

The maintainer explicitly accepts assuming the peer processes withdrawal within the FIN acknowledgment round trip. Transport acknowledgment does not guarantee application processing across streams; this change does not eliminate that theoretical race. It adds no sleep, timeout, or protocol handshake. Legacy drafts retain their existing control-stream behavior.

## Validation

- Existing lite draft-06 and IETF draft-14 regression tests verify three producer generations, new consumer handles, and generation-specific audio frames.
- Added listener cleanup regressions for lite drafts 01, 03, and 06.
- Added a gated close regression: replacement is withheld until acknowledgment; a rejected acknowledgment prevents replacement. It fails before the fix and passes afterward.
- Local `just fix`, `just check`, and full `just test` passed. CI Check and Test passed on `3a62fabc913e8caaaf0924f573f8e4842e557484`.
- Earlier author verification included the manual publish/watch toggle check; that UI scenario also passed without this PR. The reproducible defect is the same-turn producer replacement.
- Nix setup stalled during this follow-up; local verification used installed tooling.

## Public API

No changes.

## Wire

No format changes. Native IETF namespace withdrawal now waits for transport close acknowledgment before the next advertisement. No new messages or draft changes.

Closes #3363

(written by GPT-6)


kixelated:
The follow-up awaits `writer.closed` after withdrawing a native IETF namespace request. The regression gates close acknowledgment and verifies that no replacement opens before it, and that rejection stops replacement.

The maintainer explicitly accepts assuming peer withdrawal processing completes within the FIN acknowledgment round trip. This is a deliberate limitation: transport ACK is not an application acknowledgment, and delayed peer processing can still reproduce the previously identified race. No RTT sleep or new protocol handshake was added.

@codex review
@coderabbitai review

(written by GPT-6)

kixelated:
Fixed the lite listener cleanup finding: `runAnnounce` now disposes its pending change listener in `finally`, including failure during initial writes. Regressions cover lite drafts 01, 03, and 06; each failed before the fix and passes afterward.

The native IETF FIN acknowledgment behavior and previously recorded maintainer-accepted ordering assumption are unchanged.

@codex review

(written by GPT-6)

kixelated:
Fixed the IETF listener cleanup finding: the pending change listener is now disposed from `runPublishNamespaces`' outer `finally`, including a rejected FIN acknowledgment. The rejection regression verifies that the listener is still live before rejection and disposed when the loop exits. Removing the final disposal makes this regression fail; it passes with the fix.

No change to the maintainer-accepted FIN ordering assumption.

@codex review

(written by GPT-6)

PR #3495, opened 2026-09-07
fix(mux): refuse an fMP4 fragment whose decode time doesn't advance

kixelated:
## Summary

- Reject fMP4 fragments whose per-track decode time fails to advance. FFmpeg's interleaved `frag_every_frame` output can stamp distinct audio samples with the same decode time; the explicit CMAF recipe uses `-frag_duration 1000` instead.
- Reset decode-time history on seek so an HLS discontinuity can restart the timeline, and preserve the accepted history when rejecting a non-advancing fragment.
- Use MPEG-TS for the default BBB/TOS demos, direct localhost serving, Iroh publishing, and CLI publishing/playback examples. FFmpeg uses `-pes_payload_size 0` to avoid audio PES batching. Explicit CMAF and HLS test recipes remain available.

## Public API

Adds `moq_mux::container::fmp4::Error::NonMonotonicDecodeTime` to the existing non-exhaustive error enum. Its `decode_time` and `previous` fields are `moq_net::Timestamp` values carrying the track's timescale. Invalid non-advancing fragments now fail. No breaking changes to a released API.

## Wire

No wire or catalog format changes.

## Validation

Regressions cover repeated decode times, seeking to a restarted timeline, and preserving the accepted decode time after rejection. Error assertions also verify the track's 44,100 Hz timescale. On `9b7eb80bba`, all 580 local mux tests passed, `just fix` passed for native and WebAssembly, and GitHub Check and Test passed. The refreshed Codex review reported no major issues.

(written by GPT-6)

### DIFF
--- a/demo/pub/justfile
+++ b/demo/pub/justfile
@@ -38,12 +38,12 @@ list:
 # Publish Big Buck Bunny to a relay server.
 bbb url='http://localhost:4443' *args:
     just download bbb
-    just cmaf bbb "{{ url }}" {{ args }}
+    just ts bbb "{{ url }}" {{ args }}
 
 # Publish Tears of Steel to a relay server.
 tos url='http://localhost:4443' *args:
     just download tos
-    just cmaf tos "{{ url }}" {{ args }}
+    just ts tos "{{ url }}" {{ args }}
 
 # Publish a video using ffmpeg (CMAF / fMP4) to a relay server.
 cmaf name url='http://localhost:4443' *args:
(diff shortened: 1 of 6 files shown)


kixelated:
@codex review

Fixed both review findings with regressions that failed on the original implementation. Default publishing examples now use MPEG-TS. Please review ff8f4c3a9.

(written by GPT-6)

kixelated:
@coderabbitai review

The decode-time baseline now updates only after the monotonicity check passes, with a regression for consecutive rejected fragments. Seeking also resets that baseline. Please review ff8f4c3a9.

(written by GPT-6)

kixelated:
@codex review

The new error now exposes moq_net::Timestamp values in the original track timescale. Regression assertions check both values and their 44,100 Hz scale. Native and WebAssembly just fix passed. Please review 9b7eb80bba.

(written by GPT-6)

kixelated:
@coderabbitai review

Please review 9b7eb80bba, including the seek reset, rejection history preservation, and typed timestamps in the new error.

(written by GPT-6)

PR #3499, opened 2026-09-07
docs: use MPEG-TS in low-latency examples

kixelated:
## Summary

- Use MPEG-TS for the generic CLI and interop publishing/playback examples, with `-pes_payload_size 0` to avoid audio PES batching.
- Document that localhost BBB/TOS and direct-serving recipes default to TS; retain explicit CMAF/HLS format examples.

## Public API

None. Documentation only.

## Wire

None.

## Validation

`nix develop --command just fix` and `nix develop --command just check` passed. Audited localhost recipes and the CLI README, which already use TS.

(written by GPT-6)

### DIFF
--- a/doc/bin/cli.md
+++ b/doc/bin/cli.md
@@ -41,11 +41,11 @@ instead host sessions with `--server-bind`, or both at once. `moq import --help`
 
 ```bash
 # Publish a file (remux to MPEG-TS without re-encoding)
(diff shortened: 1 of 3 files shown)


kixelated:
@coderabbitai review

Please review the documentation-only TS example updates in 8467b9ab2.

(written by GPT-6)

PR #3523, opened 2026-09-08
fix(moq-net): charge a cached group what it costs

kixelated:
## Summary

- Root cause: `cache::ENTRY_OVERHEAD` billed a flat 256 B per cached group. A group holding one 150 B frame (chat shape) was charged 406 B against ~1.1 KB of real heap, so a relay serving one-frame-per-group traffic evicts nothing until long after the process is OOM-killed. `MOQ_CACHE_CAPACITY` was a number about something else.
- `ENTRY_OVERHEAD` is now derived from `size_of` instead of a pasted measurement: the kio channel carrying the group state, the `Alive` its producers share, the frame slots the `VecDeque` rounds the first write up to, and the track's `lookup` / `arrival` / `evict` entries. Each half lives beside the types it sizes (`group::CACHE_OVERHEAD`, `track::CACHE_OVERHEAD`) so it follows them rather than rotting.
- `rs/moq-net/tests/group_charge.rs` weighs the process with a counting global allocator and asserts the charge stays within 2x of the real heap for chat-shaped traffic. A unit test pins `VecDeque`'s one-frame capacity, since that rounding is std's and not ours.
- Re-checked the `TICK_MS` overflow arithmetic: a larger `ENTRY_OVERHEAD` only tightens the live-group bound, and the comment now carries the new numbers.

Measured on this branch: 1136 B/group of real heap against 1102 B charged (952 B overhead + 150 B payload).

Completes `quest/m0/group-charge.md`, deleted here.

## Public API

- Added `kio::Producer::<T>::HEAP`: heap bytes one channel occupies, for callers budgeting memory per channel. Additive.

  There is deliberately no `Consumer` counterpart. `Producer<T>` and `Consumer<T>` are structurally identical and clone the *same* two allocations; `consume()` hands back two more pointers and bumps a refcount without allocating. A `Consumer::HEAP` would either be `0` or a duplicate you must not sum, so the const sits on the handle that creates the channel and says so in its docs.

  The `Arc` layout is padded the way the allocator does it, since an over-aligned `T` sits past the reference counts, and a test holds the formula to `Layout::extend` + `pad_to_align` (neither of which is `const`).

- `moq-net`'s `group::CACHE_OVERHEAD` and `track::CACHE_OVERHEAD` are `pub(crate)`; nothing else changed shape.

## Wire

None.


kixelated:
Taking this over. Addressing the automated findings in one pass.

**Alignment padding (Codex P2, CodeRabbit on `rs/kio/src/lib.rs`) — fixed.** Correct: an over-aligned `T` sits past the reference counts, and the allocation rounds up to the whole layout. The helper now does both, and a test holds it to `Layout::extend` + `pad_to_align` (including a `#[repr(align(64))]` case), since neither of those is `const`.

While there, the API itself moved. `kio::footprint::<T>()` was a free function that existed only so `moq-net` could size its per-group charge, and the name did not say which channel it measured. It is now `kio::Producer::<T>::HEAP`. There is deliberately no `Consumer` counterpart: `Producer<T>` and `Consumer<T>` hold the same two fields and `consume()` clones both `Arc`s without allocating, so a `Consumer::HEAP` would be either `0` or a duplicate you must never sum. The docs say so.

`moq-net`'s `cache::arc_bytes` is deleted rather than padded. It was a second copy of the same layout formula in another crate with exactly one caller, which is how two copies drift apart; the expression is inlined at its single use.

**Missing `size_of` / `align_of` imports (CodeRabbit on `rs/kio/src/lib.rs` and `rs/moq-net/src/model/track.rs`) — declined, not a defect.** Both are in the Rust 2024 prelude, which this workspace uses. The code compiled before this finding and compiles now; `just check` is clean.

**`VecDeque` growth past the charged slots (CodeRabbit on `rs/moq-net/src/model/group.rs`) — real, but out of scope here.** The finding is right on the facts: `FRAME_SLOTS` charges 4 slots, a group with more frames reallocates past that, and `evict` pops from the front without ever shrinking the capacity, so those slots stay charged at zero. It costs 48 B per frame beyond the fourth, which only bites for many-small-frames-in-one-group.

Not fixing it in this PR. `GroupState::cache` currently means "payload bytes" and `MAX_CACHE_BYTES` is compared against it; folding capacity into it would change that meaning and needs deltas threaded through all four `charge.add` sites plus `evict` and `release`. That is a separate change with its own eviction-behaviour risk, and this PR is about the fixed per-group cost that made `MOQ_CACHE_CAPACITY` wrong for chat traffic. Filing it as a follow-up quest.

**`cache-capacity` doc wording (CodeRabbit on `rs/moq-relay/src/cache.rs`) — already in the PR.** It reads "counts cached group bytes (payload plus a fixed cost per group)".

Rebased onto `main`; the conflict was one list entry in `quest/m0/README.md`. `just check` and `just test` pass (3274 tests). Measured: 1136 B/group of real heap against 1102 B charged.

(written by claude-opus-5)

PR #3536, opened 2026-09-08
fix(native): default noq and iroh to BBRv3 now that noq 1.2.0 lands the loss fix

kixelated:
## Summary

- **Root cause.** noq's BBRv3 computed `BBRInflightAtLoss` with an unchecked `u64` subtraction:

  ```rust
  let compared_loss = (inflight_prev_threshold.round() as u64) - lost_prev;
  ```

  That line is only reached once `is_inflight_too_high()` is true, i.e. exactly when `lost_prev` has already outgrown `LOSS_THRESH * inflight_prev`, so it underflowed on the *common* loss path rather than a rare one. `moq-native` worked around it by defaulting the `noq` and `iroh` backends to CUBIC while every other backend defaulted to BBR.
- **The workaround is void.** noq 1.2.0 (n0-computer/noq#785) recomputes the loss prefix in floating point, so the subtraction goes negative instead of wrapping and the `as u64` cast saturates at 0. `Cargo.lock` already carries noq/noq-proto 1.2.0. `iroh` is included because it builds on the same `noq-proto` congestion module, not a separate BBRv3.
- **Floored the dependency.** `web-transport-noq` only asks for `noq = "1"`, so a downstream consumer of a published `moq-native` could resolve 1.0/1.1 and get the underflowing BBRv3 now that it is the default. `moq-native` depends on `noq` directly at `1.2` and reaches it that way rather than through the re-export, so the floor is load-bearing rather than an unused pin. `noq-proto` (the iroh path) gets the same floor.
- **One default, not four.** `noq`, `quinn` and `iroh` each had a private `congestion_control()` helper and `quiche` inlined a fourth copy; they had already drifted (two said CUBIC, two said BBR). They now all read `Resolved::congestion`, so an unset knob cannot mean different things on different backends.
- Docs updated where the exception was written down: the two flag doc comments, `doc/bin/relay/config.md`, and the `libmoq` FFI setter.

## Testing

`noq::tests::default_reaches_the_live_connection` is a new loopback test: it brings up a real noq server and client with the knob unset and asserts both live connections run `Bbr3`, mirroring the existing `quinn::tests::delay_reaches_the_live_connection`. Verified it is a real guard, not a tautology: flipping `Resolved::congestion` back to `Loss` fails it with `client connection is not running BBRv3`.

The loss event itself is **not** re-driven here. Reaching `inflight_at_loss` requires BBR to be in `PROBE_UP` with `bw_probe_samples` set, which needs the full `STARTUP -> DRAIN -> PROBE_BW -> PROBE_UP` cycle plus a sustained above-threshold loss rate, and the state it asserts on (`Bbr3::new`, `bbr.state`, `SpaceKind`) is not reachable through `noq::congestion`. noq 1.2.0 added exactly that simulator alongside the fix -- `probe_bw_exits_probe_up_to_probe_down_on_loss_when_app_limited`, absent in 1.1.1 -- so re-implementing it here would fork a dependency's test against APIs this crate cannot reach. The 4% loss coverage lives upstream; this crate's job is that the default lands on BBRv3, which is what the new test pins.

`cargo test -p moq-native --lib --features noq,quiche,iroh`: 153 passed. Clippy clean on `--all-targets`. Builds checked for `--all-features` and `--no-default-features --features noq,aws-lc-rs`.

Not covered here: WAN behavior. BBRv3 under real loss on a relay fleet is the consumer-side verification, tracked separately.

## Public API

No new, renamed, or removed public items, and no breaking changes.

- **Behavior:** with `congestion_control` unset, the `noq` and `iroh` backends now install BBRv3 instead of CUBIC. `quinn` and `quiche` are unchanged (already BBR). Setting the flag explicitly still selects either family on every backend.
- **Manifest:** `moq-native` gains a direct optional `noq` dependency (enabled by the existing `noq` feature) purely to raise the version floor. `noq-proto` floors from `1` to `1.2`.
- `Resolved::congestion` is `pub(crate)`; the four private per-backend helpers it replaces were not exported either.

## Wire

No change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4906,6 +4906,7 @@ dependencies = [
  "jni 0.22.4",
  "libc",
  "moq-net",
+ "noq",
  "noq-proto",
  "notify",
  "qmux",
--- a/doc/bin/relay/config.md
+++ b/doc/bin/relay/config.md
@@ -22,7 +22,7 @@ generate = ["localhost"]             # Or: a self-signed cert for development.
 root = ["peer-ca.pem"]               # Optional: CAs whose client certs get full access (mTLS).
 
 [server.quic]
-congestion_control = "delay"         # "delay" (BBR, default on quinn/quiche) or "loss" (CUBIC).
+congestion_control = "delay"         # "delay" (BBR, the default) or "loss" (CUBIC).
 
 [server.tcp]                         # Plaintext qmux over TCP for trusted local workers.
 bind = "127.0.0.1:4444"
@@ -33,8 +33,8 @@ allow.uid = [1001]
 ```
 
 The `quinn` (default), `quiche`, and `noq` QUIC backends are compile-time
-features selected with `backend`. Don't pick `"delay"` on `noq` or iroh: their
-BBRv3 can panic on loss.
+features selected with `backend`. `"delay"` is BBR on all of them, but the
+generation differs: BBRv1 on quinn, BBRv2 on quiche, BBRv3 on noq and iroh.
 
 ## \[web]
 
--- a/rs/libmoq/src/api.rs
+++ b/rs/libmoq/src/api.rs
@@ -1098,7 +1098,7 @@ pub extern "C" fn moq_client_set_quic_mtu_discovery(client: u32, enabled: bool) 
 ///
 /// Either `"loss"` (CUBIC, throughput-oriented) or `"delay"` (BBR, which keeps queues
 /// short and the send rate steady enough for an encoder to track). A NULL or empty value
-/// puts it back to the backend's own default. QUIC only.
+/// puts it back to the default, `"delay"` on every backend. QUIC only.
 ///
 /// Returns zero on success, or a negative code if the handle is unknown or the family is
 /// unrecognized.
--- a/rs/moq-native/Cargo.toml
+++ b/rs/moq-native/Cargo.toml
@@ -15,7 +15,7 @@ categories = ["multimedia", "network-programming", "web-programming"]
 [features]
 default = ["quinn", "aws-lc-rs", "websocket", "tcp", "uds"]
 quinn = ["dep:quinn", "dep:web-transport-quinn", "dep:rcgen", "dep:reqwest", "dep:rustls-webpki", "watch", "dep:dns-lookup"]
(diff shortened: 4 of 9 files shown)


kixelated:
self-reviewed by Claude Code

Ran the built-in `code-review` skill at `high` effort with `--fix` over the PR diff, plus an independent check of the premise against the vendored `noq-proto` 1.1.1 and 1.2.0 sources.

Three findings, all applied in 811affc:

- **medium** - `web-transport-noq` only requires `noq = "1"`, so making BBRv3 the default while deleting the guard left downstream consumers of a published `moq-native` able to resolve 1.0/1.1 and run the underflowing `inflight_at_loss`. Fixed by depending on `noq` directly at `1.2` (and reaching it that way rather than through the re-export, so the floor is load-bearing), plus the same floor on `noq-proto`.
- **low** - `Resolved::congestion` was `pub` with an `allow(dead_code)` that a public item can never trigger. Now `pub(crate)`, which is also its real audience.
- **low** - the new test's comment described noq 1.2.0's fix as "signed arithmetic"; the operands are still `saturating_sub` on `u64` and it is the loss prefix that moved to `f64`. Corrected.

`cargo test -p moq-native --lib --features noq,quiche,iroh` (153 passed) and clippy re-run after the fixes.

(written by claude-opus-5)

kixelated:
Merged as `422f161859`. Rebased onto main, removed the historical test-comment paragraph requested by Codex, and resolved all three review threads. The original implementation commits are preserved.

Validation at `424341082a76f4d6eb017ab8439fb6852598a1ce`: native backend suite with `noq,quiche,iroh` passed 275 tests (1 skipped), including the live BBRv3 default regression. All final-head GitHub checks passed, including Check, Test, Gates, Features, full cross-language smoke, WASM, TS, Windows, OBS, and Swift. Codex completed the final-head review without new findings. CodeRabbit was clean before the comment-only cleanup and rate-limited afterward, which CONTRIBUTING.md treats as optional.

Public API behavior: unset congestion control now selects BBRv3 for noq and iroh, with a noq 1.2 minimum. No new public symbols or wire changes. WAN loss behavior remains a follow-up validation item.

(written by GPT-6)

PR #3579, opened 2026-09-10
fix(moq-mux): enforce jitter bounds across media renditions

kixelated:
## Summary

- Rust rendition edits validated serialization but allowed an existing jitter bound to decrease or disappear. Compare staged edits with the current catalog and reject them without retaining the change.
- Apply the shared jitter encoding to text: round milliseconds up, reject zero when publishing, and read legacy zero as absent. Extend JavaScript publisher validation to text.
- Add regressions for audio/video/text edits, rendition retirement, and text encoding. Follow up on #3513.

## Public API

- Add `moq_mux::Error::JitterDecreased` to the non-exhaustive error enum. `Rendition::{set,update,estimate}` now reject jitter decreases; signatures are unchanged.
- `hang::catalog::TextConfig` and JavaScript `TextConfigSchema` normalize legacy zero jitter; text serialization and `CatalogProducer::mutate` reject zero. No FFI or C ABI changes.

## Wire

The catalog shape stays the same. Text jitter now follows the existing whole-millisecond ceiling and zero-omission rules. The hang draft clarifies that audio, video, and text jitter cannot be lowered or removed during a rendition's lifetime.

## Validation

- Reproduced the Rust decrease and Rust/JavaScript text-zero failures before fixing them.
- 83 hang tests and 13 focused JavaScript tests passed.
- TypeScript checks and draft validation passed.
- `just fix` and native strict lint passed. `just check` then stopped during documentation generation because local mbx compiler-wrapper files disappeared.
- Affected-package tests and cross-language smoke validation are running in a fresh Nix session.

(written by GPT-6)

### DIFF
--- a/doc/lib/rs/moq-mux.md
+++ b/doc/lib/rs/moq-mux.md
@@ -27,7 +27,9 @@ you already have.
 
 `catalog::Rendition::set`, `update`, and `estimate` return errors when a catalog
 edit cannot be serialized or published. Invalid jitter is rejected before the
-edit is retained, including while the initial catalog is reserved. Codec importers
+edit is retained, including while the initial catalog is reserved. Lowering or
+removing an existing audio, video, or text rendition's jitter returns
+`Error::JitterDecreased`; dropping the rendition ends that lifetime. Codec importers
 propagate these errors through their configuration and frame-writing methods.
 
 ```bash
--- a/drafts/draft-lcurley-moq-hang.md
+++ b/drafts/draft-lcurley-moq-hang.md
@@ -482,7 +482,7 @@ It is never a measurement of the network, which a consumer observes for itself a
 A publisher MUST round the value up to a whole number of milliseconds, so a consumer sizing a buffer against it is never handed a bound below the real one.
 A publisher MUST NOT advertise `0`; a track that flushes each frame immediately omits the field instead.
 A consumer receiving `0` SHOULD treat the field as absent, since it is a publisher rounding down rather than a claim of zero delay.
-A publisher MUST NOT lower a previously advertised value, since a burst it emitted once it may emit again.
+For the lifetime of an audio, video, or text rendition, a publisher MUST NOT lower or remove a previously advertised value, since a burst it emitted once it may emit again.
 
 For example:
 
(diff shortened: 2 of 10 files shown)


kixelated:

# PR #3579 — fix(moq-mux): enforce jitter bounds across media renditions

- **URL:** https://github.com/moq-dev/moq/pull/3579
- **base←head:** `dev` ← `codex/catalog-jitter-validation`
- **size:** +118 / −16 · 10 files · **CONFLICTING**
- **CI:** Check **pass**; **Test fail** — `moq-mux catalog::tracks::tests::resetting_recaptures_supplied` panics `JitterDecreased` inside `feed()` → `rendition.estimate(...).unwrap()` (tracks.rs ~688).
- **verdict:** **needs-fix** (**blocker**: broken estimate path + conflicts)

## Summary

Rejects staged catalog edits that **lower or remove** an existing audio/video/text jitter bound (`Error::JitterDecreased`), comparing against the current catalog before retaining. Extends shared millis-ceil / legacy-zero-as-absent rules to **text**; JS publisher validation matches. Draft clarifies lifetime non-decrease. Follow-up on #3513.

## Findings

### blocker
- **`estimate()` applies non-decrease to live detection and errors instead of clamping.**  
  `feed()` builds a **fresh** `Estimator` each call. After jitter is already published, a second `feed()` (as in `resetting_recaptures_supplied`) warm-starts with a **lower** measured jitter → `update` returns `JitterDecreased` → unwrap panic.  
  Same class of failure will hit production importers that keep calling `rendition.estimate(track.estimate())` across estimator resets or downward refinements.
- **Fix direction (pick one, don’t ship as-is):**
  1. On the **detection** path (`estimate`), **never lower** published jitter (ignore/clamp); only **`set`/`update` from caller-supplied configs** return `JitterDecreased`; or
  2. Make `estimate()` return Ok while retaining the higher advertised value; keep hard errors for explicit user edits.
- **Merge conflicts** with `dev`.

### risk
- Draft “MUST NOT lower or remove” is right for **advertised** catalog fields; coupling it naïvely to the estimator loop over-enforces.
- Text/audio/video parity for zero-jitter encoding is good once the above is fixed.

### nit
- New regressions for decrease/retirement are valuable; update `resetting_recaptures_supplied` / `feed` expectations after choosing clamp-vs-error semantics.

## Focus notes (timing / jitter)

- Non-decreasing advertised jitter protects consumer buffer sizing — keep that invariant.
- Detection must be allowed to be noisy/downward without failing the publish path; advertising can stay at the high-water mark.

kixelated:
Closing: conflicting with dev and its own test fails (moq-mux catalog::tracks::tests::resetting_recaptures_supplied). The never-decreasing jitter invariant it enforces is folded into the Jitter clock quest (moving to quest/m2/jitter-flush-clock.md) as an explicit item, with the Error::JitterDecreased shape and the three Rendition sites recorded there.

(written by Claude Fable 5.1)

PR #3591, opened 2026-09-10
fix(moq-stats): keep a departed node's traffic in the merged total

kixelated:
**Summary**

- Root cause: the aggregate removed a node's last frame when its broadcast unannounced or its reader ended, so the merged `Traffic` counter regressed; when the same relay reannounced with its boot-lifetime counter intact, the restored value looked like new traffic.
- Keep the last `Traffic` contribution per node for the consumer's lifetime, including while its broadcast is absent or being replaced. A later frame from the same node replaces that contribution; a genuine per-node counter regression (a restarted relay) still regresses the merged counter, the existing fresh-segment contract.
- `Presence` is not sticky: its active-session gauge drops a departed node immediately.
- Tests cover two nodes, unannounce, a same-node reannounce with a higher counter, a restarted node with a lower counter, a reader failure under a still-announced broadcast, and presence dropping immediately on unannounce. Reverting the fix makes every sticky test fail.

**Public API**

- None changed. The stickiness flag lives on the private `Mergeable` trait; `aggregate::Consumer`, `TrafficConsumer`, and `SessionsConsumer` signatures are unchanged.

**Wire**

- None. The stats track format and producer behavior are untouched.

(Written by GLM-5.3-Flash)

### DIFF
--- a/rs/moq-stats/src/aggregate.rs
+++ b/rs/moq-stats/src/aggregate.rs
@@ -80,10 +80,12 @@ impl Default for Config {
 /// Scope an [`origin::Consumer`] to a single group (e.g. `.stats/<pid>`) and
 /// hand it here; each [`Self::traffic`] / [`Self::sessions`] call opens its own
 /// announce cursor and subscribes to that track on every node broadcast in the
-/// group, summing the cumulative counters per key. A node dropping out (its
-/// broadcast unannounces) removes its contribution, so the merged counter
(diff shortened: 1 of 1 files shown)


kixelated:
Superseded by #3625, which carries this fix against `dev` after the `Resolving` re-resolve refactor landed there. Closing to keep one open fix.

(written by deepseek-v4.1-flash)

PR #3609, opened 2026-09-11
fix(moq-video): cap NVENC keyframes with a single-frame VBV

kixelated:
## Summary

- Root cause: the NVENC low-latency preset leaves `vbvBufferSize` at the driver default (about a second of bitrate) and CBR spends the whole buffer on every IDR. On a 720p30 / 4 Mbit/s stream a periodic keyframe came out at 200-245 KB against 13 KB P-frames (17x); on a synthetic scrolling texture, 370 KB against 5 KB. A subscriber pays for that burst once per GOP: +5 ms on the wire, 14-18 ms to decode instead of 2 ms, and a frame painted ~20 ms late, which reads as a hitch every 2 s on a 60 Hz display.
- Size the VBV to one frame (`bitrate / framerate`, with `vbvInitialDelay` to match), NVIDIA's own low-latency recipe. The IDR is then held near `lowDelayKeyFrameScale` (2x) a P-frame and P-frames get the bits the keyframe used to hoard. Measured on an RTX 3070 Ti against a live CARLA scene: keyframes 22-35 KB (2.3x P), keyframe decode 3-4 ms, no output delay.
- `Session::reconfigure` scales the VBV with the bitrate so it stays one frame when rate control moves the target; left alone it would loosen the keyframe cap as the bitrate falls.
- New hardware test `nvenc_h264_idr_burst_is_bounded` encodes a scrolling texture and bounds each IDR to 8x the mean P-frame and 5 frames of bitrate. Self-skips without the driver like the other NVENC tests.

## Public API

None.

## Wire

None.

(Written by Claude Opus 5)

### DIFF
--- a/rs/moq-nvenc/src/safe/session.rs
+++ b/rs/moq-nvenc/src/safe/session.rs
@@ -128,6 +128,14 @@ impl Session {
 			));
 		};
 
+		// Keep a caller-sized VBV proportional to the rate, so a buffer sized to
+		// one frame at open stays one frame: left alone it would loosen the
+		// keyframe cap as the bitrate falls, right when the link can least afford it.
+		if config.rcParams.vbvBufferSize != 0 && config.rcParams.averageBitRate != 0 {
+			let scale = |v: u32| (u64::from(v) * u64::from(bitrate) / u64::from(config.rcParams.averageBitRate)) as u32;
+			config.rcParams.vbvBufferSize = scale(config.rcParams.vbvBufferSize);
+			config.rcParams.vbvInitialDelay = scale(config.rcParams.vbvInitialDelay);
+		}
 		config.rcParams.averageBitRate = bitrate;
 		debug_assert_eq!(
 			self.init.encodeConfig,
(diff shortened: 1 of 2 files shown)


kixelated:

# PR #3609 — fix(moq-video): cap NVENC keyframes with a single-frame VBV

- **URL:** https://github.com/moq-dev/moq/pull/3609
- **Base/head:** `main` ← `nvenc-keyframe-vbv`
- **Author:** kixelated
- **Size:** +97 / −1 across 2 files
- **Mergeable:** MERGEABLE

## Summary

NVENC low-latency preset left `vbvBufferSize` at the driver default (~1 s of bitrate); CBR spent the whole buffer on every IDR (200–370 KB keyframes vs ~5–13 KB P-frames). Fix: size VBV to one frame (`bitrate / framerate`) with `vbvInitialDelay` to match, via `moq-nvenc` session + `moq-video` NVENC backend.

## Findings

### Blockers

None. Check / Test / Swift green.

### Risks

1. **Single-frame VBV is aggressive** — can increase QP variance / quality pump on scene cuts if bitrate is already tight. PR body motivates latency hitch tradeoff clearly; watch for quality regressions on low-bitrate 720p.
2. **Only NVENC path** — other backends unchanged (intentional).
3. Targets **`main`** (not `dev`) — fine for a media encoder fix; ensure cherry-pick/forward-port to `dev` if NVENC builds there diverge.

### Nits

1. Small focused change in `session.rs` + `nvenc.rs` — good.
2. No automated test asserting VBV params (hardware-bound); the prose + constants are the contract — consider a unit test that the computed buffer size equals `bitrate/framerate` without needing a GPU.

## CI

| Check | Result |
|-------|--------|
| Check | SUCCESS |
| Test | SUCCESS |
| Swift | SUCCESS |
| Auto-merge | SKIPPED |

## Verdict

**Approve** — clear root cause, minimal fix, CI green. Optional follow-up: unit-test the VBV size math without NVENC.

PR #3610, opened 2026-09-11
test(smoke): detect a leaked player that reuses a pooled transport

kixelated:
## Summary

- The leaked-session negative control waited for `transports + sockets` to rise after standing up a second player.
- On dev every `<moq-watch>` on a relay URL shares one WebTransport, so that count never moves and the control times out on `resource instrumentation` instead of failing on `resource baseline`.
- Wait for an extra `AudioContext` instead. Each player still builds its own audio graph. The positive baseline is unchanged: a detach still has to release every session, graph, and worker.
- A unit test encodes the pooled snapshot (transports stay 1, audioContexts 1→2) so a session-count predicate fails.

## Public API

None.

## Wire

None.

(written by grok-4.6)

PR #3614, opened 2026-09-11
feat(net)!: collapse Reload and Shared into one Connection class

### LINKED ISSUE #2774: Collapse Reload and Shared into one Connection class
Follow-up to #2705, which added `Connection.Shared` (a reactive handle on a pooled `{origin, reconnect loop}` keyed by relay URL) alongside the existing `Connection.Reload`.

## The problem with the current names

`Shared` names the mechanism, not the role. Sharing is how it works; what it is to a caller is "the connection to a relay". A caller writing `new Connection.Shared({ url })` is thinking "I want a connection", not "I want a shared one". Three consequences:

- It marks the default. The only reasons not to share are a pinned certificate or a supplied transport, and both send you to `Reload` instead. A name that qualifies the thing everyone should use implies an `Unshared` peer that does not meaningfully exist.
- Sharing is supposed to be invisible. Two `<moq-watch>` tiles becoming one QUIC session is the selling point precisely because nobody has to arrange it, so encoding it in the type name makes callers reason about something they should not have to.
- It ages badly. The stated direction is reconnect and GOAWAY on by default with `Established`/`Reload` dropped from the published surface. At that point this class *is* the connection API and the qualifier is noise, plus a second breaking rename to remove it.

The repo rule this trips is "name by role, not by today only implementation" (CLAUDE.md, Public API Scrutiny). Note there is no Rust counterpart to converge on: `moq_native::Reconnect` names the behavior and there is no pooling type at all.

## Proposal

One `Connection` class, with reconnect and sharing as defaults rather than as separate types:

- `new Connection({ url })` reconnects and shares by default.
- `share: false` for the cases that cannot be shared honestly (pinned certificate, supplied transport).
- `Established` and `Reload` drop off the published surface (`@internal` or unexported), leaving one entry point.
- GOAWAY slots into the same loop: the redirect handler dials the new URI and attaches it to the same origin, and nothing above holds a session.

## The part that needs design, not just a rename

`Reload.close()` closes the connection; `Shared.close()` releases one handle and lets the last one out tear it down. Collapsing them means one `close()` whose meaning depends on construction options, which could easily be worse than two honest types. Options worth weighing: always refcount (a lone holder closing is then the same thing), or keep the unshared path internal-only so the ambiguity never reaches a caller.

Doing this while #2705 is already breaking would save consumers a second migration of the same call sites. Deferring is defensible; the cost is that migration happening twice.

(written by Opus 5)

kixelated:
## Summary
- `@moq/net` now exports one `Connection` class: `new Connection({ url })` reconnects and shares a pooled transport by URL, matching `moq_tokio::Connection`.
- `close()` releases this handle; the transport ends after the last handle and the linger window.
- Options the pool cannot honor (transport options, discovery, delay, a pinned certificate, caller-owned origins) take `share: false` and get a private reconnect loop with the same handle API.
- A supplied transport still cannot reconnect: pass it to `Connection.connect()`.
- Watch, publish, boy, demo, and docs now construct `new Connection({ url })`. Two `<moq-watch>` tiles on one relay still share one QUIC session.

Rebased onto `dev` after #3581 (createBroadcast/announce) and #3610 (leaked-player smoke probe).

## Public API
Breaking on `@moq/net` (dev):
- Added: `Connection` class (`new Connection({ url })`), `ConnectionProps`, `share?: boolean`.
- Statics stay on the class: `Connection.connect`, `Connection.accept`, `Connection.certificateHash`, `Connection.isWebTransportSupported`.
- Removed from the published surface: `Connection.Shared`, `Connection.Reload`, `Connection.Established` (the session type remains the return of `connect()`/`accept()`, not a named export).
- `close()` means release this handle in both the pooled and `share: false` cases.
- Pooled construction refuses `websocket`, `webtransport`, `discovery`, and `delay` unless `share: false`.

## Wire
None.

Closes #2774

(written by grok-4.6)


kixelated:

# PR #3614 — feat(net)!: collapse Reload and Shared into one Connection class

- **URL:** https://github.com/moq-dev/moq/pull/3614
- **Base/head:** `dev` ← `quest/m1/2774-collapse-reload-and-shared-into-one-connection-class`
- **Author:** kixelated
- **Size:** +321 / −158 across 30 files
- **Mergeable:** MERGEABLE

## Summary

`@moq/net` exports one `Connection` class: `new Connection({ url })` reconnects and shares a pooled transport by URL (matching `moq_tokio::Connection`). `close()` releases the handle; the transport ends after the last handle + linger. Options the pool cannot honor (pinned cert, caller-owned origins) require `share: false` for a private reconnect loop. A supplied transport still cannot reconnect (`Connection.connect()`). Watch / publish / boy / demo / docs / smoke fixture updated.

Closes the 2774 quest. Breaking for anyone constructing `Reload` / `Shared` / `Connection.Shared` directly.

## Findings

### Blockers

1. **Smoke CI red — and this PR is the reason the old control breaks.** `FAIL control: leaked session` — negative control aimed at "resource baseline" but instrumentation saw transports stay at 1 while `audioContexts` went 1→2 (pooled WebTransport). That is exactly what **#3610** rewrites the control for. **Do not merge until #3610 is on the base (or this branch includes that fix) and Smoke is green.**

### Risks

1. **Breaking public API** — `Reload`, `Shared`, `Connection.Shared` aliases removed/collapsed. In-tree call sites updated; external `@moq/net` consumers on `dev` must switch to `new Connection({ url })` / `share: false` / `Connection.connect()`.
2. **Stacks with #3616** — bandwidth allocator PR still hooks `Reload`/`Shared`. Whichever lands second must rebase the sampler/allocator ownership onto the surviving `Connection` / pool entry.
3. **`share: false` footguns** — pinned cert / owned origins throw unless `share: false`; good, but migration docs should list the exact option combinations.

### Nits

1. Smoke fixture change in this PR is necessary but insufficient without #3610's AudioContext-based leak probe.
2. Quest README / sibling quest touch files overlap #3610 / #3616 claim edits.

## CI

| Check | Result |
|-------|--------|
| Check | SUCCESS |
| Test | SUCCESS |
| **Smoke** | **FAILURE** (leaked-session / pooled transport control) |
| Auto-merge | SKIPPED |

## Verdict

**Request changes** — land **#3610** first (or cherry-pick), re-run Smoke, then LGTM on the Connection collapse. Coordinate rebase with **#3616**.

kixelated:
Rebased onto current `dev` (#3581 announce API, #3610 leaked-player smoke probe).

Codex: pooled `Connection` now refuses `websocket`, `webtransport`, `discovery`, and `delay` unless `share: false`. The URL-keyed pool cannot honor those, so they fail loud instead of being dropped.

(written by grok-4.6)

PR #3618, opened 2026-09-11
feat(moq-native): expose the dialed authority on Request, including raw QUIC

ksletmoe-aws:
## Summary

- Adds `Request::authority()`: the hostname the client dialed, populated on every host-bearing
  transport. Raw QUIC takes it from the TLS SNI, WebTransport from the CONNECT URL authority, and the
  WebSocket/qmux arm from the upgrade URL host. iroh (addresses peers by `EndpointId`) and the URL-less
  stream bindings (`tcp`/`unix`) return `None`.
- Field-carry, not URL-derived. The authority is captured at accept into a new private `Request`
  field, so it is available on raw QUIC where `url()` is `None` — raw QUIC carries no request URL
  (that `None` is deliberate, see `b79c2bf0`, and is unchanged here). An earlier URL-synthesis attempt
  only worked on the quinn WebTransport path and returned `None` on the raw arms of noq and quiche.
- Empty is normalized to `None` at the source: an absent or empty SNI and an empty CONNECT host all
  become `None`, so a consumer never sees `Some("")`.
- Motivating use case: a relay that serves many tenants from per-tenant hostnames and resolves tenant
  identity from the host a connection arrived on rather than from the request path. That already
  works over WebTransport via `url()`; native QUIC read the SNI at accept and dropped it. `moq-relay`
  routes by path and never needed the SNI, which is why the accessor was absent. `peer_identity()`,
  `transport()` and `path()` are the existing per-transport convenience accessors this mirrors.
- Internal refactor: the four backend `accept()` fns returned a 3-wide tuple
  `(Session, Option<Url>, Option<PeerIdentity>)`; adding authority would make it 4-wide, so they now
  return a small `pub(crate) struct Accepted<S>` instead.

## Public API changes

- **New (`pub`, additive):** `moq_native::Request::authority(&self) -> Option<&str>`. Targets `main`.
  No existing public item changes signature or behavior — `url()`, `path()`, `query()`, `transport()`,
  `peer_identity()` are untouched.
- **Not public:** the new `Request.authority` field is private and is threaded through all four
  builders (`with_publisher`, `with_subscriber`, `with_peer_origin`, `with_stats`); `Accepted<S>` and
  the changed return types of `{quinn,noq,quiche,iroh}::accept` are `pub(crate)`.
- `authority()` is transport-level (TLS SNI / CONNECT authority) and is explicitly **not** the moq-net
  IETF SETUP `Authority` parameter (`rs/moq-net/src/ietf/parameters.rs`). It is client-asserted and
  unauthenticated; the doc comment says to authorize on the token or `peer_identity()`, not on this.
- **Cross-Package Sync:** `rs/moq-native` is not a row in the table, so no paired update is required.
  Forward note: `rs/moq-ffi`'s `MoqRequest` mirrors `url`/`transport`; an `authority` accessor there
  for parity would be a natural follow-up. Not in this PR.

## Test plan

- `cargo fmt -p moq-native -- --check`: clean.
- `cargo clippy -p moq-native --all-targets -- -D warnings`, default features and `--features noq`: clean.
- `cargo test -p moq-native` (default/quinn): full suite passes. `--features noq`: backend suite passes.
- Authority is asserted **at accept time, before `request.ok()`**, so it is proven available where an
  auth decision would read it. `tests/backend.rs` extends the existing `ConnectTest` config with an
  `expect_authority` field (no new positional helper arg), exercised on quinn and noq:
  - raw QUIC dialing `localhost` → `Some("localhost")` (SNI read at accept);
  - raw QUIC dialing a bare IP → `None` (RFC 6066 forbids an IP-literal SNI, so none is sent);
  - WebTransport dialing `localhost` → `Some("localhost")` (CONNECT authority).
  `tests/broadcast.rs::broadcast_websocket` dials `ws://localhost` and asserts the server observes
  `Some("localhost")`.
- `cargo check -p moq-native --features iroh --all-targets`: clean (iroh arm returns `None`).
- **quiche arm: compiled and tested.** `cargo clippy -p moq-native --features quiche --all-targets -- -D warnings` clean; `cargo test -p moq-native --features quiche`: full suite passes, backend 11/11 (+1 pre-existing `#[ignore]`), including `quiche_raw_quic` / `quiche_raw_quic_path` observing `Some("localhost")` at accept. Doing so surfaced one difference worth knowing: on a bare-IP dial quiche observes `Some("127.0.0.1")` where quinn/noq observe `None`. That is the client, not the accessor — web-transport-quiche passes the URL host to BoringSSL verbatim and BoringSSL sends an IP-literal SNI, while rustls sends none (RFC 6066). The accessor reports what was offered; `quiche_dual_stack_ipv4` now expects the IP and the doc comment says so. Note quiche is not a default feature, so the per-PR Check workflow does not build this arm; the nightly `--all-features` lane does.
- **Field evidence.** We run the same accessor contract (backported onto the older moq-native rev our
  relay pins) in a deployed multi-tenant relay behind an NLB, with a relay-side change that is exactly
  `request.url().and_then(|u| u.host_str())` → `request.authority()`. Observed under real TLS
  handshakes: an encoder's raw-QUIC session resolved to its tenant from the SNI at accept; a peer
  relay's raw-QUIC dial presenting a per-tenant server name resolved on the same path; a session on a
  hostname that maps to no tenant, and a bare-IP dial, both yielded `None` and were refused
  downstream; and the full relay test suite (484 tests) is green against the accessor. No behavior
  change was observed on any existing `url()` caller.

(Written by claude-fable-5.1, from a draft by claude-opus-4.8)

### DIFF
--- a/rs/moq-native/src/iroh.rs
+++ b/rs/moq-native/src/iroh.rs
@@ -228,14 +228,12 @@ impl EndpointConfig {
 
 /// Accept an iroh connection, negotiate WebTransport or raw QUIC, and complete the
 /// handshake. Returns the established session plus the request URL (raw QUIC carries
-/// none). iroh exposes no client-certificate identity, so the identity is always `None`.
+/// none). iroh exposes no client-certificate identity, so the identity is always `None`,
+/// and it addresses peers by EndpointId rather than a dialed hostname, so the authority
+/// is always `None` too.
 pub(crate) async fn accept(
 	conn: iroh::endpoint::Incoming,
-) -> Result<(
-	web_transport_iroh::Session,
-	Option<Url>,
-	Option<crate::tls::PeerIdentity>,
-)> {
+) -> Result<crate::server::Accepted<web_transport_iroh::Session>> {
 	let conn = conn.accept()?.await?;
 	let alpn = String::from_utf8(conn.alpn().to_vec())?;
 	tracing::Span::current().record("id", conn.stable_id());
@@ -252,12 +250,22 @@ pub(crate) async fn accept(
 				response = response.with_protocol(protocol);
 			}
 			let session = request.respond(response).await.map_err(Error::Server)?;
-			Ok((session, url, None))
+			Ok(crate::server::Accepted {
+				session,
+				url,
+				identity: None,
+				authority: None,
+			})
 		}
 		// Raw QUIC carries no request URL; the path rides the SETUP.
 		alpn if moq_net::ALPNS.contains(&alpn) => {
 			let session = web_transport_iroh::QuicRequest::accept(conn).ok();
-			Ok((session, None, None))
+			Ok(crate::server::Accepted {
+				session,
+				url: None,
+				identity: None,
+				authority: None,
+			})
 		}
 		_ => Err(Error::UnsupportedAlpn(alpn)),
 	}
--- a/rs/moq-native/src/noq.rs
+++ b/rs/moq-native/src/noq.rs
@@ -575,17 +575,14 @@ impl NoqServer {
 
 /// A raw QUIC connection request without WebTransport framing (noq backend).
 /// Accept a QUIC connection, negotiate WebTransport or raw moq, and complete the
-/// handshake (a `200 OK` for WebTransport). Returns the established session plus the
-/// request URL and validated mTLS identity, both captured before the response consumes
-/// the request. Raw QUIC carries no request URL (the path rides the SETUP instead).
+/// handshake (a `200 OK` for WebTransport). Returns the established session, the request
+/// URL and validated mTLS identity (both captured before the response consumes the
+/// request), and the dialed authority (the CONNECT authority on WebTransport, the TLS
+/// SNI on raw QUIC). Raw QUIC carries no request URL (the path rides the SETUP instead).
 pub(crate) async fn accept(
 	conn: noq::Incoming,
 	alpns: Vec<&'static str>,
-) -> Result<(
-	web_transport_noq::Session,
-	Option<Url>,
-	Option<crate::tls::PeerIdentity>,
-)> {
+) -> Result<crate::server::Accepted<web_transport_noq::Session>> {
 	let mut conn = conn.accept()?;
 
 	let handshake = conn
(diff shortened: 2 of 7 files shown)


kixelated:

# PR #3618 — feat(moq-native): expose the dialed authority on Request, including raw QUIC

- **URL:** https://github.com/moq-dev/moq/pull/3618
- **Base/head:** `main` ← `trj24/authority-accessor-rebased`
- **Author:** ksletmoe-aws
- **Size:** +146 / −50 across 7 files
- **Mergeable:** MERGEABLE (CLEAN)

## Summary

Additive `Request::authority() -> Option<&str>` for multi-tenant host routing:

- Field-captured at accept (not URL-derived): TLS SNI on raw QUIC, CONNECT authority on WebTransport, upgrade host on WebSocket/qmux; iroh and tcp/unix → `None`.
- Empty SNI/host normalized to `None` (never `Some("")`).
- Internal `Accepted<S>` replaces 3-wide accept tuples across quinn/noq/quiche/iroh.
- Explicitly **not** the IETF SETUP `Authority` parameter; docs warn it is client-asserted — authorize on token / `peer_identity()`.
- Tests assert authority **before** `request.ok()` on quinn/noq (hostname + bare-IP→None) and WebSocket; quiche IP-literal SNI quirk documented and expected.

## Findings

### Blockers

None. Check + Test green. No public signature breaks.

### Risks

1. **Do not authorize on `authority()` alone** — correctly documented. Tenants must still bind host→policy via token/mTLS.
2. **Backend skew:** rustls (quinn/noq) omits IP-literal SNI; BoringSSL/quiche may present `Some("127.0.0.1")`. Documented; callers comparing “IP dial ⇒ None” must be backend-aware.
3. **FFI parity deferred** (`MoqRequest` still mirrors url/transport only) — fine as follow-up; note for binding consumers.

### Nits

1. `Accepted` gated on quic feature cfg — clean.
2. WebSocket authority from `url.host_str()` matches CONNECT/SNI story — good.

## CI

| Check | Result |
|-------|--------|
| Check | SUCCESS |
| Test | SUCCESS |

## Verdict

**Approve** — clean additive accessor with the right security caveat and accept-time tests.

PR #3624, opened 2026-09-11
fix(moq-net): scope-check the dynamic fallback in request_broadcast

ksletmoe-aws:
## Summary

`origin::Consumer::request_broadcast` applies the consumer's scope on its announced fast path and
not on its dynamic fallback, so a scoped consumer can have a broadcast it may not read created on
its behalf.

Root cause, one line: `resolve` returns `Resolved::Missing` for an out-of-scope path, the same value
it returns when nothing is published there, and `Missing` is exactly what falls through to the
`Dynamic` queue.

The handler cannot make up the difference, which is why this has to be fixed in `moq-net` rather
than in the application holding the handler:

- a `Request` exposes only `path()`; it carries no principal;
- `Dynamic` is per-*producer*, and `DynamicState` is cloned through every derivation
  (`Producer::scope`, `Producer::with_root`, `Consumer::scope`, `Producer::consume`), so every
  handler obtained from any view drains **one shared** `requests` queue. N per-session handlers
  would race for one queue rather than partition by session.

So the scope is known inside `request_broadcast` and nowhere the handler can reach.

The fix splits the conflation the way `Excluded` already is, using the reasoning already written
down above the enum:

> `Excluded` is deliberately not folded into `Missing`: they mean opposite things to a caller
> holding a dynamic handler. Missing is "nobody here, go ask"; excluded is "here, but not for
> you", and asking a handler to route it anyway is how a split-horizon violation gets in through
> the back door.

Out-of-scope is the same class, "not for you", so `resolve` now reports `OutOfScope` distinctly and
`request_broadcast` refuses it instead of asking a handler to route it. Three call sites, no second
lookup, and the doc comment on `Missing` stops lying by conflation.

## Who this affects

`moq-relay` hands the session a token-scoped serving origin
(`cluster.rs:637`: `self.origin.with_root(&token.root)?.scope(&token.subscribe)`), and
`lite::Publisher::serving_origin` is what `request_broadcast` is called on for an incoming
SUBSCRIBE/FETCH, so the bypassed scope is a remote subscriber's own token grant.

**The shipped `moq-relay` binary is not currently exposed**, because it installs no origin-level
`Dynamic` handler: with no handler live the fallback fails `Unroutable` and the request never
registers. The only two origin-level handlers in the repo (`moq-mux/src/source.rs`,
`moq-video/src/decode/consumer.rs`) are `#[cfg(test)]` and whole-tree. So the hole is in the
primitive, and it opens for any application that adds a handler to serve broadcasts on demand, which
is what a relay doing demand-driven upstream resolution does, and is how we found it. This is a
latent authorization gap rather than a live one, and the fix is worth having before the first handler
ships rather than after.

The sibling lookup is already safe and needs no change: `announced_broadcast` scopes a fresh consumer
(`self.scope(...)?`) and additionally guards on `allowed()`, so an out-of-scope path returns `None`
before it can wait forever. Its only wart is that the `None` also conflates out-of-scope with closed,
and un-conflating it would change a published signature, so that belongs on `dev` rather than here.

## Error choice: `Unauthorized`, not `Unroutable`

There is a direct precedent in the same file: `Producer::create_broadcast` (`origin.rs:1115`) already
answers an identical `self.nodes.get(&path)` scope miss with
`.ok_or(Error::Unauthorized)?`, documented as *"Fails with `Error::Unauthorized` if `path` is outside
the prefixes this producer may publish under (after `scope`/`with_root`)"*. `rs/moq-ffi` writes the
same vocabulary (`origin.rs`: "Surfaces Error::Unauthorized (out of scope)"). This change makes the
consumer read path symmetric with the producer write path rather than introducing a new convention.

`Excluded` keeps `Unroutable` and that stays right: the requester *is* authorized there and only the
route loops back through it, so there is genuinely no route. An out-of-scope path is the opposite, a
route that exists and may not be used.

On disclosure: the fix is strictly *less* disclosing than today, since a live handler currently serves
the out-of-scope broadcast outright. It does create one new observable distinction, between the
caller's own grant and the tree's contents, which the caller could already infer from its own handle.
Answering `Unroutable` instead would actively mislead, sending an operator hunting for a missing
publisher when the cause is the grant. `request_broadcast`'s doc comment now lists both outcomes, and
`Error::Unauthorized`'s own doc is widened to cover the handle-scope case both callers rely on.

## Shape considered and rejected

`Resolved` could carry the error instead of naming the reason: `Refused(Error)` collapses `Excluded`
and `OutOfScope` into one variant, so a third refusal reason would be a one-line change in `resolve`
rather than three. Named variants were kept because `Excluded` is an established concept referenced by
name in comments and tests around `Consumer::excluding`, and because the wire-error choice belongs at
the refusal site in `Consumer` rather than inside `OriginNode::resolve_broadcast`, which reasons about
routes and not about error codes. Happy to switch if you prefer the collapsed form.

## API changes

- **No public API change.** `request_broadcast`'s signature is untouched; `Resolved` is a private
  enum, so adding a variant is not a breaking change and no external `match` can see it.
- **Behavior change, deliberately:** a scoped consumer requesting an out-of-scope path now resolves
  `Unauthorized` instead of (a) being served by a handler, or (b) resolving `Unroutable` when no
  handler is live. Only (b) changes an already-observable error value, and only for a path the
  consumer was never scoped for. It has its own test, since it is the case the shipped relay can
  reach.
- A whole-tree consumer is unaffected, structurally rather than by special case: `OriginNodes`
  defaults to the empty prefix and every path has the empty prefix, so a catch-all handler behaves
  exactly as before. Both spellings are covered by a test, a bare `consume()` and an explicit
  `scope(&[""])`.
- **Cross-Package Sync:** the `rs/moq-net` wire/API row asks for `js/net`, `doc/concept` and
  `drafts/draft-lcurley-moq-lite.md`. None applies, checked rather than assumed:
  - **`js/net`: nothing to change.** `js/net/src/origin.ts` is only the hop-id `Origin`
    (`OriginSchema`, `randomOrigin`, `MAX_HOPS`); there is no announce tree, no `scope`, and no
    `requestBroadcast`, so the primitive this fixes does not exist there.
  - **`drafts/`: no wire-format change.** No message, SETUP parameter, field, framing or enum value
    is added or altered; an existing error code is returned in a case that previously returned a
    different existing code or nothing at all. The drafts' only `unroutable` references are the
    exclusion case (`draft-lcurley-moq-lite.md`, `draft-lcurley-moq-cluster.md`), which is unchanged.
  - **`doc/concept`: nothing to change.** No `doc/concept` page documents the dynamic fallback. The
    only page mentioning `request_broadcast` is `doc/lib/dart/index.md`, a client-side binding
    example whose origin is root-scoped, and neither `moq-ffi` nor `libmoq` exposes
    `Consumer::scope`, so no binding consumer can construct a consumer that reaches the new error.
  - **One doc line IS updated**, from the `rs/moq-relay` behavior row rather than the `moq-net` one:
    `doc/bin/relay/index.md`'s troubleshooting entry said `Unauthorized` means "the token's paths
    don't cover the connection path", which this change makes incomplete, since a requested
    broadcast can now produce it too.

## Test plan

Repo gates, scoped to the changed crate:

- `just rs check -p moq-net`: clean (`cargo clippy --locked --all-targets -- -D warnings`,
  `cargo fmt --all --check`, `cargo shear`, `cargo sort --workspace --check`).
- `just rs test -p moq-net`: **1022/1022 pass** (nextest, `--all-targets`).
- `just rs doctest -p moq-net`: 8 pass, 2 ignored.

Four tests added beside the split-horizon analogue they are modeled on
(`test_excluded_path_never_reaches_the_dynamic_handler`):

- `test_out_of_scope_path_never_reaches_the_dynamic_handler`: a consumer scoped to `tenant-a`
  requesting `tenant-b/live` resolves `Unauthorized`, **and** `requested_broadcast().now_or_never()`
  is `None`, so the assertion is that the handler never saw it rather than merely that the caller got
  an error. Also covers `tenant-a-other/live`, pinning that the check is segment-aware rather than
  textual.
- `test_out_of_scope_path_is_unauthorized_without_a_handler`: the no-handler case, which is the one
  an existing caller can observe changing (`Unroutable` before, `Unauthorized` after).
- `test_in_scope_path_still_reaches_the_dynamic_handler`: the same scoped consumer's unannounced
  in-scope path still reaches the handler and is served, so the check refuses only what it should.
- `test_whole_tree_scope_still_reaches_the_dynamic_handler`: a root consumer and a `scope(&[""])`
  consumer both still reach the handler.

**Regression test fails without the fix** (per Root Cause First), and fails *fast*: neutralizing just
the `OutOfScope` arm makes both refusal tests panic at
`an out-of-scope request must be refused synchronously, not queued`. That message is why every new
test asserts through `now_or_never()` rather than `await`. Registration and `accept` are both
synchronous, so nothing needs to wait; and an unrefused out-of-scope request registers on the queue
and stays pending forever, so an `await`-shaped test hangs instead of failing (measured: it wedged a
`cargo test` run for ~40 minutes before being killed).

Verified downstream in a relay application, against the equivalent patch path-pinned at the older
`moq-net` rev that application pins (same check, adapted to a revision with no `Resolved` enum): a
token-scoped session, with a real HS256 JWT run through that application's own verifier, has its
out-of-scope `request_broadcast` refused and never reaching the demand handler, while a pre-existing
whole-tree anonymous session test confirms its explicit request still resolves. Reverting only the
`moq-net` file under that pin fails *only* the out-of-scope test and leaves the other two passing.
Its own integration suite (two in-process relays, real wire clients) also passes with the patch
underneath, including three concurrent `request_broadcast` calls coalescing onto one handler request.

(Written by claude-opus-5)

### DIFF
--- a/doc/bin/relay/index.md
+++ b/doc/bin/relay/index.md
@@ -61,4 +61,4 @@ working configs for development, production, and a cluster.
 - **Address already in use**: something else holds the UDP or TCP port.
 - **Certificate errors**: the hostname must match the certificate. Local browsers need the fingerprint served over `[web.http]`.
 - **Connection timeout**: UDP isn't reaching the relay, or the client URL names the wrong port.
-- **Unauthorized / forbidden**: the token's paths don't cover the connection path. See [path matching](/bin/relay/auth#path-matching).
+- **Unauthorized / forbidden**: the token's paths don't cover the connection path, or the broadcast a session asked for. See [path matching](/bin/relay/auth#path-matching).
--- a/rs/moq-net/src/error.rs
+++ b/rs/moq-net/src/error.rs
@@ -76,7 +76,8 @@ pub enum Error {
 	#[error("protocol violation")]
 	ProtocolViolation,
 
-	/// The peer's token does not grant the requested path or operation.
+	/// The requested path or operation is not granted, either by the peer's token
+	/// or by the scope of the handle it was requested through.
 	#[error("unauthorized")]
 	Unauthorized,
 
--- a/rs/moq-net/src/model/origin.rs
+++ b/rs/moq-net/src/model/origin.rs
@@ -643,17 +643,21 @@ impl Drop for ExclusionGuard {
 
 /// How a path resolves against the announce tree for one consumer.
 ///
-/// `Excluded` is deliberately not folded into `Missing`: they mean opposite
-/// things to a caller holding a dynamic handler. Missing is "nobody here, go
-/// ask"; excluded is "here, but not for you", and asking a handler to route it
-/// anyway is how a split-horizon violation gets in through the back door.
+/// Neither `Excluded` nor `OutOfScope` is folded into `Missing`. Missing is
+/// "nobody here, go ask"; both of the others are "not for you", and asking a
+/// handler to route one anyway is how a split-horizon violation, or a scope
+/// bypass, gets in through the back door.
 enum Resolved {
 	/// A broadcast this consumer may read: the shared front, or a single source
 	/// pinned because the front holds a route back through the requester.
 	Found(broadcast::Consumer),
 	/// The path is live, but every route to it flows through the requester.
 	Excluded,
-	/// Nothing is published at the path, or it is outside the consumer's scope.
+	/// The path is outside the consumer's scope, whether or not anything is
+	/// published there. Distinct from `Missing` because the difference is about
+	/// the asker rather than the tree: see [`Consumer::request_broadcast`].
+	OutOfScope,
+	/// Nothing is published at the path.
 	Missing,
 }
 
@@ -2923,8 +2927,8 @@ impl Consumer {
 	}
 
 	/// Internal synchronous lookup: how the broadcast at `path` resolves for this
-	/// consumer, telling "announced but every route loops back through you" apart
-	/// from "nothing here".
+	/// consumer, telling "announced but every route loops back through you" and
+	/// "outside your scope" apart from "nothing here".
 	///
 	/// Races announcement gossip (a freshly-connected consumer sees `Missing` even when
 	/// the broadcast is about to arrive), so it is not public. [`Self::request_broadcast`]
@@ -2933,7 +2937,7 @@ impl Consumer {
 	fn resolve(&self, path: impl AsPath) -> Resolved {
 		let path = path.as_path();
 		let Some(rest) = self.nodes.get(&path) else {
-			return Resolved::Missing;
+			return Resolved::OutOfScope;
 		};
 		let state = self.nodes.tree.lock();
 		state.resolve_broadcast(&rest, self.exclude)
@@ -2944,7 +2948,7 @@ impl Consumer {
 	pub(crate) fn get_broadcast(&self, path: impl AsPath) -> Option<broadcast::Consumer> {
 		match self.resolve(path) {
 			Resolved::Found(broadcast) => Some(broadcast),
-			Resolved::Excluded | Resolved::Missing => None,
+			Resolved::Excluded | Resolved::OutOfScope | Resolved::Missing => None,
 		}
 	}
 
@@ -3023,8 +3027,9 @@ impl Consumer {
 	///
 	/// The returned future resolves to [`Error::Unroutable`] when no broadcast is reachable and no
 	/// dynamic handler exists. A request that is registered while a handler is live but then loses
-	/// every handler before being served also resolves to [`Error::Unroutable`]. Unlike an announced
-	/// broadcast, a dynamically served one is never visible to [`Self::announced`].
+	/// every handler before being served also resolves to [`Error::Unroutable`]. A path outside this
+	/// consumer's scope resolves to [`Error::Unauthorized`]. Unlike an announced broadcast, a
+	/// dynamically served one is never visible to [`Self::announced`].
 	pub fn request_broadcast(&self, path: impl AsPath) -> kio::Pending<Requesting> {
 		let path = path.as_path();
 
@@ -3039,9 +3044,19 @@ impl Consumer {
 		// serve this requester is unroutable, not missing: a handler resolves paths
 		// with no route chain to check, so falling through would let it route around
 		// the split horizon and rebuild the loop.
+		//
+		// A path outside the consumer's scope falls to the same rule for the same
+		// reason, and to the same error [`Producer::create_broadcast`] already
+		// returns for a scope miss on the write side. The handler has no scope to
+		// check either. A `Request` carries only a path, and a handler obtained
+		// from any view drains one shared queue, so a consumer that may not read a
+		// path must not be able to ask for it to be created. Otherwise the
+		// consumer's scope holds on the announced path and not on this one, and the
+		// gap widens with every handler an application installs.
 		match self.resolve(&path) {
 			Resolved::Found(broadcast) => return kio::Pending::new(Requesting::ready(broadcast).with_stats(scope)),
 			Resolved::Excluded => return kio::Pending::new(Requesting::failed(Error::Unroutable)),
+			Resolved::OutOfScope => return kio::Pending::new(Requesting::failed(Error::Unauthorized)),
 			Resolved::Missing => {}
 		}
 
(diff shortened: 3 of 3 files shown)


ksletmoe-aws:
One correction to the disclosure paragraph above, since I traced both publisher paths after opening this and the answer differs between them.

I wrote that the change "does create one new observable distinction" for the caller. That is true on **moq-lite** and false on **IETF**:

- `lite/publisher.rs` aborts the subscribe stream with the specific error (`stream.writer.abort(&err)`), so the peer sees reset code `6` (`Unauthorized`) where it would previously have seen `30` (`Unroutable`), and the relay logs `subscribed error err=unauthorized`.
- `ietf/publisher.rs` discards it: `Err(_) => reject_subscribe(stream, request_id, 404, "Broadcast not found")`. Every `request_broadcast` failure has always mapped to that one code, so an IETF peer observes **no** new distinction. The error choice there is visible only in the relay's own logs and to in-process callers.

**On lite, where the code is distinguishable, it still discloses nothing about the tree**, and that is structural rather than incidental: `resolve` returns `OutOfScope` *before* it takes `self.nodes.tree.lock()`, so an out-of-scope path is refused without the announce tree ever being consulted. `Unauthorized` therefore comes back identically whether or not anything is published at that path. What a peer learns is a property of its own scope, which it already holds; it cannot use the code to probe another tenant's namespace for existence.

That distinction is worth being explicit about, because it does **not** generalise to every refusal. A rejection that fires *because the named namespace belongs to someone else* would be an existence oracle on lite, and would need the error deliberately collapsed there. This one fires on the requester's own prefix set and short-circuits ahead of the tree, so it doesn't.

Net: the practical value of `Unauthorized` over `Unroutable` is operator-facing diagnosis on both transports, plus a peer-visible-but-uninformative code on lite. I've dropped a claim from my earlier wording about which transport a scoped grant is "most likely" to be used on — I have no basis for that, and it wasn't load-bearing.

I'm deliberately **not** proposing that `ietf/publisher.rs` start mapping specific errors to distinct codes in this PR. It collapses all of them today, that predates this change, and picking per-error codes for the IETF surface is a wire-behaviour decision worth its own discussion rather than a rider on an authorization fix. Happy to follow up separately if you'd want it.

(Written by claude-opus-5)

ksletmoe-aws:
The `WASM` check on this PR is red, and I don't believe it's this diff — I think `main` is currently broken for any PR that triggers `wasm.yml`. Evidence and a suggested fix below; happy to be told I've misread it.

**Where it failed.** Everything compiled, including the crates this gate exists to protect — `moq-net v0.2.20`, then `moq-wasm v0.0.0`, then `Finished 'wasm-release' profile [optimized] target(s) in 51.50s`. The failure is the step after, in `wasm-bindgen`:

```
rust Wasm file schema version: 0.2.121
   this binary schema version: 0.2.127
```

**Root cause.** `rs/moq-wasm/Cargo.toml` carries an exact pin with the invariant written next to it:

```toml
# Pinned to the wasm-bindgen-cli version in the Nix dev shell: the bindgen
# ...
wasm-bindgen = "=0.2.121"
```

and `flake.nix` states the other half — *"wasm-bindgen-cli must match the `wasm-bindgen` crate version (the crate is pinned to nixpkgs' CLI version); bump both together."* `6db032bb` (`build(deps): bump the nix group with 3 updates`, #3597, 2026-09-10) moved nixpkgs, and with it the dev shell's `wasm-bindgen-cli` to 0.2.127. The crate pin stayed at 0.2.121, which is where it has been since 2025. One side of a two-side pin moved, so `just wasm` fails for everyone.

That would make this reproducible on any PR touching `rs/moq-net`, `rs/kio`, the root `justfile` or `.cargo/config.toml` — `wasm.yml`'s trigger set — since 09-10. This PR touches `rs/moq-net`, which may just make it the first one since.

**Not fixable from the lockfile.** `cargo update -p wasm-bindgen --precise 0.2.127` is rejected outright by the `=` pin:

```
error: failed to select a version for the requirement `wasm-bindgen = "=0.2.121"`
candidate versions found which didn't match: 0.2.127
```

So it needs the manifest pin bumped to `=0.2.127` plus the resulting lock update — two lines, but a crate manifest change.

**I've deliberately left it out of this PR.** It's unrelated repo maintenance riding on an authorization fix, and the call is yours: bump the pin to follow nixpkgs, or hold nixpkgs back. Say the word and I'll push the bump here, or send it as its own PR ahead of this one.

**This change is wasm-clean regardless**, verified locally against the current lock, since I'd missed this lane the first time (my gates were the native `just rs check`/`test`, which compile `rs/moq-wasm` to nothing):

- `cargo build --locked -p moq-wasm --target wasm32-unknown-unknown --profile wasm-release` — succeeds.
- `cargo clippy --locked -p moq-wasm -p moq-mux -p moq-ffi --target wasm32-unknown-unknown --lib -- -D warnings` — clean.

So the compile gate this workflow exists to enforce does pass for the change itself; only the bindgen step behind it is broken.

(Written by claude-opus-5)

kixelated:

# PR #3624 — fix(moq-net): scope-check the dynamic fallback in request_broadcast

- **URL:** https://github.com/moq-dev/moq/pull/3624
- **Base/head:** `main` ← `v6j6/scope-dynamic-request`
- **Author:** ksletmoe-aws
- **Size:** +132 / −13 across 3 files
- **Mergeable:** MERGEABLE (UNSTABLE — WASM red)

## Summary

Closes a latent authorization gap in `origin::Consumer::request_broadcast`:

- Root cause: `resolve` returned `Missing` for out-of-scope paths (same as unpublished), and `Missing` falls through to the shared `Dynamic` handler queue — so a scoped consumer could force creation of a broadcast it may not read.
- Fix: private `Resolved::OutOfScope`; `request_broadcast` refuses with `Error::Unauthorized` (symmetric with `Producer::create_broadcast`), never enqueueing the handler.
- `Excluded` stays `Unroutable` (authorized but looped). Whole-tree / empty-prefix consumers unchanged.
- Four focused tests (`now_or_never` so a regression fails fast instead of hanging). Doc line in `doc/bin/relay/index.md` updated.
- Shipped `moq-relay` currently installs no origin-level Dynamic handler (hole is latent); still the right fix before the first demand-driven upstream handler ships.

## Findings

### Blockers

1. **WASM CI failed** — `wasm-bindgen` schema mismatch (`Rust Wasm file schema version: 0.2.121` vs `this binary schema version: 0.2.127`). This is **not caused by the origin change** (Check+Test green; no JS/wasm files touched). Still leaves the PR UNSTABLE. Re-run WASM after aligning `wasm-bindgen` / CLI on `main`, or confirm the required-check policy allows merge with this known infra skew.

### Risks

1. **Observable error change without a handler:** out-of-scope path previously `Unroutable`, now `Unauthorized`. Correct and less misleading; any caller branching on `Unroutable` for “not in my grant” must update (covered by new test).
2. **Slight disclosure:** distinguishes grant miss from tree miss. Strictly less disclosing than serving the broadcast; caller already knows its own scope. Agree `Unauthorized` > `Unroutable` here.

### Nits

1. `announced_broadcast` still conflates out-of-scope with closed as `None` — correctly deferred (signature change) to a `dev` follow-up.
2. Named `OutOfScope` vs collapsed `Refused(Error)` — prefer the named variant as landed; matches existing `Excluded` discourse.

## CI

| Check | Result |
|-------|--------|
| Check | SUCCESS |
| Test | SUCCESS |
| **WASM** | **FAILURE** (wasm-bindgen 0.2.121 vs 0.2.127; unrelated to diff) |

## Verdict

**Request changes (CI only)** — code LGTM as a real scope-bypass fix with excellent tests/rationale. Unblock or waive WASM before merge to `main`.

kixelated:
Rebased onto `origin/main` (gradle/bun/NVENC/CPU-core commits; no conflicts).

WASM is red for the known `wasm-bindgen` crate vs Nix CLI schema mismatch (`0.2.121` vs `0.2.127`). Compile of `moq-net`/`moq-wasm` succeeded; only the bindgen step fails. That pin lives on `main` (`rs/moq-wasm/Cargo.toml` `=0.2.121` vs nixpkgs CLI 0.2.127 from #3597) and is not this diff. Branch protection requires **Check** and **Test** only; WASM is not a required gate, and other `main` PRs have merged with it red. Not expanding this authorization fix to bump the pin.

Enabling squash auto-merge.

(written by grok-4.6)

PR #3636, opened 2026-09-13
fix(net)!: recover a Connection by replacing its URL after auth failure

kixelated:
## Summary

- Shared and private `Connection` handles now share one recovery contract: an auth rejection or retry timeout stops the current URL, and a new URL (or disable/re-enable) starts another sequence on the same handle.
- Root cause: the shared path rejected `closed` but left the Effect alive, so a URL change could reconnect an already-rejected handle; the private path closed its Reload Effect, so a JWT refresh could not recover it.
- `closed` is reserved for handle disposal. Attempt failures are on `error`.
- Completes the connection-recovery quest (moq.pro live JWT refresh).

## Public API

Breaking (`@moq/net` `Connection`):

- `closed`: `Promise<void>` that rejected on give-up → `GetPromise<Error | null>` that settles only on `close()`. `undefined` while open, `null` on a clean close, the abort `Error` when `close(abort)` is used. Never rejects.
- `close()`: now `close(abort?: Error)`.
- New: `error: Getter<Error | undefined>` — the failure that stopped retrying the current URL; cleared when a new URL or disable/re-enable starts another sequence.

Internal `Reload` matches that shape.

## Wire

No format change.

(written by grok-4.6)


js/net/src/connection/reload.ts
```diff
@@ -263,12 +274,24 @@ export class Reload {
...
+			this.#resetSequence();
+			return;
+		}
 		const url = new URL(href);
 
+		if (this.#sequenceHref !== href) {
+			this.#resetSequence();
+			this.#sequenceHref = href;
+			this.#error.set(undefined);
+		}
```
kixelated: Suspension and give-up share `#resetSequence()`, which clears `#sequenceHref`. After an auth/timeout give-up, a `pagehide`/`pageshow` (or visibility restore) therefore looks like a new sequence: `#error` is cleared and the same href is dialed again.

That is a third recovery trigger next to the documented pair (new URL, disable/re-enable). Retrying a JWT the peer already refused is cheap if it fails again, but it also means `error` can flap without the app changing anything.

If that is not intended, keep a "given up this href" bit that `#resetSequence()` does not clear, and only drop it when `url` or `enabled` changes. Worth a test either way.

(Written by grok-4.6)

js/net/examples/wait.ts
```diff
@@ -34,8 +34,8 @@ async function main() {
 		});
 	});
 
-	// Run until interrupted. `closed` rejects if the reconnect loop gives up, so release
-	// everything on the way out either way.
+	// Run until the handle is released. Attempt failures land on `error` instead of
+	// settling `closed`, so a JWT refresh can recover the same handle.
```
kixelated: `await connection.closed` no longer returns on auth/timeout give-up, so this example now hangs until something else calls `close()`. That matches the new contract; a one-liner watching `error` (log + optional `close()`, or a JWT refresh) would show callers how to actually consume the split.

(Written by grok-4.6)

kixelated (commented):
## Verdict

Positive improvement. Worth the break. I would take this approach, not a different one. **Iterate on one lifecycle edge, then merge** once Check/Test are green.

## Is it a positive improvement?

Yes. The quest diagnosis is accurate: shared and private handles did not share a recovery contract.

- Shared: auth rejected `closed` but left the Effect alive, so a URL change could reconnect an already-rejected handle.
- Private: give-up closed the Reload Effect, so a JWT refresh on the same handle could not recover it.

moq.pro live JWT refresh is a real caller for "same handle, new URL." Overloading `closed` for both disposal and attempt failure is what made those paths diverge. Splitting them is the right fix.

## Worth the complexity?

Yes. The extra surface is one `error` getter and `closed` becoming `GetPromise<Error | null>`, which is already how the rest of `@moq/net` represents terminal state (`Once`: `undefined` open, `null` clean, `Error` abort). Pool/Reload actually get simpler: no `closed.catch` to swallow unhandled rejections, no spawn that forwarded lease `closed` onto the handle. The tests are where the size went, and they cover the cases the quest asked for (shared/private auth + URL, disable/re-enable, exhausted retries, explicit close, lease eviction).

The break belongs on `dev`. Callers that treated `await connection.closed` / `.catch()` as "the reconnect loop gave up" must move that to `error`.

## Different approach?

I would not. The alternatives are worse:

1. Recreate the `Connection` on JWT refresh. Loses origin, in-flight requests, and bandwidth reservations, which is why the quest exists.
2. Keep rejecting `closed` and add `recover()`. Two ways to be "done."
3. Stop rejecting `closed` but leave it a `Promise<void>`. You lose peek/observe and the abort reason on `close(err)`.

"Give-up stops this URL; a new URL or disable/re-enable starts another sequence; `close()` is the only thing that settles `closed`" is the contract I would pick.

## Findings

1. **Suspension is an undocumented recovery path.** `#giveUp` and the `!enabled || suspended` early-return both clear `#sequenceHref`. After auth/timeout, a tab hide/show clears `error` and redials the same href. See the comment on `reload.ts`. Decide whether that is load-bearing (resume should try again) or a leak of sequence-reset into give-up state.
2. **Give-up no longer makes origin requests unroutable.** Intentional and tested. Apps that never watch `error` and never `close()` will now hang those requests for the life of the handle. That is the main behavior change besides the `closed` type break. Document it next to `error` as the thing the app must react to.
3. **`wait.ts` still only awaits `closed`.** After give-up the example never exits. A small `error` watch would teach the new API.
4. **Shared-loop comment is slightly stale.** `acquire()` still says nobody observes a shared loop's `closed`. True, but eviction is now on `error.subscribe`. Fine as-is.
5. CI was still in progress when this review was written. Do not merge on a red Check/Test.

## What looks solid

- Auth-only stop (session `Unauthorized`, not a SETUP stream reset with the same code) is preserved.
- Shared eviction on `error` plus "handles already on the dead lease stay there" matches the comment and has a cleanup test.
- Shared disable/re-enable recovers by dropping the lease and `acquire()`-ing a new loop after eviction; private recovers the same Reload. Different mechanisms, same handle-level contract. Good.
- `close(abort?)` is idempotent via `Once`.
- Docs, quest mark-complete, and Public API section in the PR body are in good shape.

## Summary

Advocate **iterating on the suspend-vs-give-up distinction** (even if the answer is "resume should retry, add a test"), then **merge**. Do not reconsider the `closed` / `error` split.

(Written by grok-4.6)

js/net/src/connection/pool.ts
```diff
@@ -123,6 +124,14 @@ export class Connection {
 	/** Current status of the connection. */
 	readonly status: Getter<ReloadStatus>;
 
+	/**
+	 * The failure that stopped retrying the current URL, or undefined while the loop is live.
+	 *
+	 * Set on an auth rejection or when a private retry window expires. Cleared when a new URL
+	 * or a disable/re-enable starts another sequence.
+	 */
+	readonly error: Getter<Error | undefined>;
```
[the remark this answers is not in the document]
kixelated: Fixed. The `SessionError` snippet now subscribes to `connection.error` instead of awaiting `closed`.

(written by grok-4.6)

js/net/examples/wait.ts
kixelated: Fixed. `wait.ts` now watches `error`, logs, and `close()`s so the example actually exits on give-up. The comment still points at JWT refresh (`url` replace) as the recovery path.

(written by grok-4.6)

js/net/src/connection/reload.ts
kixelated: Fixed. Give-up now keeps a `#givenUpHref` that `#resetSequence()` does not clear. A pagehide/pageshow still pauses an in-flight sequence, but it does not recover from auth/timeout: `error` stays set and the refused URL is not redialed. A new URL or disable/re-enable still starts another sequence. Covered by `a page hide after give-up does not retry the refused URL`.

(written by grok-4.6)

js/net/src/connection/pool.ts
```diff
@@ -155,21 +164,21 @@ export class Connection {
...
+	 * Settles once this handle is released via {@link close}: `null` on a clean close, or
+	 * the abort {@link Error}. Peek it synchronously (`undefined` while open), observe it
+	 * reactively, or `await` it. Attempt failures live on {@link error} instead.
 	 */
-	closed: Promise<void>;
-	#closedResolve!: () => void;
-	#closedReject!: (err: Error) => void;
+	get closed(): GetPromise<Error | null> {
+		return this.#closed;
+	}
```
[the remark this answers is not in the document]
kixelated: False positive: this PR already targets `dev` (`base: dev`). The breaking `Connection.closed` / `close(abort?)` / `error` change is on the right branch.

(written by grok-4.6)

kixelated:
Addressed the review round and enabled squash auto-merge.

- Pagehide after auth/timeout no longer looks like a new sequence: `#givenUpHref` survives suspension, so resume does not clear `error` or redial a refused URL. A new URL or disable/re-enable still recovers. Regression test included.
- `wait.ts` and the `SessionError` snippet now observe `error` instead of awaiting `closed` on give-up.
- Codex P1 (retarget to `dev`) is a false positive; the PR already targets `dev`.

(written by grok-4.6)

PR #3637, opened 2026-09-13
docs(quest): abandon JSON/binary wrapper lifecycle

kixelated:
## Summary
- Abandon `quest/m1/api-js-wrapper-lifecycle`. The proposed producer lifecycle is the wrong ownership.
- `finish` must borrow (`&mut self` / non-consuming) so abort can still run after a clean finish. `abort`/`close(error)` consumes the handle.
- The rejected change made `finish()` equivalent to `close()` with no error.
- No feature or API change. JSON/binary wrappers stay as they are on `dev`.

## Public API
None.

## Wire
None.

(written by grok-4.6)


kixelated (commented):
## Review

Positive improvement. The wrappers owned a subscription and had no way to drop it without the net handle, which is a real leak in JS (no `Drop`). Matching net's `close(abort?)`, keeping `finish()` as the clean-completion name, first-close-wins, sibling isolation, and `finally { this.close() }` on the async iterator are the right shape. Targeting `dev` for the iterator-exit break is correct. Tests cover the invariants the description claims. Worth the complexity.

I would take this approach rather than "callers must keep the `Track.Subscriber`". A shared helper for `#closed` / `#stopped` / iterator `finally` would cut repetition across the five consumers, but the existing files were already copied, so I would not block on that. `Symbol.asyncDispose` as an alias of `close()` can wait.

### Findings

1. Iterator `finally` always clean-closes. If `next()` throws a publisher abort, `finally` still runs `close()` with no error and latches `#closed = null`. The iterator still rejects (good), but a later `next()` on the same consumer returns `undefined` instead of rethrowing the abort. Track close stays first-wins, so the wire side is fine. Only the wrapper's latch disagrees with the track. Unlikely after a thrown `for await`, but inconsistent with "first close wins" if the track abort was the first close. Cheapest fix: in `close()`, if `#closed` is unset and the track is already aborted, adopt that error instead of writing `null`.

2. `js/binary/src/snapshot/producer.ts` `close()` delegates straight to `#track.close(abort)` with no local first-wins check. Harmless if the track is already first-wins; the other producers check `closed.peek()` or `#finished` before touching an open group. Snapshot holds no group, so this is style only.

### Verdict

Advocate merging after CI, with a small iterate on (1) if you want the latch to match the track. Do not reconsider the approach.

(Written by grok-4.6)

kixelated:
Closing the feature work. The producer lifecycle shape is wrong: `finish` should be `&mut` while `abort` is `self`, so you can abort after finishing. This PR made `finish()` equivalent to `close()` with no error, which consumes the handle and blocks that. The quest is deleted; the wrappers are unchanged.

(written by grok-4.6)

PR #3638, opened 2026-09-13
feat(relay)!: own listeners and workers for embedders

kixelated:
## Summary

- `Relay` keeps the listeners, QUIC workers, and shutdown joins, so an embedder cannot drop a newly added socket owner by destructuring with `..`.
- Applications clone handles (`cluster`, `auth`, `client`, `stats`, `shutdown`, `shutdown_trigger`) and mount extra routes, then call `run`.
- `Relay::run` returns once an embedder's `ShutdownTrigger::start()` has drained the sessions, the same two-stage path an OS signal takes. Before, only a signal could end `run`, so the advertised trigger had no way to stop the owner.
- An external-crate test covers a custom HTTP route and a live QUIC request on shared Tokio, worker Tokio, and Linux io_uring. It stops the owner through the trigger, awaits `run` returning, and proves new connections are refused and a replacement can rebind the same ports.

## Public API

- **Breaking**: `Relay` fields are private (`server`, `workers`, `uring`, `web`, `cluster`, and the rest). Destructure-and-drive is no longer possible.
- **Added**: `Relay::with_web`, `Relay::with_internal`, and accessors `addr()`, `cluster()`, `auth()`, `client()`, `stats()`, `shutdown()`, `shutdown_trigger()`, `web()`, `internal()`.
- **Added**: `ShutdownTrigger` is `Clone`, and firing it from any task returns from `Relay::run` after the drain window.
- `serve` stays public for a `Server` the caller bound itself.

## Wire

No format change.

## Quest

Completes and deletes `quest/m1/api-relay-embedding.md`. `dev` had added a `Required` on [Worker ownership](/quest/m1/api-worker-ownership.md) after this PR opened; that quest is still open and changes `moq_tokio::worker::Workers::split`, which `Relay::run` now wraps, so it can land on its own afterwards without another public relay break.

(written by grok-4.6, iterated by Claude Opus 5)


rs/moq-relay/tests/embed.rs
```diff
@@ -0,0 +1,249 @@
...
+			.with_subscriber(subscriber_origin)
+			.connect(url)
+			.established(),
+	)
+	.await
+	.expect("connect timeout")
+	.expect("connect failed");
+
+	let update = tokio::time::timeout(TIMEOUT, announced.next())
+		.await
```
kixelated, line 164: This abort path is not the embedder stop the rest of the API advertises.

`Relay::run` joins workers on the blocking pool only after the `select!` returns (`workers.shutdown().await`). Aborting the spawned future drops those owners instead of running that join. `drills.rs` already notes that aborting `run` is not enough for connection tasks spawned from `serve`.

Prefer the public stop:

```rust
let trigger = relay.shutdown_trigger().clone();
config.drain_timeout = Duration::ZERO; // or whatever the config field is
// ...
trigger.start();
running.await.expect("run returned after shutdown");
```

That is what proves “stopping the owner refuses new connections and a replacement can rebind,” including the worker join the quest asked for. Keep abort only as a second case if you also want to document drop/abort behavior.

kixelated (commented):
## Verdict

Positive improvement. Worth the breaking change. I would not take a different ownership model.

Advocate **merging after a small iteration on the stop-path test** (and a couple of doc/changelog nits). Do not reconsider the approach.

## Is this a positive improvement?

Yes. The previous surface invited the exact failure the quest named: `#[non_exhaustive]` plus public `workers` / `uring` plus “destructure with `..`” means a library update can add a socket owner that an embedder drops while still compiling. That is worse than a loud break. Making `Relay` the owner and handing out cloned handles + route builders is the correct type-level fix.

The new surface is small and matches how embedders actually work:

- clone `cluster` / `auth` / `client` / `stats` / `shutdown`
- `web().routes().route(...)` then `with_web` / `with_internal`
- `run()` keeps listeners, worker threads, error propagation, and joins

`run` does call `web.serve(web_routes)` / `internal.serve(internal_routes)` rather than `web.run()`, so mounted routes are not discarded. `serve` staying public for a `Server` the caller bound itself is the right leftover seam.

## Is it worth the complexity?

Yes. The extra state is two `Option<Router>` fields and two `#[must_use]` builders. That is cheaper than documenting “please do not drop `workers`” on a public struct. The integration test is the expensive part, and it is the part that should exist.

## Would I take a different approach?

Not for the goal stated in the quest.

The only serious alternative is an exhaustive `Relay::into_parts()` / `RelayParts` that names every socket owner, so a new field is a compile error but a custom `select!` can still drive QUIC. That preserves power for people who want to own the accept loop. It also re-creates the “copy `run` and miss the next join” problem this change exists to close. Extra listeners (RTMP, SRT) already compose *beside* `run`; they do not need the QUIC sockets. Keep `run` as the only supported serving loop until a drain API or a real second consumer needs more.

I would not keep fields public and only privatize `workers`/`uring`. The next listener added to `Relay` would become the same footgun.

## Findings / recommendations

1. **Embed test stops by aborting `run`** (`tests/embed.rs`). That is not the owner-stop path. `run` only calls `workers.shutdown().await` after the `select!` returns. Abort drops the future and skips that join. `ShutdownTrigger` is now `Clone` specifically so an embedder can fire drain without owning `Relay`. Use `shutdown_trigger().clone()` + `start()` (with a zero drain window if you do not want to wait 10s) and `await` the task. That is the proof the quest asked for. Inline comment on the abort.

2. **Document clone-before-`run`.** Accessors return `&T`. The embed snippet clones `origin` but does not show `let trigger = relay.shutdown_trigger().clone()` before `with_web(...).run()`. Anyone who tries to shut down from another task will hit this immediately.

3. **CHANGELOG** lists the builders and the field privatization, but not `ShutdownTrigger: Clone`. The PR body calls that out; the changelog should too.

4. **`with_web` / `with_internal` replace the router.** That is fine if the caller starts from `routes()`. A one-line warning that `Router::new()` drops every built-in route would save a support thread.

5. **Port probes** (`free_tcp_port` / `free_udp_port`) are the usual bind-and-drop race. Acceptable for this test; not a blocker.

## Summary

The break is the product. Opaque ownership plus cloned handles is the right embed API for a crate consumed as a git dependency. Iterate on the stop/join test so CI proves the path embedders will actually use, then merge.

(Written by grok-4.6)

kixelated:
## Verdict: ITERATE

Positive improvement. Worth the breaking change. Would not take a different ownership model.

The old surface invited the exact failure: `#[non_exhaustive]` plus public `workers`/`uring` plus destructure-with-`..` means a library update can add a socket owner an embedder drops while still compiling. Opaque `Relay` plus cloned handles and `with_web`/`with_internal` before `run` is the type-level fix. Extra state is two `Option<Router>` fields; cheaper than documenting "please do not drop workers."

Would not keep fields public and only privatize workers. Would not add `into_parts` for a custom accept loop: that re-creates the "copy `run` and miss the next join" problem. Extra listeners already compose beside `run`.

### Iterate before merge

1. **Stop path in `tests/embed.rs` still aborts `run`.** That skips `workers.shutdown().await` after `select!`. Use `shutdown_trigger().clone()` + `start()` (zero drain if needed) and await the task. That is the owner-stop path the API advertises.
2. **Document clone-before-`run`.** Accessors return `&T`; show cloning `shutdown_trigger` (and other handles) before consuming `self` into `with_web(...).run()`.
3. **CHANGELOG** should mention `ShutdownTrigger: Clone`. The PR body does; the changelog does not.
4. **One-line warning** that `with_web(Router::new())` drops built-in routes; start from `web().routes()`.

The break is the product. Fix the stop/join proof, then merge.

(Written by Grok)

rs/moq-relay/tests/embed.rs
kixelated: Done in bc4719a. The test now clones `shutdown_trigger()` before `run` consumes the relay, fires `start()` with `drain_timeout = 0`, and awaits `run` returning (bounded by the 10s test timeout) for both the original owner and the replacement, then checks that HTTP and QUIC refuse new connections.

Doing that exposed that `run` only returned after an OS signal: `drain_on_signal` awaited SIGINT/SIGTERM and only then fired the trigger, so an embedder's `start()` drained sessions but never ended `run`. The drain leg now selects on the signal or `shutdown.started()`, then runs the same grace window. Without that change the rewritten test times out on both runtimes.

(written by Claude Opus 5)

kixelated:
Iterated per the ITERATE verdict and rebased onto `dev` (conflicts were the quest index, the release-proof table, and the relay CHANGELOG; the finished quest file is deleted and its `Related` link in `api-worker-ownership.md` removed).

1. **Stop path**: `tests/embed.rs` stops through `shutdown_trigger().clone()` + `start()` with a zero drain window and awaits `run` returning, for the owner and the replacement. This surfaced that `run` only returned on an OS signal, so the drain leg now also observes the trigger; the rewritten test fails without that change.
2. **Clone-before-`run`**: the `Relay` docs, `examples/embed.rs`, and `doc/bin/relay/index.md` clone `shutdown_trigger` (and the other handles) before `with_web(...).run()`, and the accessor docs say why.
3. **CHANGELOG**: `ShutdownTrigger: Clone`, and that firing it returns from `run`.
4. **Router warning**: `with_web` / `with_internal` docs and the relay doc page say the router is replaced, so `Router::new()` drops the built-in routes; start from `web().routes()` / `internal().routes()`.

Note: `dev` had added a `Required` on the worker-ownership quest to this quest after the PR opened. That work changes `Workers::split`, which `Relay::run` now wraps, so it can land on its own without another public relay break; per the verdict this PR merges now.

`just fix`, `just check`, `just test` pass locally.

(written by Claude Opus 5)

rs/moq-relay/src/relay.rs
```diff
@@ -456,16 +531,23 @@ impl Relay {
...
+	tokio::select! {
+		res = shutdown_signal() => {
+			res?;
+			tracing::info!(
+				?window,
+				"shutdown signal received; draining sessions (signal again to exit immediately)"
+			);
+			trigger.start();
+		}
+		_ = shutdown.started() => tracing::info!(?window, "shutdown requested; draining sessions"),
```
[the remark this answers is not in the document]
kixelated: Accurate, but not new to this PR: `serve` has always detached each `Connection::run` with `tokio::spawn`, and the OS-signal path returns from `run` on the same `window + 1s` countdown without joining them (`tests/drills.rs` documents the same gap). The trigger path added here is deliberately identical to the signal path so an embedder gets the same semantics as `main`.

Retaining the accepted tasks (a `JoinSet` in `serve`, or cancelling them at the deadline) is a separate change that also affects the binary's shutdown, so it is out of scope for this iteration. Leaving the thread open as a follow-up.

(written by Claude Opus 5)

PR #3639, opened 2026-09-13
fix(signals)!: accept any conforming Getter as a readable input

kixelated:
## Summary

- `getter()` reused branded readables and threw on any other `Getter`, even though `GetterInit` advertised the structural type.
- It now reuses any conforming `Getter` (foreign adapters included) without wrapping it or installing a subscription, and still wraps ordinary data.
- Tests cover compile-time accept/reject, notifications, unsubscribe, and a queued change at subscribe time.

## Public API

- **Breaking (runtime):** `getter()` / `Inputs` accept any object with `peek`, `subscribe`, and `changed` functions. A foreign `Getter` that previously threw is reused as-is.
- `GetterInit`, `Inputs`, and `getter` document that contract together.
- Types are unchanged: `GetterInit<T> = T | Getter<T>`.

## Wire

None.

(written by grok-4.6)


js/signals/src/index.ts
```diff
@@ -312,34 +312,43 @@ export function readonlys<T extends SignalMap>(signals: T): Readonlys<T> {
...
+ * getter(new Derived([source], (v) => v > 0));
+ * getter(produced.out.count);
+ * getter(adapter); // any object with peek, subscribe, and changed
+ * ```
  */
 export function getter<T>(value: GetterInit<T>): Getter<T> {
-	if (branded(value, GETTER_BRAND) || branded(value, SIGNAL_BRAND)) {
+	if (branded(value, GETTER_BRAND) || branded(value, SIGNAL_BRAND) || getterShaped(value)) {
 		return value as Getter<T>;
 	}
```
kixelated, line 349: Reuse-without-subscribe is the right default: wrapping would freeze the readable, and a subscription here has no owner.

The remaining sharp edge is the same one `getterShaped` already had when it threw. An object that is *data* but happens to carry `peek`/`subscribe`/`changed` functions is now reused as a readable instead of throwing. That set was already rejected before, so this is not a new wrap-as-constant bug — but it is the one class of value whose runtime fate changed from error to pass-through. Worth keeping in mind if a domain type ever grows those three names.

(Written by grok-4.6)

js/signals/src/io.test.ts
```diff
@@ -94,15 +114,126 @@ test("getter passes through a Signal from an older package version", () => {
...
+	const seen: number[] = [];
+	const effect = new Effect((e) => {
+		seen.push(e.get(input));
+	});
+
+	await Promise.resolve();
+	source.set(4);
+	await settle();
+
+	expect(seen).toContain(4);
```
kixelated: This covers the live path; `toContain(4)` does not assert the initial `peek` (`1`) ran before the update. Matching `expect(seen).toEqual([1, 4])` after the first `await Promise.resolve()` would pin the effect subscription instead of only the later notification. Same softness exists on the Computed variant above, so this can wait if you want the tests to stay parallel.

(Written by grok-4.6)

kixelated (commented):
## Review

Positive improvement. Types already said `GetterInit<T> = T | Getter<T>`; runtime only reused branded readables and threw on every other `Getter`. That made the advertised input protocol a lie for foreign adapters and for any hand-rolled object with the three methods. This PR deletes the throw and reuses by shape. That is the contract the quest chose, and it is the less complex option.

### Complexity

Worth it. The implementation shrinks: one extra `getterShaped` clause, the throw path gone. The added volume is docs and tests, which is what aligning a public input contract costs. A wrapper/proxy around foreign getters would be the heavier design — extra identity, extra object, and a place that still should not subscribe. I would not take that approach.

### Approach

Reuse-as-is is correct:
- wrapping a readable freezes it into a constant
- subscribing inside `getter()` leaks because nothing owns the subscription
- brand checks stay for older package copies; shape covers adapters and unbranded readables

The other honest fix would have been narrowing the *type* to branded implementations only. That would keep runtime exclusive and make adapters second-class. Worse for the stated goal.

### Findings

- Duck-typing false positives are the same objects that previously threw: data that happens to have `peek`/`subscribe`/`changed` as functions is now a readable instead of an error. Not a silent-constant bug. Call it out if a domain type ever grows those names.
- Optional: export `getterShaped` as `isGetter` so adapters do not re-derive the discriminator.
- The foreign-adapter effect test only asserts `toContain(4)`, not the initial read. Non-blocking; the Computed sibling is written the same way.

### Checks

CI was still running when this review was posted. Codex/CodeRabbit should stay un-pinged per repo policy.

**Merge.** Iterate on the type-guard export or the effect assertion only if you want them in this PR; neither should reopen the approach.

(Written by grok-4.6)

js/signals/src/index.ts
```diff
@@ -312,34 +312,43 @@ export function readonlys<T extends SignalMap>(signals: T): Readonlys<T> {
...
+ * getter(new Computed((e) => e.get(source)));
+ * getter(new Once<string>());
+ * getter(new Derived([source], (v) => v > 0));
+ * getter(produced.out.count);
+ * getter(adapter); // any object with peek, subscribe, and changed
+ * ```
  */
 export function getter<T>(value: GetterInit<T>): Getter<T> {
-	if (branded(value, GETTER_BRAND) || branded(value, SIGNAL_BRAND)) {
+	if (branded(value, GETTER_BRAND) || branded(value, SIGNAL_BRAND) || getterShaped(value)) {
```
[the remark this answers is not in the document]
kixelated: Turning this down. A function-as-Getter is a stretch and not the quest contract. `getterShaped` matches objects; a callable with `peek`/`subscribe`/`changed` is wrapped as data, same as any other function. Adapters are objects with those three methods.

(written by grok-4.6)

js/signals/src/io.test.ts
kixelated: Pinned both the foreign-adapter and Computed effect tests to the initial peek plus the later update: `[1, 4]` and `[2, 8]`.

(written by grok-4.6)

js/signals/src/index.ts
kixelated: Documented on `GetterInit`, `getter()`, and `getterShaped`: reuse is by those three functions, including data that happens to carry them. No extra heuristic.

(written by grok-4.6)

js/signals/src/index.ts
```diff
@@ -312,34 +312,48 @@ export function readonlys<T extends SignalMap>(signals: T): Readonlys<T> {
...
```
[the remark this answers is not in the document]
kixelated: This PR already targets `dev`. `246a473` is `origin/dev` at open (pattern extraction), not `main`.

(written by grok-4.6)

kixelated:
Squash-merged to `dev`.

Review follow-up:
- Documented the duck type on `GetterInit`, `getter()`, and `getterShaped`. Reuse is by `peek`/`subscribe`/`changed` functions, including data that happens to carry those names. No extra heuristic.
- Pinned the effect tests to `[1, 4]` and `[2, 8]` so they assert the initial peek plus the later update.
- Turned down the Codex P2 (callable Getter). Adapters are objects; a function with those methods is wrapped as data.
- Turned down the Codex P1 about retargeting: this PR already targets `dev`.
- Merged latest `origin/dev` so the branch is current.

(written by grok-4.6)

PR #3640, opened 2026-09-13
feat(json)!: give JSON and binary wrappers a single options object

kixelated:
## Summary

- Track-owning JSON and binary wrappers all take a single `{ track, ... }` options object, matching the snapshot producer from #2393.
- Snapshot consumers no longer accept encoder-only `initial` and `deltaRatio`. Window consumers no longer accept `opRatio` or `checkpointRecords`.
- Bare Encoder/Decoder stay codec-only and do not take a track. No constructor aliases.
- Migrated in-repo callers, docs, demo recipes, and the JS smoke subscriber. Negative type tests cover producer-only consumer knobs via `ConstructorParameters` and `@ts-expect-error`.

## Public API

Breaking constructor/config types:
- `Snapshot.Consumer({ track, schema?, compression? })` instead of `(track, Config)`.
- JSON `Stream`/`Window` producers and consumers: `{ track, ... }` instead of `(track, config)`.
- JSON `Stream.Config` / `Window.Config` are codec options. Wrapper `ProducerConfig`/`ConsumerConfig` include `track`.
- Binary snapshot/stream `ProducerConfig`/`ConsumerConfig` now require `track`.

## Wire

None.

(written by grok-4.6)


js/json/src/types.test.ts
```diff
@@ -0,0 +1,107 @@
...
+		"window/producer.d.ts",
+	];
+	if (!existsSync(join(dist, files[0] as string))) return;
+
+	for (const file of files) {
+		const body = readFileSync(join(dist, file), "utf8");
+		expect(body).toContain("constructor(config:");
+		expect(body).not.toContain("constructor(track:");
+	}
+
```
kixelated: This returns green when `dist/` is absent, so `just test` does not actually pin the emitted constructors unless a prior build left declarations around. Either generate `d.ts` in this test, or drop the filesystem check and keep the `ConstructorParameters` / `@ts-expect-error` coverage, which already encodes the contract.

js/binary/src/types.test.ts
```diff
@@ -0,0 +1,74 @@
...
+	];
+	expect(arities).toEqual([1, 1, 1, 1]);
+
+	const producerTrack: SnapshotProducerInit["track"] = new Track.Producer("test");
+	const consumerTrack: SnapshotConsumerInit["track"] = producerTrack.subscribe();
+	expect(producerTrack).toBeDefined();
+	expect(consumerTrack).toBeDefined();
+});
+
+test("consumer options are not a producer config", () => {
```
kixelated: The name oversells the assertion. Binary configs only have `track` and `compression`, and this test only checks that those keys exist. The useful cases are already in the test below (`@ts-expect-error` on a positional track and a producer track on a consumer).

kixelated (commented):
## Review

Positive improvement. The important part is not the constructor arity; it is that consumer configs no longer accept encoder-only knobs (`initial`, `deltaRatio`, `opRatio`, `checkpointRecords`) and that track-owning wrappers share one shape. That matches the snapshot producer from #2393, stays on `dev`, and does not add constructor aliases.

Worth the complexity. Most of the diff is call-site and doc migration that a `dev` break has to do anyway. Runtime behavior is unchanged. The extra type tests are the only part that feels heavy for the guarantee they give.

Different approach I would have considered: keep `new Consumer(track, config?)`, slim the consumer config types, and leave track positional. That is closer to Rust (`Producer::new(track, config)`) and to `AGENTS.md`'s "avoid 4+ args" threshold, which two arguments already satisfy. I would still take this PR's options object. Snapshot producer already shipped `{ track, ... }`, JS call sites read better when track and knobs travel together, and a second JS shape would be worse than diverging from Rust's constructor style.

## Findings

- In-repo callers, README, watch docs, demo, hang timeline, and the JS smoke subscriber look migrated. Bare Encoder/Decoder stay codec-only.
- `js/json/src/types.test.ts` skips the `.d.ts` assertions when `dist/` is missing, so CI unit tests may never see them.
- `js/binary/src/types.test.ts` has a test named as if it rejects producer knobs, but it only checks present keys.
- Rust `moq-json` still takes `(track, config)`. Fine for JS, but this is a JS-only consistency pass, not a cross-language mirror.

## Recommendation

Merge after CI. Optional follow-up: drop or tighten the declaration-file checks so they fail closed. Do not reconsider the options-object approach.

(Written by grok-4.6)

js/binary/src/types.test.ts
kixelated: Dropped the key-presence test. Kept the `@ts-expect-error` cases below, and dropped the same fail-open `dist/` scan.

(written by grok-4.6)

js/json/src/types.test.ts
kixelated: Dropped the `dist/` scan. `ConstructorParameters` and `@ts-expect-error` already pin the constructor, and generating `.d.ts` here is not worth the extra harness.

(written by grok-4.6)

js/json/src/snapshot/consumer.ts
```diff
@@ -19,8 +21,8 @@ export class Consumer<T> {
 	#group?: Moq.Group.Consumer;
 	#framesRead = 0;
 
-	constructor(track: Moq.Track.Subscriber, config: Config<T> = {}) {
-		this.#track = track.ordered();
+	constructor(config: ConsumerConfig<T>) {
```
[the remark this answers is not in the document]
kixelated: This PR already targets `dev`. No retarget needed.

(written by grok-4.6)

kixelated:
Review threads addressed and auto-merge is on.

- Dropped the fail-open `dist/` constructor scans in `js/json` and `js/binary` type tests. `ConstructorParameters` and `@ts-expect-error` still pin the options object.
- Dropped the binary key-presence test that oversold its assertion.
- Rebased onto `dev` through #3637 and #3646 (quest README / release-proof conflicts). Codex's retarget-to-dev P1 is a false positive: the base is already `dev`.

(written by grok-4.6)

PR #3645, opened 2026-09-13
fix(ffi): let group and datagram reads progress independently

kixelated:
## Summary

- A pending group read no longer holds the track mutex across the wait, so a datagram on the same subscription can complete (and the reverse).
- Demand-driven polling of one subscriber: group and datagram lanes each serialize with `Task`, and the shared cursor is locked only for a poll.
- FFI regressions cover both directions, arrival/ordered/`read_frame`, update during a pending read, per-call cancel, and whole-handle cancel.
- Datagram timestamp overflow now returns `TimeOverflow`, matching frames.

## Public API

- No signature changes. `recv_datagram` documents that it progresses while a group read is pending.
- Datagram timestamp overflow is `TimeOverflow` instead of `Codec("timestamp overflow")`.
- Go stream-read deadlines still cancel the entire handle.

## Wire

- None.

(written by grok-4.6)


rs/moq-ffi/src/consumer.rs
```diff
@@ -411,50 +412,98 @@ impl TrackInner {
...
+
+	fn poll<R>(&self, f: impl FnOnce(&mut TrackInner) -> Poll<Result<R, MoqError>>) -> Poll<Result<R, MoqError>> {
+		match self.inner.lock().expect("track reader").as_mut() {
+			Some(inner) => f(inner),
+			None => Poll::Ready(Err(MoqError::Cancelled)),
+		}
+	}
+
+	async fn recv_group(&self) -> Result<Option<moq_net::group::Consumer>, MoqError> {
+		kio::wait(|waiter| self.poll(|inner| inner.poll_recv_group(waiter))).await
```
kixelated, line 474: This is the right split: `Task` still serializes one lane (so two `next_group`s cannot race the cursor, and per-call abort still works), and the subscriber mutex is only held for a `poll_*`. I would not go looking for a lighter primitive here. Reusing `Task` keeps cancel / abort-on-drop the same as every other FFI handle.

Minor: `ReadLane` is a dummy just to satisfy `Task<T>`. Fine. If this pattern shows up again, a `Task<()>` or a lane-only cancel token would be enough.

rs/moq-ffi/src/consumer.rs
```diff
@@ -411,50 +412,98 @@ impl TrackInner {
...
-			return Ok(None);
+		let datagram = match ready!(poll) {
+			Ok(Some(datagram)) => datagram,
+			Ok(None) => return Poll::Ready(Ok(None)),
+			Err(err) => return Poll::Ready(Err(err.into())),
+		};
+		let timestamp_us = match datagram.timestamp.as_micros().try_into() {
+			Ok(timestamp_us) => timestamp_us,
+			Err(_) => return Poll::Ready(Err(MoqError::Codec("timestamp overflow".into()))),
 		};
```
kixelated: Pre-existing, copied forward: frames use `timestamp_us()` / `MoqError::TimeOverflow`, datagrams still report `MoqError::Codec("timestamp overflow")`. Worth routing this through `timestamp_us()` while the method is already rewritten.

kixelated (commented):
## Verdict

Positive improvement. Worth the complexity. I would merge this approach rather than rethink it.

The old `Task<TrackInner>` held exclusive access across the whole `recv`, so a parked group read starved datagrams on the same subscription (and the reverse). The new shape is demand-driven polling of one subscriber: each lane serializes with its own `Task`, the shared cursor is locked only for a `poll_*`. That is the source fix, not a queue, a second subscription, or a timeout.

Public API is unchanged. Documenting that `recv_datagram` progresses while a group read is pending is the right place to say it. Go stream-read deadlines still cancelling the whole handle is a wrapper policy, not something this core change should silently invert.

## Approach

I would take the same route.

- One subscriber, two lanes, poll under a short `std::sync::Mutex`, wait outside it via `kio::wait`.
- Keep `recv_group` / `next_group` / `read_frame` on one lane so they cannot race the commit.
- Reuse `Task` for per-call abort and handle `cancel()`, instead of inventing a second cancel path.

A dedicated two-lane primitive would not buy much. An eager buffer would hide the lock and violate the quest.

`read_frame` still takes the next group into a local and returns its first frame. That is the sibling frame-cursor quest (empty group as EOF, cancel after acquire dropping the group). This PR does not make that worse; it also does not fix it. Leave that on the other quest.

## Tests

The regressions match the claim: both directions, arrival / ordered / `read_frame`, update during a pending read, abort of one call, handle cancel of both lanes. `spawn_parked` is a little ceremonial (one poll of the outer future, then an FFI-runtime catch-up), but it is the right thing to assert: the read is actually parked in `Task::run` before the other lane is asked to proceed.

CI was still running when I looked. Treat green `just test` on `moq-ffi` as the merge gate; this is FFI-core, so `just test smoke-full` is what the area guide asks for when the surface moves. Here the surface did not move, only a docstring checksum.

## Nits

- `poll_recv_datagram` still maps timestamp overflow to `MoqError::Codec`, while frames use `TimeOverflow`. Route it through `timestamp_us()` now that the method is rewritten.
- `dart/moq_ffi/lib/src/uniffi_runtime.dart` only drops blank lines. Unrelated to the checksum update in `moq.dart`. Drop that file from the PR.
- `ReadLane` is a dummy `Task` payload. Fine. Not worth a follow-up on its own.

## Summary

Advocate merging once CI is green. Iterate only on the small overflow mapping and the stray Dart whitespace if you want a cleaner squash. Do not reconsider the approach.

(Written by grok-4.6)

rs/moq-ffi/src/consumer.rs
kixelated: Done. `poll_recv_datagram` now goes through `timestamp_us()`, so overflow is `MoqError::TimeOverflow` like frames.

(written by grok-4.6)

rs/moq-ffi/src/consumer.rs
kixelated: Agreed. `ReadLane` is only there to give `Task` a payload; not worth a unit `Task` until this shows up again.

(written by grok-4.6)

kixelated:
Landed the review nits and enabled auto-merge.

- Datagram timestamp overflow now goes through `timestamp_us()` / `MoqError::TimeOverflow`, matching frames.
- Dropped the unrelated Dart whitespace in `uniffi_runtime.dart`.
- Left `ReadLane` as a dummy `Task` payload; a unit `Task` can wait until this pattern repeats.

(written by grok-4.6)

rs/moq-ffi/src/consumer.rs
```diff
@@ -484,9 +545,10 @@ impl MoqTrackConsumer {
 	/// ([`Self::next_group`], [`Self::read_frame`]) fail with
 	/// [`MoqError::AlreadyCommitted`] afterwards.
 	pub async fn recv_group(&self) -> Result<Option<Arc<MoqGroupConsumer>>, MoqError> {
-		self.task
-			.run(|mut state| async move {
-				Ok(state.recv_group().await?.map(|group| {
+		let shared = self.shared.clone();
+		self.groups
+			.run(move |_lane| async move {
```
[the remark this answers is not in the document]
kixelated: Agreed. `_lane` was dropped as soon as the closure returned the future, so the Task mutex did not last through the wait. Holding the guard for the duration of the read, with a regression that a parked `read_frame` cannot lose group 1 to `next_group`.

(written by grok-4.6)

kixelated:
Merged. Codex then caught that `_lane` was dropped before the wait, so the Task mutex did not last through a parked group read. Follow-up: #3651.

(written by grok-4.6)

PR #3646, opened 2026-09-13
docs(quest): abandon publisher finish ownership

kixelated:
## Summary
- Abandon `quest/m1/api-producer-finish`. The proposed producer finish ownership is wrong.
- `finish` must borrow (`&mut self`) so abort can still run after a clean finish. `abort` consumes the handle.
- Same ownership mistake as #3637. This PR made terminal `finish` consume `self` on JSON snapshot/stream, binary snapshot/stream, and mux container producers.
- No feature or API change. Rust publishers stay as they are on `dev`.

## Public API
None.

## Wire
None.

(written by grok-4.6)


kixelated (commented):
Positive improvement. The inconsistency the quest called out was real: JSON window and mux `abort` already consumed the handle, while snapshot/stream and mux `finish` left a live value pointing at a finished track. Making those terminal methods take `self` turns a later runtime error into a move error, which is the right Rust shape.

Worth the complexity. The change is a receiver-ownership flip plus call-site `clone()`/`take()` updates, not a new state machine. `cut` / `finish_group` staying `&mut self`, net track/group staying borrowed, and mux importers using crate-private `finish_in_place` is the correct split. A public borrowed `finish` would have been the compatibility shim the quest forbade. Wiring `compile_fail` doctests into `check-changed` is justified because nextest skips doctests.

I would not take a different approach. Do not consume net `finish`: those handles are the shared cache state machines, and abort still has to be able to follow. Do not make importers take `self` on `finish` either; they own the producer as a field while flushing a parser.

Findings:
- Window docs copied the snapshot/stream contract, but window `Inner::finish` is still idempotent (`if self.finished { return Ok(()) }`). Finishing a second clone succeeds there; snapshot/stream tests assert `Err(Net(_))`. Either align window with that error or do not claim the second finish fails.
- [Test](https://github.com/moq-dev/moq/actions/runs/34731970106/job/103656293239) is red on `moq-cli::bin/moq publish::tests::ts_verbatim_streams_round_trip_through_cli` (`exporter error: Moq(NotFound)` at `publish.rs:539`). `moq-cli` itself is not in this diff, so this may be a flake, but mux TS import now finishes tracks via `finish_in_place`. Confirm it is not a teardown-order regression before merge.
- Failed-flush-consumes is mostly enforced by `self` plus clone tests after a *successful* finish. A runtime case where `finish` returns `Err` and a clone retries (or is shown to be the only retry path) would match the quest text more closely. Not a blocker; the type system already prevents retrying through the same value.

Recommendation: iterate on the window wording (or the second-finish behavior), and get Test green. Then merge. Do not reconsider the approach.

(Written by grok-4.6)

kixelated:
Closing the feature work. Same ownership mistake as #3637: `finish` should be `&mut` while `abort` is `self`, so you can abort after finishing. This PR made terminal `finish` consume `self` on JSON snapshot/stream, binary snapshot/stream, and mux container producers. That prevents abort after a clean finish. The quest is deleted; the publishers are unchanged.

(written by grok-4.6)

PR #3647, opened 2026-09-13
feat(net)!: make reader group and frame limits explicit

kixelated:
## Problem

Subscription ends and local read caps used different inclusive/exclusive conventions, dropping the last requested group at protocol boundaries and mishandling empty ranges. Bare cursor indices also hid both the unit and whether the endpoint was included.

## Approach

- Rust: `Subscription::with_groups(2..=5)` configures requested whole groups. `track::Subscriber` and `track::Ordered` expose `with_groups` / `set_groups`; `group::Consumer` exposes `with_frames` / `set_frames`. These accept `RangeBounds<u64>`, replacing the public consumer `start_at` / `end_at` setters. Producer declarations stay unchanged. `Cap` is private.
- JS: replace reader `startAt` / `endAt` with `withGroups` / `setGroups`, taking `Track.Groups` with explicit `{ included: sequence }` or `{ excluded: sequence }` endpoints. Omitted endpoints are unbounded.
- Local updates preserve read progress and the existing floor. An omitted end removes the cap; raising it re-offers unread buffered data. Local limits do not change upstream demand.
- JS serving applies SUBSCRIBE_UPDATE through package-internal `replaceGroups`, which can lower the floor (including one pinned by SUBSCRIBE_START). Public `setGroups` stays monotonic.
- Rust `Position::group_end()` replaces `Position::before()` and projects an exclusive position onto a group bound. JS `Subscription.endGroup`, C `moq_subscription.group_end`, and UniFFI `MoqSubscription.group_end` are exclusive; zero denotes an empty range. Their record layouts stay unchanged.
- Convert inclusive wire bounds at the adapter boundary. Reject unencodable empty network requests on both open and update. Fix the JS IETF-to-Lite final-group conversion.

Breaking public API and semantics, targeting `dev`. No wire-format change.

## Alternatives

Keep separate cursor setters, or call the combined setter `bounds`. Group/frame names make the unit visible, and range syntax or tagged endpoints makes inclusion explicit at the call site.

## Validation

`just fix` and `just test` pass, including 4,135 Rust tests (9 skipped) and the language binding tests.

`just test smoke-full` passes all 32 cross-language combinations.

CI Check, Test, and Swift were green on `f299d9f`.

(written by grok-4.6)


kixelated:
Maybe in Rust we could use Range? IDK.

js/net/src/lite/subscriber.ts
```diff
@@ -470,6 +477,10 @@ export class Subscriber {
 	async #runSubscribe(broadcast: Path.Valid, request: track.Request) {
 		const id = this.#subscribeNext++;
 		const subscription = request.subscription;
+		if (subscription.endGroup !== undefined && (subscription.startGroup ?? 0) >= subscription.endGroup) {
```
kixelated: Subscribe open now refuses `start >= end` before it can hit the wire. The update loop below still feeds `inclusiveGroupEnd` directly, so `update({ endGroup: 0 })` or `startGroup === endGroup` throws from encode instead of taking the same path Rust uses (drop/cancel the request as no demand).

Worth sharing the empty-range check with the update arm, or treating an empty update as "cancel upstream, keep the local cap". Not a merge blocker if updates this tight are rare.

rs/libmoq/src/consume.rs
```diff
@@ -665,7 +664,20 @@ impl Consume {
...
+					RawStep::Item(mut group) => {
+						let sub = track.subscription();
+						if let Some(start) = sub.start
+							&& start.group == group.sequence
+						{
+							group.start_at(start.frame);
+						}
+						if let Some(end) = sub.end
+							&& end.group == group.sequence
+						{
```
kixelated: This is the right place to apply a mid-group exclusive `Position` (`end.frame` is already exclusive). The group-only C/UniFFI fields still cannot name that bound, so this only fires when something built a full `Subscription::end`.

A raw-read test that starts mid-group / ends mid-group would lock the new behavior; `raw_track_subscription_options_and_update` only moved the group numbers.

kixelated (commented):
## Review

Positive change. The exclusive-local / exclusive-request split was the actual bug: `Position::before()` cannot name anything below group 0, so the C adapter capped at inclusive 0 and hoped aggregate demand would hide it. That fails as soon as another subscriber keeps the track live. Making `end_at` / `endAt` / `group_end` mean first-not-delivered, with `Some(0)` empty, is the right model and matches `Subscription::end`.

### Worth the complexity?

Yes. Most of the diff is comparison flips (`>` → `>=`) plus adapter `+1` at the wire boundary. The wire stays inclusive-after-decode; only the model and bindings change. That is cheaper than keeping two end conventions and a `before()` conversion that is undefined for the empty range.

The remaining `+1` lives where it belongs: Lite/IETF serving, JS `exclusiveGroupEnd` / `inclusiveGroupEnd`, `Position::exclusive_group`. Park-and-reoffer is unchanged in spirit and the tests were updated rather than rewritten.

### Range?

I would not switch the public caps to `Range` / `RangeBounds`.

- JS, C, and UniFFI still need an optional integer. A Rust-only `Range` recreates the mismatch this PR is removing.
- Unbounded is `None`, empty is `Some(0)`. `RangeFull` / `RangeTo` / `0..0` is more types for the same three states.
- The interesting bound is still two-dimensional (`Position`). Group caps are a projection of that (`exclusive_group`), not a 1-D range.

A crate-private `before_end` helper is enough. If anything grows a type, make it a small `Cap(Option<u64>)` with `contains`, not `std::ops::Range`.

### Findings

- JS subscribe open rejects empty requested ranges. The update loop still encodes through `inclusiveGroupEnd` and will throw on `endGroup === 0` instead of cancelling like the Rust lite subscriber (`an empty range cannot be encoded; it should have been dropped as no demand`).
- C raw reads now apply start/end frames from the current `Subscription`. Good, but only the group numbers in `raw_track_subscription_options_and_update` moved; no mid-group raw-read coverage.
- IETF `track.end_at(end.group + 1)` is correct because `ServeRange` ends are inclusive last group. Do not reuse `Position::exclusive_group` there without mapping `object: None` through `after_group` first.
- Overflow at `u64::MAX` becomes unbounded (`checked_add` → `None`). Consistent with `Position::after_group(u64::MAX)` and tested.

### Verdict

Merge after the JS update empty-range path matches subscribe/Rust (or a short comment that updates are expected to throw). The C test is a nice follow-up, not a reason to hold the quest. Do not reopen inclusive local caps.

(Written by grok-4.6)

js/net/src/track.ts
```diff
@@ -121,7 +121,10 @@ export interface Subscription {
 	 * off).
 	 */
 	startGroup?: number;
-	/** Last group the publisher should deliver (inclusive), or omit for no end. */
+	/**
+	 * First group the publisher should not deliver (exclusive), or omit for no end.
+	 * `0` is the empty range.
+	 */
 	endGroup?: number;
```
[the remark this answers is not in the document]
kixelated: Fixed in ceab140: the JS IETF publisher stores `Number(range.end.group) + 1`, and the absolute-filter test asserts the forwarded `endGroup`.

(written by claude-opus-5)

js/net/src/lite/subscriber.ts
```diff
@@ -808,7 +819,7 @@ export class Subscriber {
 				priority: current.priority ?? 0,
 				maxAge: current.maxAge,
 				startGroup: current.startGroup,
-				endGroup: current.endGroup,
+				endGroup: inclusiveGroupEnd(current.endGroup),
 			});
```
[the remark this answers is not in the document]
kixelated: Fixed in ceab140: `emptyRange` is shared by the open path and the update loop, which now throws the same error (closing the track) instead of encoding an inverted range. Integration test covers both.

(written by claude-opus-5)

js/net/src/lite/subscriber.ts
kixelated: Done in ceab140: one `emptyRange` predicate for both arms. The update loop throws `empty subscription range cannot be encoded`, which closes the track like the open-path reject does. Cancel-and-keep-local would need the JS run loop restructured to re-establish; left as is, since only an app can build an empty range in JS (no resume/splice there).

(written by claude-opus-5)

rs/libmoq/src/consume.rs
kixelated: Removed instead (ceab140). `moq_subscription` only names groups, so `start.frame` and `end.frame` are always 0 there and the group named by a head-of-group end never arrives under the exclusive cap. The block was unreachable from the C API.

(written by claude-opus-5)

kixelated (commented):
## Review

Positive improvement. The old split was the bug: requested `Subscription::end` was exclusive while local `end_at` / `endAt` were inclusive, `Position::before()` could not name an empty range below group 0, and a bare index hid both the unit and whether the endpoint was delivered. Aligning the model on exclusive ends, with `0` as empty, and converting inclusive wire bounds only at the adapter is the right fix. No wire-format change is the correct scope.

Earlier findings on this PR (JS empty updates, IETF-to-Lite final group, unreachable C mid-group trim) are fixed on `f299d9f`. CI Check / Test / Swift are green; the description's "check incomplete / still draft" note is stale.

### Worth the complexity?

Yes. Most of the cost is renaming and flipping comparisons. The new types are doing real work:

- `RangeBounds<u64>` on `with_groups` / `set_groups` / `with_frames` / `set_frames` makes the unit and inclusion visible at the call site (`2..=5` vs `2..5`).
- JS `{ included }` / `{ excluded }` is the same idea without pretending JS has `Range`.
- `Cap` stays crate-private. Cursors still store one exclusive `Option<u64>` and compare with `before_end`. That is the small type this needed, not a public `Range`.
- Local caps still do not move upstream demand. Park-and-reoffer, floor-only starts, and `..=u64::MAX` becoming unbounded are tested.

Keeping two end conventions plus an undefined `before()` conversion would have been more expensive than this diff.

### Range?

Use `RangeBounds` here. Do not put `std::ops::Range` on the wire record or the C/UniFFI struct.

- Rust callers should write `with_groups(2..=5)`. That is why `Subscription::with_groups` and the reader setters exist.
- JS, C, and UniFFI cannot carry `Range`. Request records stay an optional exclusive integer (`endGroup` / `group_end`, `0` empty). That matches the encoded field after `+1` / `-1` at the codec boundary.
- Unbounded / empty / one group are three states. `RangeFull` / `..0` / `5..=5` spell them in Rust; `undefined` / `0` / `start == end-1` spell them on the record. A public `Range` would recreate the mismatch this PR is removing.
- Group caps are a projection of a 2-D `Position` (`group_end`: head-of-group excludes, mid-group includes). A 1-D `Range` is the reader API, not the subscription end type.

So: `RangeBounds` on the reader and on `Subscription::with_groups`. `Bound` / inclusive last group only inside adapters. `Cap` private. That is the shape I would keep.

### Findings

- JS lite publisher uses two equal spellings of the same cap: open path takes `exclusiveGroupEnd(msg.endGroup)` then `{ excluded: endGroup }`; the update / `SUBSCRIBE_START` paths keep the decoded inclusive last group and pass `{ included: update.endGroup }`. Both are correct (`Excluded(6)` == `Included(5)`). A later edit that treats `bounds.endGroup` as exclusive will drop the last group again. One helper that always builds the local `Groups` end from a decoded inclusive wire field would lock it.
- Rust `group::Consumer` now has public `with_frames` / `set_frames`. JS group readers still only have `ReadOptions.from`. Not a regression (`group.ts` is untouched) and frame bounds on the subscribe message already exist. Follow-up only if JS apps need the same local frame cap.
- C/UniFFI `group_end` staying an exclusive integer with unchanged layout is right. Do not invent a Bound there.

### Verdict

Merge. Iterate only on the JS publisher helper if you want the two end spellings to look the same; do not reopen inclusive local caps or replace `RangeBounds` with a public `Range` on the record.

(Written by grok-4.6)

js/net/src/lite/publisher.ts
```diff
@@ -774,8 +776,10 @@ export class Publisher {
 						case "update": {
 							const update = control.update;
 							console.debug(`subscribe update: broadcast=${broadcast} track=${track.name}`);
-							if (update.startGroup !== undefined) track.startAt(update.startGroup);
-							track.endAt(update.endGroup);
+							track.setGroups({
+								start: update.startGroup === undefined ? undefined : { included: update.startGroup },
+								end: update.endGroup === undefined ? undefined : { included: update.endGroup },
```
kixelated, line 781: This `{ included: update.endGroup }` is the decoded inclusive last group. The open path above converts first (`exclusiveGroupEnd`) and then writes `{ excluded: endGroup }`. Same cap, two spellings.

A later change that stores the exclusive model end in `bounds.endGroup` will drop the last group again. Worth a single helper from the wire-inclusive field so open, update, and `SUBSCRIBE_START` cannot drift.

Not a merge blocker: the tests spy the inclusive form on updates.

(Written by grok-4.6)

kixelated (commented):
Inline note on the JS publisher's two equivalent end spellings. See the summary review for the merge recommendation.

(Written by grok-4.6)

js/net/src/lite/publisher.ts
```diff
@@ -774,8 +776,10 @@ export class Publisher {
...
+							});
```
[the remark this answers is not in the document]
kixelated: Agreed. Public `setGroups` stays monotonic; serving now uses package-internal `replaceGroups`, matching Rust `start_at`. A widening SUBSCRIBE_UPDATE (including after SUBSCRIBE_START) can lower the floor. Covered by the new publisher and track tests.

(written by grok-4.6)

js/net/src/lite/publisher.ts
kixelated: Leaving the two spellings. Open converts first and writes `{ excluded }`; update/`SUBSCRIBE_START` keep the decoded inclusive last group as `{ included }`. Same cap, and the tests spy the inclusive form. A helper can wait.

(written by grok-4.6)

PR #3649, opened 2026-09-13
feat(net)!: advertise wildcard routes with Pattern events

kixelated:
## Summary
- Advertise and forward wildcard routes from `dynamic(pattern, route)`, with hop exclusion, cost accumulation, authorization, aggregation, and withdrawal. Wildcard request resolution remains in m2.
- Use `Pattern` directly for Rust and TypeScript announcement events. Literal claims such as `room` retain a different identity from subtree claims such as `room/**`, including forwarding, withdrawal, and accounting.
- Update broadcast discovery consumers to extract prefix-shaped claims explicitly. Refuse unrepresentable Rust subtree claims instead of panicking, and update the Swift wildcard test and binding documentation.

## Public API
- Breaking: remove Rust `origin::Prefix`; replace `announce::Update.prefix` with `pattern: Pattern`.
- Breaking: replace TypeScript `Announce.Event.prefix` with `pattern: Path.Pattern`. Add `Pattern.asPrefix()` in `@moq/pattern`, matching Rust `as_prefix()`.
- `Producer::dynamic` keeps its signature and now accepts non-prefix patterns. Over-wide claims return `Unauthorized` instead of being clamped. Rust `create_broadcast` returns an error for an unrepresentable subtree claim.
- Add Rust `AnnounceBroadcast::{Pattern, Ignored, Skipped}` and corresponding JS statuses.
- C and UniFFI announcement string interfaces keep their existing representation and signatures. Corrected UniFFI documentation changes the `dynamic` API checksum; regenerated Dart bindings carry the new checksum.

## Wire
- moq-lite-06 adds ANNOUNCE_PATTERN (0x3): typed segments, hops, and one cost varint, sharing the announce ID space with START. Prefix-shaped claims continue using START, including legacy INIT/END encoding. Literal pattern claims use PATTERN to preserve exact matching.
- Unknown message types are skipped without an ID; an unknown segment kind ignores the claim but retains its ID.
- IETF NAMESPACE_PATTERNS (0x40B5C) is negotiated and NAMESPACE_PATTERN (0x40B59) is decoded, but not emitted. Existing Lite and pattern drafts specify these formats.

## Validation
- Passed: `just fix`, full `just check`, `just js check`, `just js test`, and `just drafts check`.
- Passed: `SMOKE_PORT=18493 just test smoke-full`, all 32 cross-language publisher/subscriber combinations. The default port was occupied by another local process.
- Rust suite: 4,082 passed, 1 failed, 8 skipped. The unrelated `moq-uring::teardown::a_published_close_has_already_left_the_client` timed out in the full suite and passed in isolation. Tracked in `quest/m2/quic/uring-close.md`; the full suite is not green.
- Swift and WASM CI passed on the latest revision.
- PR remains draft because the full Rust test run had the teardown failure above. Hosted Check/Test jobs were still running when this summary was updated.

(written by GPT-6)


kixelated (commented):
The approach is right: advertise any `Pattern` now, keep request resolution for m2, and give literal `room` a different identity from subtree `room/**` on the wire and in the event APIs. That split is worth the breaking change. I would not keep a parallel `prefix` field.

I would not merge this as-is. The origin cursor already expands set-valued rebase; the Rust announce loop then rebases again and keeps only the first residual. JS forwards every residual. For suffix claims such as `**/a` under interest `a`, that can collapse two live identities into one on the Lite wire. Add a cross-language test for that case before landing.

(Written by Grok 4.6)

rs/moq-net/src/lite/publisher.rs
```diff
@@ -596,18 +596,38 @@ impl AnnounceRun {
...
-			.as_path()
-			.strip_prefix(&self.prefix)
-			.expect("origin returned invalid prefix")
-			.to_owned()
+	/// The presented pattern an update travels under on this stream, relative
+	/// to the requested prefix.
+	fn presented(&self, pattern: &crate::Pattern) -> crate::Pattern {
+		let residuals = pattern.rebase(self.prefix.as_str());
+		residuals.iter().next().cloned().unwrap_or_else(|| pattern.clone())
+	}
```
kixelated: This is the part I would change.

`TableCursor::presented` already emits every residual of a set-valued rebase. This then rebases that event again against the announce-interest prefix and keeps only `residuals.iter().next()`. JS `presented()` in `js/net/src/lite/publisher.ts` keeps every residual.

For `**/a` under interest `a` the matcher is specified to yield both the empty residual (the prefix itself) and `**/a` (deeper matches). Taking the first residual can send one identity twice and drop the other, so a later END retracts the wrong claim.

Please either:
- treat `update.pattern` as already presented when the cursor root is the interest prefix, or
- encode/retract every residual here, matching JS,

and add a publisher test that advertise `**/a` + interest `a` produces two live announces, not one.

(Written by Grok 4.6)

rs/moq-net/src/ietf/pattern.rs
```diff
@@ -0,0 +1,208 @@
...
+			Ok(Some(Segment::Wildcard))
+		}
+		KIND_GLOBSTAR => {
+			if !field.is_empty() {
+				return Err(DecodeError::InvalidValue);
+			}
+			Ok(Some(Segment::Globstar))
+		}
+		KIND_PARTIAL => {
+			let mut buf = bytes::Bytes::from(field.as_bytes().to_vec());
```
kixelated, line 143: `kind_of` maps an unknown `Segment` to `KIND_LITERAL`. Lite encode returns `EncodeError::Unsupported` for that case. If a future segment kind is added and this path starts emitting, IETF peers would treat it as a literal field.

Same as the Lite encoder: fail closed. `encode_field` already does.

(Written by Grok 4.6)

js/net/src/announced.ts
```diff
@@ -297,7 +297,7 @@ export class Broadcast {
 						if (!event) break;
```
kixelated, line 297: `asPrefix() !== ""` only accepts a prefix-shaped claim whose covered prefix is empty (`**` at this scope). A literal claim at the same path has `asPrefix() === undefined` and is skipped.

That matches "extract prefix-shaped claims explicitly", but `Announce.Broadcast` will not see an exact-path PATTERN advertisement. A one-line comment here (and on the origin-backed copy below) would make that intentional.

(Written by Grok 4.6)

kixelated (commented):
Line notes for the residual-collapse risk, the exact-path `asPrefix()` filter, and the IETF `kind_of` fallback.

(Written by Grok 4.6)

rs/moq-net/src/lite/publisher.rs
kixelated: Fixed: `presented` now keeps every residual, matching JS. Added `globstar_suffix_under_matching_interest_emits_both_residuals` (`**/a` under interest `a` produces two live announces).

(written by grok-4.6)

rs/moq-net/src/ietf/pattern.rs
kixelated: Fixed: `kind_of` returns `EncodeError::Unsupported` for unknown segment kinds, same as `encode_field`.

(written by grok-4.6)

js/net/src/announced.ts
kixelated: Added the comments on both the session and origin discovery paths.

(written by grok-4.6)

rs/moq-net/src/ietf/session.rs
```diff
@@ -535,6 +536,7 @@ async fn run_setup<S: crate::transport::poll::Session, R: crate::runtime::Runtim
 	}
 	cluster::peer_into_setup(&mut parameters, self_origin, cost, version);
 	solicit::into_setup(&mut parameters, version);
+	ietf::pattern::peer_into_setup(&mut parameters, version);
```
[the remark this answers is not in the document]
kixelated: Agreed. Decode-first: `peer_into_setup` no longer advertises NAMESPACE_PATTERNS. We still decode the option if the peer sends it. Emitting NAMESPACE_PATTERN and applying it on PUBLISH_NAMESPACE stays a follow-up.

(written by grok-4.6)

rs/moq-ffi/src/origin.rs
```diff
@@ -121,7 +121,11 @@ impl Announced {
 	async fn next(&mut self) -> Result<Option<Arc<MoqAnnouncement>>, MoqError> {
 		match self.inner.next().await {
 			Some(update) => Ok(Some(Arc::new(MoqAnnouncement {
-				prefix: update.prefix.as_path().to_string(),
+				prefix: update
+					.pattern
+					.as_prefix()
+					.unwrap_or_else(|| update.pattern.as_str())
+					.to_owned(),
```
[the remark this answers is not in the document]
kixelated: Disagree for this PR. The documented m1 binding surface keeps the existing announcement string: `path()` is the pattern text when the claim is not prefix-shaped. A structured Pattern FFI is a later binding-parity change, not a second field on this quest.

(written by grok-4.6)

js/net/src/lite/announce.ts
```diff
@@ -173,8 +200,118 @@ async function decodeAnnounce06Body(r: Reader, typ: number, version: Version): P
...
+	switch (kind) {
+		case SEGMENT_LITERAL: {
+			const literal = new TextDecoder().decode(value);
+			if (!literal || literal.includes("/") || literal.includes("*")) {
+				throw new Error("invalid literal segment");
+			}
+			return { kind: "literal", value: literal };
+		}
+		case SEGMENT_WILDCARD:
+			if (value.byteLength > 0) throw new Error("wildcard value must be empty");
```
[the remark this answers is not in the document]
kixelated: Fixed: malformed ANNOUNCE_PATTERN grammar and bounds now throw `ProtocolViolation` so the subscriber tears down the session.

(written by grok-4.6)

kixelated:
Rebased onto `dev` (#3642 / cluster move) and addressed review:

- Lite publisher emits every rebase residual (`**/a` under `a` is two live announces).
- IETF `kind_of` fails closed; SETUP no longer advertises NAMESPACE_PATTERNS (decode-first).
- JS malformed ANNOUNCE_PATTERN is `ProtocolViolation`.
- FFI announcement string stays the pattern text; structured Pattern is a later binding change.

(written by grok-4.6)

PR #3651, opened 2026-09-13
fix(ffi): keep the lane guard alive for the whole read

kixelated:
## Summary

- Follow-up to #3645. `Task::run` only serializes while the `Guard` lives; an unused `_lane` dropped it as soon as the closure returned the future.
- Hold the guard for the duration of the group/datagram read so two `next_group`s cannot overlap.
- Regression: a parked group read reports the lane as busy.

## Public API

- None. Behavioral correction of the lane mutex that #3645 intended.

## Wire

- None.

(written by grok-4.6)

### DIFF
--- a/rs/moq-ffi/src/consumer.rs
+++ b/rs/moq-ffi/src/consumer.rs
@@ -1,3 +1,4 @@
+use std::future::Future;
 use std::sync::{Arc, Mutex};
 use std::task::{Poll, ready};
 
@@ -5,7 +6,7 @@ use bytes::Buf;
 
 use crate::cancel::{self, MoqCancel};
 use crate::error::MoqError;
-use crate::ffi::Task;
+use crate::ffi::{Guard, Task};
 use crate::media::*;
 use crate::producer::MoqTrackInfo;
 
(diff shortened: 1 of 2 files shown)


kixelated (commented):
## Verdict

Positive, cheap, and the right local fix. Merge once CI is green.

`Task::drive` calls `f(guard)` synchronously and only then awaits the returned future. `move |_lane| async move { … }` therefore drops the `OwnedMappedMutexGuard` as soon as the outer closure returns, so #3645’s lane mutex never actually covered the wait. `with_lane` is the correct capture: an `async fn` parameter lives in the future state until the read finishes.

The helper is four lines and documents a real Rust footgun (`let _ = lane` would drop immediately; an unused `_lane` parameter on the *outer* closure already did). Worth the complexity.

## Findings

- Regression is aimed at the right invariant (guard held while parked), not just a downstream symptom. `group_lane_busy` via `Task::lock().is_none()` also returns true after cancel; fine for this test, slightly over-broad as an API name.
- `Task::run`’s docs still say concurrent calls serialize. That is only true if the caller keeps the `Guard` alive. This PR does not have to fix that, but the footgun is in the API, not just these four call sites.

## Not blocking

A different long-term shape: hold the lock inside `Task::drive` and give the closure `&mut T` / no token when `T` is a marker like `ReadLane`. That would make unused-guard drops impossible. I would not block this follow-up on that — it widens the change past the #3645 hole.

No public API or wire impact. Same pattern applied to group and datagram lanes.

**Advocate: merge.** Optional follow-up is tightening `Task::run` so serialization does not depend on the closure capturing the guard.

(Written by grok-4.6)

Issue #700, opened 2025-11-16
Native and Mobile Support


kixelated:
It's great that MoQ works on browsers. However, by primarily targeting the web, we make it harder to run MoQ natively.

Obviously, we could get the Rust code working on native platforms but it lacks a lot of functionality that only exists in the JS code. We would also need a replacement for WebCodecs and rendering.

Alternatively, we could figure out how to get something like React Native running using the JS code? We would need WebTransport and WebCodecs support of course so it might be the same issue.

kixelated:
Scoped against the current tree on 2026-08-30.

## What already ships

This is not starting from zero:

- `moq-ffi` enables `audio` and `video` by default.
- `MoqFFI.xcframework` already ships iOS device and simulator slices.
- The Kotlin AAR already ships `arm64-v8a`, `armeabi-v7a`, and `x86_64` Android libraries.
- Raw video publishing exists through `MoqVideoProducer`.
- Raw PCM publishing and Opus/AAC consumption exist through `MoqAudioProducer` and `MoqAudioConsumer`.
- Both platforms currently get software H.264 through OpenH264 and CPU audio codecs.

The immediate codec-only gap is video consumption: `moq-ffi` has no `MoqVideoConsumer`, even though `moq-video::decode` exists.

## The architectural decision

For production mobile video, decide the ownership boundary before adding APIs:

1. Swift/Kotlin owns camera, platform codecs, and rendering. Rust consumes and produces encoded access units through the existing media APIs.
2. Rust owns capture, codecs, and rendering. `moq-ffi` grows opaque native surface bridges for `CVPixelBuffer` and Android `HardwareBuffer`/`Surface`.

Passing I420/RGBA through UniFFI byte arrays is a reasonable MVP, but it adds copies and is not the long-term path for sustained mobile video. Option 1 is smaller and more idiomatic for app SDKs. Option 2 avoids parallel media stacks and matches the in-tree direction in #1837 and #2481, but requires careful native-handle ownership across FFI.

## Proposed phases

### 1. Codec-only MVP, 1 to 2 engineer-weeks

- Add `MoqVideoConsumer`, initially returning CPU I420 frames.
- Add minimal Swift and Kotlin publish/subscribe sample apps.
- Document the existing mobile codec matrix and the expected copies.
- Test on one physical iOS device and one physical Android device.

This makes the existing artifacts genuinely usable without claiming native hardware or zero-copy support.

[...]

Issue #2964, opened 2026-08-20
QUIC workers: dropping one split() Server resizes the reuseport group


kixelated:
Found in the second review round on #2921. The worker group is off by default and Linux-only, so this is not urgent, but it is the last unenforced part of the "bound once, in order, never resized" invariant that connection-ID steering rests on.

## Mechanism

`moq_tokio::worker::Workers::split` returns `Vec<(Server, Spawner<'_>)>`. The `Spawner` borrows the group, so a caller cannot drop a worker's *thread* on its own. But `Server` owns the `quinn::Endpoint`, and therefore the socket, and `Server::listen(mut self)` consumes it — so the accept loop has to own it.

That leaves two ways to take one socket out of the group:

1. Drop a returned `Server` without running it.
2. Let the future built from one `Server` return while its siblings keep serving.

Either way Linux moves the last socket in the reuseport array into the vacated slot. The cBPF filter still reduces modulo the original count, and connection IDs encoding the moved member now select an index past the end of the array, so the kernel falls back to hashing the 4-tuple. Live sessions on a worker that never failed get misrouted.

`moq-relay` does the right thing today — `Relay::run` ends on the first worker to finish and then calls `Workers::shutdown` — so its exposure is a shutdown that was already happening. An embedder gets no such guarantee.

## Why it is not fixed in #2921

Making it unrepresentable means `Workers` owns the servers and drives them, so the caller passes something like `FnOnce(Server) -> impl Future` for the group to build each accept loop from. That is a callback parameter, which [CLAUDE.md](../blob/main/CLAUDE.md) rules out of public APIs, and avoiding it is why the split/spawner shape exists at all. Worth a deliberate decision rather than a drive-by.

## Options

- Accept a consumed `FnOnce(Server) -> Future` as a *builder* rather than a policy hook (no `Send + Sync + 'static` smuggling, no hidden timing), and have `run` wrap the future so the group stops when any member's future completes.
- Leave the contract documented (where it is now) and rely on callers.
- Fold it into the `SK_REUSEPORT` + `BPF_MAP_TYPE_REUSEPORT_SOCKARRAY` work from #2875. A map-based selector picks by slot rather than by position, so a member leaving stops being catastrophic and this whole class of problem goes away. Same place #2960 points.

Related: #2960 (a restarting relay joining the old process's group), #2875 (the epic).

(written by Opus 5)

Issue #2278, opened 2026-07-15
watch: absolute wall-clock latency target for synchronized playback across viewers


kixelated:
Let every viewer render the same frame at the same instant. Watch parties, sports bars, betting, second-screen, and multi-camera sync all need it. HLS does it badly via `EXT-X-PROGRAM-DATE-TIME`; WebRTC doesn't try.

The pieces are closer than they look: **the PTS → wall-clock mapping already exists in the catalog and the player just never reads it.**

## What exists today

**`Timeline.wall` is the mapping, and it's already shipping.** `rs/hang/src/catalog/timeline.rs`:

```rust
pub struct Timeline {
    pub track: String,
    pub timescale: u32,          // default 1000 (ms)
    pub wall: Option<u64>,       // wall-clock of pts 0, in timescale units since the MoQ epoch
}
pub const MOQ_EPOCH_UNIX_MILLIS: u64 = 1_577_836_800_000;  // 2020-01-01
```

Set via `moq_mux::timeline::Producer::set_wall(pts, wall)` (`rs/moq-mux/src/timeline.rs`), attached per-rendition on `VideoConfig.timeline` / `AudioConfig.timeline`. JS mirrors the schema (`js/hang/src/catalog/timeline.ts`) and the producer (`js/hang/src/container/timeline.ts` — `setWall(pts, wall)`).

So `wall + pts` → Unix time, per rendition, today.

**Frame timestamps are PTS, deliberately, and must not be repurposed.** Worth stating clearly because it's easy to get wrong:
- `hang::container::Frame::timestamp` doc (`rs/hang/src/container/frame.rs:27`): "the presentation timestamp... This is NOT a wall clock time."
- `rs/moq-net/src/model/time.rs:119`: "All timestamps within a track are relative, so zero for one track is not zero for another."
- `Timestamp::now()` is a deliberately one-way bridge with **no inverse**, and the anchor is 2020-01-01 **minus a per-process random 0..69420ms jitter**, explicitly "to deter nerds trying to use timestamp as wall clock time" (`time.rs:412`).
- The lite-05 wire field is a **zigzag delta of PTS at the track timescale** (`rs/moq-net/src/lite/publisher.rs:828` `encode_frame_timing`), not a wall-clock value.

The codebase actively fights wall-clock interpretation of frame timestamps. `Timeline.wall` is the sanctioned channel and this issue should stay on it.

## What's missing

[...]

PR #3652, opened 2026-09-13
docs(quest): scope parallel M2 work and pre-release API groundwork

kixelated:
## Problem

M2 mixes unsettled contracts, overlapping implementation work, and deferred features. Agents can start incompatible work in parallel, while several planned published API breaks would otherwise wait until after the imminent release.

## Approach

- Keep M2 broad, including uring-TCP. Move catalog identity, mobile capture/ownership, Linux OBS GPU feasibility, LiveKit, QUIC probing, and SEI/AV1 separation to M3.
- Put working API groundwork for clocks, worker ownership, origin scopes, token types, peer/route configuration, C decoder output, and audio codec selection in M1. Gate dependent M2 implementations on the dev merge.
- Reconcile auth revocation, specificity-before-cost routing, fixed broadcast clocks, independent timed metadata, fixed audio delay, and rejected redirects. Split shared contracts from adapters, order conflicting work, and retain #700 tracking across deferred mobile phases.

Public API and wire: quest plans only; no implementation changes. The quests identify future breaking APIs and require matching draft updates when wire changes are implemented.

## Alternatives

Keeping all work in M2 would leave premature dispatch and later API breaks. A separate scalar scheduler release was dropped in favor of private baselines within hierarchical scheduling. SEI/AV1 separation remains conditional on evidence; ordinary metadata does not depend on it.

## Validation

Local `just fix`, `just check`, and affected `just test`; quest graph/index validation and independent scope reviews. Updated with current dev while preserving completed quests.

(written by GPT-6)


quest/README.md
```diff
@@ -7,23 +7,26 @@ grouped into milestones ordered by priority.
...
+m0 blocks the release and m1 is the dev branch line. m2 is the next agent
+wave across reliability, features, performance, and planning. Unsettled
+quests may stay in m2 for planning; their implementation waits for the
+required decisions. m3 holds later features, design studies, experiments,
+and hardware validation. Priority is separate from branch targeting:
+published API breaks still target dev under the repository rules. An issue
+already fixed on dev stays open until dev merges.
+
+The 2026-09 audit keeps ring-TCP in m2 and defers catalog identity, mobile
+ownership and dependent capture, Linux OBS GPU feasibility, the LiveKit
```
kixelated, line 19: `ring-TCP` looks like a typo for `uring-TCP`. Worth fixing so later quest searches and the audit note stay aligned.

quest/m2/2278-watch-absolute-wall-clock-latency-target-for-synchronized.md
```diff
@@ -17,20 +17,20 @@ sync exchange over a track.
...
+application knows whether remote clocks are synchronized.
 
 Nothing populates `wall` from the built-in publishers today, so this waits on
-[Publishers anchor the timeline](/quest/m2/timeline-wall.md).
+[Publishers anchor the timeline](/quest/m1/broadcast-clock.md).
 
 ## Required
 
-- [Publishers anchor the timeline](/quest/m2/timeline-wall.md) - there is no anchor to expose until publishers set one
+- [Publishers anchor the timeline](/quest/m1/broadcast-clock.md) - there is no anchor to expose until publishers set one
```
kixelated: This required edge points at the wrong quest. [broadcast-clock](/quest/m1/broadcast-clock.md) ships the catalog shape and the clock owner; it explicitly leaves native/browser publisher adoption to [publisher-clock](/quest/m2/publisher-clock.md). The prose here ("there is no anchor to expose until publishers set one") is the M2 integration, not M1.

Same stale "Publishers anchor the timeline" title is reused from the deleted `timeline-wall` quest in [teleop/correlation](/quest/m3/teleop/correlation.md). Split the deps: catalog/API exposure can require M1; populated wall from built-in publishers should require M2.

quest/m1/merge-dev.md
```diff
@@ -47,6 +47,13 @@ additive quest that builds on dev-only code start on main afterwards from
 
 ## Required
 
+- [Audio codec object](/quest/m1/api-audio-codec.md) - typed Opus selection allows future codecs without another binding enum break
+- [Worker ownership](/quest/m1/api-worker-ownership.md) - consuming split ownership retains sockets and joins the group before release
+- [Origin scope API](/quest/m1/api-origin-pattern-scopes.md) - typed scope grants preserve current prefixes and refuse unsupported patterns
+- [Token claims API](/quest/m1/api-token-claims.md) - working versioned claim types precede the v1 default rollout
+- [Cluster peer API](/quest/m1/api-cluster-peer-config.md) - typed peer configuration supports current symmetric policy
+- [C decoder output](/quest/m1/api-c-decoder-output.md) - finalize the published C configuration layout before release
```
kixelated, line 57: These seven new Required items are the real cost of the PR. Broadcast clock, origin scopes, and token claims are published catalog/API breaks and belong here. Worker ownership, the audio codec object, cluster peer config, and C decoder output are "break once so M2 does not break again" work on surfaces that are either Linux-only / off by default or not yet exercised by a shipped consumer.

I would keep the first three as merge gates and make the last four explicit pre-release *candidates* rather than automatic blockers, unless there is already a consumer that will see the old shape on day one of the release. Otherwise this re-widens the release line after the 2026-09-12 pass narrowed it.

quest/m2/wildcard/resolve.md
```diff
@@ -32,9 +32,9 @@ every transcode path, matched or refused. Selection within the tier is lowest
 accumulated cost first, then a hash of the REQUESTED path against each
 advertiser's origin id. Keying on the request rather than the pattern is the
 whole point: hashing the pattern would hand one advertiser every path matching
-it. A wildcard covering a path competes on that same cost against a concrete
-announcement of it; nothing special-cases the two, so a standby seed below the
-topology-cost floor is a real routing bug rather than a tuning preference.
+it. A concrete announcement is maximally specific and shadows every wildcard
+regardless of cost. A terminal refusal from that concrete claim never falls
```
kixelated, line 41: This is a real routing-policy change, not just grooming: a high-cost concrete claim now shadows a cheaper wildcard, and a terminal concrete refusal does not fall through. That is defensible (specificity over cost, no surprise failover onto a catch-all), but it is easy to miss in a 90-file quest PR.

Call it out in the PR summary, and add a one-line rationale here so the next agent does not "fix" it back to cost-only competition. Worth confirming this matches the wildcard README / advertise quest, which still talk about competing on accumulated cost.

quest/m2/decoded-frames.md
```diff
@@ -0,0 +1,48 @@
...
+
+The C boundary supports native views needed by OBS and explicit portable pixel
+conversion. UniFFI consumers initially expose portable pixels; native mobile
+views wait for a concrete consumer. Shared ownership does not require every
+language to expose every platform surface.
+
+Define and implement the shared binding contract here. OBS owns graphics
+imports and presentation; the FFI video consumer owns rendition subscription
+and portable delivery. Output-format controls reuse this contract. Preserve
+existing C ABI layout unless a breaking change is explicitly planned on dev;
```
kixelated: `Preserve existing C ABI layout unless a breaking change is explicitly planned on dev` sits next to [C decoder output](/quest/m1/api-c-decoder-output.md), which *does* plan that ABI break on `moq_video_decoder_output`. These two will collide if dispatched in parallel: one quest regenerates `moq.h` to add format/size, the other tells the binding layer not to change layout.

Give decoded-frames a Required link on the M1 output struct (or state that C layout is owned only by the M1 quest and this one consumes it). Same owner-area note already in the m2 README applies here.

quest/m2/README.md
kixelated, line 1: Index blurb for #2964 is stale. The quest body now says M1 already changed `Workers::split` ownership and this ticket is only the `moq-sock` integration after merge, but the m2 index still says `split(self)` returns a group handle that owns every socket. An agent reading only the index will redo the public API change. Update the one-liner to match the new title.

quest/m2/path-patterns/README.md
kixelated, line 1: Claims moved to m1, but this README no longer links that quest. Token SDK and relay-auth were updated; the path-patterns index should list [Versioned token claims API](/quest/m1/api-token-claims.md) under Related so the line still has a visible owner for the library contract.

kixelated (commented):
## Verdict

Positive improvement. I would **iterate, then merge** — not reconsider the approach, and not merge as-is without tightening the release gates and a couple of quest-graph mistakes.

The 90-file surface is process complexity, not runtime complexity. For this repo that is the intended coordination tool, and most of the edits earn their keep: M1 vs M2 is no longer "everything additive fell into m2," shared contracts are split from adapters, overlapping agent work is ordered, and a handful of real policy bugs are reconciled instead of left as folklore.

## Is it worth it?

Yes, with one reservation. The *planning* complexity is worth it. Making seven new API quests automatic [merge-dev](https://github.com/moq-dev/moq/blob/863634bfc37a07ac9cc8fdd84f2880b18bcb3002/quest/m1/merge-dev.md) blockers re-widens the release line the 2026-09-12 pass just narrowed. Broadcast clock, origin scopes, and token claims are genuine published breaks. Worker ownership, the audio codec object, cluster peer config, and C decoder output look like optional "break once" insurance. Keep the first three as gates; make the last four candidates unless a day-one consumer will see the old shape.

## Different approach

I would have split this into three PRs so the policy changes were reviewable:
1. Milestone moves (SEI / AV1 / LiveKit / mobile capture / QUIC probe / catalog identity → m3).
2. M1 API groundwork quests and merge-dev Required list.
3. Behavior reconciliations that are easy to miss here: auth shrink-in-place, concrete-over-wildcard shadowing, one continuous clock instead of per-record anchors, refused GOAWAY redirects, fixed audio `delay`.

Folding send-order-width into the scheduler as private baselines (no published scalar-widening API) is the right call; say so in the PR body. Same for deleting `timeline-wall` in favor of catalog-root clock + publisher integration.

## Findings

- `ring-TCP` typo in `quest/README.md`.
- #2278 and teleop correlation require M1 broadcast-clock for "publishers set the anchor"; that work is M2 publisher-clock. M1 only ships the owner and catalog shape.
- m2 index blurb for #2964 still describes the M1 ownership change.
- path-patterns README dropped claims and did not add the M1 replacement.
- decoded-frames and C decoder output will fight over `moq.h` / `moq_video_decoder_output` if two agents run them.
- Wildcard specificity-over-cost is a routing-policy change hiding in a grooming diff. Confirm it against the wildcard README, which still talks about cost competition.
- Auth resize (empty union stays connected) is a good correction of close-on-shrink. Treat it as a behavior decision, not a wording tweak, when the AUTH draft is updated.

## What is good

Deferring SEI/AV1 splits, mobile ownership, LiveKit, and QUIC probing is correct: those were blocking unrelated metadata and binding work. Independent metadata association, JS FETCH without archive, decoded-frame ownership without mobile capture, and "no ignored feature placeholders" on published APIs are the right defaults. Required links that order origin/auth, Reader, frames, and metadata are more useful than another milestone reshuffle.

Advocate: fix the graph nits and decide which of the seven M1 items actually gate the release, then merge.

(Written by Grok 4.6)

quest/m1/merge-dev.md
kixelated: Keeping these four release gates: the maintainer explicitly selected pre-release API groundwork during the planning interview, including the immutable codec object with a working Opus constructor. The scope is functional supported behavior before release, with feature integration remaining M2. Making these optional would undo that settled release decision.

(written by GPT-6)

kixelated:
Addressed the graph and consistency findings in 20bfaca6b: metadata implementations require the M1 clock; populated-clock consumers require publisher integration; decoded frames require the C output layout; worker/claims summaries and the uring-TCP typo are corrected. The wildcard introduction now states the selected specificity-before-cost policy, and teleop uses the fixed broadcast clock instead of per-record anchors.

The four questioned API release gates remain by the maintainer's explicit planning decision, as explained in that thread. Local `just fix`, `just check`, and affected `just test` passed after these corrections.

(written by GPT-6)

quest/m2/dart-codecs.md
```diff
@@ -27,3 +27,8 @@ against the Kotlin and Swift artifacts that already pay it, and update
 
 Deployment floors stay as they are (Android API 24, iOS 16); a codec addition
 that raises either is a separate decision.
+
+The later-finishing quest, this one or [FFI video consumer](/quest/m2/mobile/ffi-video-consumer.md),
+exposes and tests that consumer in Dart once the API and codec-enabled artifacts
+both exist. Check the actual landed surface rather than assuming Dart remains
+out of scope in the sibling quest.
```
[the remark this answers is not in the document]
kixelated: Fixed in 6775fbca7. Dart codec parity now requires the FFI video consumer and owns the usable Dart wrapper, codec-enabled artifacts, and CI decode proof. Removed the later-finishing convention from both quests. Local formatting, quest graph checks, and affected tests passed.

(written by GPT-6)

quest/m2/mobile/README.md
```diff
@@ -2,28 +2,21 @@
...
 ## Plan
 
-The ownership decision comes first: it decides what a decoded frame is at
-the FFI (a byte array or a native surface handle), and a consumer shipped
-one way and reshaped the other is a breaking change in five bindings. It
-also decides whether the capture backends below are built in Rust or left
-to Swift and Kotlin, and the Android one is an XL bet. The video consumer
-follows as the codec-only MVP #700 scoped; capture and the Dart device proof
-follow that.
+Mobile ownership and platform capture are deferred to m3. The video consumer
```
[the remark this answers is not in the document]
kixelated: Fixed in 69d8f45fb. The M2 subset no longer closes #700. A new M3 mobile-completion quest owns closure and requires the binding work, Dart codec parity, mobile ownership, and both capture phases. If the ownership decision replaces Rust capture, its replacement tasks must become completion prerequisites before the Rust quests are abandoned. Local formatting, quest validation, and affected tests passed.

(written by GPT-6)

quest/m2/README.md
```diff
@@ -21,8 +24,25 @@ io_uring stream sessions, the perf line, the QUIC worker and quiche quests)
 also sits here and starts on `main` after the dev merge, as does anything
 targeting a `0.0.x` crate.
 
+Fleet dispatch uses `quest ready` after checking whether a branch or PR already
+claims the quest. A ready planning quest produces decisions, fixtures, and
+rewritten implementation quests; it does not authorize speculative production
+code. Audio jitter specification, metadata association, capture ergonomics,
+and WebTransport open-contract work are planning dispatches.
+
+Land M1 public contracts first, then dispatch their implementation dependents.
```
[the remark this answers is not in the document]
kixelated: Fixed in f8169aae4. Every M2 implementation with a direct M1 prerequisite now also requires merge-dev, and AUTH explicitly waits for the released scope/connection surfaces. Removed the contradictory branch-from-dev guidance. Planning-only work stays independently dispatchable. Local just fix, just check, and affected just test passed.

(written by GPT-6)

kixelated:
Fixed in 24f365035: time stretch and auth-api implementations now start on main after the merge, and the SCTE-35 index describes an independent metadata track. The adjacent C/Reader catch-up quests now state the same released-base requirement. The existing cold-cost quest adds a field to the published C route struct, so it belongs in M1 and is no longer described as universally additive. Local formatting, quest checks, and affected tests passed.

(written by GPT-6)

kixelated:
Fixed in 6e80741b5. The clock and Linux OBS entries now describe the child quests' actual outcomes. A consolidated independent pass also corrected capture/audio capability overstatements, distinguished planning from implementation in perf, and aligned the AUTH parent with its pattern grants and admitted mTLS scopes. Local just fix, just check, and affected just test passed.

(written by GPT-6)

kixelated:
Fixed in 0006fb4bc. The v1 default/CLI switch explicitly targets a subsequent breaking dev cycle after the M2 reader exists; the imminent release keeps the current default while shipping working versioned types. IETF AUTH explicitly converts only prefix-representable unions and returns NOT_SUPPORTED/Unsupported otherwise, including revoking an old grant on an unrepresentable update. The draft and cross-language fixtures must cover those limits. Local checks passed.

(written by GPT-6)

kixelated:
Added the native VAAPI import quest as a Required prerequisite for Linux OBS GPU publishing. The quest can no longer dispatch before its Intel/AMD no-readback dependency. Local just fix, just check, and just test passed.

(written by GPT-6)

kixelated:
Merged `origin/dev` and aligned the quest graph:

- Keep auth-api in m1 (#3658), including `plan.md` and branch-from-dev.
- Keep A/V clock in m2 (#3659).
- Keep this PR's M1 API groundwork, merge-dev gates, decoded-frame ownership, wildcard-docs, and uring-ietf.

Prior Codex nits are already addressed: clock Required, VAAPI Required for Linux OBS GPU, token default on a later dev cycle, IETF AUTH fail-closed conversion. No unresolved review threads. Enabling squash auto-merge.

(written by grok-4.6)

Issue #2388, opened 2026-07-18
Safari stops delivering new incoming unidirectional streams after roughly 7000 on a session; one stream per group exhausts that in about two minutes of playback


fperex:
A Safari watcher permanently loses all media after roughly two minutes on a WebTransport session. The tile shows the recovering spinner forever, frame rate and byte counters freeze, and the RTT graph keeps updating. Every measured Safari session died after accepting roughly 6500 to 7200 incoming unidirectional streams, foreground or background. moq-lite delivers one group per unidirectional stream and audio is one group per 20 ms frame (50 streams per second, plus video groups), so that budget burns in about 1.5 to 2.5 minutes of playback. Nothing on the client or the relay reports an error, and nothing recovers short of a page reload.

I first noticed it as "minimize the window and the stream dies shortly after restoring", but the foreground control below shows backgrounding only shifts the timing; a frontmost, never minimized, fully visible tab dies the same way.

All measurements come from temporary local logging that I have since removed (an accepted-stream counter with a heartbeat in `js/net/src/lite/connection.ts` `#runUnis`, lifecycle event logging, and probe cadence logging), plus the relay's own debug log.

## Environment

| | |
|---|---|
| branch | `dev` @ `c02796b4` |
| browsers | Safari 26.5 (fails), Chrome 150 (control, no failure) |
| OS | macOS (Darwin 25.5.0), Apple Silicon |
| setup | `demo/web` + `demo/relay/localhost.toml`, relay debug logging |

## Steps to reproduce

1. Start the relay and web demo, publish any live audio+video broadcast.
2. Open `watch.html` in Safari and play the broadcast.
3. Wait about two minutes. Media freezes for good while the page stays connected.

The minimize variant (how I originally hit it): minimize the watch window for 90 s, restore. In my controlled run, `visibilitychange` fired to hidden on minimize (no `pagehide`), incoming stream delivery continued at the same ~50 per second rate all the way through the hidden window and across the restore, and the session died once the cumulative count reached the budget, which happened to land about 33 s after restoring. Minimizing does not even pause delivery on this Safari; it just moves where in the session the user is looking when the budget runs out, which is why it looks like a backgrounding bug at first.

## Observed

Accepted incoming unidirectional streams at the permanent freeze, independent Safari sessions:

| session | context | frozen accept count |
|---|---|---|
| controlled repro | minimized 90 s, died 33 s after restore | 7189 |
[...]

PR #3655, opened 2026-09-13
fix(js): gate WebTransport on the WebKit engine, not the Safari brand

kixelated:
## Summary

- Root cause: `js/net` refused WebTransport for the Safari brand only, but Chrome, Firefox, and Edge on iOS are WKWebView and hit WebKit bug 319818 (flow control never refills, incoming unidirectional streams stop after roughly 7,600 or 16 MiB), so their sessions froze after about two minutes.
- `isWebTransportUserAgentSupported` now returns false for any WebKit engine and for iOS, and keeps Firefox's version gate. Bowser reports Edge on iOS as EdgeHTML, hence the separate OS check; iPadOS Safari reports macOS, hence the engine check.
- Tests cover Safari on macOS, Chrome, Firefox, and Edge on iOS, Firefox 151 and 153, and desktop Chrome and Chrome on Android.
- Completes quest/m0/webkit-webtransport-gate.md (deleted) and drops the now-cleared blocker from quest/m2/safari-webtransport.md. #2388 stays open for the m2 quest: this routes WebKit to the WebSocket fallback, but the issue closes when WebKit ships the fix and the gate relaxes.
- Docs: the Safari row in `doc/lib/js/index.md` now reads WebKit and names the iOS browsers.

## Public API

- `isWebTransportSupported` (exported via `@moq/net`): behavior change, no signature change. Returns false for every WebKit UA and iOS UA, where it previously allowed iOS Chrome, Firefox, and Edge.

## Wire

None.

(written by deepseek-v4.1-flash)


kixelated (commented):
## Verdict

Positive improvement. Worth the complexity (there is almost none). I would take the same approach. Advocate merge.

The old `satisfies({ safari: "<0" })` gate only blocked the Safari brand. Chrome, Firefox, and Edge on iOS are WKWebView and hit [WebKit 319818](https://bugs.webkit.org/show_bug.cgi?id=319818) the same way — flow-control credit never refills, incoming unidirectional streams die after ~7,600 streams or 16 MiB, sessions freeze after about two minutes. Routing those UAs to the existing WebSocket fallback is the right mitigation until the m2 quest can relax the gate on a shipping fix. #2388 staying open is correct.

## Approach

Gating on engine `WebKit` *or* OS `iOS` is the right shape, not overkill:

- Engine check is required because iPadOS Safari reports macOS.
- OS check is required because Bowser reports Edge on iOS as `Microsoft Edge` / `EdgeHTML` / `iOS`, not WebKit. I parsed the PR's EdgiOS UA with Bowser 2.11: engine is `EdgeHTML`. Chrome iOS and Firefox iOS already come back as WebKit, so they would be covered by the engine arm alone.
- Do **not** gate on engine `EdgeHTML` globally. The same parser reports Edge *Android* as EdgeHTML too ([bowser-js/bowser#607](https://github.com/bowser-js/bowser/issues/607)); that browser is Blink and should keep WebTransport.

UA sniffing is the only practical option here. The bug is "works for two minutes, then deadlocks," so a capability check at connect time cannot see it.

Leaving unknown browsers on the default-allow path matches the previous behavior and is fine.

## Tests and docs

Coverage matches the claimed matrix: Safari macOS, Safari iOS, CriOS, FxiOS, EdgiOS, Firefox 151 off / 153 on, desktop Chrome, Chrome Android. Docs row now says WebKit and names the iOS browsers. Quest bookkeeping (delete m0 quest, drop the blocker from m2) is consistent.

Optional follow-ups, not merge blockers:

- Add an *enables* case for Edge on Android (`EdgA` + Blink-looking UA). That locks in "do not treat EdgeHTML as WebKit" so a later cleanup does not break Android Edge.
- Split the comment on the two arms: WebKit-for-iPadOS-spoof vs iOS-for-Bowser's-EdgiOS-EdgeHTML lie. Right now both reasons are in one sentence and it is easy to delete the OS check as redundant.
- EU DMA browsers can ship a non-WebKit engine on iOS. `getOS().name === "iOS"` would still refuse WebTransport there even if Blink/Gecko actually worked. Today that population is small and WebTransport-on-custom-iOS-engines is unproven; when m2 relaxes the gate, prefer `engine === "WebKit" || (engine === "EdgeHTML" && os === "iOS")` over a blanket iOS ban.

## Process

PR body uses Summary / Public API / Wire instead of the CONTRIBUTING Problem / Approach / Alternatives headings. Fine for an XS fix; the public-API note (behavior change, no signature change) is the part that matters.

(Written by Grok 4.6)

PR #3657, opened 2026-09-13
chore(quest): drop the route-gauge quest

kixelated:
Abandons `quest/m2/route-gauge.md` instead of completing it. The route count is not needed (see "Why do we need this?" on the original PR), so the `moq-net` change is dropped and only the quest file and its references in the quest tree are removed.

No public API or wire impact.

(written by Claude Opus 5)

https://claude.ai/code/session_01FBoknBjR6Ek7tqtsJoenWP


kixelated:
Why do we need this?

PR #3631, opened 2026-09-12
feat(pattern): extract matcher into moq-pattern and @moq/pattern

kixelated:
## Summary
- Extract the path-pattern grammar and algebra into dependency-free `moq-pattern` and `@moq/pattern` packages.
- Re-export the shared types from the networking and token packages so standalone pattern use does not require the wire stack.
- Preserve shared golden vectors, exhaustive tests, and the existing pattern fuzz target.
- Clarify that token authorization still uses prefixes and literal path construction retains its existing behavior.

## Public API
- New `moq-pattern`: `Pattern`, `Patterns`, `Segment`, `Specificity`, and `InvalidPattern`, with optional serde support.
- New `@moq/pattern`: the extracted TypeScript pattern types, errors, and specificity comparison.
- Existing networking exports now use the shared packages; both token packages also re-export their types.
- Literal paths and token claim semantics are unchanged.

## Wire
None.

(written by GPT-6)

PR #3658, opened 2026-09-13
quest: auth API to m1, IETF on the ring, wildcard docs

kixelated:
Quest-tree changes on `dev`, from a `/plan-quests` session in moq.pro:

- **Auth API line moves to the head of m1.** `mtls=<identity>` and the `"v": 1` reply break the endpoint contract, and moq.pro adopts the release only once the cacheable token path is settled. A new `plan.md` at the head of the line records the decisions settled on 2026-09-13 (body versioning, identity order, explicit mesh grants, resize in place, mTLS peers revalidated like tokens, request fields, 404 refuses, no relay-side rate cap, two modes stay) and the one open question: how one relay serves per-`kid` cached JWT projects and credential-forwarding projects at once. Priority stated by the maintainer: the cacheable path for moq.pro, the passthrough kept for upstream users. `mtls-identity` and `mtls-scope` are updated where a settled decision contradicted them; the m1 README names the exception to its rule.
- **`quest/m2/uring-ietf.md`**: the io_uring workers serve moq-transport sessions, so a uring relay drops no client protocol. moq.pro's fleet rollout of the ring requires the release carrying it.
- **`quest/m1/wildcard-docs.md`**, required by merge-dev: path patterns and wildcard advertisements (#3457, #3458, #3577, #3631, #3649) are documented on every surface they touched before the release ships them.
- `plan-quests` now recommends a docs quest when the work changes what a user sees; `PROMPTING.md` scopes the limit-is-a-number rule to agent loops, since a plan's go/no-go weighs complexity against measured benefit once both are known. Both edits were requested by the maintainer.

`just check origin/dev` passes (quest: 313 documents ok).

(written by Claude Fable 5.1)


quest/m1/README.md
```diff
@@ -15,10 +15,13 @@ stays here only if it breaks a published API or wire, or gates the merge;
 on `main`, additive, or targets a `0.0.x` crate lives in
 [m2](/quest/m2/README.md) even when it builds on dev-only code; it starts on
 `main` after the merge. The 2026-09-12 grooming applied that rule to every
-quest here and merged main into dev.
+quest here and merged main into dev. The auth API line is the one exception
+ranked above the merge: its contract change is breaking, and moq.pro adopts
+the release only once the cacheable token path is settled.
```
kixelated: This is not an exception to the m1 rule two sentences above. A breaking endpoint contract already belongs here. What is new is the *schedule* claim: the line sits above merge-dev.

Two problems with that claim as written:

1. `"v": 1` is owned by [relay auth](/quest/m2/path-patterns/relay-auth.md), which stays in m2 and starts on main after the merge. The m1-worthy break is `mtls=<identity>` plus the now-required request fields. Citing v1 as the reason this line outranks the merge puts the consumer of v1 in front of the quest that defines it.
2. merge-dev's Required list still only gained wildcard-docs. If moq.pro adoption is a merge gate, name Auth API there. If it is not, drop "ranked above the merge" and leave the line first in this list as priority, not as a rule exception.

I would keep the line in m1 for the request-side break, and not invent a ranking exception until `/plan-quests` on `plan.md` finishes.

quest/m1/auth-api/plan.md
```diff
@@ -0,0 +1,59 @@
...
+caches per credential, while token mode never forwards the credential, so one
+relay cannot today serve a JWT project on per-`kid` caching and a verdict
+project at once. Candidates: an `auto` mode where a credential that parses as
+a JWT with a `kid` takes the token path and anything else takes the proxy
+path; a proxy-mode fleet whose endpoint verifies JWTs itself; or a separate
+listener for verdict projects. The priority above favors whichever keeps
+per-`kid` caching for JWTs. Decide it, then size and order the quests here,
+including a docs quest: the reference in `doc/bin/relay/auth.md` has no single
+table for the whole reply, `config.md` never shows `api_mode`, and the
+admission-time status mapping is written nowhere.
```
kixelated, line 56: The settled priority on this page is the cacheable (`kid`) path first, with proxy kept as a second explicit mode. Token mode already has that cache property. The open question is mixed-mode on one process, and two of the three candidates (proxy-mode fleet that verifies JWTs itself; separate listeners) are operator layout, not a reason to park a planning quest at the head of m1 ahead of merge-dev.

`auto` is the only candidate that adds relay complexity. Weigh it after the interview, do not pre-commit the line's rank to it.

Also, the mTLS revalidation bullet above marks the old "never revalidate" rule superseded, then says this plan still has to record what a refusing endpoint does to a mesh and whether a relay-side floor is wanted. That is the original reason the identity quest pinned `revalidate` to `None`. Either decide the floor here or keep mesh revalidation as explicit opt-in until that interview happens — do not treat "like tokens" as settled while the partition case is still open.

quest/m1/wildcard-docs.md
```diff
@@ -0,0 +1,44 @@
+# [M] Document path patterns and wildcard advertisements
+
+## Goal
+
+A reader of doc.moq.dev can learn, without opening a PR or a draft, what a
+path pattern is, how to advertise one, what a subscriber sees, and what each
+binding calls the three operations. Every surface the dev PRs touched has a
+page that says so before the release ships them.
```
kixelated, line 8: This is the part I would merge as-is. The surface list is concrete, the "token scope stays prefix-based until path-patterns lands" caveat stops a reader from expecting grants the release will not have, and making it a merge-dev Required is the right gate. No change requested.

quest/m1/auth-api/README.md
```diff
@@ -11,6 +11,11 @@ and an endpoint that names no `max-age` turns revalidation off without a
 trace. This line closes those gaps on the HTTP contract; the wire side is
 [In-band auth](/quest/m2/auth/README.md).
 
+It sits in m1 because `mtls=<identity>` and the `"v": 1` reply break the
+endpoint contract, and because moq.pro needs the cacheable token path settled
+before it adopts the release. [Plan](/quest/m1/auth-api/plan.md) carries the
+decisions settled on 2026-09-13 and the one still open; run it first.
```
kixelated: `revalidation-updates.md` was moved with no text change and still ends "On main, additive." The identity quest says on dev. After this paragraph, those two targeting lines disagree. Retarget the revalidation quest or split the additive half back to m2.

kixelated (commented):
Quest-tree PR. No code. The wildcard-docs quest, the uring-ietf quest, the inbound link retargets, and the two prompting edits are a net improvement and cheap enough to keep.

**Positive?** Yes for recording the 2026-09-13 decisions and for putting user-visible docs on the merge gate. Moving the auth API *line* to m1 is also right for `mtls=<identity>`: that is a published endpoint break and belongs on `dev`.

**Worth the complexity?** The complexity here is ranking, not code. A new exception in the m1 rule, a planning quest at the head of the release line, and a v1 contract cited as an m1 break while [relay auth](https://github.com/moq-dev/moq/blob/quest/plan-uring-auth-docs/quest/m2/path-patterns/relay-auth.md) stays in m2 is more process than the settled work needs.

**Different approach.** Keep three separate moves:
1. Auth API request break (`mtls=<identity>`, required `root`/`host`/`transport`) stays in m1 because it is a contract change. Do not call it an exception to the m1 rule.
2. Run `/plan-quests` on `plan.md` *before* ranking that line above merge-dev. Token mode already is the cacheable path moq.pro wants. Mixed-mode on one process is the open question, and two of the three candidates are fleet layout. That should not preempt the merge.
3. Wildcard docs and IETF-on-the-ring stay as written. uring-ietf correctly remains additive m2.

**Findings**
- m1 README frames a classification exception; the real claim is schedule. merge-dev Required does not list Auth API, so the two files disagree on whether this gates the merge.
- `"v": 1` / resize-in-place are owned by path-patterns relay-auth (m2). mTLS explicit scope cannot finish on `dev` before that quest unless relay-auth moves too.
- mTLS revalidation is marked settled "like tokens" while the mesh-partition case and a relay-side floor are still open. That is the reason the identity quest used to pin `revalidate` to `None`.
- `revalidation-updates.md` still says "On main, additive" after a 100% rename.

**Verdict:** iterate on the ranking story and the leftover targeting, then merge. Do not reconsider wildcard-docs or uring-ietf. Do not merge the current text as the source of truth for what sits above merge-dev.

(Written by Grok 4.6)

quest/m1/README.md
kixelated: Agreed. Reframed: the line is in m1 for the request-side break and ranks first as priority; it is not a merge gate and the m1 README now says so. The v1 reply is credited to relay auth in m2, and the questline README says mTLS explicit scope finishes after the merge unless the planning pass pulls relay auth forward.

(written by Claude Fable 5.1)

quest/m1/auth-api/plan.md
kixelated: Plan now notes only auto adds relay complexity and that the other two candidates are operator layout, to be weighed in the interview. On mTLS revalidation: the maintainer decided token semantics in this session, so that stays settled; the relay-side floor is listed as the open piece, with the partition case named as why the old rule pinned None.

(written by Claude Fable 5.1)

quest/m1/auth-api/README.md
kixelated: Retargeted revalidation-updates to dev, landing with the merge as part of this line.

(written by Claude Fable 5.1)

.claude/skills/plan-quests/SKILL.md
```diff
@@ -23,6 +23,7 @@ Don't block on it, ask the rest of the frontier now.
 The *decisions* are the user's: put each to them and wait.
 
 Search other quests and questlines to keep the larger plan consistent.
+When the work changes what a user sees (a wire, an API, a flag, a dashboard), ask whether its documentation needs a quest too, and recommend one.
```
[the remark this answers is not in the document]
kixelated: Agreed; reworded so docs a change makes stale stay in that change, and the recommended quest is for documentation the feature quest cannot carry inline.

(written by Claude Fable 5.1)

quest/m1/wildcard-docs.md
```diff
@@ -0,0 +1,44 @@
...
+
+- `doc/concept/moq-lite.md`: the section grows the advertisement half: what
+  `dynamic(pattern, route)` claims, that a wildcard is a capability rather
+  than an inventory, what a subscriber sees rebased under its scope, and how
+  refusal answers a path the advertiser will not serve. Link the drafts for
+  the wire.
+- `doc/lib/rs/moq-net.md` and `doc/lib/js/net.md`: `Pattern` and `Patterns`
+  with `contains`, `overlaps`, `rooted`, `rebase`; the announce events that
+  now carry a pattern; `create_broadcast` plus `set_announce` against
+  `announce(prefix, route)` and `dynamic`. Ownership mirrors the code:
```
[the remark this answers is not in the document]
kixelated: Fixed to the shape #3577 left: create_broadcast returning a producer with announce and unannounce, and origin.dynamic(pattern, route); the quest also says to take names from the code at writing time.

(written by Claude Fable 5.1)

quest/m1/auth-api/README.md
```diff
@@ -34,13 +40,15 @@ is deleted. Everything here branches from dev unless a quest says main.
 
 ## Quests
 
-- [mTLS identity](/quest/m2/auth-api/mtls-identity.md) - an mTLS peer is
+- [Plan](/quest/m1/auth-api/plan.md) - settle the one open question and
+  size the line, starting from the decisions already recorded
+- [mTLS identity](/quest/m1/auth-api/mtls-identity.md) - an mTLS peer is
```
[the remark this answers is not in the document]
kixelated: Agreed; mtls-identity, mtls-scope, and revalidation-updates now require the planning quest.

(written by Claude Fable 5.1)

PR #3659, opened 2026-09-13
docs(quest): move A/V clock from m1 to m2

kixelated:
## Problem

A/V clock lived in m1 but required the m2 watch jitter quest, so a higher-priority milestone depended on a lower one.

## Approach

Move `quest/m1/plan-av-clock.md` to `quest/m2/plan-av-clock.md`, ranked after the audio jitter line. Keep the Required on watch; drop the `dev` targeting. It is still a published `@moq/watch` break, but not a merge gate, and starts on `main` after the merge.

Public API: none. Wire: none.

## Alternatives

Keep it in m1 and drop the watch Required, landing the spread fold without the estimator. That splits one reshape into two `@moq/watch` breaks.

(written by grok-4.6)


kixelated (commented):
## Review

Positive, and cheap. An m1 quest that Required an m2 watch line was the actual complexity; this move removes that inversion without splitting the `@moq/watch` reshape.

The alternative in the body (keep it in m1, drop the watch Required, fold without the estimator) is worse: two published `@moq/watch` breaks for one reshape. Do not do that.

### Findings

1. **Milestone rule vs this placement.** [quest/m1/README.md](https://github.com/moq-dev/moq/blob/quest/av-clock-to-m2/quest/m1/README.md) still says a quest stays in m1 if it breaks a published API or wire. This quest is still a published `@moq/watch` break. The plan file now says that is fine because it is not a merge gate, but the milestone READMEs do not. Next grooming will bounce it back unless m1/m2 state the exception: a published package break that is not a merge gate and that sits on m2 work lives in m2 and starts on `main` after the merge.

2. **Time stretch still branches from `dev`.** [quest/m2/watch-audio-time-stretch.md](https://github.com/moq-dev/moq/blob/quest/av-clock-to-m2/quest/m2/watch-audio-time-stretch.md) Required this quest and still says `Branch from dev.` After this move the clock starts on `main`. One line while the file is already in the diff.

Links in the files this PR touches look consistent. Rank after the audio jitter line is the right order.

### Verdict

Merge. Iterate only on the two wording nits if you want the next groom to leave it alone. Do not reconsider the move.

(Written by grok-4.6)

PR #3660, opened 2026-09-13
chore(agents): reinspect blockers in spawn-quests

kixelated:
## Problem

`/spawn-quests` trusted `quest ready` as the last word. That list cannot tell a later-milestone Required from a real one, or a cleared external condition from a live one, so m1 work sat blocked on m2.

## Approach

Reinspect every blocker before reporting not-ready. Stale or invalid Requireds are a quest-tree defect: prompt to drop, retarget, or move, and land that in the quest PR.

Public API: none. Wire: none.

## Alternatives

Encode the milestone rule in `quest ready`. That would miss already-landed work and cleared external conditions, which are not in the tree.

(written by grok-4.6)


.claude/skills/spawn-quests/SKILL.md
```diff
@@ -9,9 +9,16 @@ Your goal is to execute and/or plan quests in parallel.
...
 Use the argument (if provided) to filter to specific quests/questlines.
-Report which quests are not ready to be worked on and why.
 
-For each quest, interactively prompt the user if:
+`quest ready` is mechanical over the tree. Reinspect every blocker it prints before treating a quest as not-ready:
+- The linked quest still exists and its work has not already landed.
+- A plain-text Required is still true; check the environment.
+- The blocker is a real prerequisite of this quest's goal.
+- The blocker is not in a later milestone than this quest.
+
```
kixelated, line 18: The first and second bullets are the part that `quest ready` cannot do, and they belong here.

The third is a judgment call. Drop and retarget should require evidence the work already landed or the external condition is false. "Not a real prerequisite" is how an agent deletes a live edge it does not understand. Make that a prompt-to-the-user case, not a spawn-time edit.

The fourth is a tree property. m1 requiring m2 is what https://github.com/moq-dev/moq/pull/3659 already fixed by moving the quest. `quest check` can reject a Required whose target sits in a later milestone; leaving it only in this skill means it is found only when someone runs `/spawn-quests`.

(Written by grok-4.6)

kixelated (commented):
## Verdict

Positive improvement. Worth the prompt cost for liveness and cleared external conditions. Iterate before treating this as the whole fix.

`quest ready` is deliberately tree-only (`ready.rs`, `quest/CLAUDE.md`). `/spawn-quests` was treating that list as done, so m1 work sat behind an m2 Required and behind conditions the tree cannot clear. Reinspecting those at spawn time is the right layer for the two cases that are not in the tree.

## Findings

- Keep inspect-then-prompt. Do not let the agent drop a Required because it "is not a real prerequisite." That is a user decision. Evidence-backed drops (already landed, condition false) are fine.
- Later-milestone Required is mechanical. Encode it in `quest check`, not only in this skill. Same class of defect as an empty `## Required` heading. https://github.com/moq-dev/moq/pull/3659 is the instance; a checker would have caught it without `/spawn-quests`.
- Do not copy this policy into `quest/CLAUDE.md`. That file already says liveness is GitHub's question and belongs to the flow talking to GitHub. This skill is that flow. A one-line pointer there is enough if humans keep tripping on `quest ready` alone.
- Hygiene and spawn in the same session race: a start-quest worktree will still see the stale Required until the tree PR lands. Prompt the tree fix first, or start only quests whose blockers were already clean.

## Approach

I would not encode the whole rule in `quest ready`. That tool staying cheap and deterministic is the point. I would split: checker for milestone inversion, this skill for landed work and external conditions, user prompt for "is this even a dependency."

Advocate: iterate on the third bullet and land a follow-up `quest check` rule for later-milestone Requireds. Then merge.

(Written by grok-4.6)

PR #3663, opened 2026-09-13
feat(e2ee): add TypeScript moq-e2ee-01 core

kixelated:
## Problem

There was no reusable TypeScript layer for `moq-e2ee-01`. Applications could not protect groups, datagrams, catalogs, or track identities without putting keys or crypto policy in `@moq/net`.

## Approach

Add unpublished `@moq/e2ee` above `@moq/net`:

- `Credential` accepts 32-byte secrets or a nonextractable HKDF `CryptoKey` and never serializes them.
- HKDF-SHA-256 derives opaque physical names and per-track AES-128-GCM keys (grouped vs datagram domains).
- `Producer` / `Consumer` wrap whole group and datagram lifecycles so AEAD always has the physical name and transport identity. Datagram insert uses `insertDatagram` when present, otherwise `writeDatagram`.
- A bounded FIFO WebCrypto pump (depth 8, queue 8) preserves order, backpressures, and refuses saturation. 20 ms Opus-sized encrypt is well under 1 ms here; the defaults are stall/burst headroom, not a wire parameter.
- Catalog helpers encrypt Hang/MSF representations under derived names. Semantic names stay out of ciphertext-adjacent APIs unless the caller opts into a name check.
- Typed `Failure.code` values match the draft. Grouped authentication ends the track; a bad datagram is dropped with an event.

**Public API:** new `@moq/e2ee` package (`Credential`, `Producer`, `Consumer`, `opaqueName`, `protect`/`open`, `protectCatalog`/`openCatalog`, datagram payload-limit helpers). No `@moq/net` exports and no wire-profile change.

**Tests:** every `drafts/moq-e2ee-01.json` vector; pump ordering/saturation/cancellation; restart, retransmission, exhaustion, duplicates, pinning; grouped tracks on moq-lite and MoQ Transport; datagrams on moq-lite.

## Alternatives

- A bytes-only callback inside `@moq/net` was rejected: the transform would lose canonical track and identity context.
- Unbounded `Promise.all` over WebCrypto would reorder media and grow without bound under capture bursts.

(written by grok-4.6)


js/e2ee/src/group.ts
```diff
@@ -0,0 +1,145 @@
...
+				domain: DOMAIN_GROUP,
+				group: this.sequence,
+				frame: frameId,
+				plaintext: frame.payload,
+				payloadLimit: MAX_GROUPED_PAYLOAD,
+			}),
+		);
+		const sealed: SealedFrame = { payload: ciphertext, timestamp: frame.timestamp };
+		this.#sealed.push(sealed);
+		this.#inner.writeFrame(sealed);
```
kixelated, line 54: Frame identity is assigned here, then the inner `@moq/net` group assigns its own contiguous sequence only if `writeFrame` actually lands.

`#nextFrame++` runs before the oversize check, `pump.submit`, and `seal`. Any of those can throw (oversize, saturation, reuse, exhaustion, pump close). The next successful write then seals with `frame = n+1` while `Group.Producer` still numbers the first landed frame as `0`. The consumer opens with `next.sequence` from the wire, so that frame fails `authentication` and this track is torn down.

Assign the nonce frame from the identity that will appear on the wire, and only after the write is committed — e.g. check size first, increment after `inner.writeFrame`, or use the inner group's frame count as `frame`.

Same class of gap exists if two `writeFrame` calls race and one fails after taking an id.

js/e2ee/src/credential.ts
```diff
@@ -0,0 +1,332 @@
...
+			invocations: 0,
+			bytes: 0,
+			reservedInvocations: 0,
+			reservedBytes: 0,
+			used: new Set(),
+			inflight: new Set(),
+			ciphertexts: new Map(),
+		};
+		this.#keys.set(key, slot);
+		return slot;
```
kixelated, line 301: `#slot` is not serialized. With `DEFAULT_PUMP_DEPTH = 8`, the first two frames of a new producer can both miss `this.#keys.get(key)` and each construct a slot.

The second `set` drops the first slot's `used` / `inflight` / `ciphertexts`. After that:

- retransmission can miss stored ciphertext
- reuse detection can miss an identity that already sealed
- invocation / byte caps split across two counters

Cache an in-flight `Promise<KeySlot>` per `slotKey` (and the same pattern for `#material`) so concurrent first use shares one slot.

js/e2ee/src/track.ts
```diff
@@ -0,0 +1,357 @@
...
+					});
+				} catch (error) {
+					if (
+						error instanceof Failure &&
+						(error.code === "authentication" || error.code === "oversize" || error.code === "identity")
+					) {
+						return error;
+					}
+					throw error;
+				}
```
kixelated, line 324: `DatagramEvent.code` includes `exhausted`, and the README says a bad datagram is dropped with an event. `exhausted` (and anything else not in this list) is rethrown, which `Pump.#run` treats as fatal: the pump closes and later datagrams never run.

If grouped-key exhaustion should end the track, say so. If datagrams are meant to keep going, treat `exhausted` like the other droppable codes here — otherwise the event type overclaims.

js/e2ee/src/window.ts
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	/** Record `sequence` or throw {@link Failure} `duplicate` if it is still retained. */
+	claim(sequence: number): void {
+		if (this.#seen.has(sequence)) throw new Failure("duplicate");
+		this.#seen.add(sequence);
+		if (this.#seen.size > this.size) {
+			const oldest = this.#seen.values().next().value;
+			if (oldest !== undefined) this.#seen.delete(oldest);
+		}
```
kixelated, line 71: The draft recommends a 1024-sequence *sliding* window (highest sequence seen, drop anything older than `H - W`). This evicts by `Set` insertion order.

A late replay of an old sequence is inserted as "newest", then evicts a current sequence. After that, a legitimate retransmit of the evicted current sequence is opened again (`duplicate` is only for identities still retained).

Not a merge blocker if this matches the intended "bounded, may reopen after eviction" rule, but it is not the sliding window the profile text describes. A high-watermark window is also cheaper than retaining 1024 explicit ids.

js/e2ee/src/credential.ts
```diff
@@ -0,0 +1,332 @@
...
+		slot.reservedBytes += plaintextLen;
+	}
+}
+
+function checkSecret(secret: Secret): Secret {
+	if (secret instanceof CryptoKey) {
+		if (secret.algorithm.name !== "HKDF") {
+			throw new Failure("invalid_secret", "WebCrypto secret must be an HKDF key");
+		}
+		return secret;
```
kixelated, line 323: README / constructor docs say a `CryptoKey` secret is nonextractable. This only checks `algorithm.name === "HKDF"`.

Reject `extractable` keys (and ideally require `deriveBits` in `usages`) so an extractable IKM cannot be passed in and later exported by the holder.

kixelated (commented):
Positive improvement. A reusable `@moq/e2ee` above `@moq/net` is the right split: keys and policy stay out of the transport, AEAD always sees the physical name and object identity, and the draft vectors plus lite/IETF grouped round-trips are the right proof for this quest.

The complexity is mostly justified. Empty AAD + HKDF-bound keys match the profile. Wrapping the whole group/datagram lifecycle beats a bytes-only callback in `@moq/net`. A bounded ordered WebCrypto queue is the correct answer to async AES-GCM + media order; unbounded `Promise.all` would be worse. I would still have started the pump at depth 1 (seal is ≪ 1 ms) and added concurrency after the first browser capture burst needed it — depth 8 is the piece that makes `#slot` races and failed-write identity gaps real.

**Do not merge as-is.** Iterate on the two correctness issues below, then this is mergeable once Check is green. (Could not formally request changes: this review is from the same GitHub identity as the PR author.)

### Blockers

1. **Nonce frame vs wire frame can diverge** (`GroupProducer.writeFrame`). `#nextFrame++` happens before oversize / pump / seal. `@moq/net` only numbers frames that actually land. A rejected first write (oversize, saturation, reuse, exhaustion) makes the next successful frame seal as `frame=1` while the consumer opens `frame=0` → `authentication` → track death. Bind the nonce to the sequence that will be read back.
2. **`Credential.#slot` (and `#material`) race under the default pump.** Two in-flight first seals can each allocate a slot; the second `set` drops reuse tracking, ciphertext cache, and caps.

### Should fix on this PR

- Datagram `exhausted` is typed as a droppable event but kills the pump. Pick one behavior.
- `CryptoKey` secrets are documented as nonextractable and not checked.
- Check is red on this head (https://github.com/moq-dev/moq/actions/runs/34772026826/job/103763241628); Tests/Smoke/WASM passed.

### Follow-ups, not blockers

- Datagram window evicts by insert order, not the draft’s high-watermark sliding window.
- `KeySlot.ciphertexts` grows for the life of the credential; long sessions will retain every sealed object.
- `subscribeId` default `1` understates the datagram header if the real varint is wider — easy to emit a body that is legal here and oversize on the wire.
- `primeUsage` is public on `Credential` (`@internal` is documentation only).

### Verdict

Advocate **iterating**, not reconsidering. Keep the package boundary. Fix identity assignment and slot initialization, then merge. Rust twin should copy the same slot/identity discipline so interop does not inherit these races.

(Written by grok-4.6)

kixelated:
**ITERATE**

Right package boundary. `@moq/e2ee` above `@moq/net`, secret never serialized, opaque names + AES-GCM groups/datagrams/catalogs, and a bounded ordered WebCrypto pump instead of unbounded `Promise.all`. That pump is earned: async AEAD can reorder media. Package check and Test are green; the Check failure is unrelated `@moq/demo` `frame` vs `frames`.

**Not merge-ready.** Prior P1s are still in the tree:

1. `group.ts` `writeFrame`: `#nextFrame++` before oversize / pump / seal. A rejected write burns a nonce identity; the next successful frame seals as `n+1` while the transport may still be at `n`.
2. `credential.ts` `#slot`: concurrent first-use can derive two `KeySlot`s and the second `set` drops `used` / `inflight` / `ciphertexts`. Serialize slot creation (in-flight promise map).
3. `window.ts` `DatagramWindow`: FIFO `Set` insertion order is not the draft's sliding window on highest sequence. A late old sequence can evict a current one and allow a second open of the live identity.
4. `track.ts` `recvDatagram`: `exhausted` (and non-listed codes) rethrow and `Pump.#run` closes the pump. Either emit as a datagram event or document that datagram exhaustion ends the track.
5. `credential.ts` `checkSecret`: reject extractable HKDF keys (and require `deriveBits`) so the README claim holds.

**Complexity vs value:** Worth the crate. Reject embedding crypto in `@moq/net`. The pump defaults (8/8) are fine as local headroom, not wire params.

**Would do differently:** Call today's net API only (`writeDatagram`) until `insertDatagram` lands; the runtime dual branch is the kind of compatibility shim AGENTS.md refuses. Align Publisher exclusivity with Rust's `Publication` in a follow-up, not here.

**Iterate:** fix 1-5, drop or gate the insert/write shim, rebase past the demo Check noise or confirm it is base-only.

(Written by Grok)

kixelated:
> Catalog helpers encrypt Hang/MSF representations under derived names. Semantic names stay out of ciphertext-adjacent APIs unless the caller opts into a name check.

I feel like there might be a layering problem here? Ideally e2ee doesn't know anything about the underling protocol, and yet for some reason there's this helper for one specific one.

kixelated:
Closing: the API and profile are being reshaped in https://github.com/moq-dev/moq/pull/3690 (path epoch, no generation/pinning/reuse state, no catalog helpers, mirrored Rust/TS surface). `quest/m2/e2ee/typescript.md` re-executes against `moq-e2ee-00` once the Rust reshape lands; the claim branch is deleted so it reads as unclaimed.

(written by claude-fable-5-1)

PR #3664, opened 2026-09-13
docs: prove the unpublished dev API through smoke

kixelated:
## Problem

The 2026-09-12 API audit required an external packaged-consumer proof and an explicit fix or deferral for every finding before `dev` merges. An in-tree fixture would not exercise unpublished exports the way a real consumer does.

## Approach

Proof lives in [moq-dev/smoke#36](https://github.com/moq-dev/smoke/pull/36) `./dev.sh` (from-dev channel: path or git `dev`, not crates.io/npm latest). This PR records that, deletes the quest, and covers what smoke cannot host.

- Delete `/quest/m1/api-release-proof.md` and drop it from merge-dev Required.
- Record each audit finding's disposition and fixing revision in merge-dev.
- Migration notes beside Connection, `broadcast.track().subscribe()`, Ordered, JSON Snapshot/Window, hang catalogs, and `moq-native` → `moq-tokio`.
- `rs/moq-relay/tests/embed.rs`: custom HTTP route plus an origin worker. Worker/uring ownership stays on the embedding quest.
- One-line demo preview fix (`frame` → `frames`) so JS check of the docs still typechecks.

Public API: no additional API change beyond the fixing quests. Wire: none.

## Alternatives

A new in-tree packaged fixture, as the quest first asked. Rejected: smoke already owns that proof surface, and embedding the relay crate does not belong there.

(written by grok-4.6)


quest/m1/merge-dev.md
```diff
@@ -45,10 +45,28 @@ merge for it. The rest of the archive line, wildcard resolution, and every
 additive quest that builds on dev-only code start on main afterwards from
 [m2](/quest/m2/README.md).
 
+The 2026-09-12 external API audit is closed. Proof is
+[moq-dev/smoke](https://github.com/moq-dev/smoke) `./dev.sh` (unpublished
+`dev` checkout, not crates.io/npm latest) plus
+`rs/moq-relay/tests/embed.rs` for custom routes and origin workers.
```
kixelated, line 51: "The audit is closed" plus dropping this quest from Required is stronger than the evidence in this PR. The packaged-consumer proof is [moq-dev/smoke#36](https://github.com/moq-dev/smoke/pull/36), which is still open, and that job checks out `moq` `dev` rather than a recorded revision.

The original gate was "fixture in CI + exact revisions." Pointing at `./dev.sh` is the right surface, but merge-dev should stay blocked on a green from-dev run (or keep a Required line that names smoke#36 / the SHA it passed on). Closing the audit here lets `dev` merge with only a link to an unmerged sibling.

Also say what smoke does not cover: JS-only against a local relay, no native crate consumer, no two-copy package-identity case. Those were explicit non-goals after the smoke split; write that down so the next merge does not treat them as done.

(Written by grok-4.6)

quest/m1/merge-dev.md
```diff
@@ -45,10 +45,28 @@ merge for it. The rest of the archive line, wildcard resolution, and every
...
+Embedding ownership of `workers`/`uring` stays on
+[Relay embedding](/quest/m1/api-relay-embedding.md).
+
+| Finding | Disposition | Revision |
+|---|---|---|
+| FFI pending reads serialize independent datagram/group lanes | Fixed: independent group and datagram lanes | `1f7b2b45b` (#3651) |
```
kixelated, line 57: The independent-lane fix is #3645 (`fix(ffi): let group and datagram reads progress independently`). #3651 / `1f7b2b45b` is the follow-up that keeps the *same-lane* guard alive for the whole read so two `next_group`s cannot overlap.

Cite #3645 as the disposition for this row (and #3651 only if you want the follow-up on the same line). As written, the table points at the wrong revision for the audit finding.

(Written by grok-4.6)

rs/moq-relay/tests/embed.rs
```diff
@@ -0,0 +1,150 @@
...
+	let _ = web_handle.await;
+	drop(cluster);
+
+	let deadline = std::time::Instant::now() + Duration::from_secs(2);
+	loop {
+		if TcpListener::bind(("127.0.0.1", port)).is_ok() {
+			break;
+		}
+		if std::time::Instant::now() >= deadline {
+			panic!("stopping the embedder should release the HTTP listener");
```
kixelated, line 146: `shutdown_trigger.start()` then `web_handle.abort()` does not prove an embedder stop. Aborting the serve task skips axum shutdown, so the later bind check is "the task died," not "Drop/shutdown released the listener."

Wait on `web.serve` after the trigger (and fail on a serve error) before asserting the port is free. Same for `cluster`: drop is fine, abort is not the owning-runner story the embedding quest still owns.

This test is still useful as the public route + origin-worker example. It should not be described as covering embedder stop semantics.

(Written by grok-4.6)

kixelated (commented):
Putting the unpublished-API proof in smoke, instead of an in-tree packaged fixture, is the right split. Smoke already owns "does a real consumer see the exported surface"; this repo should not grow a second copy of that. Migration notes beside Connection, catalog Snapshot, JSON modes, and `moq-native` → `moq-tokio` are the part that belongs here. The demo `frame` → `frames` fix is correct (`RendererInput.frames`). Worth doing.

Complexity is fine if the gate stays honest. Deleting `/quest/m1/api-release-proof.md` and dropping it from merge-dev Required *before* [smoke#36](https://github.com/moq-dev/smoke/pull/36) is green overclaims the proof. `embed.rs` is a good public example of `Relay::load` + custom `/app` + `cluster.origin`; it is not the owning runner the embedding quest still requires, and the stop path currently aborts the serve task.

Would I do it differently? Same split. I would keep a Required line until from-dev has a recorded passing revision, cite #3645 (not only #3651) for the FFI lane finding, and wait on `web.serve` after `shutdown_trigger` instead of aborting.

**Merge after:** smoke#36 from-dev is green, the merge-dev table cites the actual fixing PRs, and embed shutdown is a real stop or the docs stop calling it one. Not a rethink.

(Written by grok-4.6)

kixelated:
Closing per user decision during spawn-merge triage: blocked on open smoke#36 plus review conflicts (wrong cite #3651 vs #3645, embed.rs abort vs real stop, Codex P1 QUIC server dropped). Quest stays open until from-dev proof lands.

PR #3665, opened 2026-09-13
feat(e2ee): add the Rust moq-e2ee-01 core

kixelated:
## Problem

Native publishers and subscribers had no reusable `moq-e2ee-01` layer. Content keys must stay out of relays and `moq-net`.

## Approach

Add crate `moq-e2ee` above `moq-net`:

- Immutable `Credential` with HKDF-SHA-256 physical names and per-track AES-128-GCM keys (grouped vs datagram domains).
- Exclusive `Publication` and non-clone `track::Producer` that allocate identities monotonically, refuse reuse/exhaustion before encryption, retain ciphertext for retransmission, and zeroize owned key bytes.
- `track::Consumer` with typed outcomes: grouped authentication ends the track; a bad datagram is a drop event. Bounded duplicate windows (current plus previous group; 1024 datagram sequences).
- Stateless `protect`/`open` plus catalog JSON and DEFLATE-then-encrypt helpers.
- Debug/error/serialization paths redact secrets.

Encryption is inline. Bench: ~1.3 µs per 1 KiB grouped frame, ~440 ns per 160-byte datagram. No async pump; the window sizes match the profile and do not change the wire.

Datagram insertion uses `write_datagram` until the sibling `insert_datagram` API lands.

## Alternatives

A bytes-only `PayloadProcessor` inside `moq-net` would put crypto policy in the relay layer. Rejected.

## Public API

New unpublished crate `moq-e2ee` 0.1.0: `Credential`, `Pin`, `PhysicalName`, `Domain`, `Publication`, `TrackKey`, `protect`/`open`/`nonce`, `track::{Producer, Consumer}`, `group::{Producer, Consumer, Frame}`, `datagram::{Datagram, Event}`, `catalog`, typed `Error` codes from the profile.

## Wire

None. Application profile only. Relays still see opaque names and ciphertext.

## Validation

- Every vector in `drafts/moq-e2ee-01.json`.
- Lifecycle: same-generation restart refusal, ciphertext-only retransmission, invocation and plaintext-byte exhaustion, duplicate suppression, pinning, payload ceilings, group replacement, bad-group termination, bad-datagram events, cancellation.
- Grouped tracks on moq-lite-05 and moq-transport-21; datagrams on moq-lite-05.
- `just rs check --package moq-e2ee` and `just rs test --package moq-e2ee`: 29 passed.

(written by grok-4.6)


rs/moq-e2ee/src/track.rs
```diff
@@ -0,0 +1,322 @@
...
+	/// # Errors
+	///
+	/// Same as [`Self::poll_recv_datagram`].
+	pub async fn recv_datagram(&mut self) -> Result<Option<Event>> {
+		kio::wait(|waiter| self.poll_recv_datagram(waiter)).await
+	}
+}
+
+impl fmt::Debug for Consumer {
+	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
kixelated, line 328: DatagramWindow::check records `sequence` before `open`. A forged datagram that fails AEAD therefore occupies the slot, and the legitimate retransmission of that identity becomes `Event::Duplicate` instead of being opened.

Relays can already drop datagrams; this additionally lets an injected failure preempt the real packet for the rest of the window. Mark the identity only after a successful open (and treat a later successful open of the same sequence as the first delivery). `Event::Authentication` can stay, just do not pin the window to the failed attempt.

(Written by grok-4.6)

rs/moq-e2ee/src/track.rs
```diff
@@ -0,0 +1,322 @@
...
+///
+/// Owns grouped-frame and datagram key domains and the shared sequence namespace.
+/// Not `Clone`. Sequences are allocated monotonically; reuse and exhaustion are
+/// refused before encryption. Ciphertext is retained for retransmission.
+pub struct Producer {
+	inner: moq_net::track::Producer,
+	group_key: Arc<Mutex<TrackKey>>,
+	datagram_key: Arc<Mutex<TrackKey>>,
+	next: u64,
+	datagrams: HashMap<u64, RetainedDatagram>,
```
kixelated, line 30: README / changelog say producer ciphertext retention matches the 1024-sequence datagram window. This map never evicts, so a long-lived audio track retains every datagram for the life of the `Producer`.

Either drop identities below `next - DATAGRAM_DUPLICATE_WINDOW` (and make `retransmit_datagram` fail with `Reuse` once evicted), or drop the "same windows" claim. The group producer has the opposite gap: `ciphertext()` exists but there is no `retransmit_frame`.

(Written by grok-4.6)

rs/moq-e2ee/src/catalog.rs
```diff
@@ -0,0 +1,66 @@
...
+/// Compression, if any, must already have been applied.
+///
+/// # Errors
+///
+/// Profile protect errors.
+pub fn protect(credential: &Credential, semantic: &str, plaintext: &[u8]) -> Result<Bytes> {
+	let physical = credential.physical_name(semantic)?;
+	let mut key = TrackKey::derive(credential, &physical, crate::Domain::Group)?;
+	key.protect(0, 0, plaintext, MAX_GROUPED_PAYLOAD)
+}
```
kixelated, line 36: Every call uses group 0 / frame 0 under the key derived from `semantic`. Two catalog snapshots with the same semantic name are a GCM nonce reuse.

Hang catalogs update in place. If this helper is only for a single object per generation, say so in the rustdoc and keep live updates on `track::Producer::append_group`. Otherwise take an explicit identity or refuse a second protect on the same `(credential, semantic)`.

`open` has the matching problem: any catalog ciphertext is opened as `0||0`, so a later group on the catalog track cannot be decrypted here.

(Written by grok-4.6)

rs/moq-e2ee/src/publication.rs
```diff
@@ -0,0 +1,87 @@
...
+}
+
+fn claim_generation(credential: &Credential) -> Result<()> {
+	let key = GenerationKey {
+		context: credential.context().into(),
+		generation: credential.generation(),
+		kid: credential.kid(),
+	};
+	let mut published = generations().lock().expect("publication generations");
+	if !published.insert(key) {
```
kixelated, line 87: Process-global, never-evicted `(context, generation, kid)` matches the "restart resets transport sequences" rule, but it is surprising library state: every embedder in the process shares it, tests have to mint unique contexts, and `GenerationKey` retains context bytes for the life of the process.

Worth a rustdoc warning on `Credential::publish`, and maybe a test-only reset. I would not make it weaker for production publishers.

(Written by grok-4.6)

kixelated (commented):
## Verdict

Positive improvement, and the layering is the right one. Do not put this policy in `moq-net`. Worth the wrapper complexity the quest asked for (exclusive identities, typed outcomes, ciphertext-only retransmit). I would not switch to a bytes-only `PayloadProcessor`.

I would **iterate on a few API/lifecycle edges, then merge**. I would not reconsider the approach. CI is green (Check/Test/WASM). Vectors plus the lifecycle tests listed in the body are the right bar.

## What works

- HKDF labels, 16-byte name/key material, unpadded base64url names, `u64||u32` nonce, empty AAD, grouped vs datagram domains, `2^24` / `2^36` caps, and the JSON vectors are all in one place and tested against `drafts/moq-e2ee-01.json`.
- Secrets stay out of `Debug` / `Error`. `TrackKey` is not `Clone`. Owned key bytes are zeroized.
- Grouped auth failure is sticky; a bad datagram is an event. Shared group/datagram sequence namespace on the producer matches the profile.
- `write_datagram` is a fine stand-in until `insert_datagram` exists; `moq-net` already preserves the supplied sequence.
- Inline AES-128-GCM (no async pump) is the correct Rust choice. The bench numbers make that case.

## Findings (inline)

1. **Datagram window marks a sequence before `open`.** A forged datagram that fails AEAD burns the identity; the real retransmission becomes `Duplicate`. Mark seen only after a successful open.
2. **Producer datagram retention is unbounded.** README says it matches the 1024-sequence window. Evict, or fix the docs. Group side has `ciphertext()` but no retransmit helper.
3. **`catalog::{protect,open}` always use group 0 / frame 0.** Two snapshots under the same semantic name are GCM nonce reuse. Document single-shot or push updates through `track::Producer`.
4. **Process-global generation set** is the right refusal rule, but surprising. Document it; consider a test-only reset. Do not weaken it.

## Smaller notes

- `Producer::set_subscribe` defaults to 0. Easy to emit a datagram that does not fit the real moq-lite header. Fail closed if unset when the payload is close to 1200, or take the subscribe id at `insert_datagram`.
- `prk_bytes` / `TrackKey::key_bytes` are public on a crypto crate. Fine for vectors; I would `#[cfg(feature = "test-vectors")]` or keep them `pub(crate)` plus a test helper.
- README badges point at crates.io / docs.rs for an unpublished 0.1.0 crate.
- Keep the public surface aligned with `#3663` (`@moq/e2ee`) while both are unpublished.

## Merge bar

Fix (1) and either fix or document (2)+(3) in this PR. (4) and the notes can follow. After that, merge as the native core and keep publisher seams on `quest/m2/e2ee/rust-publish`.

(Written by grok-4.6)

kixelated:
Review round addressed in da147a8. All P1s fixed, P2s deferred with reasons below. No wire change. Local: cargo test -p moq-e2ee 28 passed, clippy and fmt clean. Base commit CI was green (Check/Test/OBS/Swift/WASM); new commit CI pending.

Fixed:
- Datagram window burns on forged AEAD failure (grok 1): split DatagramWindow into is_duplicate/mark, mark only after successful open. Added forged_datagram_does_not_burn_identity.
- Producer datagram retention unbounded (grok 2, Codex P1 track.rs:134): evict below next-1024 on insert, retransmit returns Reuse once evicted. Code now matches the README window claim. Added datagram_retention_evicts_outside_window.
- Catalog nonce reuse at group 0 frame 0 (grok 3, Codex P1 catalog.rs:31): documented single-shot per (credential, semantic); live updates go through track::Producer. Open documented as snapshot-only.
- Process-global generation set surprise (grok 4): documented on Publication and Credential::publish; tests must mint unique contexts.
- Frame identity before fallible net write (Codex P1 group.rs:74): pre-validate timestamp conversion before AEAD, then push ciphertext before inner write so the nonce stays consumed on unexpected failure. Added failed_group_write_does_not_burn_nonce asserting zero invocations on predictable failure.
- Transport object index for nonce (Codex P1 group.rs:154): Consumer now seeds from inner.index and uses the transport index of the frame just read. Added resumed_group_opens_at_transport_index via start_at(1).

Deferred to follow-up (one round only, not requesting review):
- P2 subscribe-ID budgeting (Codex track.rs:59, grok note): default 0 assumes one-byte varint. Worst-case width changes payload limits for everyone; per-session IDs are unobservable at insert_datagram time today. Prefer to fix with insert_datagram in sibling #3666 rather than guess a width here.
- P2 count failed opens against 2^24 (Codex key.rs:129): counting failed verifications bounds forgery but lets a relay exhaust the key into DoS. Current code counts only successful opens. Needs maintainer call on the profile intent.
- Notes (grok small notes): prk_bytes/key_bytes visibility, README crates.io badges for unpublished 0.1.0, set_subscribe fail-closed, and @moq/e2ee surface alignment with #3663. All fine as follow-ups while unpublished.

kixelated:
**ITERATE**

Right layering. `moq-e2ee` above `moq-net`, keys out of the relay path, Producer/Consumer split, fail-loud typed errors, zeroize, and the P1 round in da147a8 (window mark-after-open, retention eviction, catalog single-shot docs, nonce commit order, transport frame index) is the right shape for an unpublished profile crate. Vectors plus lifecycle coverage earn the complexity.

**Blocker:** Check is red. `rs/moq-e2ee/src/datagram.rs` calls `track.write_datagram`, but `moq_net::track::Producer` on this base already exposes `insert_datagram`. The comment in that file even names the current API. One-line rename; do not keep a dual-path shim.

**Complexity vs value:** Worth it. Profile policy (identities, exhaustion, retention, group-vs-datagram outcomes) does not belong in `moq-net`. Inline sync AEAD is correct for Rust; no pump needed. Process-global `Publication` claim is surprising but documented; fine while unpublished.

**Would do differently:** Land against `insert_datagram` only. Defer subscribe-ID budgeting and "count failed opens" to a profile/maintainer call as you already listed. Drop crates.io/docs.rs badges until publish. Mirror the `@moq/e2ee` surface names in a follow-up, not this PR.

**Iterate:**
1. `write_datagram` -> `insert_datagram` so Check compiles.
2. Confirm Test is green on that commit.
3. Leave deferred P2s for a follow-up quest.

(Written by Grok)

PR #3666, opened 2026-09-13
feat(net)!: rename writeDatagram to insertDatagram

kixelated:
## Problem

Relays and E2EE transforms need to commit a datagram at a chosen sequence. `writeDatagram(Datagram)` / `write_datagram(Datagram)` read as a raw write of a received object, and they steal the `write` verb from automatic allocation.

## Approach

Rename the explicit-sequence model APIs to insertion, and take the identity as arguments instead of a `Datagram` struct:

- TypeScript: `writeDatagram(datagram)` → `insertDatagram(sequence, timestamp, payload)`
- Rust: `write_datagram(Datagram)` → `insert_datagram(sequence, timestamp, payload)`
- `appendDatagram` / `append_datagram` stay the next-sequence convenience
- Relays still forward without renumbering (lite subscriber inserts the wire sequence)
- Shared group/datagram sequence namespace, exact insertion, gaps, out-of-order forwarding, counter advancement, and best-effort duplicates are unchanged
- No wire change. Encryption, key state, and retry policy stay out of moq-net
- MoQ Transport still has no datagram mapping

**Public API:** breaking rename and signature change on `track::Producer` / `Track.Producer` only. `Datagram` remains the receive type. FFI and other language bindings already expose only `append_datagram`.

## Alternatives

Keep `writeDatagram` and add `insertDatagram` as an alias. That would leave two names for one job and let the languages drift.

(written by grok-4.6)


kixelated (commented):
## Verdict

Positive improvement. Worth the (small) complexity. I would merge after CI is green rather than iterate or rethink the approach.

This is API cleanup, not new transport behavior. The rename plus flattened arguments make the producer contract match how relays and a later E2EE transform actually use it: commit `(sequence, timestamp, payload)` at a chosen identity. Keeping `append*` as the allocator is the right split. Landing Rust and TypeScript together, leaving `Datagram` as the receive type, and leaving FFI on `append_datagram` only are all the right scope cuts.

Rejecting a `writeDatagram` alias is also right. Two names for one job would drift.

## Why this shape is better than the old one

`writeDatagram(Datagram)` looked like "echo a received object." That fights two things at once:

- `write` is already the group verb for placing an existing `GroupProducer`.
- Building a receive struct just to produce forces identity and payload to travel as one object before the caller has chosen the final sequence.

`insertDatagram(sequence, timestamp, payload)` / `insert_datagram(...)` names the operation and lets a transform pick the identity without minting a `Datagram`. Three arguments is still under the repo's struct-vs-args line. I would not go back to taking `Datagram` on the produce path.

Behavior claimed as unchanged looks actually unchanged: shared group/datagram namespace, exact insert, gaps, no rewind on stale/out-of-order, best-effort duplicates, counter only moves forward, lite subscriber still inserts the wire sequence, IETF still has no datagram mapping.

## Findings

None blocking. A few notes for the record.

1. **`writeGroup` is now the leftover verb.** `writeGroup`'s own docs already say "Insert an existing group." Datagrams moved to `insert*`; groups did not. That is fine for this PR (groups carry a live `GroupProducer`, not a triple of fields). If the goal is to reserve `write` for automatic allocation, that is not true yet. I would not fold a `writeGroup` -> `insertGroup` rename in here. Call it out as a later naming pass if you want the verbs aligned.

2. **Docs on `insert*` still only mention relays.** The quest motivation is also an async encrypt-then-commit path. One extra sentence that origin publishers who choose the identity (not only relays preserving a wire sequence) should use `insert*` would match why this quest existed.

3. **JS IETF non-delivery test uses `sleep(50)`.** The Rust counterpart uses `now_or_never`, which is the better "must not appear" check. The JS race is consistent with nearby tests and with `MAX_DATAGRAM_AGE_MS`, but it is the only new timing flake surface. Not a reason to hold the PR.

4. **Rust IETF test inserts datagram sequence `0` after `write_frame` already consumed `0`.** Harmless for a non-delivery assertion; a sequence past the group counter would make the setup less confusing.

5. **Pre-existing JS/Rust produce-path differences are untouched** (Rust converts timestamp into the track timescale and enforces `final_sequence`; JS does not). Out of scope, and the new tests do not pretend they converged.

## Tests

The added matrix is the valuable part of the diff, not the rename itself: exact id, gap, out-of-order, duplicate, stale-after-append, close/abort, size cap, cloned/replaced producers, lite wire survival, IETF non-delivery. That matches the quest plan. Changelog in `rs/moq-net` is updated. Quest file is closed out.

Public API impact is as stated: breaking rename and signature on `track::Producer` / `Track.Producer` only. No wire change. Targeting `dev` is correct.

## Recommendation

Merge. Do not add an alias. Do not expand this into a group-verb rename or a JS/Rust semantics unification.

(Written by grok-4.6)

kixelated:
Merged via auto-merge (squash) onto `dev`. CI green at merge time: Check/Test/WASM SUCCESS, merge state CLEAN/MERGEABLE.

Summary of changes (breaking, no wire change):
- Renamed produce-path `writeDatagram(Datagram)` / `write_datagram` to `insertDatagram(sequence, timestamp, payload)` / `insert_datagram` on `track::Producer` / `Track.Producer`, keeping `append*` as the allocator. Relays and future E2EE transforms commit at a chosen identity without minting a receive-side `Datagram`.
- Updated lite subscribers (JS/Rust) to insert the wire sequence, plus changelog and quest close-out.
- Added produce-path test matrix: exact id, gap, out-of-order, duplicate, stale-after-append, close/abort, size cap, cloned/replaced producers, lite wire survival, IETF non-delivery.

Reviews: grok positive with non-blocking notes only (no changes made per scope: no `writeDatagram` alias, no `writeGroup` rename, no JS/Rust semantics unification). Codex inline asking to retarget to `dev` was already satisfied (base is `dev`); no suggestions on the latest commit.

PR #3674, opened 2026-09-13
feat(libmoq)!: select decoded pixel format and size from C

kixelated:
## Problem

`moq_video_decoder_output` carried only `max_age_ms`: C callers got I420 at native size with no way to ask for anything else, and its layout had to be settled before release since any added field is an ABI break.

## Approach

- `moq_video_decoder_output` gains `format` (`moq_video_pixel_format` discriminant), `width`, and `height`. Zeroed (I420, 0x0) preserves the current default: tightly-packed I420 at native size. The stale additive-layout claim is removed; the struct is versioned by recompilation against the shipping `moq.h`.
- `moq_decode_video` validates the request before resolving anything: unknown formats fail with `InvalidCode`, half-set/odd/unrepresentable sizes with `InvalidConfig`. Accepted requests deliver exactly the requested layout: the backend resize hint is passed through (free on scalers like NVDEC) but the delivery loop enforces the size on the CPU with `Frame::resize` and converts with `into_i420`/`to_rgba`, failing on the terminal callback instead of delivering silently.
- New `rs/libmoq/c-tests/decode-output.c` fixture, compiled and run by the `video_decoder_output_c_fixture` unit test against this build's `moq.h`/`libmoq.a`: layout static-asserts, session-free refusals, and an in-process publish plus three decodes (I420 default, native RGBA, resized I420). Rust e2e tests cover RGBA, resize, and refusals in the standard lane.
- `doc/lib/c/index.md` documents the new knobs. OBS needs no change: the plugin decodes encoded tracks itself and does not call `moq_decode_video` (verified by grep).
- Quest `quest/m1/api-c-decoder-output.md` deleted on completion; its references in `quest/m1/README.md`, `quest/m1/merge-dev.md`, `quest/m2/decoded-frames.md`, and `quest/m2/libmoq-fetch.md` unwired (stale wording fixed inline).

Public API: `moq_video_decoder_output` layout breaks (new fields); `moq_decode_video` gains synchronous validation failures; `moq_video_frame` pixels follow the requested format (I420 `w*h*3/2`, RGBA `w*h*4`); new internal `DecoderOutput` threads the request to the delivery loop. No wire change. `moq.h` regenerates from source via build.rs; no checked-in copy.

ABI break: C callers must rebuild against the new `moq.h`; the struct grew from 8 to 24 bytes. Recorded here for the M1 release proof.

## Alternatives

- Zero-copy GPU frame handles: deferred to the M2 decoded-frames quest; this keeps one CPU frame representation per the plan.
- Compatibility shim or second frame struct: rejected by the plan; the break lands on `dev` before release instead.

(Written by Muse Spark)


kixelated:
Verified for merge: base is `dev`, correct for the intended breaking C ABI (`moq_video_decoder_output` gains `format`/`width`/`height`). Not a draft, merge state CLEAN/MERGEABLE, Check/OBS/Test all SUCCESS. Codex review completed with no findings; no other bot or human reviews to address. No changes made; enabling auto-merge.

(written by muse-spark)

PR #3688, opened 2026-09-14
feat(relay): admit every session through a moq-auth lease

### LINKED ISSUE #3058: moq-relay: a revalidation re-check cannot update a session's tier, and changes its alias only by closing it
Found while reviewing #3041, which merged as 7047347. Not a regression - it is what the new re-check does and does not do with a reply whose *non-scope* fields changed.

`Auth::recheck` scores a reply by `Scope::covered_by`, which compares root, subscribe and publish. The resulting `AuthToken` is then dropped and only the `CacheHints` propagate:

```rust
Fetched::Ok { resp, hints } => match self.authorize(&grant.params, &resp) {
    Ok(token) if grant.scope.covered_by(&token) => Recheck::Valid { hints },
```

So two fields the auth API can legitimately change behave inconsistently:

**`tier` is silently ignored.** An endpoint that re-buckets a connection - moving it to a named billing tier, or off one - has no effect on a live session. The session keeps the tier it was admitted with until it reconnects. That matters because the tier decides which meter pays.

**`alias` closes the session.** The alias becomes the token's `root`, so `covered_by`'s `self.root == token.root` fails and the session is revoked. That is arguably correct - the broadcast would otherwise keep announcing under a root the API no longer assigns - but it is a silent hard disconnect for what may be a benign rename, and it is not obviously the intended contract.

The general question is what a re-check is allowed to *update in place* versus what forces a reconnect. Scope narrowing is settled (close, and the client reconnects into the narrower grant). `tier` and `alias` are not.

`tier` looks the harder half: the session's stats handle is built from the admission token in `connection.rs`, so applying a new tier mid-session means rebuilding that handle, and usage already recorded under the old tier stays there. Worth deciding deliberately rather than inheriting whichever behaviour fell out.

(written by Claude Opus 5)

### LINKED ISSUE #3087: relay: mTLS peers bypass --auth-api-mode, so proxy grants can't refuse or scope them
`--auth-api-mode proxy` (#3044) lets an auth endpoint decide every connection - except mTLS ones, which bypass the mode entirely.

`Auth::verify_mtls` calls `resolve_mtls`, which builds its own request and reads only `alias` and `tier` off the reply. A `200 {}`, or a reply carrying a deliberately narrow `grant`, still becomes an unrestricted publish-and-subscribe token. So an operator delegating authorization to their endpoint cannot refuse or scope an mTLS client through the documented grant response. `resolve_mtls` also hardcodes `host: None`, so host-routed tenants dialing the same path produce indistinguishable lookups.

## Shape

Stop special-casing mTLS in the authorization path. It should build its request through `api_request` (sending `mtls=true`, and `host` in proxy mode) and resolve through `authorize`, like any other connection:

- **token mode**: `mtls=true` satisfies "has a credential" without a JWT or `key` - the cert *is* the token. Absent a grant the peer stays unrestricted, exactly as today, so existing deployments returning `{alias, tier}` are unaffected.
- **proxy mode**: the endpoint returns a `grant` like anyone else, and no grant is a refusal - consistent with the rest of the mode.

## Deliberately NOT in scope: revalidation

mTLS peers keep `revalidate: None`. Not because mTLS is precious, but because a deployed endpoint sending a blanket `Cache-Control: max-age` on every reply would silently arm revalidation on a production relay mesh the moment the relay ships - gating fleet interconnect on that endpoint staying reachable, with no one having chosen it.

If mesh revalidation is wanted later it should be its own change, with the endpoint opting in deliberately for `mtls=true`, and a **relay-side floor** on the staleness window for those requests so an operator who forgets the header doesn't get a fleet that partitions on the first auth blip. Note `stale-if-error` alone is not sufficient protection: it only applies when the endpoint *errors*, so an endpoint that successfully answers "no" still partitions the mesh instantly.

(Written by Claude Opus 5)

### LINKED ISSUE #3603: moq-dev/moq: a client certificate's identity never reaches the authorization decision
## Summary

When a relay is configured with `--server-tls-root`, a peer presenting any certificate that chains to
that root is admitted to **whatever path it dials**, with `publish` and `subscribe` both unscoped
beneath it. The authorization endpoint is told that *some* certificate was verified — `mtls: true` —
and never which one. So an endpoint cannot distinguish one certificate holder from another, and
certificate-scoped entitlement is not merely unimplemented but **not expressible** through the
`--auth-api` contract.

The part that is easy to miss is that the root is not a property of the certificate. It is chosen by
the client, in the URL:

```rust
// rs/moq-relay/src/auth.rs
pub async fn verify_mtls(&self, path: &str, transport: Option<Transport>) -> Result<AuthToken, AuthError> {
    let (root, tier) = self.resolve_mtls(path, transport).await?;
    let mut token = AuthToken::unrestricted(Path::new(&root).to_owned());
    token.tier = tier;
    Ok(token)
}
```

and `unrestricted` grants everything beneath that root:

```rust
pub fn unrestricted(root: PathOwned) -> Self {
    Self {
        root,
        subscribe: PathPrefixes::from(vec![Path::new("").to_owned()]),
        publish: PathPrefixes::from(vec![Path::new("").to_owned()]),
        ...
    }
}
```

`resolve_mtls` does consult the auth API, and it fails closed on an API error, so the endpoint is not
powerless — it can alias the path, set a tier, or refuse. But the only thing it can key any of that on
is the path the client asked for. **The party being scoped picks the scope.**

## What we measured

A relay with `--server-tls-root <affiliate CA>` and `--auth-api`, an estate of two tenants, and one
client holding a valid affiliate certificate and **no token at all**. Each cell is a 30 s subscription;
the figure is payload bytes delivered, taken from the subscriber's capture rather than from a relay log
line.

| Dialled | Result |
|---|---|
| `wbd/cnn` — the channel this affiliate licenses | 7,260,560 bytes |
| `wbd/cnn-intl` — licensed by no affiliate | 6,736,040 bytes |
| `wbd/tnt` — licensed by other affiliates | 7,263,380 bytes |
| `wbd/nobody` — licensed by nobody | 2,408,656 bytes |
| **`rival/cnn` — the other broadcaster's tenant** | **7,093,804 bytes** |

Every cell served. The same client with a *token* instead of a certificate is refused on four of those
five, with zero payload bytes, exactly as the grants say it should be — so the token path is working
and it is the certificate path that bypasses it.

## Why this is worth changing rather than documenting

The comment in `resolve_mtls` states the rationale plainly — "mTLS peers are already trusted (the cert
is the credential)" — and for the case it was written for, cluster peers dialling `/`, that is right.
The difficulty is that the same switch is the natural one to reach for when an operator wants mutual
authentication for *subscribers*, and there it silently converts a per-affiliate credential into a
master key for every tenant on the relay. Nothing warns at startup, nothing appears in the session, and
the authorization endpoint — the one component that holds the licensing matrix — cannot see that it has
been bypassed.

It also fails in a direction that is hard to detect from the outside: everything works. The affiliate
gets its channel. The misconfiguration only becomes visible when someone dials a path they should not
have, which is exactly the case nobody tests.

## What we would suggest

Minimally, and without changing any default:

1. **Carry the peer's certificate identity in `AuthApiRequest`.** A subject or a SAN, or the leaf's
   SHA-256 fingerprint, as one more optional query parameter beside `kid`. That alone makes
   certificate-scoped entitlement expressible: the endpoint can return a narrower `public.subscribe`
   for a subscriber certificate, or refuse outright, using the same reply shape it already has.
2. **Consider not granting `publish` unscoped to an mTLS peer by default**, or making the grant
   configurable separately from `subscribe`. A subscriber certificate that can publish into the tenant
   it is reading is a larger grant than the deployment probably intends.
3. **Warn at startup** when `--server-tls-root` is set alongside `--auth-api`. The combination is not
   wrong, but it means the API's answer is advisory for one class of peer and binding for the other,
   and that is worth one line on stderr. We are **not** claiming a precedent for this: an earlier
   version of this issue said the relay already warns about configurations where nothing can
   authenticate, and that was wrong. It does not. Every `tracing::warn!` in `Auth::new` on
   `fd4f5d82e` is a deprecation notice for a superseded flag (`--auth-tls-*`, `--auth-key-dir` with a
   URL, `--auth-public-api`), and the remaining checks are `anyhow::ensure!` hard errors on mutually
   exclusive flags. There is no existing diagnostic for a configuration that is merely hazardous, so
   this asks for a new class of warning rather than one more instance of an established one — which
   may make it a larger ask than (1) or (2), and it should be judged on the hazard rather than on
   consistency with something that isn't there.

We are happy to prepare the `AuthApiRequest` change if the direction is acceptable; the reply-side
contract needs no change for (1), only the request.

## What we are not claiming

- This does not contradict the documented split of mTLS for data-plane peers and JWTs for subscribers.
  It is the cost of ignoring that split, and the point is that the relay does not make the split
  enforceable.
- We have not tested a cluster deployment. Every measurement above is one relay, and how this
  interacts with cluster-peer trust is not something we can speak to.
- Certificate *revocation* was not measured. An all-or-nothing credential has no per-channel
  de-provisioning behaviour to measure, so we cannot say how quickly withdrawing a certificate takes
  effect.

kixelated:
Completes `quest/m1/auth/relay.md`, the last quest of the Auth server line, and deletes the questline. Rebased onto `dev` after #3686 and #3684 merged.

Closes #3087, closes #3603, closes #3058.

## Problem

The relay admits sessions three ways that are easy to confuse: a JWT it verifies itself, an mTLS peer admitted unscoped to whatever it dials (#3603), and `--auth-api` in token or proxy mode with a `Cache-Control` cache an operator has to know about (#3087, #3058). The contract, the server, and the package now exist; the relay still holds all the old policy and an HTTP client of its own.

## Approach

- **`rs/moq-relay/src/auth.rs`** is 6k lines shorter: `AuthConfig` is `url` plus `public`/`public_subscribe`/`public_publish` patterns, validated as exactly one. `Auth` is `Mode::Server(moq_auth::Client)`, `Mode::Public(Grant)`, or `Mode::Refuse` (a `--cluster-lan`-only `moq` process). `Auth::admit` returns a `Lease` (the consumer plus, for a static grant, its own producer) and an `AuthToken` reduced from the grant through `Pattern::as_prefix`, refusing any other pattern by name. `request_for` builds the `moq_auth::Request` from a `moq_tokio::Request`; `peer` turns a `PeerIdentity` into certificate facts.
- **Every accept path** goes through it: `connection.rs` (the tokio listener), `uring.rs` (the io_uring workers), `websocket.rs` (the axum upgrade, `ConnectInfo` for the peer address), and the one-shot `/fetch` and `/announced` routes, which are admitted as `transport: http` sessions; `/announced` ends its lease when the handler returns, `/fetch` holds it in the response body so a refusal or expiry mid-transfer cuts the stream and the `end` event carries the outcome. The LAN mesh credential stays local: `Cluster::lan_peer_grant` feeds `Auth::admit_fixed`. `MtlsPeer` carries the identity so HTTPS and WebSocket report it too.
- **`supervise`** selects on `lease.changed()` and `lease.closed()` beside the session: a changed root or a grant that no longer covers the session closes it with `Unauthorized` (the re-check is rebuilt from the dialed path the token keeps, so a grant that drops its `root` alias resolves back to what was dialed), a revocation closes it with the reason, the session's own close is reported through `lease.close` so the `end` event carries it, and the transport's byte totals are read once at close for the counters. A changed tier is logged and applies to the next connection (see Alternatives).
- **moq-tokio**: `Request::remote_addr`, `local_addr`, `server_name`, `alpn` (a `Link` filled by every backend; the WebTransport and WebSocket sub-protocol counts as the negotiated protocol), `PeerIdentity::name` and `issuer` via x509-parser.
- **`moq --listen`** takes the same two flags, refuses to serve with neither or both, and admits each client through `serve_client` with the process origin scoped from the grant; the server-wide origins are gone so an ungranted side is a no-op origin rather than everything.
- **Deleted**: the shared HTTP cache, in-flight coalescing, `Cache-Control` parsing, `verify_mtls`, `AuthToken::unrestricted`, `--auth-key`, `--auth-key-dir`, `--auth-public-api`, `--auth-domain`, `--auth-api`, `--auth-api-mode`, `--auth-mtls-tier`, `--auth-tls-*`, `jsonwebtoken`, `wiremock`, `hex`, `sha2`. `http_client.rs` stays for `--cluster-connect-api` polling (see Alternatives).
- **Docs**: `doc/bin/relay/auth.md` rewritten around the contract; `config.md` `[auth]`, `cluster.md`, the relay README, `packaging/moq-relay/relay.toml`, the demo and test TOMLs, `nix/modules/moq-relay.nix` (a `moq-auth` unit behind `auth.enable`, `MOQ_AUTH_URL`; `MOQ_AUTH_PUBLIC` as a pattern otherwise), and the relay, CLI, and tokio changelogs.
- **Tests**: `tests/auth_lifetime.rs` rewritten against a scripted server and `moq_auth::serve::Server`: admit with every request fact, the `/announced` and `/fetch` routes admitted as `http` sessions with the fetch lease held while the body streams and a refusal cutting it, 403/5xx/garbage refusing at connect, a moved tier keeping the session, a narrower grant closing over TCP and WebSocket, a refusal closing with `end: refused`, an outage surviving until `expires` with `end: expired`, `end` with duration and bytes over QUIC, and an mTLS peer refused with no grant and scoped by a narrow one over QUIC and WebSocket. `released_cli.rs` drops the deleted spellings. The uring mTLS test runs a `moq_auth::serve::Server`; cluster, GOAWAY, smoke, and drills fixtures use `--auth-public '**'`. Config merge tests cover `auth.url` and the pattern lists from CLI, env, and TOML. Exercised the built binaries by hand: `moq auth serve` + `moq-relay --auth-url` + `moq import/export` over TCP with a token, a bad token refused, anonymous admitted under the public rules, and both startup guards. `just test smoke --publishers rust --subscribers rust,js-native-bun` passes.

## Impact

- Breaking on every relay auth flag: `--auth-url` or `--auth-public` (patterns), nothing else. Setting both or neither fails at startup. A client certificate admits what the server grants; a mesh needs `moq auth serve --mtls-publish '**' --mtls-subscribe '**'` or `--auth-public '**'`.
- Breaking on the embedding surface: `Connection` holds a `Lease`, `supervise` is public and takes it, `AuthToken::new(path, &Grant)`, `MtlsPeer(PeerIdentity)`, `Auth::refuse`.
- `moq --listen` refuses to serve without an auth flag; `--cluster-lan` alone admits peers only.
- `moq_auth::Transport::Http` added for the relay's one-shot HTTP routes (`@moq/auth` schema too).
- moq-tokio: additive request accessors and `Link`; `PeerIdentity` gains `Debug`, `name`, `issuer`.
- Wire: none.

## Alternatives

- **Tier change on revalidate.** The quest asks for an in-place retag; moq-net's stats carriers resolve `(path, tier)` counters once per subscription with lock-free meters, so that is a moq-net redesign. Per the maintainer's call in this session: keep the admitted tier, log the change, and file `quest/m2/stats-retier.md`.
- **io_uring link facts.** `moq_uring::quic::Connection` exposes no peer address or SNI, so the workers report `alpn` only; `quest/m2/uring-link-facts.md` covers the rest. The tokio listener reports everything.
- **Dropping `reqwest-middleware`/`http-cache-reqwest`.** The cluster's `--cluster-connect-api` polling still uses the caching client, so they stay; only the auth uses are gone.
- **`Transport::Http`** rather than reporting `/fetch` as `tcp`: the server should be able to tell a one-shot HTTP request from a MoQ session.
- **`/fetch` body as one owned stream.** The hand-rolled `http_body::Body` rebuilt its future on every poll, dropping the kio waiters a live group registers, so a fetch of an open group only woke at the deadline. `Body::from_stream(unfold(..))` owns the lease and the group for the body's life; the new test exercises it.


kixelated:
Rebased onto dev (post-#3686) and verified the Codex round. All three findings were valid and are addressed by 88b0b0c80/e18c35d4f already on this branch: the token keeps the dialed path and rechecks rebuild from it (unit-tested in auth.rs), the /fetch body owns its lease with mid-transfer refusal cutting the stream (covered by http_routes_hold_a_lease), and the HTTPS root docs plus demo TOMLs describe the certificate as a fact.

My changes on top of that stack:

- d0c03ac04/12622e7f4 follow-through was already here; I added c87eacd21 which gates the /announced assertion on a subscriber seeing the broadcast first. Without it the test raced gossip and failed on dev with an empty body (200 with "" instead of "test").
- Restored this branch to 12622e7f4 after my earlier force-push overwrote it mid-rebase (apologies, parallel sessions crossed); the only content of mine kept is the test stabilization above.

Verified: just fix clean, just check origin/dev green, just test default origin/dev 4578 passed / 9 skipped. Auto-merge (squash) enabled.

(written by Muse Spark)

PR #3694, opened 2026-09-14
fix(moq-net): a remote source never displaces a local publisher

kixelated:
## Summary

- A client that shares one origin between its publish and subscribe halves sees its own announce echoed back by a relay that does no loop detection. Base moq-transport carries no hop ids, so an anonymous relay's echo arrives with an `UNKNOWN` first hop, which the exposure-based reflection check (`FrontState::excluded`) cannot recognize.
- Root cause: `attach_source` treated that echo as a rival publisher and let it take the path. The local publisher was parked, its `PUBLISH_NAMESPACE` withdrawn, and the relay's `SUBSCRIBE` served from the echo route, i.e. the client subscribing to the relay for its own broadcast (`replacing a live broadcast from a different publisher` followed by `parking this source until it ends`, seen against Cloudflare's moqt-16 relay).
- Fix: a front held by an announced local source (empty hop chain) parks any remote source instead of yielding to it. The process publishing a path is that path's origin. Regression test fails without the gate.
- `dev`'s origin model already resolves a local exact-path broadcast ahead of any route, so this is `main`-only.

## Public API

None.

## Wire

None.

(Written by Claude Opus 5)


kixelated:
Addressed the Codex P1 (parked remotes never retried after the local hold released): the parked wait in run_source now snapshots holds_local_publisher and wakes when the hold changes shape, so a local unannounce or linger-detach lets the waiting announced remote retry and take the path. Added regression test test_parked_remote_retries_when_local_unannounces, verified it fails without the fix (path stays dark, next blocked) and passes with it. just fix, just check, and just test all clean (3272 passed).

Public API: none. Wire: none.

(written by Muse Spark)

PR #3699, opened 2026-09-15
fix(moq-net): expire cached groups in order

kixelated:
## Summary

- Root cause: the expiry sweep was one rotating window of `EVICT_SCAN` (4) entries per write over an eviction order padded with dead entries. At a few groups per second a position is revisited only after newer groups have already expired, so a group past the retention window outlived them. The cache became a contiguous tail with stragglers below a hole, and a `group_start=0` subscriber replaying history got the stragglers, a gap, then the tail (seen as `first 339..347, GAP 348..389, 390..` on a 5 msg/s chat track).
- Fix: drain the front of the eviction order (oldest first, stopping at the first fresh entry) before the rotating window, which stays so entries behind a fresh one still expire. Regression test simulates 110s at 3 groups/s and asserts the cache is contiguous after every write; it fails without the drain at write 22.
- Subscriber behavior is untouched: the cursor still starts exactly where asked, the cache just no longer has holes.
- `dev` budgets delivery by media timestamp against the live edge and sweeps the pool separately, so this is `main`-only.

## Public API

None.

## Wire

None.

(Written by Claude Opus 5)


kixelated:
You can review this, but please abandon the PR if there's any issue, especially on the performance side, that can't be easily fixed. Dev is fixing this.

kixelated:
Closing per the maintainer's abort condition on this PR.

The new front drain in `evict_expired` (`while !self.evict.is_empty()` draining from index 0 under the TrackState write lock in `commit_group`) is an unbounded O(backlog) walk on the publish write path: after any idle period longer than max_age a single write aborts the entire expired queue, causing a latency spike. Bounding it per write reintroduces the straggler holes this PR fixes, so there is no easy fix that preserves the contiguity goal. Dev already resolves expiry differently (media-timestamp budgets + pool sweep), so main does not need this.

Abandoning without merging.

(written by Muse Spark)

PR #3702, opened 2026-09-15
docs(cluster): revise the cluster draft to -01 and plan its implementation

### LINKED ISSUE #3693: moq-cluster: setup option should be even
```
RELAY_HOPS Setup Option {
  Option Key (vi64) = 0x40B55
  Option Value Length (vi64)
  Hop ID (vi64)
}
```

The length fields is redundant.  Should be this?

```
RELAY_HOPS Setup Option {
  Option Key (vi64) = 0x40B56
  Hop ID (vi64)
}
```

### LINKED ISSUE #3695: moq-cluster: RELAY_COST is one-dimensional and client-only
There are many ways to measure the cost of a link, latency, hop count, dollars.  Yes it could be boiled to a single metric, but routing algorithms may want to get all the data they can and run some fancy local algorithm.

Further, 

> Both endpoints add the client's value to the ROUTE_COST of every advertisement they receive over the connection, so the link is priced the same in both directions.

There's probably examples where this assumption is not true, so it would be good to be able to express that asymmetry.  It seems like the value is expressing the "upstream" (e.g. towards the publisher) cost, so it shouldn't be hard to have it be asymmetric?

### LINKED ISSUE #3697: moq-cluster: supporting multiple publishers of the same namespace
> they are distinct publishers reusing a namespace: a receiver MUST NOT treat them as interchangeable and SHOULD treat the later as *replacing* the earlier.

moq-transport explicitly allows multiple publishers of the same namespace, and requires relays to send e.g. SUBSCRIBE to each of them if they both match.  I'm not saying this is my favorite feature or that I've implemented it.

Even if e.g. the application's use of the namespace are scoped to a single broadcast, there are still use cases like fully redundant ingest for high value broadcasts, where there truly are two publishers with identical content.

### LINKED ISSUE #3698: moq-cluster: use REQUEST_UPDATE to update cost on PUBLISH_NAMESPACE
No need to allow a second PUB_NS on a single stream.

The namespace case does require sending it a second time (without NS_DONE) or perhaps we add a NS_UPDATE message to carry the cost update.

kixelated:
Answers afrind's review of draft-lcurley-moq-cluster-00 (#3693, #3695, #3697, #3698). The draft moves now; the two wire changes get m1 quests.

## Draft changes (`drafts/draft-lcurley-moq-cluster.md`)

- **RELAY_HOPS becomes HOP_ID** on the even key 0x40B54, so the value is a bare varint instead of a length-prefixed one (#3693).
- **PUBLISH_NAMESPACE updates use REQUEST_UPDATE** on the request stream; an omitted parameter keeps its value, so a warm relay sends ROUTE_COST as an explicit 0. HOP_PATH and ROUTE_COST are registered for REQUEST_UPDATE. NAMESPACE is a response and is still re-sent on its stream (#3698).
- **One publisher per namespace**: a new section states that a session advertises a namespace at most once, a subscription is served from one source, and a second publisher replaces the first, with the reasons (SUBSCRIBE cannot name a publisher, no shared history, per-publisher advertising only moves the choice downstream). Closes #3697 as won't-do.
- **Cost model**: RELAY_COST stays one dimensionless scalar per direction, citing OSPF's per-interface output cost (RFC 2328), RIP (RFC 2453), IS-IS wide metrics and their unimplemented companions (RFC 5305), OSPF's deleted TOS metrics (RFC 2178), and BGP MED's same-neighbour comparability (RFC 4271) for the trust-boundary clamp. Closes #3695 as won't-do for multi-dimensional cost; the asymmetry ask was already met by the per-direction text.
- Changelog appendix added for -01.

## Quests

`quest/m1/cluster-01/`: [S] HOP_ID option (closes #3693 when done, branch from main) and [M] Cost update (closes #3698 when done, branch from dev, publishes -01 as its last step).

## Public API and wire impact

No code changes. The draft's wire changes (option key and value encoding, update message) land with the quests; nothing deployed negotiates -00, and a -00 peer meeting a -01 peer sees an unknown Setup Option on each side and simply does not negotiate the extension.

The draft on dev has diverged (pattern extension, selection); the new section is written to hold under either text, and the merge keeps dev's facts.

`just check` passes (drafts and quest checks).

Closes #3695
Closes #3697


kixelated:
Merged. Verified: HOP_ID on even key 0x40B54 as bare varint with matching registry and parity text, REQUEST_UPDATE for PUBLISH_NAMESPACE with explicit-0 repricing while NAMESPACE stays re-sent, one-publisher-per-namespace section with valid anchors, cost model with RFC citations and trust-boundary clamp. All referenced quest files exist. Local `just drafts check` and `cargo run --package quest -- check` (320 docs ok) pass; CI Check and Test pass. No code changes, so no fix needed.

(written by Muse Spark)

PR #3703, opened 2026-09-15
fix(relay): gate the embed test UDP port helper to Linux

kixelated:
**Problem**: `just check` fails on macOS with `function free_udp_port is never used` at `rs/moq-relay/tests/embed.rs:27`. The helper is only called from tests gated on `target_os = "linux"`, so under `-D warnings` the whole check aborts. Linux CI never sees it.

**Approach**: Gate `free_udp_port` with `#[cfg(target_os = "linux")]` and drop the now-unused `UdpSocket` import. Verified locally: clippy fails on macOS without the change and passes with it.

**Impact**: none, test-only.

**Alternatives**: making the worker tests portable is not possible since the mode relies on `SO_REUSEPORT`.

(written by Claude Opus 5)


kixelated:
Superseded: the same gate already landed on dev in d5621226b via the main->dev merge.

(written by Claude Opus 5)

PR #3715, opened 2026-09-17
draft(flate): specify group-scoped DEFLATE as an IETF draft

kixelated:
`moq-flate` / `@moq/flate` implement a wire format that was only written down inside the hang draft, in a section about hang's metadata tracks. This lifts it into `drafts/draft-lcurley-moq-flate.md` so any track can use it, and cuts hang's copy down to a cross-reference.

It is a payload encoding, not a transport extension: nothing is added to the wire, no code point is requested, and the application declares a track compressed the same way it already declares the container and codec.

## What it says

- **Scope**: each subgroup is a separate DEFLATE stream, its payloads sharing one window in order, starting cold. moq-lite has no subgroups, so each group is one stream of frames. Datagrams are refused, being neither ordered nor reliable.
- **Stream**: raw DEFLATE (RFC 1951, no zlib/gzip wrapper), a sync flush after each payload, the fixed `00 00 ff ff` marker stripped and re-appended (the RFC 7692 permessage-deflate trick), no length prefix, zero-length payloads skipped. The stream is never terminated, so a consumer decodes incrementally and does not read the absent end of stream as truncation. This matches what both implementations already emit byte for byte.
- **Declaring it**: the application, never the transport. The `.z` track-name suffix is the convention; a catalog may declare it instead. Sniffing is forbidden, since raw DEFLATE has no magic number.
- **Security**: the CRIME/BREACH size side channel of a shared window (a secret MUST NOT share a subgroup with attacker-influenced data), decompression bombs (bound per object, abandon the subgroup), and per-subgroup window memory.

## Why no transport property

An earlier revision of this branch registered a `FLATE` track property, first in the first-come-first-served range and then in the Mandatory Track Property range. Both are worse than the application declaring it, and the draft now says so:

- An ignorable property lets a consumer predating the document hand compressed bytes to its application. A mandatory one makes every relay on the path understand it before a compressed track can traverse at all.
- More fundamentally, a transport-level declaration makes the compression window the transport's business. A relay drops objects and re-frames what it forwards, so it would be on the hook for emitting a stream that still decodes, which means carrying a DEFLATE implementation. Keeping the encoding inside the payload keeps that responsibility with the endpoints: the relay forwards opaque bytes and a consumer that receives an incomplete subgroup simply cannot decode the rest of it.

## Public API and wire impact

None of either. No code changes, and the format lives entirely inside object payloads, which moq-transport treats as opaque: no message, parameter, property, or version. IANA Considerations requests no registrations.

Hang's `# Compression` section keeps the `.z` convention and now cross-references the new draft instead of restating the mechanics, which had already drifted: hang was silent on empty frames where both implementations skip them.

## Checks

`kramdown-rfc --v3` + `xml2rfc --preptool` over every draft (the `just drafts check` pipeline, run with the tools installed directly since the nix shell is not available in this environment) and `bun run --cwd doc check`, which renders the new `/draft/moq-flate` page and the edited hang page through VitePress. No new kramdown construct, so `doc/.vitepress/drafts.ts` needed no case.

https://claude.ai/code/session_01SRMBGVHohyvUtYg4N2hG8E


drafts/draft-lcurley-moq-flate.md
```diff
@@ -0,0 +1,161 @@
...
+A consumer that is missing an object MUST NOT decompress any later object in that scope; every later object in it is unrecoverable, and the consumer SHOULD abandon the scope (in {{moqt}} terms, stop reading the stream) rather than deliver corrupt payloads.
+
+The compression level, block types, and flush strategy beyond the mandatory per-object sync flush are a publisher's choice.
+Any {{RFC1951}}-conformant stream decodes.
+A sync flush is `Z_SYNC_FLUSH` in zlib and its ports, so this format needs no DEFLATE implementation of its own.
+
+
+# Declaring Compression {#declare}
+The FLATE property declares that every object payload on a track is compressed per this document.
+It is a track-level Key-Value-Pair ({{moqt}} Section 2.5), carried with the track's properties and delivered in SUBSCRIBE_OK or TRACK_STATUS ({{moqt}} Section 12).
```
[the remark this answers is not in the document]
kixelated: Confirmed and fixed in `bc2335b6d`.

Checked against draft-ietf-moq-transport-21 rather than only the implementation: Track Properties are the final field of `SUBSCRIBE_OK` (Fig. 11), `PUBLISH` (Fig. 12), `FETCH_OK` (Fig. 16), and `REQUEST_OK` where they are populated only for `TRACK_STATUS_OK` (Section 8.4). The draft now names all four. `TRACK_STATUS` alone was wrong; that phrasing came from the timestamp draft, which is against older numbering.

The same read turned up something better for the underlying hazard you describe. Section 3.6 reserves `0x4000`-`0x7FFF` for Mandatory Track Properties: an endpoint that does not understand one MUST NOT process or forward the track, and refuses with `UNSUPPORTED_EXTENSION` on `PUBLISH`, `SUBSCRIBE_OK`, and `FETCH_OK`. That is exactly this property, so FLATE moved from `0xF1A7E` (first-come-first-served, ignorable) to `0x7F1A`. A consumer predating this document now fails loudly instead of decoding garbage, whichever message it learned the track through. The cost, that a relay predating the document declines to forward a compressed track at all, is stated in the security considerations. Section references are updated to draft-21 numbering (8.3, 8.4, 3.6, 10.7, 16.8).

---

drafts/draft-lcurley-moq-flate.md
```diff
@@ -0,0 +1,161 @@
...
+The payloads of a scope, concatenated in object order, are compressed into a single raw DEFLATE stream {{RFC1951}}, with no zlib or gzip wrapper.
+
+A publisher MUST perform a sync flush after each payload: an empty stored block ({{RFC1951, Section 3.2.4}}) that ends the DEFLATE block, byte-aligns the output, and retains the window.
+Each object carries exactly the bytes the DEFLATE stream produced for its payload, with no length prefix; the transport already frames objects.
+The result is that each object is independently addressable on the wire while still compressing against its predecessors.
+
+A sync flush always ends with the fixed empty-block marker `0x00 0x00 0xff 0xff`.
+A publisher MUST omit this trailing marker from each object and a consumer MUST append it before decompressing, the same trick as permessage-deflate ({{RFC7692, Section 7.2.1}}).
+
+A zero-length payload is carried as a zero-length payload; a publisher MUST NOT feed it to the DEFLATE stream and a consumer MUST NOT feed it to the inflater.
```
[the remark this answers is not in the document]
kixelated: Agreed, fixed in `bc2335b6d`. This was the open question in the PR description, and option 2 (the one I recommended) is what landed: hang keeps the `.z` convention and cross-references this draft for the format, so the mechanics exist in one place.

On the divergence you name, the two did not strictly contradict each other, since hang says nothing about empty frames, but it is underspecified there and both implementations skip them, so the point stands either way and the cross-reference settles it.

Hang references the new draft through an inline reference entry with a datatracker target, matching how moq-lite references qmux-websocket: `I-D.lcurley-moq-flate` has no bibxml until this is submitted, so the build would fail on the fetch.

@kixelated if you would rather hang stay self-contained until this draft is published, `git revert bc2335b6d` drops only the hang change plus the property rework, so say the word and I will split it.

---

drafts/draft-lcurley-moq-flate.md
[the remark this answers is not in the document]
kixelated: Superseded by `ac492a471`: there is no `FLATE` property any more, so the message list this thread was about is gone with it.

The maintainer's call, and it is the right one. A transport-level declaration makes the compression window the transport's business, and a relay drops objects, serves a FETCH over a subgroup it holds only part of, and re-frames what it forwards. It would be on the hook for emitting a stream that still decodes, which means carrying a DEFLATE implementation. The mandatory range also forces every relay on the path to understand the property before a compressed track can traverse at all, which is the opposite of what a relay should need to know about a payload.

Compression is declared by the application now, by the `.z` suffix or a catalog field, and IANA Considerations requests no registrations.

---

drafts/draft-lcurley-moq-flate.md
```diff
@@ -0,0 +1,140 @@
...
+
+Compressing a whole track or session captures that redundancy but is incompatible with how MoQ delivers.
+Groups are dropped under congestion, arrive out of order, and a subscriber joins at an arbitrary point.
+A window spanning data the subscriber never received cannot be reconstructed, so a single drop would end the track.
+
+This document takes the middle position: the window is scoped to what the transport already delivers ordered and complete, and resets at every boundary.
+A dropped or late group costs nothing beyond itself, and joining mid-track costs only the current group.
+Compression stays invisible to the transport, so a relay routes, caches, and drops a compressed track exactly as it does any other.
+
+{{hang}} already compresses its metadata tracks this way.
```
kixelated: remove the hang reference.

drafts/draft-lcurley-moq-flate.md
```diff
@@ -0,0 +1,140 @@
...
+This document specifies the format independently so that any track can use it, and so that an implementation has something to conform to that is not tied to one catalog format.
+
+
+# Compression Scope {#scope}
+The compression scope is the **subgroup**: the object payloads of one subgroup, in ascending object order, share one DEFLATE window.
```
kixelated: require using sub-group 0 only.

drafts/draft-lcurley-moq-flate.md
```diff
@@ -0,0 +1,140 @@
...
+A subgroup is the largest unit {{moqt}} delivers reliably and in order, which is exactly the requirement.
+
+Each scope is independent and begins with a cold window.
+An endpoint MUST NOT carry DEFLATE state across scopes, and MUST key its state by the enclosing track, group, and subgroup rather than by the stream that happens to carry them: a FETCH delivers objects from many subgroups over one stream.
```
kixelated: This is kinda verbose/dumb.

Like duh, otherwise you can't decompress.

drafts/draft-lcurley-moq-flate.md
kixelated: Done in `8671849b7`. A publisher MUST send a compressed track in Subgroup 0 only and MUST NOT send it in datagrams; a consumer MUST reject a track that does either.

This turned out to delete most of the section. With one subgroup the scope is the group on every transport, so the rule about keying state by track/group/subgroup is gone (it only existed because a FETCH interleaves subgroups on one stream), the datagram scope is refused rather than defined, and moq-lite needs one line instead of a mapping.

---

drafts/draft-lcurley-moq-flate.md
kixelated: Fair. Cut in `8671849b7`, along with the rest of the prose that restated what the rules already imply.

The whole scope section is now four sentences, and the ordering rule survives only as the consequence worth stating: a consumer missing an object MUST abandon the rest of the group. The draft is 106 lines, down from 138.

---

drafts/draft-lcurley-moq-flate.md
kixelated: Done in `fa941845b`. Removed both citations and the informative reference entry, so the draft no longer cites hang at all: hang references this draft now, and pointing back tied a general format to one catalog.

---

drafts/draft-lcurley-moq-flate.md
```diff
@@ -0,0 +1,138 @@
...
+
+# Compressed Stream {#stream}
+The payloads of a scope, concatenated in object order, are compressed into a single raw DEFLATE stream {{RFC1951}}, with no zlib or gzip wrapper.
+
+A publisher MUST perform a sync flush after each payload: an empty stored block ({{RFC1951, Section 3.2.4}}) that ends the DEFLATE block, byte-aligns the output, and retains the window.
+Each object carries exactly the bytes the DEFLATE stream produced for its payload, with no length prefix; the transport already frames objects.
+The result is that each object is independently addressable on the wire while still compressing against its predecessors.
+
+A sync flush always ends with the fixed empty-block marker `0x00 0x00 0xff 0xff`.
+A publisher MUST omit this trailing marker from each object and a consumer MUST append it before decompressing, the same trick as permessage-deflate ({{RFC7692, Section 7.2.1}}).
```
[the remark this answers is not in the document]
kixelated: Confirmed and fixed in `dc3da7ba6`, in one sentence:

> The stream is never terminated: no object ends in a final block, so a consumer decompresses incrementally and MUST NOT treat the absent end of stream as truncation.

That is the incremental-decoder half of your suggestion. Not the terminal-block half: a final block would have to be written when the group ends, which the publisher does not always get to do (a group can be reset or abandoned mid-flight), and a consumer would then have to tolerate its absence anyway. Requiring the tolerance and nothing else is the shorter rule and the one both implementations already satisfy.

---

drafts/draft-lcurley-moq-flate.md
```diff
@@ -0,0 +1,108 @@
...
+    fullname: Luke Curley
+    email: kixelated@gmail.com
+
+normative:
+  moqt: I-D.ietf-moq-transport
+  RFC1951:
+  RFC7692:
+
+informative:
+  moql: I-D.lcurley-moq-lite
```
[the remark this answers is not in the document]
kixelated: Right that the classification and the `MUST NOT` were inconsistent, fixed in `9510ed87c` by taking your second option, since the requirement was redundant anyway.

A publisher already MUST NOT send a compressed track in datagrams, on any transport. moq-lite's datagram form of a group is a datagram, so restating it there said the same rule twice, against the drafts convention to state each rule once. The line is now descriptive, so `moql` is correctly informative:

> {{moql}} has no subgroups: a group is already one ordered stream of frames.

---

kixelated:
Dropping `--- note_Note_to_Readers` rather than adding translator support here.

The leak is real: this branch's `drafts.ts` has no `--- note_*` case, so the marker would show up as raw text on `/draft/moq-flate`. The construct and its translator already live in #3728, which is the PR that adds the note to every draft. Landing the translator here would duplicate that and conflict; leaving the untranslated marker would break the page until #3728 merges.

This draft goes back to no new kramdown construct. #3728 can add the note after this lands.

(written by grok-4.6)

kixelated:
Squash auto-merge enabled.

This lands `drafts/draft-lcurley-moq-flate.md`: group-scoped raw DEFLATE as an application payload encoding, declared with `.z` (or a catalog field), not a transport property. Hang's compression section now cross-references it instead of restating the mechanics. No public API or wire change.

Rebased onto `origin/main`. Dropped the untranslated `note_Note_to_Readers` so the doc site does not leak kramdown; that construct lands with #3728.

(written by grok-4.6)

PR #3732, opened 2026-09-17
fix(obs): let each dock tab scroll instead of forcing the dock's height

kixelated:
## Problem

The MoQ dock could not be made shorter than its Encoding tab. A `QTabWidget`'s minimum height is its tallest page, and Encoding (mode, note, latency, plus the five-row custom-encoding group box) is taller than a typical dock slot, so OBS refused to shrink the dock row and the preview above it collapsed to a sliver as soon as the dock was shown.

## Approach

Wrap each tab page in a frameless `QScrollArea` (`Scrollable()` in `moq-dock.cpp`): the tab's minimum becomes a few rows, the rest scrolls, and the dock takes whatever height the layout gives it. The viewport keeps the dock's own background rather than the scroll area's base color; the horizontal bar is off since the forms already wrap rows.

## Impact

- None on the wire or the C ABI. UI only.

## Alternatives

- Shrink the Encoding page (collapse the group box behind a toggle). Hides settings instead of fitting them; the scroll area keeps every control reachable at any dock height.

## Follow-ups

- None.

## Verified

Built natively on macOS (Xcode, OBS 32.2.2) and loaded into OBS: `Docks → Reset Docks`, then re-adding the MoQ dock no longer collapses the preview (the dock sits at the Audio Mixer's height), and a floating dock resizes down to roughly half the Encoding form with a scrollbar. Publish (Go Live) and the MoQ Source still work against a local relay. `just obs check` and `just obs compile` pass.

(written by Claude Opus 5)


kixelated:
**MERGE**

Clear UI root cause: `QTabWidget` minsize tracks the tallest page, and Encoding is taller than a normal dock slot. Wrapping each tab in a frameless, resizable `QScrollArea` is the simple fix; collapsing the Encoding group would hide controls instead of fitting them.

Complexity is low and local to the dock. Verified behavior (dock height, floating resize, Go Live still works) matches what the change intends. No wire or C ABI impact.

(Written by Grok)

kixelated:
Squash auto-merge enabled.

Rebased onto `origin/main` (`81b987d82`, #3738). Codex had no findings. CodeRabbit had no actionable comments; the docstring coverage warning is noise for this C++ dock UI change.

(written by grok-4.6)

PR #3733, opened 2026-09-17
fix(msf): compile the extension collision check in release builds

kixelated:
## Problem

`nix build .#moq-cli` (and any `--release` build of `moq-msf`, so the CLI and relay too) fails on `dev` since #3720:

```
error[E0425]: cannot find function `colliding_member` in this scope
   --> rs/moq-msf/src/lib.rs:259:4
```

`colliding_member` is `#[cfg(debug_assertions)]`, but `debug_assert!` expands to `if cfg!(debug_assertions) { assert!(..) }`, so its arguments are still compiled in release. Nothing on `dev` builds release: the release workflows run on tags only.

## Approach

Run the check under a `#[cfg(debug_assertions)] if let Some(name) = ..` block, which also computes the collision once instead of twice. The existing `colliding_typed_extension_trips_the_debug_assertion` test still passes in debug; `cargo check --release -p moq-msf` passes again.

## Impact

- No public API or wire change.

## Alternatives

- Drop the `cfg` on `colliding_member` and let release builds serialize the extension twice per catalog: the check was meant to cost nothing to ship.

## Follow-ups

- Nothing on `dev` or `main` compiles a release profile before a tag. A `cargo check --release` for the shipped binaries in `just check` would have caught this; worth a quest if it recurs.
- Stacked: the nix cacert fix for `moq-cli` is verified on top of this branch.

https://claude.ai/code/session_01CpMVbmhAWE3JwrXF3Cto1o

(written by Claude Opus 5)


kixelated:
**MERGE**

Classic Rust footgun: `debug_assert!` still type-checks its arguments in release, so a `#[cfg(debug_assertions)]` helper is invisible there and the build dies. Wrapping the call in a `cfg` block (and computing the collision once) is the correct, cheap fix.

Complexity is tiny; value is restoring `--release` for `moq-msf` and everything that depends on it. Dropping the `cfg` on `colliding_member` would tax every serialize for a check that should cost nothing to ship.

Worth a follow-up quest to put `cargo check --release` for shipped bins into `just check`, as the description notes. Also: GitHub currently reports this as not mergeable; rebase onto current `dev` before landing.

(Written by Grok)

kixelated:
Rebased onto current `origin/dev` (`cd9aad115`).

The only conflict was `rs/moq-msf/src/lib.rs`. #3723 cfg-gated the `debug_assert!`, then #3718 dropped `#[cfg(debug_assertions)]` on `colliding_member` so the helper still type-checked. That already compiles in release, but it leaves the helper in the ship binary and still serializes twice in debug. Kept this PR's cheaper shape: restore the `cfg` on the helper and run the check in a `cfg` `if let` once.

Locally: `just check`, `just test`, and `cargo check --release -p moq-msf` are green. Squash auto-merge is on.

#3734 stays stacked on this branch; it was not retargeted or rewritten.

(written by grok-4.6)

PR #3734, opened 2026-09-17
fix(nix): give the moq-cli check phase a trust store

kixelated:
## Problem

`nix build .#moq-cli` fails its check phase on `dev`:

```
test tests::cluster_connect_api_http_attaches_client_tls ... FAILED
client: unexpected error: No CA certificates were loaded from the system
```

The test builds the connect TLS config, which loads native roots and (since the rustls-native-certs bump) errors when none are found. The nix sandbox has no system trust store. `moqRelayArgs` already carries the fix; `moqCliArgs` never got it.

## Approach

Add the same two lines to `moqCliArgs`: `cacert` as a native build input and `SSL_CERT_FILE` pointing at its bundle. Verified with `nix build .#moq-cli` before (fails as above) and after (passes) on top of #3733.

## Impact

- No public API or wire change. Nix packaging only.

## Alternatives

- Skip the test under nix (`doCheck = false`): the package would stop running the CLI's tests, which the existing comment above `src` deliberately keeps.

## Follow-ups

- Stacked on #3733: without it the release build never reaches this test. Retarget to `dev` once that merges.

https://claude.ai/code/session_01CpMVbmhAWE3JwrXF3Cto1o

(written by Claude Opus 5)


kixelated:
**MERGE**

Positive, focused, and the right root-cause fix. The sandbox has no system trust store; pointing `SSL_CERT_FILE` at `cacert` is exactly what `moqRelayArgs` already does, and skipping the check would paper over a real failure mode.

Complexity is near zero; value is unblocking `nix build .#moq-cli`. No alternate approach is better.

Land #3733 first (this is stacked on it), then retarget/rebase onto `dev` as the description says.

(Written by Grok)

kixelated:
Rebased onto `origin/dev` after #3733 merged (`d6a28bfc`). The stacked msf commit dropped as already upstream; this PR is now just the two-line `moqCliArgs` trust-store fix (`cacert` + `SSL_CERT_FILE`), matching `moqRelayArgs`. Codex has no findings on `21a11a7`. Squash auto-merge is enabled.

(written by grok-4.6)

PR #3740, opened 2026-09-17
fix(hang): refuse a malformed text catalog section in JS too

kixelated:
## Problem

`@moq/hang` wrapped the catalog `text` section in a blanket `z.catch`, so a captions rendition with a bad field (no `format`) silently vanished in the browser while `rs/hang` `deserialize_section` rejected the same catalog. Completes `quest/m0/js-text-catalog-fallback`.

## Approach

- `js/hang/src/catalog/root.ts`: route `text` through the same narrow `section(TextSchema, "renditions")` fallback as `json` and `binary`. Absent or not-a-section still decodes as absent; a section with a malformed rendition rejects the catalog.
- Tests on both sides of the line: JS (`root.test.ts`) covers absent/foreign parses to empty and a malformed rendition rejects; Rust (`root.rs`) gains the malformed-rendition counterpart its legacy test lacked.
- Fix the stale `js/net` group expiry message, which named the retired `Latency` type; it now says `max age budget` like Rust.
- Fix the Rust doc comment that pointed at the `z.catch`.

## Impact

- Public API: none.
- Wire: none.
- Behavior: a JS consumer now refuses a catalog whose `text` section carries a malformed rendition, matching Rust. A non-section `text` value is still dropped without failing the catalog.
- The `js/net` expiry error message text changed (`latency budget` -> `max age budget`).

## Alternatives

Keep the blanket catch and warn. Rejected: it hides a publisher bug and leaves the two parsers disagreeing.

## Follow-ups

None.

(written by Claude Opus 5)


kixelated:
## Verdict: MERGE

**Positive improvement?** Yes. JS wrapped catalog `text` in a blanket `z.catch`, so a captions rendition missing `format` vanished in the browser while Rust rejected the same catalog. Routing `text` through `section(TextSchema, "renditions")` aligns parsers and stops silently dropping publisher bugs.

**Worth the complexity?** Yes — net complexity drops (shared helper instead of a one-off catch). Behavior stays lenient for absent/foreign keys and strict for malformed sections, matching `json`/`binary`.

**Different approach?** Keeping the catch-and-warn was correctly rejected: it hides publisher bugs and leaves JS/Rust disagreeing. Full reject of any non-section `text` would break pre-captions extension keys.

**Notes:** Tests cover both sides of the line (absent/foreign → empty; malformed → reject) plus the Rust counterpart. The `js/net` “latency budget” → “max age budget” rename is a small quest-noted cleanup; fine bundled here.

(Written by Grok)

kixelated:
Enabling squash auto-merge.

JS hang catalog `text` now uses the same `section()` fallback as json/binary, so a malformed rendition rejects instead of silently vanishing. Matches Rust. Bundled: js/net expiry message rename (`latency budget` → `max age budget`) and a Rust doc fix.

No public API or wire change. Codex had no findings; Check/Test are green.

(written by grok-4.6)

Issue #3532, opened 2026-09-08
moq-net/moq-ffi: no blind subscribe, so a relay that does not announce (Cloudflare draft-16) is unreachable from every non-JS binding


jmvillagra:
### Summary

`moq-net` (and therefore `moq-ffi` and the Kotlin/Swift/Python bindings) can only reach a broadcast the peer has **announced**. `origin::Consumer` offers `request_broadcast` (needs a live route) and `announced_broadcast` (waits for an announcement), and the IETF subscriber only ever creates a `broadcast::Producer` from an incoming `PUBLISH_NAMESPACE` (`ietf::Subscriber::start_announce`).

The JS stack does not have this restriction. `js/net/src/connection/established.ts` exposes:

```ts
/**
 * Consume the broadcast at the given path, immediately.
 *
 * The subscription is reset if nobody publishes the path, so use
 * {@link announcedBroadcast} instead when the broadcast may not be online yet.
 */
consume(path: Path.Valid): broadcast.Consumer;
```

and `js/watch/src/broadcast.ts` uses it with the comment *"No announcement gate: subscribe immediately"*, alongside a `skipDiscovery()` helper that warns *"relay does not support broadcast discovery; subscribing to siblings blind."*

There is no Rust equivalent, so **any relay that does not announce is unreachable from every non-JS binding** — including Cloudflare's provisioned `moq-transport` relays, which is the deployment `README`/[the blog post](https://moq.dev/blog/first-cdn/) point at.

### Reproduction

Subscriber against a Cloudflare draft-16 relay (token in the URL path, as Cloudflare requires):

```
https://draft-16.cloudflare.mediaoverquic.com/<subscribe_token>   broadcast: timing.hang
```

Transport and version negotiation are perfect — this is purely the discovery step:

```
DEBUG web_transport_quinn::connect: sending CONNECT request request=ConnectRequest {
  path: "/<subscribe_token>",
  protocols: ["moq-lite-05", "moq-lite-04", "moq-lite-03", "moql", "moqt-20", "moqt-19",
              "moqt-18", "moqt-17", "moqt-16", "moqt-15", "moq-00"], headers: {} }
DEBUG web_transport_quinn::connect: received CONNECT response response=ConnectResponse { status: 200, protocol: Some("moqt-16") }
 INFO moq_native::client: connected version=moq-transport-16
[...]

PR #3741, opened 2026-09-17
fix(tokio): fail a connect on auth only once every transport has

kixelated:
Closes #3532

**Problem**

`moq_tokio::Client::connect` races a QUIC dial against the WebSocket fallback, and either arm's auth error ended the race on the spot. A WebTransport-only endpoint (Cloudflare's relays answer every non-WebTransport request with 403) refused the fallback while the QUIC arm was still in flight, so the whole connect failed as `Forbidden`, indistinguishable from a bad token, and the reconnect loop treated it as terminal.

**Approach**

- `race_transport_connect` no longer returns early on `is_auth()`. Both arms run to completion, and the pair folds with auth awareness: an auth error only when both arms refused; a mixed pair reports the non-auth failure, which stays retryable; both-non-auth stays `TransportRace` so the log names both.
- `ws://` and `wss://` URLs connect over WebSocket alone instead of racing a QUIC arm that no backend can dial. Without this, that arm's `InvalidScheme` would be the "non-auth" failure that masks a real 401 on the fallback (the existing `websocket_unauthorized_handshake_is_explicit` and `reconnect_stops_on_websocket_unauthorized` tests caught it). This mirrors `websocket::race_handle` yielding `None` for `moqt://` and `moql://`.
- Tests: the four `race_transport_connect_*` unit cases from the quest (QUIC auth then WebSocket success, WebSocket 403 then QUIC success, both refused, WebSocket 403 then transient QUIC), plus an end-to-end `websocket_forbidden_does_not_end_a_quic_connect` in `tests/broadcast.rs` that answers the fallback's TCP dial with a 403 on the QUIC port; it fails 3/3 without the fix.
- Deletes the quest and its references. m0 stays: #3753 added Idle consumer.

**Impact**

- Public API: none. `Error::is_auth()` and `connect_error()` keep their signatures; the observable change is which error a mixed race reports.
- Wire: none.

**Alternatives**

- Always build `TransportRace` when both arms fail and teach `Error::connect_error` to require both halves be auth. Keeps both messages in every log line, but the quest asked for the retryable arm's error to surface directly so `status()` and `connect_error()` read the arm that matters without a special case.
- Race the QUIC arm for `ws://` too and special-case `InvalidScheme` in the fold. Dialing a transport the URL rules out is wasted work and a growing list of "errors that don't count".

(written by Claude Opus 5)


kixelated:
## Verdict: MERGE

Positive, focused fix for a real connect/reconnect failure mode against WebTransport-only endpoints (e.g. Cloudflare returning 403 on the WebSocket fallback while QUIC is still in flight). Complexity is low and justified.

### Why this is good
- Removing the `is_auth()` early returns so both arms finish, then folding with auth awareness, matches how reconnect treats auth as terminal: a single arm's 403 must not abort a still-viable (or merely transient) other arm.
- Fold cases are correct: both-non-auth → `TransportRace`; mixed → non-auth (retryable); both-auth → auth. Preferring the QUIC error when both are auth is fine given the goal that `status()` / `connect_error()` read the arm that matters without special-casing `TransportRace`.
- Short-circuiting `ws://` / `wss://` to WebSocket-only is the right companion change — racing a QUIC arm that can only fail with `InvalidScheme` would otherwise mask a real WebSocket 401. Mirrors the existing `moqt://` / `moql://` behavior on the other side.
- Tests cover the quest matrix (unit) plus an e2e that reproduces the zero-delay 403-on-fallback case. Quest cleanup is tidy.

### Worth the complexity?
Yes. The alternative (always `TransportRace` + teach `connect_error` to require both halves auth) would bury the retryable error behind a wrapper; surfacing the non-auth arm directly is simpler for reconnect. No public API or wire change.

### Different approach?
Not needed. Optional follow-up only: when both arms are auth, prefer keeping both messages in the log somehow — not blocking.

CI Check/Test are green.

(Written by Grok)

kixelated:
Rebased onto `origin/dev` (DIRTY after #3739/#3670 and later landings). Only conflict was `quest/m0/README.md`: this was the last m0 quest, so the empty m0 questline is deleted and merge-dev no longer requires it.

Local `just check origin/dev` (quest/clippy/fmt) and `just test default origin/dev` passed, including `websocket_forbidden_does_not_end_a_quic_connect` and the four `race_transport_connect_*` cases. Codex completed with no findings. Auto-merge enabled.

(written by grok-4.6)

kixelated:
Rebased again after #3753 landed. m0 is not empty (Idle consumer), so the empty-questline deletion was dropped. `quest/m0/README.md` now lists only Idle consumer. Auto-merge still on.

(written by grok-4.6)

PR #3749, opened 2026-09-18
fix(relay): compile the io_uring listener after the auth module move

kixelated:
## Problem

#3670 moved `peer` into `moq_relay::auth` and made `moq_auth::serve::{Policy, Rules}` non-exhaustive, but the io_uring listener and its test still use the old spellings. Nothing in `just check` compiles that feature unless moq-bench is in the diff, so dev is green while `bench check` and the nightly `rs uring` are not. Found while #3745 (which touches moq-bench) hit it.

## Approach

Point `uring.rs` at `crate::auth::peer`, build the test policy through `Policy::default()` and `Rules::new` like `auth_lifetime.rs` does, and give the one clippy `type_complexity` finding on the same path a local alias so `rs uring` lints clean.

## Impact

- Public API: none.
- Wire: none.

## Alternatives

Folding this into #3745: rejected, it is unrelated to that rename and dev should build on its own.

## Follow-ups

None. The nightly `rs uring` job is the gate that would have caught this earlier; it runs on `dev` only if the schedule points there.

(written by Claude Opus 5)


kixelated:
## MERGE

**Positive improvement?** Yes. After #3670 moved `peer` under `moq_relay::auth` and made `Policy`/`Rules` non-exhaustive, the io_uring listener and its test were left on the old spellings. That only shows up when the `uring` feature is compiled (`bench check` / nightly `rs uring`), so `dev` looked green while those paths were broken. This restores them.

**Worth the complexity?** Yes, and there is almost none. Three focused edits: `crate::auth::peer`, build the test policy the same way `auth_lifetime.rs` already does (`Policy::default` + `Rules::new`), and a local `Set` alias for the existing clippy `type_complexity` finding on that path.

**Different approach?** Folding into #3745 would hide an unrelated compile break inside a rename PR. Shipping this alone is the right call so `dev` builds on its own without waiting on that work.

Head SHA reviewed: `e81cb8d08f5b3d61548bf234f821a7b35dc64827`

(Written by Grok)

kixelated:
Landing this as-is. Check and Test are green, the PR is MERGEABLE, and Codex finished with no findings. No merge conflicts or review fixes.

Squash auto-merge is enabled. The change points the io_uring listener at `crate::auth::peer` after #3670, builds the uring test policy through `Policy::default()` and `Rules::new`, and aliases the clippy `type_complexity` finding on that path.

(written by grok-4.6)

PR #3752, opened 2026-09-18
test(mux): stop copies_share_one_epoch racing the clock

kixelated:
`copies_share_one_epoch` compared two live `micros()` readings, which differ whenever a microsecond ticks between the calls. It flaked in CI with `left: 16, right: 17` (https://github.com/moq-dev/moq/actions/runs/35295667959/job/105447584451).

The test now builds the clock with `Clock::new_at` and compares the copies' `epoch` and `wall` anchors directly, which is what "share one epoch" actually means.

No public API or wire impact; test only.

Verified with `cargo test -p moq-mux clock` (24 passed).

(written by Claude Opus 5)


kixelated:
**MERGE**

Positive improvement: yes. `copies_share_one_epoch` was asserting equality of two live `micros()` samples, which can disagree by a microsecond whenever the wall ticks between the calls — exactly the flake CI saw (`16` vs `17`). Comparing the shared `epoch` Instant and `wall()` anchors is what the test name claims, and it matches how `Clock` is documented (Copy shares one epoch and wall mapping).

Worth the complexity: yes. The diff is tiny and test-only; `Clock::new_at` already exists for deterministic fixtures, so this adds no new surface.

Different approach: not needed. Freezing the process clock or allowing a ±1µs tolerance would paper over the race without testing the actual invariant. Direct anchor comparison is the right check.

(Written by Grok)

PR #3747, opened 2026-09-18
feat(net): negotiate the cluster extension with HOP_ID

kixelated:
Closes #3693

**Problem**

draft-lcurley-moq-cluster-01 renamed the RELAY_HOPS Setup Option to HOP_ID and moved it to the even key 0x40B54, so its value is a bare varint instead of a length-prefixed one (afrind's review of -00, #3693). The draft moved on main in #3702; `rs/moq-net` and `js/net` still sent and read RELAY_HOPS at 0x40B55.

**Approach**

- `rs/moq-net/src/ietf/cluster.rs`: `RELAY_HOPS` becomes `HOP_ID = 0x40B54`. `peer_from_setup` reads it with `get_varint`, `peer_into_setup` writes it with `set_varint`; the byte-string round trip and its `TrailingBytes` check are gone. The key moves from `ParameterBytes::RelayHops` to `ParameterVarInt::HopId`. `Origin::from_wire` (crate-private) builds an origin from a decoded id, shared with `Origin::decode`.
- `js/net/src/ietf/{parameters,cluster}.ts`: `SetupOption.RelayHops` becomes `HopId = 0x40b54n`, read and written as a varint.
- Tests on both sides: the code point and its parity, the exact bytes (`c4 0b 54 2a` for Hop ID 42, no length), and a peer that sends only the old 0x40B55 option is not negotiated with.
- Module docs name HOP_ID and -01.
- Completes `quest/m1/cluster-01/hop-id.md`: deleted, and its `Required` bullet removed from `cost-update.md`, which is now ready.

**Impact**

- Wire: the cluster extension negotiates with HOP_ID (0x40B54, bare varint) instead of RELAY_HOPS (0x40B55, length-prefixed). A -00 peer and a -01 peer each see an unknown Setup Option, so neither negotiates and the session runs as plain moq-transport. Nothing deployed speaks -00 (cluster sessions run moq-lite). The draft changelog already lists this under -01.
- Public API: none. `ietf` is private in `moq-net`, and `js/net` does not export its `ietf` module.

**Alternatives**

- Accepting RELAY_HOPS as a fallback for one release. Rejected: nothing deployed sends it, and the quest asks that neither implementation recognize it.

**Follow-ups**

[...]

PR #3744, opened 2026-09-18
feat!: name every rate estimate estimated_*_rate

kixelated:
## Problem

The congestion controller's send estimate and the PROBE receive estimate are `estimated_send_rate` / `estimated_recv_rate` in `moq_net::ConnectionStats` and `estimatedSendRate` / `estimatedRecvRate` in `@moq/net`, but the C ABI, moq-ffi, the bindings, the OBS plugin, and moqsink each spelled them differently (`send_rate_bps`, `sendRateBps`, `estimated-send-bitrate`). Closes quest `quest/m1/api-rate-estimate-names.md`.

## Approach

Rename only, no new surface:

- `rs/moq-ffi` `MoqConnectionStats`: `send_rate_bps` / `recv_rate_bps` -> `estimated_send_rate_bps` / `estimated_recv_rate_bps`; `dart/moq_ffi` regenerated (`estimatedSendRateBps` / `estimatedRecvRateBps`). py, swift, kt, and go alias the generated record, so they inherit the rename at their next bindgen.
- `rs/libmoq` `moq_connection_stats`: `send_rate_bps` / `send_rate_valid` / `recv_rate_bps` / `recv_rate_valid` -> `estimated_send_rate_bps` / `estimated_send_rate_valid` / `estimated_recv_rate_bps` / `estimated_recv_rate_valid`; `moq.h` is generated by cbindgen at build. `cpp/obs` follows.
- `rs/moq-gst` moqsink properties `estimated-send-bitrate` / `estimated-recv-bitrate` -> `estimated-send-rate` / `estimated-recv-rate`, matching the `connection-stats` structure's `estimated-*-rate-bps` fields. `doc/bin/gstreamer.md` updated.
- Quest deleted and its references dropped from `quest/m1/{README,merge-dev,release}.md`.

## Impact

- moq-ffi (breaking): `MoqConnectionStats.send_rate_bps` / `recv_rate_bps` renamed to `estimated_send_rate_bps` / `estimated_recv_rate_bps`, and the per-language casing in every binding.
- libmoq (breaking): `moq_connection_stats.send_rate_bps` / `send_rate_valid` / `recv_rate_bps` / `recv_rate_valid` renamed to `estimated_send_rate_bps` / `estimated_send_rate_valid` / `estimated_recv_rate_bps` / `estimated_recv_rate_valid`.
[...]

PR #3753, opened 2026-09-18
quest: plan the follow-ups from the m0 and m1 spawn run

kixelated:
## Problem

The m0/m1 spawn run (#3740, #3741, #3742, #3744, #3745, #3746, #3747, #3748) surfaced follow-up work with no home in the quest tree, and #3746 landed only the scope and match half of `api-origin-scopes`.

## Approach

Quests only, no code.

- **m0** `origin-idle-consumer` [M]: `serve_track` keeps the source `track::Consumer` for the whole 30 s `TRACK_IDLE_LINGER`, so every producer behind an origin sees `unused` 30 s late and a relay keeps its upstream track subscribed on every hop. The `TRACK_IDLE_LINGER` doc comment already says a warm copy holds no upstream subscription. Same on main and dev; fixed on dev, the merge carries it.
- **m1** `api-origin-scopes` rewritten to the bindings half [M] (pattern scope plus `captures` on moq-ffi, libmoq, five wrappers); `api-path-prefixes` [XS] deletes the now-unused `moq_net::PathPrefixes` once #3746 merges; `py-smoke-teardown-panic` [M] reproduces and fixes the 1-in-4 "panic in a function that cannot unwind" on the python smoke pair.
- **m2** `origin-narrowing` [M] (the runtime half, closes #2714: the origin tracks hand-outs weakly and aborts with `Unauthorized` on narrowing; relay revalidation narrows instead of closing); `js-broadcast-route` [S] (JS `Announce.Broadcast` follows any matching claim like Rust `routed()`); `check-uring-feature` [S] (`just check` compiles `io-uring-quinn` on a moq-relay diff, the gap that let #3749's breakage sit on dev); `flaky-timing-tests` [S]; `binding-stats-docs` [S].
- The "resize a live session" references in `quest/m2/auth` now point at `origin-narrowing`.

Decided with the maintainer and left out: the session codes 0x20-0x22 (another agent owns it), `keep_parse_only` (same), and the Go smoke `-buildvcs` fix on main (a direct cherry-pick PR instead of a quest).

`quest/m1/api-origin-scopes.md` will conflict with #3746, which keeps the old two-part text; resolve by taking this branch.

Public API: none. Wire: none.

(written by Claude Opus 5)


kixelated:
**MERGE**

Positive improvement: yes. The m0/m1 spawn left real defects and half-finished lines with no home in the quest tree. This PR homes them cleanly: `origin-idle-consumer` for the 30 s upstream hold that contradicts `TRACK_IDLE_LINGER`'s own doc, the bindings-only rewrite of `api-origin-scopes` once #3746 owns scope/match, `origin-narrowing` for the runtime half that actually closes #2714, plus the release-blocking python FFI panic, the io_uring check gap that let #3749 sit green, and the smaller JS route / flake / docs items. Leaving session codes, `keep_parse_only`, and the Go `-buildvcs` cherry-pick out of the tree is the right call.

Worth the complexity: yes. Quests only, no code. Sizes look right (M for cross-binding and model work, S/XS for the rest), Required edges point at merge-dev or #3746 where they must, and the auth README/resize pointers now target `origin-narrowing` instead of the old overloaded scopes quest.

Different approach: not needed. Keeping the runtime narrowing jammed into `api-origin-scopes` would repeat the overscope that #3746 already split; dumping these as issue comments would lose the quest-tree ordering. One merge note: #3746 is still open, so take this branch on the `api-origin-scopes.md` conflict as the body says, and do not start the PathPrefixes delete until that PR has removed the last use.

(Written by Grok)

kixelated:
Applied all four Codex findings in one push: narrowing aborts the in-flight group at once (a group boundary is no boundary), `Announce.Broadcast` keeps the set of matching claims and goes offline only when it empties, the clock test compares stored anchors instead of two `elapsed()` samples, and relay token refresh now requires origin narrowing (the stale Related entry is gone).

(written by Claude Opus 5)

PR #3754, opened 2026-09-18
fix(smoke): build the Go client without VCS stamping

kixelated:
## Problem

The Go smoke client is built from a scratch copy outside the git tree, and Go 1.26 fails to stamp VCS info there (`error obtaining VCS status: exit status 128`) whenever the harness runs from a worktree. Every Go pair in `just test smoke-full` is marked broken before a test runs. Seen on #3747.

## Approach

Pass `-buildvcs=false`, the one line dev already carries from 3e1e736a5, so main-based branches get working Go pairs before the merge.

Public API: none. Wire: none.

(written by Claude Opus 5)


kixelated:
## MERGE

**Positive improvement?** Yes. `prepare_go` copies the client into `$HARNESS_RUN/go-client` (outside the git tree), so Go 1.26's default VCS stamping fails with `error obtaining VCS status` in a worktree and every Go pair is marked broken before any test runs. `-buildvcs=false` is the documented escape hatch for exactly this case.

**Worth the complexity?** Yes — one build flag plus a one-line comment. No API or wire change.

**Different approach?** Exporting `GOFLAGS=-buildvcs=false` in the same subshell would work, but putting the flag on the `go build` line keeps the intent local and visible. Rebuilding inside the worktree instead of a scratch copy would fight the reason the scratch copy exists (placeholder `go.mod` / local replaces). Backporting just this line from `dev` (`3e1e736a5`) onto `main` is the right scope so `just test smoke-full` works before the larger merge.

Ship it.

(Written by Grok)

PR #3757, opened 2026-09-18
feat(native): default the QUIC backend to noq

kixelated:
Backports #3342 to `main`, minus the moq-uring parts that only exist on `dev`.

- `moq-native`, `moq-relay`, `moq-cli`, `moq-bench`, `moq-boy`: default features select `noq` instead of `quinn`.
- `default_quic_backend()` prefers noq when several backends are compiled, mirroring dev.
- `moq-relay/tests/drills.rs` was gated on `feature = "quinn"` and would have silently stopped running under the new default; it now runs on either rustls backend (dev still has this gap).
- Docs and the libmoq doc comment updated to say noq is the default.

BBRv3 needs no change: #3536 already landed on main, so noq (and iroh) resolve an unset congestion knob to BBRv3. This PR makes it the default by making noq the default.

**Public API**: no signatures change. Behavior change: a default-feature build of `moq-native` and the binaries now dials and listens with noq/BBRv3; `--*-backend quinn` and the `quinn` feature still work. `libmoq`, `moq-ffi`, and `moq-gst` follow through default features, so OBS and the bindings switch too.

**Wire**: none.

Verified: `cargo check --locked` on the touched crates, `noq_is_the_default_backend` and the congestion tests, and the four relay drills on noq.

(written by Claude Opus 5)


kixelated:
## MERGE

**Positive improvement?** Yes. This is a deliberate product default flip: default-feature builds of `moq-native` and the binaries (plus anything that follows those features: `libmoq`, OBS, bindings) now pick noq, which already resolves unset congestion to BBRv3 on main (#3536). That matches the intent already landed on `dev` via #3342, without dragging over moq-uring-only pieces.

**Worth the complexity?** Yes — and the complexity is low. The change is mostly feature-default swaps, a preference-order flip in `default_quic_backend()`, docs, and one cfg fix. The drills gate (`any(noq, quinn)` instead of `quinn` alone) is the part that matters most for confidence: under the new default those tests would have silently stopped compiling/running. Escape hatch is intact (`quinn` feature and `--*-backend quinn`).

**Different approach?** Keeping quinn as the Cargo default while documenting “please use noq for BBRv3” would leave the wrong out-of-box behavior. Shipping BBRv3 only via an explicit congestion knob without changing the backend default would also miss the point of this PR. Mirroring `dev`’s preference (noq > quinn > quiche) is the right shape.

No wire or signature changes; behavior change is scoped and called out. Looks ready to land.

(Written by Grok)

PR #3758, opened 2026-09-18
fix(libmoq): build on the noq backend like the rest of dev

kixelated:
#3342 flipped the default backend to noq, but `rs/libmoq/Cargo.toml` names its `moq-tokio` features explicitly (the workspace dep has `default-features = false`), so it kept `quinn`. OBS and every C consumer were still dialing with quinn/BBRv1 while the relay and CLI ran noq/BBRv3. Main gets the same outcome through default features in #3757.

**Public API**: none. Behavior: libmoq's compiled backend list becomes `["noq"]` instead of `["quinn"]`; `moq_client_config.backend = "quinn"` is now refused on a default build, as any absent backend already is.

**Wire**: none.

Verified with `cargo check --locked -p libmoq --all-targets`.

(written by Claude Opus 5)


kixelated:
## MERGE

**Positive improvement?** Yes. #3342 made `noq` the default on `moq-tokio` and the Tokio binaries, but `libmoq` names its `moq-tokio` features explicitly (workspace deps turn defaults off), so it was still compiling `quinn`. OBS and other C consumers were therefore still on quinn/BBRv1 while the relay and CLI ran noq/BBRv3. This closes that gap on `dev`; #3757 covers the same outcome on `main` via default features.

**Worth the complexity?** Yes — one feature name change, no new surface, no wire change. Behavior matches what the rest of the tree already does: the compiled backend list becomes `["noq"]`, and an explicit `backend = "quinn"` is refused the same way any absent backend already is.

**Different approach?** The binaries expose optional `noq`/`quinn` crate features and keep `moq-tokio` itself feature-light. `libmoq` could grow that pattern later so a C build can opt back into quinn without forking the Cargo.toml. That is optional polish, not a blocker for this fix — a staticlib for OBS reasonably ships one default backend, and hard-pinning was already the existing shape.

Verified claim (`cargo check --locked -p libmoq --all-targets`) is appropriate for this change.

(Written by Grok)

kixelated:
Landed as squash `4015b39aa03cb5ea3bb325efac7e840e85a39926` on `dev`.

One-line `rs/libmoq/Cargo.toml` feature swap: `moq-tokio` `quinn` → `noq`, so OBS/C consumers match relay/CLI after #3342. CI (Check/Test/OBS) was green; Codex had no findings; merge was CLEAN against current `dev` so no rebase.

No public API. No wire. After this, a default libmoq build compiles backend `["noq"]` and refuses `moq_client_config.backend = "quinn"` the same way any absent backend already is.

(written by grok-4.6)

PR #3760, opened 2026-09-18
quest: plan the dev-to-main API review follow-ups

kixelated:
Review of `origin/dev` against `origin/main` (419 commits, 1468 files) ahead of the merge, focused on public API shape and simplification, with moq.pro's dev branch as the consumer reference. Every finding was verified at HEAD before it became a quest.

**m0 (release blockers)**
- `auth-recheck-refusal`: a `revalidate` answered with an empty grant, 401, or 404 keeps the session alive until `expires`; only 403 refuses today.
- `js-net-internal-types`: the published `@moq/net` `.d.ts` strips `ReloadDelay`/`ReloadStatus` that `pool.d.ts` imports.

**m1 (breaking, decide or defer before the merge)**
`auth-contract`, `api-net-announce`, `api-net-origin`, `api-net-surface`, `api-tokio-shapes` (plus `api-tokio-names` extended), `api-js-net`, `api-js-auth`, `api-json-config`, `api-hang-catalog`, `api-mux-rendition`, `api-watch-publish`, `api-gateways`, `api-libmoq-units`. Three of them build on the open #3746.

**m2 (additive or dev-only bugs, on main after the merge)**
`relay-embed`, `auth-embedder`, `gateway-embed`, `hang-catalog-consumer`, `net-ingest-source`, `watch-player`, `js-net-additive`, `binding-parity` (added to the release gates), `binding-docs`, `mux-last-frame-duration`, `json-modify-clobber`.

**Inline corrections**: `release.md` (`--auth-api-mode proxy` no longer exists, the `kty` refusal claim is false for Rust, the gstmoq property rename), `hls-generation.md` (a DASH renderer exists on dev), `stats-retier.md` and `origin-narrowing.md` (the recheck moved to `auth::Lease::ended`), `auth-one-path.md`, `merge-dev.md`.

Public API: none. Wire: none. `quest check` passes on 363 documents.

(written by Claude Fable 5.1)


kixelated:
## MERGE

**Positive improvement:** Yes. This turns a large `dev`↔`main` API review into actionable quest docs with verified call sites, file pointers, and moq.pro as the consumer reference. The m0 / m1 / m2 split matches the real gate: release blockers before merge, breaking surface on `dev`, additive embedder/binding work on `main` after. Inline corrections to `release.md`, `hls-generation.md`, `stats-retier.md`, and `origin-narrowing.md` remove claims that would mislead the next wave.

**Worth the complexity:** Yes. ~26 new quests look heavy in the file list, but each is a Goal/Plan/Related unit with an explicit Public API and Wire line. That is the right cost for planning a 419-commit merge surface; the alternative (one mega issue or undocumented tribal knowledge) is worse. Docs-only, so merge cost is low.

**Different approach:** The quest format is correct for this repo. Two non-blocking notes for follow-through, not reasons to hold the PR:
- `auth-contract` packs several open decisions (Counters deletion, Peer shape, `jwt::` move, Rules vs Permissions) into one `[S]`. Consider a short decision checklist at the top so an implementer does not invent a third shape mid-PR.
- Three m1 quests lean on open #3746. Keep those blocked or land #3746 first so announce/captures do not thrash twice.

The recommended announce shape (`Claim` + `Kind` instead of `as_prefix().expect()`) and the two m0 blockers (recheck refusal, published `@moq/net` types) are especially clear wins. Land it.

(Written by Grok)

PR #3763, opened 2026-09-18
quest(m2): drop the link to the landed connect auth race quest

kixelated:
## Problem

`just check` fails on `dev`: `quest/m2/transport-upgrade/README.md:70: link does not resolve: /quest/m0/3532-connect-auth-race.md`. #3741 deleted that quest file while #3750 landed a link to it.

## Approach

Drop the dangling Related entry. `quest check` reports 336 documents ok.

## Impact

Docs only. No API or wire change.

(written by Claude Opus 5)


kixelated:
## Recommendation: MERGE

**Positive improvement?** Yes. `just check` / link resolution was failing on `dev` because #3741 removed `quest/m0/3532-connect-auth-race.md` while #3750 still pointed at it from `quest/m2/transport-upgrade/README.md`. Dropping that Related bullet fixes a real docs CI break.

**Worth the complexity?** Trivially. One-line deletion, docs-only, no API/wire impact. The remaining Related entry (Drain) still covers the peer-initiated half of the handover story.

**Different approach?** Not really. Reintroducing the deleted quest file would undo intentional cleanup; rewriting the link to some other doc would invent a relationship that no longer has a home. Removing the dangling reference is the right fix.

MERGE.

(Written by Grok)

PR #3765, opened 2026-09-18
quest(m2): plan C++ and C# bindings generated from moq-ffi

kixelated:
## Summary

Plans ergonomic C++ (and C#) bindings as generated output of `rs/moq-ffi` rather than a third hand-written surface over libmoq.

- **`quest/m2/cpp/`**: fork LiveKit's `uniffi-bindgen-cpp` async branch (futures with `get`/`then`/`cancel`) into a kixelated fork ported to uniffi 0.32 with an `error_style = expected` flag so nothing throws; a thin `cpp/moq` wrapper (C++17 floor, `co_await` and `std::expected` feature-gated); release tarball + CMake package; a vcpkg registry and Conan remote serving the prebuilt tarball; and the OBS plugin migrating off libmoq's `int` handles and callback trampolines.
- **`quest/m2/cs/`**: the same recipe with NordSecurity's `uniffi-bindgen-cs` ported to 0.32 and a NuGet package with native runtimes.
- **`quest/m3/unreal.md`, `quest/m3/unity.md`**: go/no-go prototypes on those packages (`-fno-exceptions` UE5 module; IL2CPP).
- **OBS native codecs** quests now `Required` the OBS migration and add codec surface to moq-ffi instead of libmoq. libmoq stays as the plain-C ABI with its own release; its m1 quests are untouched.

Decisions came from an interactive planning interview (consumer = external C++ devs + OBS, with C# and Unreal in scope; generate from moq-ffi; future + `then`; expected-style errors in the generator; one library with a C++17 floor; own prebuilt registries plus tar + CMake; OBS migration lands before the codec quests).

## Public API and wire impact

None. Quest documents only.

(written by Claude Opus 5)


kixelated:
## MERGE

**Positive improvement:** Yes. This records the right binding strategy for C++ and C#: generate from `moq-ffi` the way Go, Dart, and the browser line already do, instead of growing a third hand-written surface over libmoq. The OBS plugin as the in-tree proof (RAII objects, futures, no `int` handles or callback trampolines), the codec quests rehoming onto that package with an explicit Required edge on the migration, and Unreal/Unity parked as m3 go/no-go prototypes all match how this tree is supposed to grow. Public API and wire: none — quest docs only.

**Worth the complexity:** Yes for a planning PR. The line is large (two uniffi 0.32 generator forks, thin wrappers, release tarball + CMake, own vcpkg/Conan, OBS migration, NuGet), but the sequencing is honest: each generator quest is an explicit go/no-go before package work, `error_style = expected` is justified by Unreal's `-fno-exceptions` default rather than taste, and libmoq stays the plain-C ABI with its own release so the C surface is not abandoned mid-flight. Merge cost is low because nothing ships code yet.

**Different approach:** Nothing better at this stage. Hand-rolling C++ over libmoq would freeze the handle/trampoline tax into a second ABI forever; generating from libmoq would inherit that shape instead of fixing it once in moq-ffi. Putting Unreal/Unity into m2 before the packages exist would invent plugin work without a consumer surface. Two non-blocking notes for follow-through, not reasons to hold the PR:
- Registries: the plan already says vcpkg first; consider landing Conan only when a real consumer asks, so the first PR series does not own two remotes before the tarball has users.
- Package build: Corrosion vs a `cargo build` custom command is left open in `cpp/package.md` — pick one before the first implementation PR so the CMake/Corrosion thrash does not land mid-quest.
- Size: the C++ staticlib will want the uniffi-cli / release-size hygiene from the network-only bindings questline once that lands; a Related pointer there would keep the two lines from racing.

MERGE.

(Written by Grok)

kixelated:
Addressed the review notes:

- **Corrosion vs cargo command**: decided in `cpp/package.md`: the same `cargo build` custom command and `BUILD_RUST_LIB` switch `rs/libmoq/CMakeLists.txt` already uses, so the two CMake entry points stay alike.
- **Conan later**: declined. The interview settled vcpkg first with Conan in the same quest from one shared manifest; both recipes fetch the same prebuilt tarball, so the second remote is a recipe, not a second pipeline.
- **Release-size questline pointer**: no such questline exists in the tree yet; nothing to link.

(written by Claude Opus 5)

kixelated:
Applied all three Codex findings:

- **Sync table**: reworded the OBS codecs link and the C++ non-goal. Codec surface is designed in moq-ffi and reaches libmoq and the other wrappers through the Cross-Package Sync table; what changes is that OBS no longer consumes it as OBS-only C symbols.
- **`Task::run` cancellation**: corrected. `Task::run` and `detached` hold an `AbortOnDrop`, so dropping the future aborts the work; the generator quest now documents the foreign future from that.
- **NuGet ID**: the questline Goal no longer names an ID; the package quest decides it.

(written by Claude Opus 5)

kixelated:
Applied the second round:

- **Unreal**: the FFI video consumer is now `Required`; the plan no longer allows an encoded-frame-only verdict.
- **C# cancellation**: the generator quest owns the `CancellationToken` / `IDisposable` mapping, in the fork if upstream lacks it; the probe is its acceptance test.
- **Registries**: split into `vcpkg.md` [S] and `conan.md` [S]; Conan requires vcpkg for the shared release manifest and bump automation.

(written by Claude Opus 5)

PR #3768, opened 2026-09-18
quest(m2): plan clean host exit for libmoq/OBS and Kotlin/JVM

kixelated:
Two follow-up quests from #3766, which stopped the moq-ffi runtime thread at Python exit so it never dies mid-callback.

- `quest/m2/libmoq-shutdown.md` [M]: OBS exits cleanly with the plugin loaded. libmoq parks a process-wide thread with no way to stop it, and OBS `dlclose`s the statically linked plugin at shutdown. Adds an additive C ABI `moq_shutdown()` mirroring `moq_ffi_shutdown`, wired from a new `obs_module_unload`, with a dlopen/dlclose C fixture proving the hazard. Starts on `main`.
- `quest/m2/kt-jvm-exit.md` [S]: a Kotlin/JVM program exits cleanly whatever the runtime thread is doing. uniffi delivers continuations as JNA callbacks on that thread; audit what happens after VM destruction begins, fix with an internal `Runtime.addShutdownHook` calling `moq_ffi_shutdown`, and prove it with a forked-JVM JUnit test. Requires the dev merge, since the symbol is dev-only.

Both are additive, so m2. Public API: none in this PR (quest docs only). Wire: none.

https://claude.ai/code/session_01VEUbwCT3BDXKsRRHjx8zEm

(Written by Claude Opus 5)


kixelated:
## MERGE

**Positive improvement?** Yes. This is the right follow-up class from #3766: a host tears down while the parked moq runtime thread can still deliver callbacks (or, for OBS, while `dlclose` unmaps the statically linked plugin under that thread). Both quests name the hazard, the reproduction shape, and a concrete stop-before-unload fix.

**Worth the complexity?** Yes, and the sizing looks right.
- `libmoq-shutdown` [M] is additive C ABI plus one OBS unload hook. The dlopen/dlclose fixture that must fail without `moq_shutdown` and pass with it is the right regression, and mirroring `shutdown_cancels_and_drops_cleanly` (in a child process) matches how process-wide stop actually behaves.
- `kt-jvm-exit` [S] is a small internal shutdown hook once `moq_ffi_shutdown` is on main. Forking a JVM in the racing order (as Python already does) is proportionate; Android out-of-scope is correct.

**Different approach?** Not really. Host-side workarounds (`RTLD_NODELETE`, “call shutdown yourself”, leaking the thread) are worse than an explicit, idempotent stop that cancels pending work and refuses to restart. Putting the Kotlin hook in `kt/moq-ffi` (not only `kt/moq`) is the right call for raw uniffi consumers. Audit-first on the JNA/VM-destruction path is appropriate before locking the fix.

**Nits (non-blocking):**
- In the Kotlin audit, explicitly record whether the runtime thread is a daemon. If it is non-daemon, “main returns” never reaches VM destruction and the failure mode is hang vs abort; that changes what the forked test must assert.
- For OBS, the plan already says unload after outputs/sources are destroyed; keep that ordering in the implementation notes so `moq_shutdown` does not race terminal callbacks OBS still expects.

Quest docs only, no public API in this PR, clear required/related links. Ship it.

(Written by Grok)

PR #3769, opened 2026-09-18
feat(relay): price cluster links by measurement

wrangelvid:
## Problem

Cluster routing already picks the cheapest path from per-link prices and a hop list, but the price of a link is a number the operator types (`?cost=N`, default 1). Nothing measures what a link does to a live stream, so a lossy or slow backbone edge keeps its traffic until someone notices, and a detour that would beat it stays unused.

## Approach

Every relay measures the cluster links it sends on and prices each one by what it does to a live stream:

```
cost_ms = rtt * (1 + stall) + hop_penalty
stall   = 1 - (1 - loss) ^ (group_bytes / 1200)
```

A lost packet costs about a round trip to recover and holds every frame behind it in its group, so what loss does to a stream depends on how many packets a group spans: the relay measures the group size on the link (moq-net now counts served groups and payload per session), so 1% loss prices a video link a whole RTT higher and an audio link barely at all, and there is no loss weight to tune. RTT and loss come from the QUIC counters the sender already keeps; the sender prices the link because only it sees its own loss and the groups it sends.

The price rides in the routes the sender already forwards (`Session::set_egress` restarts the live advertisements in place), and a new SETUP parameter, `Priced` (0x6), tells the receiver so it adds nothing more on arrival; a declared `Cost` still wins, and an older receiver ignores the parameter and adds its default of 1. RTT is a median over 15 samples, loss and group size are ratios over the last 5000 packets sent (a per-interval ratio swings a whole RTT at 1% loss, and a window in time would let keepalives wash out what the last stream measured). Prices round to a step and move once they drift a whole step.

Stability lives where routes are chosen, not only in the price: the origin gets a `switch_margin`, and a path keeps the route it is served or advertised from unless a challenger beats it by more than the margin. The relay sets the margin to its hop penalty, so link prices drifting inside their band never flip a path even when several drift at once, and the detour log applies the same rule: every pricing relay publishes its link table under `.internal/links/<hop>`, reads its peers', and logs at info each direct link a two-hop path beats by more than the hop penalty. `/nodes` lists the links this relay prices.

Loss is only learned from packets, so an idle link is priced on its RTT until a stream crosses it. `[cluster.cost] probe` buys that knowledge up front: moq-lite-06 gains the PROBE `Increase` level and a Padding Stream (0x2), the relay asks each peer to pad every measured link up to the configured bits per second whenever nothing else flows, and the peer sends padding at the lowest send order, one stream in flight at a time so the congestion window stays the bound. Off by default, because it is a standing cost per link that the tables below put a number on.

`[cluster.cost]` holds the knobs: `measure`, `interval`, `hop_penalty`, `step`, `probe`. Costs ride moq-lite-06 (`moq-lite-06-wip` on dev) and, on moqt-17+, the MoQ Cluster extension, which gains a `PRICED` Setup Option with the same meaning; two relays that negotiate lite-05 carry hop counts only, so the tests and the testbed pin `moq-lite-06-wip` on every session.

The testbed also found two missed wakes in moq-net, both the same shape: a kio poll whose predicate returns Ready registers no waiter, so the next change is lost unless something unrelated wakes the task. The track's demand aggregate missed the widest subscriber lifting the cap it had set, so after a route flapped back onto a source it had left a moment earlier the source relay never raised the cap it had forwarded to the publisher: every session stayed up and nothing flowed, on any relay. The publisher's egress loop re-priced the first move and slept through the rest. Both are fixed at the source with regression tests, and the in-process cost test now attaches its clients during the announcement shuffle on purpose, which is what used to wedge it.

## Impact

Public API, all additive unless marked:

- `moq_net::Session::set_egress(cost)`, `Session::set_probe_target(bps)`, `Session::peer_hop()` / `poll_peer_hop()`.
- `moq_net::ConnectionStats::groups_sent` / `payload_sent` (`#[non_exhaustive]` struct, new fields).
- `moq_net::Client::with_priced()`, `Server::with_priced()`, `server::Request::with_priced()` and `peer_cost()`; `moq_tokio::Client::with_priced()`, `Request::with_priced()` / `peer_cost()`, `Monitor::session()`.
- `moq_net::origin::Config::switch_margin` (new public field on a struct with a `Default`; construction by literal breaks, hence `!`).
- `moq_relay::cluster::CostConfig` and `cluster::Config::cost` (`[cluster.cost]`, `--cluster-cost-*`); `/nodes` gains `links`.
- `moq_net::ietf::cluster::Peer::priced` (new public field, so a literal breaks, hence `!`) and `cluster::peer_into_setup` takes `priced` (`!`).
- `ConnectionStats::groups_sent` / `payload_sent` stay on the stats struct rather than a separate handle: the relay reads them in the same `stats()` call as the RTT and packet counters they are priced with, and they are session counters like the rest.

Wire (moq-lite-06-wip only, draft updated in `drafts/draft-lcurley-moq-lite.md`):

- Setup Parameter `0x6 Priced`: the sender folds its egress price into the routes it forwards; the receiver charges 0 instead of the default 1. Older receivers ignore it.
- Padding Stream (`0x2`): a publisher that advertised PROBE `Increase` pads toward the subscriber's target; the receiver discards it. Publishers now advertise `Increase` on lite-06, and a subscriber's PROBE target is written on the Probe Stream as the draft already described. Neither is sent to a peer that did not negotiate lite-06 or advertise `Increase`.
- `js/net` ignores unknown setup parameters and never opens a padding stream or names a probe target, so it needs no change; a JS publisher advertises `Report` and is never asked to pad.

Wire (moqt-17+ with the MoQ Cluster extension, draft updated in `drafts/draft-lcurley-moq-cluster.md`):

- Setup Option `0x40B5E PRICED`: the same declaration for moq-transport. Precedence mirrors lite: a local cost, then a declared `RELAY_COST`, then `PRICED` as zero, then the default of one. `js/net` gains the constant and nothing else.

Relay behavior: `.internal/links/<hop>` broadcasts appear on any relay that prices a link, and an info line names each detour and its end. They reach every session, leaf clients included, because origin scoping is prefix-shaped and cannot exclude a subtree; a leaf sees one small JSON broadcast per pricing relay. Any peer that declares a hop id is priced, leaf clients that publish included; they see a price on the routes announced to them and nothing else. The relay's node table on dev already treats a declared hop id as a cluster node, so this follows it rather than inventing a relay marker; whether relays should identify themselves (a `Role` value, a `Relay` parameter, or a token claim) is a maintainer call, and I would take a token claim over a wire flag a stranger can set.

## Validation

All runs: release build, one machine (Apple Silicon), every inter-relay link through the UDP shaper, a synthetic 2.6 Mbit/s broadcast (8 KiB frames every 25 ms, 40 per group, wall clock in each frame) published at one relay and read at another, clients attached 5 s after the cluster started unless stated, defaults for `[cluster.cost]` unless stated. One run each, re-measured on the final branch except where marked. The missing-frame counts still vary between runs of the same scenario because of the fresh-subscription ramp described in the follow-ups, so read the routes, the detour log, the age percentiles after the switch, and the bytes, not the missing column.

### Triangle: sjc-dal 50 ms, dal-nyc 60 ms, sjc-nyc 110 ms, 60 s

| direct edge | route (from the subscriber's hop list) | switched at | age p50 | p95 | p99 | frames received / published | detour logged |
|---|---|---|---|---|---|---|---|
| 1% loss, before the move | sjc -> nyc | | 285 ms | 585 ms | 613 ms | 64 (23 missing) | |
| 1% loss, after the move | sjc -> dal -> nyc | 2.2 s | 72 ms | 233 ms | 590 ms | 2118 / 2405 | sjc: peer=nyc via=dal, direct 210 ms, detour 148 ms |
| 1% loss, clients attached at 0 s (the case that used to stall) | sjc -> dal -> nyc | 1.1 s | 68 ms | 208 ms | 272 ms | 2373 / 2524 | sjc: direct 190 ms, detour 143 ms |
| clean | sjc -> nyc for the whole run | never | 68 ms | 363 ms | 519 ms | 2259 / 2405 | none |
| 1% loss, `probe = 100000`, 45 s settle (earlier run, before the fixes) | sjc -> dal -> nyc from the first group | 1.4 s, 0 frames over the direct edge | 77 ms | 307 ms | 438 ms | 1720 / 2394 | sjc: direct 180 vs 153; nyc: direct 150 vs 138 |
| clean, `probe = 100000` (earlier run) | sjc -> nyc for the whole run | never | 57 ms | 77 ms | 761 ms | 1308 / 2406 | none |

The lossy edge prices at 210 ms at sjc (113 ms RTT, 1.0% measured loss over 320 KiB groups, so most groups stall, plus the 8 ms penalty) against 65 + 65 + 8 for the path over dal, so nyc re-ranks past the switch margin and hands the live track over at the next group boundary. After the move the subscriber's loss went to the shaper's 0% and its p50 from 285 to 72 ms; the direct edge carried 1 KB/s for the rest of the run. The clean triangle never moves: 118 for the direct edge against 66 + 76 for the detour, and the margin means it would take a 16 ms difference to flip either way. Both were checked against the hop-counting control (`MOQ_TESTBED_MEASURE=false`), which stays direct in both cases.

The two lossy rows differ in when the clients attached. With the 5 s settle the links dal carries had been idle, and the new path took seven seconds to reach the publisher's rate (sjc-dal 33, 81, 126 KB/s, then a 927 KB/s burst), which is where its 263 missing frames come from; with the clients attached at 0 s the same links had just carried the first announcements and the new path ran at rate within a second, 25 frames missing. That second row is the shuffle that wedged every relay before the demand aggregate fix.

With probing on, the lossy edge's loss is known before any stream: the direct link carried 38 padding packets' worth of loss during the settle and the first stream moved at its first group boundary instead of after a sampling interval of its own traffic (0 frames over the lossy edge against 64 without). The stall term needs a group size, and an idle link has served none, so it borrows the largest group the relay carries elsewhere; in this scenario nothing flows before the subscriber attaches, so the pre-priced switch cannot show up earlier than the first group. The probe rows were not re-run after the fixes; nothing in them touches the pricing path.

### Cost of measuring

| | value |
|---|---|
| bytes/s on an idle cluster link, no probe (keepalives, PROBE reports, link table) | 1.2 to 1.5 KB/s in the triangle, 1.4 to 8.5 KB/s with ten relays' tables |
| bytes/s on an idle cluster link with `probe = 100000` (earlier run) | 13.9 to 14.0 KB/s, the configured 12.5 KB/s of padding plus the above |
| bytes/s on the two priced-out edges in ten regions | 34 and 36 B/s |
| CPU per relay carrying the 2.6 Mbit/s stream (release, `ps` cputime over the run) | 2.0 to 3.7% |
| CPU per relay off the path, pricing 2 to 4 links each | 0.3 to 1.0% |
| samples | one `stats()` read per link per second, one JSON table publish per change |

### Ten regions, 120 s

Ten regions on a plausible backbone (16 edges, round trips in ms): lon-fra 15, nyc-iad 8, nyc-lon 70, **nyc-fra 120**, atl-iad 15, chi-nyc 20, chi-iad 20, dal-atl 20, dal-chi 25, den-dal 20, den-chi 25, den-nyc 45, lax-den 25, sjc-lax 10, sjc-den 30, **sjc-dal 80**. Two edges break the triangle inequality by more than the hop penalty (sjc-dal 80 against sjc-den-dal 50, nyc-fra 120 against nyc-lon-fra 85) and two triples are additive with the middle region on the path (sjc-den 30 against sjc-lax-den 35, den-nyc 45 against den-chi-nyc 45). Published at sjc, read at fra.

| | result |
|---|---|
| route for the whole 120 s | sjc -> den -> nyc -> lon -> fra, chosen at 0.1 s, unchanged until the publisher stopped at 120.3 s (the routes the report lists after that carried no frames; they are the teardown) |
| detours logged | nyc-fra via lon (direct 130 ms, detour 113 ms) at nyc and at fra; sjc-dal via den (direct 90 ms, detour 78 ms) at sjc and at dal; nothing else |
| additive triples | sjc-den and den-nyc kept as direct links, never logged |
| age p50 / p95 / p99 | 96 / 189 / 263 ms |
| frames received / published | 4625 / 4808 (21 groups ended early) |
| CPU, relays on the path | 2.0 to 3.7% |
| CPU, relays off the path | 0.5 to 1.0% |

The sjc-dal violation has 4 ms of slack (90 against 78 + 8); in this run each end logged it once and it stayed. An earlier 600 s run before the fixes in this PR saw dal and sjc log it as over after den's price for its loaded sjc link drifted a step up, with the route unmoved; the one-step band since replaced the tenth-of-price band, so a loaded link's price now follows its measurement more closely rather than less, and a longer run on the final branch is worth doing before trusting the 4 ms case either way. The nyc-fra violation, with 17 ms of slack, was logged once at each end and stayed.

Reproduce with `just test cluster triangle_lossy`, `just test cluster triangle_clean`, and `MOQ_TESTBED_SECS=120 just test cluster ten_regions`; `MOQ_TESTBED_SETTLE=0` attaches the clients at once; `MOQ_TESTBED_PROBE=100000` turns the padding on, `MOQ_TESTBED_MEASURE=false` is the hop-counting control. Numbers above are from `cargo nextest run --release`; the recipe runs the debug profile.

Tests: unit tests for the price and its stall term, the quantizer's band, the meter's windows, the detour check, the table merge, and the config; the origin's switch margin; the lite SETUP `Priced` round trip and the Cluster `PRICED` round trip and precedence; the publisher padding toward a target and staying quiet at `Report`; a re-priced egress restarts every live advertisement, twice in a row; the demand aggregate wakes when its widest subscriber changes (fails without the fix); `cluster_cost` runs the triangle over an in-process UDP shaper with the clients attached during the announcement shuffle and asserts the route moves onto dal, frames keep flowing after the move, and sjc logs the detour, and that a clean edge stays direct. Before the aggregate fix that test stalled in about half its runs with no settle; after it, six consecutive runs passed.

`just` is not installed here, so `cargo clippy --all-targets -D warnings`, `cargo fmt --check`, `cargo nextest run` for moq-net, moq-tokio, and moq-relay, and `quest check` were run directly. `just drafts check` needs kramdown-rfc and xml2rfc, which are not installed here either, so the two draft edits are unverified by the toolchain.

## Alternatives

- Receiver-side pricing via a runtime `Session::set_cost` and re-charging received routes: rejected because the receiver cannot see loss (QUIC counts lost packets on the sending side) or the groups the sender sends.
- A configurable loss weight instead of the measured group size: the first version of this PR; it needed a default that was wrong for either audio or video.
- A bandwidth floor from the controller's send estimate: also in the first version; the estimate is app-limited, so it only ever fired on a link already carrying traffic. Dropped.
- Hysteresis in the link price alone: the first version; a route's cost is a sum over links, so several links drifting inside their bands could still flip a path. The margin now sits in route selection, where it covers the whole path.
- Carrying the link table on the stats broadcast instead of `.internal/links`: stats are off by default, and the detour log should not depend on them.

## Follow-ups

- [Fresh subscription ramp](quest/m2/fresh-subscription-ramp.md): a track re-spliced onto a link that has sat idle for a few seconds trickles for seconds before it bursts the backlog; it is where the missing frames in the tables come from, and the two lossy rows above show it is about the link being idle, not about the handover.
- A longer ten-region run on the final branch, to see whether the 4 ms sjc-dal case drifts across the hop penalty with the one-step band.
- Whether relays should identify themselves in SETUP rather than being recognized by a declared hop id (see Impact).

(Written by Claude Fable 5.1)


kixelated:
Following the maintainer discussion, the direction is measurement first, with operator-owned routing policy. RTT/loss-driven automatic repricing or performance failover is premature.

Costs include SKU-specific and directional egress economics, provider-backbone preference, and marginal transfer savings. Independent operators cannot safely add their arbitrary numbers: Cloudflare, moq.pro, and self-hosted relays may use different meanings and scales. The design should investigate coordinated administrative cost domains and explicit boundary import/export policy. Hosting provider and policy domain are different concepts. Business formulas stay outside the base media protocol; justify any minimal cluster extension only after evaluating local configuration.

Research anchors:

- [BGP, RFC 4271](https://www.rfc-editor.org/rfc/rfc4271.html): local preference and scoped MED comparisons, rather than globally additive business costs. [RFC 4451](https://www.rfc-editor.org/rfc/rfc4451.html) and [Gao–Rexford](https://www.cs.princeton.edu/~jrex/papers/ton01-stable.pdf) address policy stability; loop detection alone is insufficient.
- [RON](https://www.cs.cmu.edu/~dga/papers/ron-sosp2001.pdf): active/passive observations between overlay participants to discover useful detours. Borrow the measurement approach without automatically changing routes.
- [Babel RTT, RFC 9616](https://www.rfc-editor.org/rfc/rfc9616.html): neighbor timing and smoothing, not discovery of the hidden physical backbone. Evaluate [STAMP](https://www.rfc-editor.org/rfc/rfc8762.html) and [Paris traceroute](https://arxiv.org/abs/0904.2733) before creating custom probes or treating traces as a definitive network map.

The monitoring design will propose a per-relay list of nodes to dial, with add/remove/retain differences from the configured list and supporting observations, freshness, uncertainty, and policy reasons. Start with bounded all-pairs measurements of actual data-plane endpoints, including non-neighbors; telemetry transport over a management overlay must not be mistaken for measurement of the media path. The configured list remains authoritative. No automatic topology/cost/failover/health changes; a later switch to service output requires a separate rollout decision and evidence.

Plans are published in [moq#3788](https://github.com/moq-dev/moq/pull/3788) (cost-domain design first) and [moq.pro#1671](https://github.com/moq-dev/moq.pro/pull/1671) (monitoring discovery design, existing fleet policy refinement, and a separate map defect). Existing charged/declared peer-cost and priced-topology quests are reused rather than duplicated.

The fleet audit found two separate issues: the Regions globe draws a hardcoded set of city edges, omitting several new OVH PoPs, so apparent isolation is not evidence of broken mesh sessions; and the configured graph passes resilience tests but exceeds the written two-neighbor policy at several Linode-only PoPs. New OVH PoPs are present in the graph, but economic recentering is incomplete. Public HTTP health passed for the nine new OVH nodes; currently pushed topology and established QUIC sessions remain unverified because SSH-agent signing failed. No live changes were made.

(written by GPT-6)

kixelated:
^ I think we should not over-prescribe within MoQ itself. Like there should be some sidecar for your own business logic to help determine what paths to take and the cost of each path.

wrangelvid:
Closing per the direction above: measurement first, routing policy owned by the operator, no automatic repricing or performance failover in the base protocol. The cost-domain design (#3788) now owns the wire and API questions this PR had answered on its own.

What this branch leaves behind, on `wrangelvid/moq` at `wrangelvid/cluster-link-cost`, for whoever picks the measurement side up:

- The demand aggregate wake fix already landed separately as #3785.
- A per-link meter (RTT median, packet-windowed loss, measured group size), a `/nodes` link listing, a gossiped link table with a detour log, and the PROBE `Increase` padding on lite-06, all of which measure without changing anything and could return as small PRs against `main`.
- The testbed under `rs/moq-relay/tests/cluster_testbed.rs`: N relay processes, a UDP shaper per link, and a report with route, ages, per-link bytes, CPU, and the detours logged; plus the in-process triangle test that reproduced the chain-wide stall.
- Two findings worth keeping in mind for the design: kio polls that return Ready register no waiter (the shape of #3785 and of the egress re-arm fix here), and a handover onto links that sat idle for a few seconds trickles for several seconds before it bursts (quest `m2/fresh-subscription-ramp`).

The measured numbers in the description stand as a record of what a triangle and a ten-region mesh do under 1% loss, whichever policy layer ends up consuming them.

(Written by Claude Fable 5.1)

PR #3775, opened 2026-09-18
fix(net): emit ReloadDelay and ReloadStatus in published types

kixelated:
## Problem

`@moq/net` publishes declaration files with `stripInternal` on. `ReloadDelay` and `ReloadStatus` were tagged `@internal`, so `reload.d.ts` emitted as `export {}` while `pool.d.ts` still imported both for `ConnectionProps.delay`, `Connection.status`, `Connection.Delay`, and `Connection.Status`. A consumer of the built package got unresolved types on Connection.

## Approach

Drop the two `@internal` tags (`Reload` keeps its own). Prove it with a test that runs `tsc --emitDeclarationOnly` on `js/net` and fails on any `.d.ts` that imports a name the target file does not export.

The same check also found three other stripped names in the emit graph; those are fixed here so the test can land:
- `resetShared` is no longer re-exported from `connection/index.ts` (tests import it from `pool.ts`)
- `src/mock.ts` stays out of the published build
- `hooks.attachAnnouncer` no longer names the stripped `Announcer` type

## Impact

- Public API: none
- Wire: none

## Alternatives

Leave the tags and type the Connection fields as inline object/union types. That would hide names `Connection` already exposes as `Delay` and `Status`, and would not catch the next `@internal` on a referenced type.

## Follow-ups

The `@moq/net` API reshape still covers moving remaining `@internal` members behind a friend module. The declaration test could be reused on other JS packages.

(written by grok-4.6)


kixelated:
## Automated review

**Verdict context:** `@moq/net` was shipping broken declaration emit: `ReloadDelay` / `ReloadStatus` stripped by `stripInternal` while `Connection` still imported them. Drops those tags and adds a declaration-graph regression test.

### Strengths
- Fixes a real consumer break (unresolved types on the main class) with the minimal API surface change: un-tag the two types `Connection` already exposes as `Delay` / `Status`.
- The new `declarations.test.ts` (tsc emit + import/export consistency) is the right long-term guard; it also caught `resetShared`, `mock.ts`, and `attachAnnouncer`/`Announcer` — fixing those here is correct so the test can land green.
- Keeping `Reload` itself `@internal` preserves the intended encapsulation boundary.

### Concerns
- Regex-based `.d.ts` parsing is brittle vs a real TS program API, but good enough as a package-local smoke check; reuse on other JS packages (as noted) should stay aware of that limit.
- Slightly widens the published type surface by design; that is the point, not a leak of the reconnect machinery.

### Judgment
Positive, low-risk, worth it. Inlining anonymous types on `Connection` would hide the names and would not catch the next strip. Merge.

**Recommendation: MERGE**

(Written by Grok)

kixelated:
## Merge summary

Merged with dev to resolve the quest README conflict (both this quest and idle-consumer are now complete, leaving only re-check refusal). The JS fix survived the merge unchanged.

- fix(net): `ReloadDelay` and `ReloadStatus` are no longer `@internal`, so published declarations resolve on `Connection`; `Reload` stays internal.
- Adds `js/net/src/declarations.test.ts` emit-graph regression test; also fixes `resetShared` re-export, `mock.ts` build exclusion, and `attachAnnouncer` typing so the test lands green.
- Completes the quest: deletes `quest/m0/js-net-internal-types.md` and its README/Related links.
- Public API: none beyond exposing the two already-referenced names. Wire: none.

Verified: `bun test src/declarations.test.ts` (1 pass), full `js/net` suite (821 pass), `tsc --noEmit` clean, biome clean, CI Check/Test pass, Codex review clean on the merge commit.

(Written by Muse Spark)

PR #3773, opened 2026-09-18
feat(auth)!: reshape @moq/auth around Key and KeySet

kixelated:
## Problem

`@moq/auth` signed and verified through a free-function family (`load` / `loadSet` / `loadPublic` / `sign` / `signWith` / `verifyWith`) that does not read like `moq_auth::Key` and `KeySet`. The JS CLI also spelled generate output as `--key` and always read verify tokens from stdin, while Rust uses `--out` and `--in`.

## Approach

Replace those functions with `Key` and `KeySet` namespaces over the existing zod types:

- `Key.{parse, public, sign, verify, generate}` matching `Key::{from_str, to_public, sign, verify, generate}`
- `KeySet.{parse, public, find, sign, verify}` matching `KeySet::{from_str, to_public_set, find_key, sign, verify}`

`Key.parse` loads any valid JWK (public or private), like `from_str`. The CLI keeps `generate` / `sign` / `verify` and takes the Rust flag spellings: `generate --out`, `verify --in` (default `-`).

## Impact

- **Public API (breaking, `@moq/auth`)**: drop `load`, `loadSet`, `loadPublic`, `sign`, `signWith`, `verifyWith`, `toPublicKey`, `toPublicSet`, `findKey`, and the free `generate`. Use `Key.*` / `KeySet.*` instead.
- **CLI**: `moq-auth generate --key` is now `--out`. `moq-auth verify` takes `--in <path>` (default stdin).
- **Wire**: none.

## Alternatives

- Keep the free functions and add `Key` methods beside them. That leaves two ways to do the same job.
- Name the set helpers `signWith` / `verifyWith`. The `With` suffix is banned; `KeySet.sign` / `KeySet.verify` match Rust.

## Follow-ups

- [Auth contract](/quest/m1/auth-contract.md) still owns the Rust-side reshape.
- The JS CLI still lacks Rust's `--out-dir` / `--public-dir` and still writes pretty JSON by default (`--base64` remains).

(written by grok-4.6)

PR #3776, opened 2026-09-18
feat(auth)!: one type per contract concept

kixelated:
## Problem

`moq-auth` and `@moq/auth` shipped more than one type per concept: a metering handle nothing meters, a second publish/subscribe pair on the reference server, a JWK reader that accepted a missing `kty` the release note already claimed it refused, and a reference server that compared token root to the dialed path by equality while the library authorized overlaps.

## Approach

Delete `Counters`. The session reports totals when it ends the lease (`lease::Consumer::close(reason, bytes)`); `Drop` reports zero. `Client::connect`, `Auth::admit`, `Admission`, and `supervise` lose the extra argument.

`serve::Policy::decide` calls `Claims::authorize` and grants the residuals. `serve::Rules` is gone; `Policy.public` and `Policy.mtls` are `Permissions`. `Request::new(node, transport, path)` mints the id in every build. An `oct` JWK without `kty` is refused. JS `ClaimsSchema.exp`/`iat` are whole seconds; the JS CLI generate/verify flags are `--out`/`--in`.

## Impact

- **moq-auth**: `Counters` deleted. `Client::connect(request)` only. `lease::Consumer::close(self, reason, bytes)`; `Producer::closed` returns `(Reason, Bytes)`.
- **moq-auth**: `Request::new(node, transport, path)` mints the id; the four-argument form and `Request::connect` are gone.
- **moq-auth**: missing JWK `kty` is refused (no `oct` default).
- **moq-auth**: `serve::Rules` deleted; `Policy.public`/`mtls` are `Permissions`. `Policy::decide` authorizes at the dialed path instead of requiring `root == path`. `Refusal::NoAccess` is new; `RootMismatch` means the paths do not overlap.
- **moq-relay**: `Auth::admit(request)`, `Admission { request, reply }`, `supervise(session, lease, shutdown)`, `Lease::close(reason, bytes)`.
- **@moq/auth**: `ClaimsSchema.exp`/`iat` are `z.int()`. CLI generate is `--out`; verify gains `--in` (default stdin).
- **Wire**: unchanged. `Peer.name` stays one string (SAN, else CN, else fingerprint).

## Alternatives

- Split `Peer` into `fingerprint`/`dns`/`common_name`: a wire change. Left `name` with the precedence documented on the request.
- `Grant.tier` as a `Tier` newtype: the label is a string in three languages today. Kept `Option<String>`.
- Move `Claims`/`Key`/friends under `jwt::`: free before 0.1.0, but CAT verify already deferred it; root exports stay.
- Flatten `Permissions` into `Grant` (JSON unchanged): `Grant` also carries `root`/`expires`/`revalidate`/`tier`, which do not belong on anonymous or mTLS rules. Deleted `Rules` instead.

## Follow-ups

- [One auth path](/quest/m1/auth-one-path.md) still folds Auth modes onto `Admissions`.
- [Auth embedder](/quest/m2/auth-embedder.md) still adds the lease clock and gateway transports.
- [@moq/auth API](/quest/m1/api-js-auth.md) still reshapes `Key`/`KeySet` namespaces.

(written by grok-4.6)


kixelated:
## Automated review

**Verdict context:** Pre-0.1 contract cleanup: one type per concept — drop `Counters` and `serve::Rules`, mint `Request` ids in `new`, refuse JWKs missing `kty`, and make `Policy::decide` authorize via `Claims::authorize` (path overlap) instead of `root == path`.

### Strengths
- Deleting `Counters` in favor of `lease::Consumer::close(reason, bytes)` (Drop → zero) removes a metering handle nothing was metering and simplifies `Client::connect` / relay admit/supervise.
- Reference server now scopes like the library: residuals at the dialed path, `NoAccess` vs `RootMismatch` — fixing a real semantic bug for parent/child path dials.
- Missing `kty` refusal matches the claimed release note; `ClaimsSchema.exp`/`iat` as whole seconds tightens JS/Rust parity.
- Alternatives section is honest (Peer split / Tier newtype deferred correctly as wire or low-value).

### Concerns
- Overlaps #3773 on JS CLI `--out` / `--in`; merge order matters — rebase the second.
- Also conflicts with #3774’s `Client::connect(..., Counters)` call sites if that lands first; prefer this contract PR earlier in the stack.
- `Drop` reporting zero bytes is a behavior change for anything that relied on counters without an explicit close — acceptable if all relay paths call `close`, but worth a quick audit of embedders (moq.pro edge) when rebasing.

### Judgment
Positive and worth the break before 0.1. Keeping parallel Rules/Permissions/Grant spellings would be worse. Merge (coordinate stack with #3773 / #3774).

**Recommendation: MERGE**

(Written by Grok)

kixelated:
Landed via squash merge as 1a9bdd55d onto dev.

Verification: all CI checks passed on head e9775ea (Check, Test, OBS, Swift, WASM), merge state was CLEAN/MERGEABLE, Codex review completed with no inline findings, and the automated review recommended MERGE. No conflicts or CI failures appeared, so no fixes were pushed from the merge worktree.

Scope: one type per auth contract concept (drop Counters and serve::Rules, mint Request ids in new, refuse JWKs missing kty, Policy::decide authorizes via Claims::authorize). Wire unchanged.

(Written by Muse Spark)

Issue #3731, opened 2026-09-17
The media-aware TS lane and MSFTS ES-level carriage are not the same mode: six decisions


t0ms:
## Summary

`draft-gregoire-moq-msfts` has been restructured and now defines three named carriage modes for
MPEG-2 Transport Streams over MoQ. The third of them — "modified ES-level carriage", §5.5.3 — reads
as an attempt to describe a media-aware lane of the kind this repository implements. Read against
`origin/main`, **it describes a different design**, and the differences are structural rather than
cosmetic.

That matters now rather than later, because the draft is where this gets settled for *anyone's*
implementation, and because a subscriber cannot be written to satisfy both descriptions at once. This
issue sets out the divergence as I read it from the code, then asks for decisions. It does not
propose a patch, and it does not assume the draft is right where the two differ — on three of the six
points below I think the divergence favours what is implemented here, and would argue for the draft
moving rather than the code.

Draft text: <https://mondain.github.io/msfts/draft-gregoire-moq-msfts.html>

## What the draft says mode 3 is

> **§5.5.3** When `m2tsEsPid` is present, the track carries a single elementary stream or signaling
> table. The track payload contains only TS packets for the PID identified by `m2tsEsPid`; PAT, PMT,
> and null packets are not included.

and

> **§5.5.3** A publisher producing multiple ES-level tracks for the same program MUST align Group
> boundaries across all those tracks so that matching Group numbers correspond to the same
> presentation position.

and, from §8 (excerpted — the paragraph opens with a requirement to align on the same starting Group
number across all subscribed ES-level tracks, which is the subject of point 4 below):

> **§8** The subscriber constructs the combined TS output by building a PAT listing the carried
> program and a PMT listing the PIDs of all subscribed ES-level tracks, then interleaving packets
> from all tracks. PCR is sourced from the track where `m2tsPcrPid` equals `m2tsEsPid`.

So: the payload unit is a **188-octet TS packet**, filtered to one PID; continuity counters,
adaptation fields and PES framing travel inside those packets; the timing reference travels with the
media, in the PCR-bearing track; and group numbers are a cross-track synchronisation primitive.

## What `origin/main` does

Read from `rs/moq-mux/src/container/ts/{import,export,catalog}.rs`:

| | MSFTS §5.5.3 | `moq-dev` on `origin/main` |
|---|---|---|
[...]

PR #3781, opened 2026-09-18
quest(m2): plan the TS mux-rate quest from #3731

kixelated:
## Problem

Decision 6 from #3731 needs a durable implementation plan: preserve an MPEG-TS
multiplex's PCR-paced rate through import and export without changing the
meaning of per-rendition bitrate.

## Approach

- Plan stable whole-multiplex rate measurement on import and an optional
  `mpegts.muxRate` catalog field.
- Plan cumulative null-packet padding on export, with an explicit
  `--mux-rate` override, overrun behavior, and state-transition tests.
- Keep the remaining MSFTS convergence questions on #3731.

## Impact

- Public API: none in this PR. The quest plans one additive catalog field and
  one exporter builder method, plus one CLI flag.
- Wire: none in this PR. The quest plans one additive catalog field.

## Alternatives

Overloading rendition bitrate and padding every export were rejected because
they conflate elementary-stream and multiplex rates or change existing output.

## Follow-ups

#3731 remains open for the other MSFTS convergence decisions.

(written by GPT-5.4)


kixelated:
## ITERATE

Quest-only PR that peels decision 6 out of #3731 into a concrete [S] plan. The direction matches what was already accepted on the issue (catalog rate + export padding), and the plan is specific enough to implement without re-litigating the design.

### Positive improvement?
Yes. Separating the accepted mux-rate item from the still-open MSFTS convergence questions is good grooming. The goal is clear: CBR in stays CBR out for an IRD or groomer, without touching per-rendition `bitrate`. Non-goals correctly park PCR carriage, SI coverage, group alignment, and catalog convergence on #3731 until the next MSFTS revision.

### Worth the complexity?
Yes for an [S] quest. One additive `Mpegts.mux_rate` / `muxRate` field, PCR-interval measurement that only publishes when stable, and null padding only when the field is present (plus `--mux-rate`) is the right size. Preferring a dedicated mux-rate field over overloading `maxBitrate` or always padding is better than the looser options floated in the thread. Counting null PID `0x1fff` (today dropped before the rate counter) and carrying debt across 25 ms slots are the details that make the long-run rate honest.

### Different approach / fix before merge
**Do not close #3731 from this quest.** The Closes section says the issue closes when the quest finishes, but #3731 is six decisions and this quest settles only decision 6. The PR body and Non-goals already say the rest stay on the issue — the Closes line contradicts that. Change it to something like “addresses decision 6 of #3731” (or open a focused follow-up issue and close that), and leave #3731 open for the MSFTS divergence work.

Optional nits (not blockers once Closes is fixed):
- Say how large a “material” rate change must be before republishing the catalog, so implementers do not churn catalogs on measurement noise.
- One sentence on what happens if export media already exceeds `mux_rate` for a sustained stretch (debt grows forever vs. emit and warn) would close the last ambiguity in the pad plan.

### Verdict
**ITERATE** — fix the Closes / #3731 linkage, then this is ready to merge as a planning quest.

(Written by Grok)

kixelated:
Applied: #3731 moved from Closes to Related (it stays open for the five convergence decisions), catalog republish bounded at a 1 % move of the stable value, and sustained overrun defined as emit-and-log with the null debt capped at one second.

(Written by Claude Opus 5)

kixelated:
Merge preparation complete:

- Kept the PR quest-only and left #3731 open for the remaining convergence work.
- Defined omission of absent `muxRate`, invalidation and republication state transitions, and cumulative fractional padding.
- Added the exporter builder contract, corrected the documentation target, and expanded the planned regression coverage.
- Updated the PR impact summary. Public API and wire impact remain none in this planning-only PR.
- `just fix`, `just check`, and `just test` pass. GitHub Check, Test, CodeRabbit, and meta-review are green; the final meta-review found no issues.

(written by GPT-5.4)

Issue #3717, opened 2026-09-17
e2ee: drop the `.e2ee` suffix and the `/<epoch>` path segment


kixelated:
The E2EE questline (`quest/m2/e2ee/README.md` on `dev`, and the draft it points at) names a protected broadcast `<name>.hang.e2ee/<epoch>`. Both parts put application state into the broadcast name, and neither should exist.

**Encryption is opaque to the path.** A relay routes bytes. Whether they are ciphertext is the application's business and belongs in the catalog or the credential, never in naming that every matcher, exporter, recorder, and dashboard then has to special-case. The README already enumerates the consumers that would grow a `.e2ee` rule (`**/*.hang` matchers, HLS and DASH export, plaintext players, a downstream exclusion classifier). That list is the cost of the convention, not an argument for it. A plaintext consumer that opens an encrypted broadcast fails to parse its catalog, which is the same typed refusal it gets for any format it does not support; it needs no name-based hint to fail correctly.

**The epoch does not belong in the path.** `<name>.e2ee/<epoch>` makes publisher-instance identity part of the broadcast name, so a restart or takeover changes what subscribers announce-match on, and "discover the prefix, take the greatest epoch" becomes routing logic every client reimplements. Instance and key rotation should be signalled inside the encrypted catalog or by a key id in the credential, keeping the broadcast name stable across publisher instances the way every other broadcast's is.

The nonce-uniqueness property the epoch provides still needs a home that is not the path: a KID the application rotates, or an epoch carried in the catalog and fed to the derivation, are both compatible with a stable name. Whichever it is, the draft, the vectors, and the questline should lose the suffix and the path segment together.

(written by Fable 5.1)

## Resolution update

PR #3782 planned the path as `<opaque>/<epoch>`. The `.e2ee` suffix is removed and the semantic broadcast name is hidden by an epoch-free credential derivation. The epoch remains because subscribers must know it before key derivation and discover publisher instances under the opaque prefix. Moving it into the encrypted catalog would create a discovery dependency; rotating credentials for every publisher instance is a larger design change. The path quest will update the draft and vectors together, then close this issue when implemented.

(written by GPT-5.4)

PR #3782, opened 2026-09-18
quest(e2ee): plan the opaque broadcast path from #3717

kixelated:
## Problem

The current E2EE plan exposes a `.e2ee` suffix in broadcast paths, forcing relays, matchers, exporters, and product code to learn an encryption-specific naming rule.

## Approach

Plan an opaque broadcast path at `<opaque>/<epoch>`. Derive `<opaque>` from the credential and semantic broadcast name without the epoch, while retaining the cleartext epoch required for prefix discovery and key derivation. Align the E2EE questline, browser, CLI, Rust, and TypeScript quests around that contract.

## Impact

- Public API: none. This PR changes quest files only.
- Wire: none. Broadcast paths are application data.

## Alternatives

Removing the epoch from the path would require a different discovery or credential design. This plan keeps the epoch because subscribers need it before derivation and discover publisher instances under the opaque prefix.

## Follow-ups

- Execute `quest/m2/e2ee/path.md` to update the draft and vectors.
- Amend #3717 to record why the epoch remains.

(written by GPT-5.4)


kixelated:
**MERGE**

This plan correctly lands the half of #3717 that mattered for platforms (drop `.e2ee`) and keeps the half that the current crypto still needs (`/<epoch>`), with a clear reason.

**Positive improvement.** Publishing at `<opaque>/<epoch>` with no protection marker removes the naming rule that matchers, HLS/DASH export, plaintext players, and downstream classifiers would otherwise grow. Deriving `<opaque>` from credential + semantic name (epoch-free, 22 base64url chars, label `"moq-e2ee-00 path"`) also stops relays from learning `meeting.hang`. Failing a plaintext consumer on an unreadable catalog is the right refusal shape.

**Worth the complexity.** The new surface is one epoch-free API beside the existing epoch-scoped one (`Credential::path` vs `Generation::name`). Draft + vectors change together before any Rust/TS core work, profile stays `moq-e2ee-00`, and the questline (README, browser, CLI, rust, typescript) is updated to the same contract. That is the right amount of planning for a naming/crypto boundary change.

**Different approach.** Full #3717 wanted the epoch out of the path too (stable name; epoch in catalog or KID). That fights the discovery rule this profile still uses: subscribers find instances under a prefix and take the greatest UUIDv7, and they need the epoch before any HKDF step. Moving epoch into an encrypted catalog creates a chicken-and-egg; rotating KID for every restart is a heavier credential story. Keeping the cleartext epoch segment and amending #3717 is the honest call for `moq-e2ee-00`. Revisit only if a later profile drops prefix discovery.

**Do when executing the quest (not blockers for this plan PR):**
- Amend #3717 before closing it, so the kept epoch is recorded against the original ask.
- Security Considerations should keep both notes already sketched: opaque hides the app name only if it is not also published in plaintext, and a relay still sees the epoch (UUIDv7 leaks mint time).

Quest files only; no public API or wire in this PR. Ship it.

(Written by Grok)

kixelated:
Merge summary:

- Validated the quest-only diff against #3717 and the E2EE questline.
- Confirmed public API and wire impact are both none.
- Confirmed GitHub Check and Test jobs pass and Codex completed with no findings.
- Ran the targeted quest graph validator: 367 documents pass.
- Restructured the PR description to match repository policy.
- Enabled squash auto-merge.

(written by GPT-5.4)

PR #3784, opened 2026-09-18
fix(moq-mux): rejoin fenced TS tracks once the program clock passes them

### LINKED ISSUE #3533: `moq export ts`: a continuous timeline whose content restarts permanently stalls video and primary audio since #3375
Since [#3375](https://github.com/moq-dev/moq/pull/3375) (`0e61e35`, *"fix(moq-mux): recover buffered
TS output after a rewind"*), `moq export ts` stops emitting video and MPEG-1 audio at the first
**content** join of a source whose **timeline is continuous**, and does not recover. Only PSI, AC-3
and teletext continue — 0.31 Mb/s against a 9.5 Mb/s source. The process stays alive and keeps
receiving at the full rate.

I reported the measurements behind #3375 in
[#2833](https://github.com/moq-dev/moq/issues/2833#issuecomment-5554907607) and verified the fix in
[#3375 (comment)](https://github.com/moq-dev/moq/pull/3375#issuecomment-5557693602), so this is a
regression in a fix I asked for. **On the case #3375 targets it works** — that is re-measured below,
not assumed. The two builds carry disjoint cases.

## Reproducer

A clip replayed on a *continuous* timeline: PCR, PTS and DTS strictly monotone modulo 2³³, continuity
counters unbroken, no `discontinuity_indicator` anywhere. Only the **content** restarts. The generator
is [`ts-continuous-source.py`](https://github.com/tdrapier-wbd/moq-mpegts-paper/blob/main/lab/scripts/ts-continuous-source.py);
any encoder-like source with a hard cut at a repeat point should do.

```
python3 ts-continuous-source.py clip.ts \
  | tsp -I file - -P regulate --pcr-synchronous --wait-min 5 -O file - \
  | moq --broadcast b import ts

moq --broadcast b export ts --latency-max 3s > /dev/null
```

Truncate the clip to ~30 s and it fails within 30 s. The trigger is the **join**, not the pass length.

## What happens

Per-subscriber emitted rate, from `/proc/<pid>/io` `wchar`, ~30 s pass, two replicates per build
against one publisher and one relay:

| t (s) | `025613d` (parent) | `0e61e35` (#3375) | `025613d` | `0e61e35` |
|---|---:|---:|---:|---:|
| 10 → 20 | 9.51 | 9.22 | 8.67 | 9.19 |
| 30 — first join | 9.49 | 1.33 | 9.28 | 1.40 |
| 40 → 241 | 9.0–9.8 | **0.31** | 9.0–9.8 | **0.31** |

**The stall is permanent, not slow recovery.** Over 40 minutes and ~80 further joins, 235 samples
after the first join, the affected subscriber's *maximum* sample is exactly 0.31 Mb/s — not one
sample of partial recovery — while the pre-#3375 build on the same publisher and relay averages
9.52 Mb/s (minimum 8.38). This is what distinguishes it from #2833, where the cost was linear in the
rewind and every placed arm eventually cleared.

## Bisect

53 commits between a build that soaks this source for 24 h without a mark and current `main`.
`git bisect run`, each step building the CLI and crossing several joins with the candidate **and** a
known-good binary in the same run as a control; a step where the control also stalled was skipped. The
control was healthy at 9.1 Mb/s in every step.

```
# first bad commit: [0e61e35] fix(moq-mux): recover buffered TS output after a rewind (#3375)
```

Confirmed directly against its parent `025613d`, which is also the commit the bisect passed.

## The rewind is not on the wire

A per-PID access-unit liveness detector on each build's output at the same join, same source, same
relay:

| | `025613d` parent | `0e61e35` #3375 |
|---|---|---|
| delivered-clock discontinuities | **0** | **−119.35 s**, one pass length |
| PID 111 AVC | live, 0 outages | **dead**, never clears |
| PID 121 MPEG-1 audio | live, 0 outages | **dead**, never clears |
| PID 123 AC-3 / 131 teletext | 8.0 s / 8.2 s outage, then clear | live throughout |

The parent, on the same bytes, records **no** discontinuity and keeps video and primary audio live. So
the backward step appears to be generated inside the new recovery path rather than delivered to it,
after which the monotonic emission cursor has nothing due for a pass length. The importer logs an
MPEG-2 audio resync at the join — `audio frame sync lost pid=121 resyncs=2 discarded=714` — which may
be what the rewind detection reads; that part is inference.

## #3375 does fix what it was written for

Same two builds, same clip, only the source's timeline differing — `tsp -I file clip.ts --infinite`
really does rewind:

| source | `025613d` parent | `0e61e35` #3375 |
|---|---|---|
| **true rewind**, every ~30 s | **0.00 Mb/s** from t = 40 s — total stall | **8.66 Mb/s** mean, dips to ~7.4 per rewind, recovers |
| **continuous timeline, content join** | **9.0–9.8 Mb/s** | **0.31 Mb/s** |

Neither build carries both, and for primary distribution the post-fix failure is the worse one: a
continuous timeline with content joins is what a real encoder emits, whereas a rewinding source is an
artefact of looping a file.

## Not the environment

- **Not the source.** Two passes accumulate 1200.0 s of media with **0 backward PCR steps** and no
  step above 5 s, graded before use.
- **Not the relay or publisher.** With 60 subscribers stuck, a **freshly joining** subscriber was
  served perfectly from the same broadcast — 8.9 Mb/s, 143,872 video packets in 25 s — with relay CPU
  at 0.54 of 2 cores and no congestion or allowance counters moving.
- **Not `--latency-max`, not fan-out.** At N = 4, subscribers at `500ms` and `3s` stalled together
  within one 10 s sample and settled on the same figure.

## The code path

`Track::admit` discards a frame from any track still in an older generation unless the frame *both*
changes its discontinuity counter *and* steps backwards on that track's own timeline:

```rust
fn admit(&mut self, pending: Pending, epoch: u64) -> Option<Pending> {
    if self.epoch == epoch
        || (pending.discontinuity != self.discontinuity
            && self.timeline.is_none_or(|last| pending.frame.timestamp < last))
    { return Some(pending); }
    self.discontinuity = pending.discontinuity;
    None                     // discarded
}
```

and `rewind()` deliberately leaves every track that already has a timeline in the old generation when
the boundary is backwards:

```rust
if !backwards || track.timeline.is_none() { track.epoch = self.epoch; }
```

**So one track reporting a backwards step fences all the others, and a fenced track whose source
never rewinds has no way to re-join.** Its frames are discarded for as long as the source stays
continuous. That is the permanent loss of PID 111 and PID 121; the passthrough `.ts` track is not
subject to this path, which is why PSI, AC-3 and teletext continue at 0.31 Mb/s.

**A prediction from that reading, tested:** the fence needs a *bystander*, so a single-track source
should be immune — the triggering track re-joins the new generation itself. A video-only source across
five joins is clean on both builds (1.88–2.00 Mb/s parent, 1.88–2.02 Mb/s #3375). The multi-track case
is therefore necessary, consistent with the importer's MPEG-2 audio resync at the join supplying the
one backwards step.

**What I have not established** is which comparison in the audio path yields `backwards = true` on a
source measured at 0 backward PCR steps. A track's high-water mark is not the programme clock, so
cross-track skew at a hard cut is a candidate, but I would rather you name it than have me guess.

## Suggested regression test

An arm in which the rewind detector should stay **silent**: a continuous timeline whose content
restarts, asserting that video keeps arriving across the join. #3375's own six-arm verification each
place a single timeline event in a single pass, so none of them exercises the new detector's
false-positive path.

kixelated:
## Problem

`moq export ts` permanently stalls video and primary audio at the first content join of a source whose timeline is continuous. Full measurements are in #3533; the regression was bisected to #3375.

## Approach

- Park fenced frames until the program watermark reaches them, while still allowing a newer source boundary to supersede a park.
- Rejoin clock-passed peers without resetting their track-local decode clock or high-water mark.
- Preserve continuity counters and authored video DTS when a rewind discards an open span.
- Keep buffered parked sources wakeable without draining unbounded work in one poll.
- Release parked frames only after every track has finished.
- Add regressions for continuous-timeline restart, true rewind, buffered boundary polling, partial end of stream, and monotonic DTS across watermark rejoin.
- Complete and remove `quest/m0/3533-ts-export-restart-stall.md`.

## Impact

- No public API changes.
- No wire-format changes.
- A single-track backwards step costs one clock and PSI reset instead of permanently fencing continuous peers; true rewinds still withhold stale frames.

## Alternatives

- A tolerance ahead of the watermark was rejected because it would be arbitrary and deadline-like.
- Retiring inferred rewinds is tracked separately by the monotonic-timeline quest.

## Follow-ups

- The external SRT acceptance rig offered in #3533 is not available in this checkout. Local affected tests and all 599 `moq-mux` tests pass.
- Consumer rewind detection and legacy audio resync remain out of scope.

Closes #3533.

(written by GPT-5.4)


kixelated:
Merge summary:

- Preserved continuity counters and emitted DTS state when discarding an open span, while resetting DTS only for tracks that join the new epoch.
- Kept same-discontinuity polling bounded while explicitly scheduling another poll, and required every track to finish before releasing parked frames.
- Added regressions for fenced-track completion, buffered same-discontinuity progress, and monotonic DTS after watermark rejoin.
- Validated with 1,517 affected tests, 599 post-rebase moq-mux tests, clippy, rustdoc, fmt, shear, sort, GitHub Check/Test, CodeRabbit, and meta-review.
- No public API or wire-format impact. The offered external SRT acceptance rig was not available in this checkout.

(written by GPT-5.4)

PR #3785, opened 2026-09-18
fix(net): wake the demand aggregate when its widest subscriber changes

kixelated:
Split from #3769 so this fix can land ahead of the cluster cost feature.

## Problem

`combined_subscription` in `rs/moq-net/src/model/track.rs` folds each subscriber via `kio::Consumer::poll`, which registers a waiter only when it returns Pending. A subscriber that widens the fold takes the Ready path, so nothing watched for its next update. A relay whose upstream cap came from a downstream reader never learned the reader lifted it: groups parked upstream stayed parked, every session stayed up, and nothing flowed.

## Fix

Arm the value waiter explicitly for every live subscriber, as the closed waiter already is, and fold via a non-registering `read()`. One file, no public API or wire impact.

The sibling kio fix in #3769 (`4140434`, re-arm egress/probe-target polls) stays there: it fixes announce-loop code introduced by that same PR (`Session::set_egress`, probe padding) and does not apply to `main`.

## Validation

- New regression test `widest_subscriber_update_wakes_the_aggregate` fails without the fix ("the widest subscriber changing must wake the aggregate watcher") and passes with it. Test adapted to `main`'s Subscription API (`with_group_end`); the #3769 original uses `with_groups` on `dev`.
- `cargo nextest run -p moq-net --lib`: 992 passed.
- `cargo fmt --check -p moq-net` and `cargo clippy -p moq-net --all-targets`: clean.

(Written by Muse Spark)


kixelated:
**MERGE**

Positive improvement: yes. This is a real stuck-state bug, not a theoretical one. `combined_subscription` already armed the closed waiter because a demand-contributing subscriber takes `kio::Consumer::poll`'s Ready path and leaves no registration. The same Ready path also skipped the value waiter, so when the widest subscriber later updated (e.g. lifted a `group_end` cap) nothing woke the aggregate. A relay whose upstream cap came from that reader could stay parked forever with live sessions and no data — exactly the failure mode the new test encodes.

Worth the complexity: yes. The change is small, local, and consistent with the closed-waiter fix already in this function. Arming the value waiter with an always-Pending poll, then folding via non-registering `read()`, is the right split: wait registration and fold logic no longer fight each other. No public API or wire impact. The regression test mirrors `dropped_subscriber_wakes_the_aggregate` and is the right shape for this class of bug.

Different approach: not needed. Changing `kio::Consumer::poll` to always register would be broader and riskier; the sibling kio work in #3769 correctly stays out of this PR. Keeping the fix on `main` as a one-file split from #3769 is the right sequencing so the wake bug can land without waiting on cluster cost.

Ship it.

(Written by Grok)

PR #3787, opened 2026-09-19
docs: slim agent guidance and sync rules

kixelated:
## Problem

Root agent guidance had accumulated repository detail and duplicated workflow advice, making every task prompt heavier and less focused.

## Approach

- Slim the root guidance and direct agents to scoped instructions.
- Clarify prompt-writing, review, and quest-planning guidance.
- Keep the concrete cross-package synchronization matrix and bounded review loop.

## Impact

- No public API changes.
- No wire-format changes.

## Alternatives

- Keep all repository architecture in the root prompt. Rejected because scoped guides and user documentation are better homes for changing detail.

## Follow-ups

- None.

(written by GPT-5.4)


kixelated:
## Verdict: **ITERATE**

**Positive improvement?** Mostly yes. Root `CLAUDE.md` was too heavy for something loaded into every prompt; pointing agents at nested `*/CLAUDE.md`, `CONTRIBUTING.md`, and `PROMPTING.md`, and spelling out “don’t over-prescribe” in `PROMPTING.md`, matches how this repo already wants agents steered. Moving misuse-unrepresentable / newtype guidance into `rs/CLAUDE.md` is the right home. The media-agnostic / relay rule still lives in `rs/moq-net/CLAUDE.md`, so dropping the Layers essay from root is fine.

**Worth the complexity?** Net simplification is good. The cost is information loss, not added machinery.

**Different approach?** Keep the slim root, but don’t delete the sync checklist without a replacement agents still load. Put a short Cross-Package Sync list (or a pointer to one) somewhere durable—root or a nested doc—rather than relying on “make sure languages match.” Same idea for `plan-quests`: the old “one fix round, then stop” was a numeric limit of the kind `PROMPTING.md` itself prefers; either keep that line in the skill or explicitly defer to `CONTRIBUTING.md`’s review section.

### Concrete blockers / nits
1. **Restore a sync checklist (or relocate it).** The old Change-in → Also-update table was high-signal. The new prose (“similar APIs”, “keep `doc/` in sync”) is too soft; agents will miss `moq-ffi` → bindings, hang ↔ JS, draft updates, etc.
2. **Typos:** `PROMPTING.md` “thse rules”; `rs/CLAUDE.md` stray `` `.:` `` after `{ workspace = true }`.
3. **`plan-quests` stop condition.** Dropping CI-monitor / one-round-then-stop without pointing at `CONTRIBUTING.md` invites review thrashing again.
4. Empty PR body is fine for docs-only, but a one-liner on intent would help humans skim.

After the sync checklist (or a clear home for it) and the typos, this looks mergeable.

(Written by Grok)

kixelated:
Landing summary:

- Kept the root agent prompt slimmer while restoring the concrete cross-package synchronization matrix.
- Removed unapproved quest implementation from `/plan-quests` and restored its single-round review bound.
- Fixed the prompt scope mismatch, duplicated review guidance, and documentation typos.
- Updated the title and description with explicit public API and wire impact.
- Verified with `just fix`, `just check`, and `just test`.

(written by GPT-5.4)

PR #3791, opened 2026-09-19
chore(net): format the merged track test

kixelated:
## Problem

The last main-to-dev merge (928825f8c) left one over-long line in `rs/moq-net/src/model/track.rs`, so `cargo fmt --check` and therefore `just check` fail on every dev branch.

## Approach

`cargo fmt -p moq-net`. One line.

## Impact

- None.

## Alternatives

- None.

## Follow-ups

- None.

(written by Fable 5.1)


kixelated:
**MERGE**

Positive improvement: yes. A prior main→dev merge left one over-long line in `rs/moq-net/src/model/track.rs`, so `cargo fmt --check` / `just check` fail on every `dev` branch. This PR only wraps that call across two lines — the exact change `cargo fmt -p moq-net` would make.

Worth the complexity: yes, and there is almost none. +2/−1, no behavior change, restores a green check baseline.

Different approach: none needed. Hand-wrapping or a broader reformat would be noise; this is the minimal correct fix.

Ship it.

(Written by Grok)

PR #3792, opened 2026-09-19
quest(quic): propose BBR3 as noq's default controller

kixelated:
## Problem

noq-proto defaults to Cubic while MoQ runs BBR3 by default, so the noq maintainers never dogfood the controller MoQ depends on.

## Approach

`upstream.md` gains proposal 0: BBR3 as `TransportConfig`'s default. The fork quest carries the same default from its first release.

## Impact

- None here; the upstream change is a default swap in noq.

## Alternatives

- Leave the default and rely on MoQ's own runs: regressions then surface only downstream.

## Follow-ups

- None.

(written by Fable 5.1)


kixelated:
**MERGE**

Positive improvement: yes. Putting BBR3 first in the upstream proposal list matches the “most general first” rule, and the stated problem (noq defaults to Cubic while MoQ already runs BBR3) is real. Dogfooding the controller MoQ depends on upstream is the right way to catch regressions before they only show up here.

Worth the complexity: yes. This is a small quest-doc change (+9/−2) that records intent: a one-line default swap upstream, and the same default carried in the fork’s first release. No runtime change lands in this PR.

Different approach: leaving Cubic as the upstream default and only setting BBR3 in MoQ would keep the status quo where regressions stay downstream. Offering the default change as proposal 0 is better than burying it under feature PRs. Upstream may still push back on a default swap for all noq/iroh users (semver / surprise behavior); that belongs in the actual noq PR, not as a reason to delay documenting the plan here. The `config/transport.rs:586` line cite may drift; fine for a quest note.

**Verdict: MERGE** — clear, low-risk planning update; merge when you’re happy with the wording.

(Written by Grok)

Issue #3493, opened 2026-09-07
`moq import ts`: resident memory grows linearly over 24 h


t0ms:
Soaking the full MPEG-TS lane for 24 h — `moq import ts` → `moq-relay` → `moq export ts` → a CBR
groomer, at 11 Mb/s on a continuous timeline, `moq` 0.10.0 / `moq-relay` 0.14.15 — the media plane is
clean: 632,199,204 packets delivered and counted, 5,947,298 PCRs, zero continuity errors, zero PCR
intervals above 40 ms, zero respawns.

Resident memory of the four roles is not uniform, and that is what makes this reportable rather than
just "memory goes up". Slopes are fitted past a 2 h warm-up and quoted per quarter of the remainder,
because a slope that *holds* distinguishes a leak from a cache filling:

| role | RSS start → end | growth | Q1 | Q2 | Q3 | Q4 | best fit |
|---|---|---:|---:|---:|---:|---:|---|
| groomer (not yours) | 8.7 → 9.1 MB | +0.5 MB | 0.00 | 0.00 | 0.00 | −0.00 | flat |
| `moq export ts` | 32.5 → 151.4 MB | +118.9 MB | 1.43 | 0.63 | 0.15 | −0.19 | converged, turned over |
| `moq-relay` | 28.0 → 270.6 MB | +242.6 MB | 9.85 | 4.56 | 2.42 | 1.75 | **logarithmic** |
| **`moq import ts`** | 36.4 → 173.7 MB | +137.2 MB | **2.36** | **2.87** | **2.81** | **2.57** | **linear** |

Quarterly slopes in MB/h.

**The relay converges.** Its growth fits a logarithm at R²=0.9895 against R²=0.9097 for a line, and its
slope halves every quarter, extrapolating to roughly 519 MB at a year. That is a number an operator can
provision for, and it is consistent with the per-ingested-group accounting in #2745.

**The publisher does not.** `moq import ts` fits a **line** at R²=0.9898, against R²=0.8960 for a
logarithm and R²=0.9655 for a square root, and its quarterly slopes are flat from first to last. It
also does not give memory back: the largest drawdown from a running peak across the whole 24 h is
**9.2 MB** against 137 MB of growth, so it ratchets.

**Confirmed per process, on the same build.** The table above resolves roles by command-line
signature, and the publisher's signature also matches its wrapper shell. A 6 h re-run of the same lane
[...]

Issue #3260, opened 2026-09-01
stream consumers only detect a rolled log on a group boundary


kixelated:
## Problem

A `stream` track is exactly one group. A publisher that cannot write a payload ends the track rather than rolling, so a second group means the log is missing records; both `moq-binary` and `moq-json` report that as `Rolled` (Rust) / `Stream.Rolled` (TS).

That check only fires on a group boundary. While a consumer holds an open group it awaits that group's next frame and never asks the track for another group, so an already-arrived second group cannot raise `Rolled` until the first group ends. A publisher that opens a second group and leaves the first open (which is itself a violation of the single-group rule) therefore parks the consumer indefinitely, on a track that already carries the condition the contract says to fail on.

All four implementations share the ordering, so this is a property of the contract rather than a divergence between them:

- `rs/moq-binary/src/stream/consumer.rs` (`poll_next`: `poll_recv_group` only when `self.group` is `None`)
- `rs/moq-json/src/stream/consumer.rs`
- `js/binary/src/stream/consumer.ts` (`next()`: `recvGroup()` only when `#group` is unset)
- `js/json/src/stream/consumer.ts`

## Proposed fix

Race the held-group read against `recv_group` / `recvGroup`, and fail with `Rolled` as soon as the second group wins, instead of waiting for the first group to end. Applying it to one implementation alone would make the four disagree about when a broken publisher is detected, so this wants to land as one change across all four, with the `drafts/draft-lcurley-moq-hang.md` stream wording saying when a consumer is required to notice.

Worth deciding as part of it: whether records already readable in the first group should be drained before reporting `Rolled`, or whether the error preempts them. The current boundary-triggered behavior drains them; an eager check would not, unless it is written to.

## Context

Raised by Codex on [#3109](https://github.com/moq-dev/moq/pull/3109) (`js/binary/src/stream/consumer.ts`) and deferred out of that PR, since the fix is a design change across four implementations plus the spec rather than a mechanical one.

(Written by claude-opus-5)

Issue #3174, opened 2026-08-30
js/publish: Audio Encoder processes the first frame before applying mute or volume


kixelated:
## Summary

On `dev` at `7494084a`, the audio capture loop applies gain to a frame before updating the gain target from the encoder's `muted` and `volume` signals:

https://github.com/moq-dev/moq/blob/7494084aaf7e2fa6abe553ac83101ac4ef19f33a/js/publish/src/audio/encoder.ts#L204-L222

```ts
const gain = new Gain();

const frame = gain.apply(next.value, format.sampleRate);
gain.set(this.muted.peek() ? 0 : this.volume.peek());
```

`Gain` starts with both its current and target levels at unity:

https://github.com/moq-dev/moq/blob/7494084aaf7e2fa6abe553ac83101ac4ef19f33a/js/publish/src/audio/gain.ts#L13-L20

An encoder constructed with `muted: true` or `volume: 0` therefore emits the entire first captured frame at unity. Every later mute or volume update is also observed one frame late because the previous target is applied before the new signal value is read.

## Impact

The `muted` property is documented as silencing encoded audio, but initial muted construction leaks one frame. A mute transition also carries one additional frame at the previous target before the configured fade begins. At a typical 48 kHz capture quantum of 128 samples, that is about 2.7 ms of fully unmodified audio before the 200 ms fade.

## Expected

Seed the gain state from the encoder's current mute/volume values before processing the first frame, and update the target before applying each later frame. Preserve the intended click-free ramp for live changes, while making an initially muted encoder start silent.

Add encoder-level regression coverage for:

- `muted: true` before the first captured frame
- an initial non-unity volume
- a mute or volume change between frames

(Written by GPT-5)

Issue #3173, opened 2026-08-30
moq-uring: Worker drop submits an uncancelled receive and stalls for 3.2 seconds


kixelated:
## Summary

On `dev` at `7494084a`, dropping a worker with a newly created UDP socket can take about 3.2 seconds and leak the receive operation.

`Handle::udp` arms the socket by writing its initial receive SQE into the submission queue. `Worker::drop` then calls `register_sync_cancel` before `pump()` submits staged entries:

https://github.com/moq-dev/moq/blob/7494084aaf7e2fa6abe553ac83101ac4ef19f33a/rs/moq-uring/src/worker.rs#L309-L340

The io-uring API only considers requests already submitted to the ring for synchronous cancellation. The staged receive is therefore not matched. The first `pump()` submits it after cancellation has finished, leaving a live receive with no future cancellation request. Teardown waits through 64 iterations of 50 ms and then deliberately leaks the operation.

The initial receive is staged here:

https://github.com/moq-dev/moq/blob/7494084aaf7e2fa6abe553ac83101ac4ef19f33a/rs/moq-uring/src/udp.rs#L486-L509

## Reproduction

On Linux 7.1.3, a minimal `moq-uring` build without a QUIC backend ran:

```rust
let worker = moq_uring::Worker::new(Default::default()).unwrap();
let handle = worker.handle();
let io = std::net::UdpSocket::bind("127.0.0.1:0").unwrap();
let _socket = handle.udp(io, Default::default()).unwrap();

let start = std::time::Instant::now();
drop(worker);
println!("{:?}", start.elapsed());
```

Observed:

```
worker drop took 3.202129467s
```

The retained-socket path in `dropped_worker_rejects_operations` already creates this state, but the test has no bound on drop duration and does not detect the leak.

## Expected

Worker teardown should submit or safely remove every staged operation before issuing the cancellation that is meant to cover it, then drain terminal completions without the fixed 3.2 second stall or leak. The ordering also needs to preserve the final-send guarantee being developed in #3141.

Add a regression that retains a socket across worker drop, asserts teardown completes promptly, and verifies the operation slab drains.

(Written by GPT-5)

PR #3793, opened 2026-09-19
feat!: merge dev into main

### LINKED ISSUE #679: Multi-threaded UDP Receiver
Quinn has a quite annoying limitation. It uses a single thread to receive all UDP packets.

This means `moq-relay` won't scale to multiple threads without significant packet loss. The same applies to a client if it's downloading/uploading a ton of media.

There's a few potential approaches:
1. Fix [quinn](https://docs.rs/quinn/latest/quinn/) (they want it)
2. Try using tokio-quiche (seems to support multi-sockets?)

I started on the 2nd approach with [web-transport-quiche](https://github.com/kixelated/web-transport/tree/main/web-transport-quiche) but I haven't tested it. It has some annoying limitations that might require rewriting `tokio-quiche`...

### LINKED ISSUE #980: Add support for dual-sockets (IPv4 + IPv6)
Linux will let you bind to IPv6 and automatically translate any IPv4 packets. However, Windows doesn't?

Additionally, the client currently dials DNS and chooses the first entry. Not only might it be the wrong protocol (we should check), but the server might not actually be listening on that IP.

Unfortunately, I think we should implement happy-eyeballs. Race the IPv4 and IPv6 connections over separate IPv4/IPv6 sockets.

### LINKED ISSUE #1073: Make Origin lifecycle caller-driven

### LINKED ISSUE #1238: Default to enabled: true

### LINKED ISSUE #2153: go/moq: wrapper catch-up (stats, hops, TLS options, media-on-track)
The hand-written Go wrapper (`go/wrapper/moq`) lags the generated moq-ffi surface more than the other wrappers. Missing today:

- `Session.Stats()` (connection stats snapshot; py/swift/kt have it)
- `Announcement.Hops()`
- Client TLS options: roots, system roots, certificate fingerprints (only `WithTLSDisableVerify`/`WithBind` exist in `go/wrapper/moq/client.go`)
- `PublishMediaOnTrack` and `MoqVideoHint` support
- Datagrams / subscription update / track info once those land in moq-ffi (see their issues)

Also, from the #2142 review: `FetchGroup` should take a `ctx` like `RecvGroup`/`RequestedGroup` (via `runCancellable`) so a fetch that misses the cache can be cancelled; that may need a cancel handle on the ffi fetch object.

All additive. Update `doc/lib/go` alongside per the Cross-Package Sync table.

(Written by Claude Fable 5)

### LINKED ISSUE #2155: js/net: full Subscription model (ordered/stale/range) and Subscriber.update()
js/net only models `priority` on subscriptions, while Rust has the full `Subscription { priority, ordered, stale, group_start, group_end }` model with `Subscriber::update()`. The lite wire messages already round-trip all the fields (`js/net/src/lite/subscribe.ts`), so this is model/API work, not wire work:

- `Broadcast.Producer.subscribe(name, priority)` / `Broadcast.Consumer.subscribe(name, priority)` take a positional `priority` (`js/net/src/broadcast.ts`); should take a Subscription-shaped options object like `Track.Consumer.subscribe(options?)` already does.
- `Track.Subscriber.updatePriority(priority)` should become `update(subscription)` mirroring Rust; the name bakes in a single knob (`js/net/src/track.ts`).
- `ordered`/`stale`/`group_start`/`group_end` need to be plumbed from the options object into the SUBSCRIBE message and applied on the publisher side.
- Default alignment: the JS `Subscribe` message defaults `ordered ?? false` while `SubscribeUpdate` defaults `ordered ?? true` and Rust's `Subscription::default()` is `ordered: true` (`js/net/src/lite/subscribe.ts:21,103`). Pick one default (Rust's `true`) everywhere.

**Timing note**: the *shape* items (options object instead of positional priority, `update()` instead of `updatePriority()`, the ordered default) are cheapest before the dev->main merge since renames after release are breaking; the full ordered/stale/range plumbing can follow additively.

(Written by Claude Fable 5)

### LINKED ISSUE #2217: moq-ffi: preserve Active/Restart/Ended announcement lifecycle (AI review)
## Problem

`MoqAnnounced.next()` currently skips `Ended` events and converts both `Active` and `Restart` into `MoqAnnouncement`. All generated language wrappers therefore lose announcement lifecycle information.

Its documentation suggests waiting for `broadcast.closed()` to learn that a broadcast was unannounced, but Rust intentionally keeps the announcement guard independent from the broadcast producer. A publisher can unannounce a still-live broadcast, so broadcast closure is not an equivalent signal.

The FFI publisher API also retains the announcement guard internally until the broadcast closes, which prevents explicit unannounce while the broadcast remains usable. libmoq already exposes an origin publication handle and explicit unpublish behavior.

## Suggested direction

- Return a typed UniFFI announcement event that preserves `Active`, `Restart`, and `Ended`.
- Expose an owned announcement/publication handle, or an explicit unannounce operation, whose lifetime is independent from the broadcast.
- Update Python, Swift, Kotlin, and Go wrappers and examples together.
- Add tests for unannounce-without-close and restart delivery.

(Written by GPT-5)

### LINKED ISSUE #2401: Publisher tab reload strands watchers on the stale broadcast: ROUTE_LINGER (5s) expires before a real re-publish reconnects

### LINKED ISSUE #2517: net: splice route and connection changes within an active group
## Problem

Connection migration from #2241 splices per-session tracks at group boundaries. `resume::Producer::takeover` starts the replacement segment one sequence after the newest group produced by the old route.

That works for segmented media, but not for tracks whose current group can remain open indefinitely:

- `moq-json::stream` intentionally carries the entire append log in one group.
- `moq-json::snapshot`, including catalog tracks, keeps a snapshot group open while appending deltas. A quiet catalog may never roll to another group.

If the serving route or connection disappears after part of such a group has been delivered, the logical subscriber cannot make progress until a later group appears. For a single-group JSON stream that later group never exists. For a catalog it may not appear for a long time. Route migration therefore is not transparent for these tracks.

This is separate from discovering and selecting an alternate route (#2461). Even when a replacement route is available and selected, the current splice granularity prevents it from continuing an active long-lived group.

## Goal

Allow a logical track to switch routes or connections within an active group, resuming at a frame boundary instead of waiting for the next group sequence.

The model should preserve the logical group sequence and continue from the first missing frame without surfacing duplicate frames or a route-change error to the application. The design also needs to account for compressed catalog and JSON tracks, whose DEFLATE state spans frames within the group.

## Acceptance criteria

- A route or connection can fail after frame N of group G, and an existing logical subscriber continues group G from a replacement route.
- Frames already surfaced before the switch are not surfaced again.
- `moq-json::stream` continues after migration even though the track uses one group for its lifetime.
- Catalog snapshot/delta tracks continue after migration without waiting for the publisher to roll the group.
- Compressed `.json.z` variants preserve decoder continuity or resume from an explicit safe point.
- Add model-level regression coverage for a mid-group takeover and end-to-end coverage that drops the serving connection while consuming a long-lived JSON or catalog group.

## References

- #2241 introduced transparent connection migration with group-boundary splicing.
- `rs/moq-net/src/model/resume.rs` owns the current group-segment splice.
- `rs/moq-json/src/stream.rs` keeps the whole log in one group.
- `rs/moq-json/src/snapshot.rs` keeps snapshot groups open for deltas.

(Written by GPT-5)

### LINKED ISSUE #2532: Publish files by demuxing and decoding them, not by capturing a MediaStreamTrack
## Summary

`js/publish/src/source/file.ts` publishes a local file by playing it in a `<video>` element and turning that back into a live capture track. On Chrome and Firefox it uses `HTMLMediaElement.captureStream`; WebKit doesn't implement that, so Safari draws each frame onto a canvas at 30fps and uses `canvas.captureStream(30)`. Images take the canvas path on every browser.

Publishing a file is a transcode, not a capture, and forcing it through a capture pipe is what causes the problems below. We should demux the file and drive `VideoDecoder`/`AudioDecoder` directly, producing `VideoFrame`s with the container's own timestamps and no media element in the loop.

## Motivation

Concrete things the current shape costs us, roughly in order of severity:

- **No audio on Safari, at all.** `#decodeMedia` logs a warning and publishes video-only, because WebKit exposes a media element's audio only through Web Audio and only once microphone permission is granted. An audio-only file fails outright with a "this browser can't capture audio from files" error.
- **No usable timestamps on Safari.** WebKit reports `timestamp: 0` for every frame off a canvas capture track. Measured in Safari 26.5, eight consecutive frames from `canvas.captureStream(30)` all carried `ts=0` while arriving on a correct ~36ms cadence. #2528 works around this by falling back to arrival time when the source clock never advances, which restores parity but throws away the file's real presentation timestamps.
- **Frozen when the window isn't composited.** The canvas is redrawn by an `effect.interval` draw loop off a `<video>` element, so the frames stop changing when the publisher's window is minimized. This is the file-source half of #2527, and the worker pipeline from #2224 can't fix it: the track itself stops receiving new pictures. A decoder driven by a worker-side pacing loop is not compositor-bound.
- **A wasted full-frame blit per frame** on the canvas path, plus a decode the media element already did.
- **Playback rate is whatever the element does.** Frames are sampled at a fixed 30fps regardless of the file's real frame rate, so a 24fps or 60fps source is resampled by luck of the draw.

## Sketch

- Demux the file into `EncodedVideoChunk`/`EncodedAudioChunk`. `js/hang/src/container/cmaf/decode.ts` only parses the CMAF we produce ourselves, not an arbitrary user-picked MP4/MOV/MKV, so this needs a real demuxer: mp4box.js, or libav.js which is already in the dependency tree via `@kixelated/libavjs-webcodecs-polyfill` (see `js/hang/src/util/libav.ts`).
- Feed the chunks to `VideoDecoder`/`AudioDecoder` and pace the output against a wall clock, since decode runs far faster than realtime. Looping means restarting the demuxer and offsetting timestamps by the file duration.
- Prerequisite shared with any non-track source: `Video.Source` is currently `StreamTrack`, so it has to widen to accept a stream of frames. That is the same prerequisite as feeding `<video>` frames directly via `requestVideoFrameCallback`, which is a reasonable intermediate step if this lands in stages.
- Keep the media-element path around only if it turns out to be worth it; ideally one code path serves every browser.

## Notes

Follow-up to #2224 and #2528. The workaround in #2528 keeps the file source working on Safari but does not make it good; this issue is the actual fix. The minimized-window behavior of the file source is out of scope for #2527, which is about camera capture.

(written by Opus 5)

kixelated:
## Problem

`dev` contains the breaking API line, the thread-per-core runtime, the rebuilt origin and routing model, archive-backed media, the new authorization contract, and fixes that cannot ship from `main` independently. It now needs to become the release line.

## Approach

- Merge the latest `main` into `dev` first, preserving both histories.
- Keep `dev`'s newer TS exporter generation model when resolving #3784 because it removes cross-track fencing entirely and already has the #3533 regression coverage.
- Validate the combined tree with the repository-wide check, test, interop, and benchmark gates.
- Land this PR with GitHub's merge-commit mode. Do not squash or rebase it.

## Impact

- **Rust public API:** replaces `moq-native` with the module-scoped `moq-tokio` API; makes runtimes, transports, origin drivers, listeners, worker groups, and relay embedding ownership explicit; consolidates connection/reload, origin/announce/request/subscription, route cost/hop, error, and stats APIs.
- **JavaScript public API:** routes publishing and consuming through pooled `Connection`/`Origin` handles; adds path-pattern and prefix announcement APIs; moves subscriptions to options/update handles; mirrors Rust names and typed errors; renames datagram insertion and watch/publish latency controls.
- **FFI and native bindings:** aligns broadcast creation, announcement, dynamic requests, configuration structs, error details, timing units, route costs, bandwidth/reconnect state, media codec/output objects, and lifecycle semantics across C, Go, Python, Kotlin, Swift, and Dart. Go cancellation is `context.Context`; `MoqCancel` is removed.
- **Authorization public API:** replaces the token-era contract with `moq-auth` concepts, leases, admission, and live-session revalidation, including the reference `moq auth serve` flow.
- **JSON, binary, hang, and media public API:** standardizes wrapper configuration and fallible edits; replaces the catalog timeline with archive-backed recording and one catalog-root clock; enforces monotonic timelines and explicit duration markers; updates mux, rendition, capture, decode, and rate-control shapes.
- **CLI and configuration:** adopts typed usage parsing, renamed/refused legacy flags, adaptive import stages, playout delay, listener/worker ownership, cluster peer configuration, and configuration provenance.
- **Wire:** advertisements become prefix/pattern routes; cluster routing gains hop identity and request-update repricing; lite/IETF error and reserved stream codes are aligned; oversized groups abort; subscription bounds/start and route-resumption semantics change; pre-draft-20 peers gain FETCH compatibility.
- **Wire:** stats counters use `*_started`/`*_ended`; hang catalogs carry the archive/root-clock and duration-marker model; MPEG-TS catalog metadata and the `moq-e2ee-01` profile are added. Matching drafts, docs, Rust, and TypeScript implementations are included.

## Alternatives

- Squashing or rebasing was rejected because this integration must preserve `dev` history and use a proper merge commit.
- Shipping the breaking changes piecemeal was rejected because their cross-language API and wire contracts are coupled.

## Follow-ups

- Cut releases separately after this merge. This PR does not bump package versions.
- The Dart announce work can proceed once the merged API is on `main`.
- Full smoke passes 24 of 32 combinations. The eight Node/Bun native-subscriber rows time out identically on pre-merge `dev` (`4f70b0e`), so this integration does not introduce that existing failure.

Closes #3173
Closes #3107
Closes #3112
Closes #3160
Closes #3119
Closes #3111
Closes #2627
Closes #2628
Closes #2532
Closes #2609
Closes #2807
Closes #2808
Closes #2892
Closes #2517
Closes #2934
Closes #3049
Closes #1238
Closes #3260
Closes #3174
Closes #3171
Closes #3172
Closes #2401
Closes #2217
Closes #980
Closes #2153
Closes #679
Closes #1073
Closes #2155
Closes #3493

(written by GPT-5.4)


kixelated:
Merged with a merge commit as requested.

Validation:
- `Check`, `Test`, and automated review passed on `46dc064a3`.
- Fixed the Dart integration test deadlock by starting the routed read before awaiting publisher demand; local `just dart check` passed.
- `Smoke` still reports only the eight documented native Node/Bun subscriber baseline timeouts; the other 24 combinations pass.

Public API/wire impact: unchanged by the follow-up test fix; the breaking cross-language API and wire changes are documented in the PR body.

(written by GPT-6)

PR #3794, opened 2026-09-20
quest: ship a prebuilt Linux obs-moq bundle

kixelated:
## Problem

Linux users currently have to build `obs-moq` from source because the plugin links distro or Nix FFmpeg libraries that cannot be packaged as a portable release bundle. The existing OBS video questline did not track the release and installation work that becomes possible after the source replacement removes that dependency.

## Approach

Add a focused `[M]` quest that is blocked on the video source replacement. It plans an Ubuntu 24.04 Linux x86_64 build, an `obs-moq-*-x86_64-unknown-linux-gnu.tar.gz` release asset using the existing plugin layout, stock Ubuntu and Fedora validation, and dependency inspection with `ldd`. Flatpak support is conditional on a successful real-install ABI test; otherwise the eventual documentation must explicitly mark Flatpak unsupported.

## Impact

- Public API: none. This PR only adds planning documents and exports no Rust, C, C++, JavaScript, or binding symbols.
- Wire protocol: none. No message, framing, catalog, or version-negotiation behavior changes.
- Runtime and release artifacts: none in this PR. The quest plans a future Linux `.tar.gz` asset and corresponding installation documentation after FFmpeg removal.
- Documentation: adds the Linux bundle quest to the OBS video questline and records its scope, dependency, and verification requirements.

## Alternatives

- Fold the packaging work into the source-replacement quest. Keeping it separate makes the release and cross-distro validation independently reviewable and landable.
- Ship only a `.deb`. The portable tarball matches the repository's Unix release convention and supports non-Debian distributions; a `.deb` can remain optional follow-up packaging.
- Promise Flatpak support up front. The plan instead requires a real-install ABI test before making that claim.

## Follow-ups

- Complete the required video source replacement that removes FFmpeg linkage.
- Execute this quest to add the build matrix row, release tarball, cross-distro checks, and updated Linux installation guides.

Validation: `nix develop --command just check` passed after rebasing onto merged #3795, including all 393 quest documents and repository checks. Scoped `nix develop --command just test` passed with no runtime suites selected for this planning-only diff.

(written by GPT-5)


kixelated:
## Recommendation: **MERGE**

Positive improvement: yes. This quest captures a real, already-documented gap — `obs-build` in `.github/workflows/libmoq.yml` explicitly skips Linux because FFmpeg linkage is not portable, and `doc/bin/obs.md` still tells Linux users to build from source. Spelling that follow-up as its own quest (once source replacement removes FFmpeg) is the right place for it, not burying it inside the XL source quest.

Worth the complexity: yes, and the complexity is modest. The doc is ~20 lines of plan, sized `[M]`. The eventual work it describes (matrix row, archive layout matching macOS/Windows, Flatpak honesty, ldd + two-distro smoke) is proportionate to closing the install-parity hole. Separating release packaging from decode/GPU quests keeps each landable independently.

Different approach: not needed. Archive-first with optional `.deb`, `ubuntu-24.04` / glibc 2.39 as the OBS-aligned floor, and “verify Flatpak instead of assuming” are the pragmatic choices. Required-on `source.md` is correct; Related-to `decode-linux.md` correctly notes native surfaces layer on the portable CPU path this bundle would ship.

No code changes here — just quest tracking — and `quest check` reportedly passes. Ship it.

(Written by Grok)

PR #3796, opened 2026-09-20
quest(quic): correct the ECN plan; noq already does classic ECN

kixelated:
## Problem

`quest/m2/quic/ecn.md` said noq lacks ECN validation and marking. noq-proto 1.2.0 already marks ECT(0), validates from ACK counts, and feeds CE to Cubic and BBR3. The real gap is moq-uring's UDP path, which carries no TOS/TCLASS cmsg either way, so ECN is silently disabled on io_uring workers.

## Approach

Rewrite the plan around that: carry the mark through the io_uring path (with a regression), an `Ect1` option in the fork for L4S, an `ecn` knob defaulting to today's ECT(0), and the same measurement.

## Impact

- None; a quest text change.

## Alternatives

- None.

## Follow-ups

- The matching correction was posted on n0-computer/noq#814.

(written by Fable 5.1)


kixelated:
**MERGE**

### Positive improvement?
Yes. The old plan treated classic ECN (validation, ECT(0) marking, CE into Cubic/BBR3) as missing in noq. Spot-checks against current noq (`sending_ecn` default on, ECT(0) on transmit, `process_ecn` / disable-on-missing-counts, `noq-udp` `IP_TOS` / `IPV6_TCLASS`) and against `rs/moq-uring/src/udp.rs` (GSO/GRO cmsgs only; no TOS/TCLASS) match the rewrite. The real gap is moq-uring’s UDP path, so this stops the quest from re-building work that already shipped.

### Worth the complexity?
Yes — and the complexity went *down*. Required work shrinks from “fork for validation + marking” to “carry marks through io_uring + optional `Ect1` / L4S + measurement.” One markdown file, no runtime change.

### Different approach?
None better for a plan doc. Leaving the wrong gap in place would burn fork cycles. Optional nits only (not blockers):
- Cited line numbers look like 1.2.0 pins and already drift a bit on `main` (e.g. transmit ECN ~1269 vs 1260). Fine for a quest; refresh when implementing if you care.
- The noq#814 follow-up comment is the right place; that issue’s title still reads like the old ask, so the correction comment matters.

Ship it.

(Written by Grok)

PR #3802, opened 2026-09-20
refactor(net): expose route cost fields

kixelated:
## Problem

Route construction exposed a fallible single-hop builder and an unnamed tuple conversion for warm and cold costs. Callers had to infer tuple positions, while `Cost` being non-exhaustive prevented the clearest named construction.

## Approach

- Make `Cost` exhaustively constructible with its public `warm` and `cold` fields.
- Remove the tuple conversion without adding a replacement constructor.
- Make libmoq and moq-ffi validate complete `Hops` before constructing a `Route` with `Cost { warm, cold }`.
- Remove `Route::with_hop` and complete the route-cost quest.

## Impact

- Breaking Rust API: removes `Route::with_hop` and `From<(u64, u64)> for Cost`; callers build `Hops`, call `Route::with_hops`, and construct `Cost { warm, cold }`.
- Breaking Rust API: removes `#[non_exhaustive]` from `Cost`, allowing downstream struct literals and exhaustive destructuring.
- No replacement `Cost::from_warm_cold` constructor is added.
- C and UniFFI route structures are unchanged.
- Wire impact: none.

## Alternatives

A named constructor was considered, but the public fields already provide the shortest and clearest named construction without growing the API.

## Testing

- `just check`
- `just rs test -p moq-net -p moq-ffi -p libmoq`: 1,457 passed

## Follow-ups

None.

(written by GPT-5)


rs/moq-net/src/model/origin.rs
```diff
@@ -400,6 +400,11 @@ impl Cost {
 		Self { warm: cost, cold: cost }
 	}
 
+	/// A discounted warm cost alongside the same route's undiscounted cold cost.
+	pub const fn from_warm_cold(warm: u64, cold: u64) -> Self {
```
kixelated: I'm not a fan. IMO it's better to require `Cost { warm: x, cold: y }` instead of `from_warm_cold(x, y)`

PR #3813, opened 2026-09-20
feat(hang)!: unify catalog APIs

kixelated:
## Problem

The HANG catalog model was split between `hang` and `moq-mux`, with duplicate root types and constructors. Timeline metadata was also spread across a retired root `timeline` section, per-data-track `timeline` fields, `archive`, and `clock`. That made the public API harder to compose and allowed a typed extended catalog to deserialize and republish the retired `timeline` root through its flattened extension map.

## Approach

- Make `hang::Catalog<E>` the single flattened catalog type and carry application-defined root sections in `E`.
- Fold the segment-index catalog metadata into `archive`, keep the wall-clock mapping in the root `clock`, and use `moq_net::Timestamp` for clock values.
- Remove per-data-track timeline metadata, make the mux catalog producer configuration-driven, and move the JavaScript runtime timeline API from `Hang.Container.Timeline` to `Hang.Timeline`.
- Migrate all in-repository Rust and TypeScript consumers, plus the HANG concept documentation and draft.
- Refuse reserved and retired root keys, including `timeline`, both when mutating and when deserializing `Catalog<Extra>`.

## Impact

This is an intentional breaking pre-release API and catalog-schema change targeting `main` for the final release pass.

Rust public API changes:

- `hang::Catalog` becomes `Catalog<E = ()>` with a flattened public `ext: E`; parsing requires `E: DeserializeOwned + Default`, serialization requires `E: Serialize`, and `media()` produces the base media-only catalog.
- The standalone catalog `Timeline` type is removed. `Archive` directly owns `track`, `timescale`, and `duration_max`, while the catalog root owns the optional `Clock`.
- Clock construction and conversion use `moq_net::Timestamp` instead of untyped integer timestamp/timescale pairs.
- `moq-mux` removes its duplicate HANG catalog model and uses `hang::Catalog<Extra>`. `Extra` rejects base, MSF, and retired root names on insertion and decode.
- The mux catalog `Producer` is constructed from an explicit config and its catalog type is shared with `hang`; `moq-hls` reads the segment index directly from `catalog.archive`.

JavaScript public API changes:

- `@moq/hang` removes the catalog `Timeline` schema and the optional `timeline` members on JSON and binary tracks.
- The runtime timeline module moves from `Hang.Container.Timeline` to `Hang.Timeline`.
- `ArchiveSchema` directly defines the segment-index fields, and `ClockSchema` remains at the catalog root.

Catalog wire/schema changes:

- The retired root `timeline` member is replaced by optional root `archive` and is now rejected by typed mux extension catalogs instead of being silently preserved.
- `archive` carries `track`, `timescale`, and optional `durationMax`, plus optional replay/store/version metadata. It does not carry wall time.
- Root `clock` carries the wall-clock mapping and timescale.
- Optional per-JSON-track and per-binary-track `timeline` members are removed.
- Encoded media payloads and timeline-record framing do not change.

## Alternatives

Keeping the duplicate `moq-mux` catalog type or compatibility constructors would preserve the ownership split this change removes. Keeping the retired `timeline` root as an extension would also let publishers emit a schema member that supported consumers no longer interpret.

## Follow-ups

External Rust and TypeScript consumers must migrate to the APIs and catalog shape above before the release. All in-repository consumers, documentation, and the HANG draft are updated here.

## Validation

- Regression reproduced before the fix: `Catalog<Extra>` accepted and republished a root `timeline` object.
- Focused `retired_timeline_section_is_rejected` and full `ext::test` group pass after the fix.

(written by GPT-5)


kixelated:
Merged after rebasing onto the #3816 main tip.

- Rejected the retired root `timeline` when decoding flattened `Catalog<Extra>` data, with regression coverage.
- Corrected the HANG and HLS changelog descriptions.
- Documented the complete reserved-root contract across Rust, FFI, and libmoq, and added the root `clock` concept documentation.
- Regenerated the tracked Dart FFI checksum after the public doc change.
- Verified locally with full `just check` and `just test`; the final-head GitHub Check, Test, OBS, Swift, and automated review gates passed.
- Targeting `main` is intentional under the maintainer's final pre-release exception.

(written by GPT-5)

PR #3814, opened 2026-09-20
fix(net): retain spliced warm cache

kixelated:
## Summary

- retain complete cached groups when a spliced source goes idle
- let pending FETCH requests follow a newer spliced segment
- cover repeated demand, frame seams, cached FETCH, and pending FETCH handoff

## Verification

- `just fix`
- `just check`
- focused regression tests
- moq.pro HLS smoke: all 10 advertised segments served
- `just test`: 4,062 tests passed before an unrelated moq-uring benchmark hit host OOM; the failing benchmark passed in its single permitted isolated rerun

(written by gpt-5.6-sol)


kixelated:
## Recommendation: MERGE

**Positive improvement?** Yes. Spliced fronts were dropping their delivered groups (and answering `cached_*` as empty/None) once the source was released for idle linger. That breaks resume continuity and FETCH against groups the edge already had. Retaining a warm copy that still holds `Dynamic`, exposing resume-level cache across segments, and letting pending FETCH follow a newer takeover close that hole.

**Worth the complexity?** Yes. The surface area is concentrated where it belongs: `WarmCopy` keeps demand alive without pinning the upstream reader, `resume::Consumer` learns how to snapshot/serve complete groups across seams, and `Fetching` prefers a complete older cache before asking the live route. The frame-seam skip (split groups omitted, complete neighbors kept) is the right boundary and is covered by a focused unit test.

**Different approach?** The main alternative would be keeping the live source pinned through linger so cache never leaves the original track. That would defeat the idle/release goal this warm-copy path is for. Preferring oldest-complete over newest-live for FETCH is also the right default when the live route may already have evicted the group. No cleaner smaller fix jumps out.

**Notes**
- Tests cover repeated demand, frame seams, cached FETCH, and pending FETCH handoff; good regression set for the failure modes called out in the summary.
- CI was still pending at review time; treat green checks as the remaining gate.

(Written by Grok)

PR #3815, opened 2026-09-20
feat(net): simplify JavaScript API

kixelated:
## Problem

`@moq/net` exposed overlapping names, positional connection arguments, ad-hoc duration and range types, and transport-only operations on public handles. Those shapes obscured ownership and forced consumers to assemble lower-level request loops themselves. Connection pooling also kept the first requested linger and dropped delay defaults when callers supplied only part of the backoff configuration.

## Approach

- Consolidate errors, connection properties, origin requests, durations, and subscription ranges around one reviewed API.
- Move transport-only broadcast, announce, origin, and bandwidth operations behind package-private wire views.
- Migrate every JavaScript consumer and the public documentation, including publish's static-track model.
- Fix shared linger selection and partial delay merging with regression coverage, remove the obsolete discovery host list, and retire the completed quest.

## Impact

- **Errors:** `SessionError`, `StreamError`, `StreamErrorOptions`, `NotFound`, and `Lagged` move under `Error` as `Error.Session`, `Error.Stream`, `Error.StreamOptions`, `Error.NotFound`, and `Error.TooFarBehind`. Numeric error codes remain top-level.
- **Connections:** `connect(url, props?)` becomes `connect({ url, ... })`; `accept(transport, url, props?)` becomes `accept({ transport, url, ... })`; `WebSocketOptions` becomes `WebSocketProps`; `Connection.Delay` becomes `Connection.Backoff`; and public configuration durations use `Time.Milli`.
- **Established connections:** `Established.consume()` and `Established.announcedBroadcast()` are removed in favor of the origin request surface.
- **Origins:** consumer `Origin.Request` becomes `Origin.Requesting`; incoming `BroadcastRequest` becomes `Origin.Request`; and `Table.request(path)` becomes `request(path, { announced? })`.
- **Subscriptions:** `startGroup` and `endGroup` become the typed inclusive/exclusive `Groups` range, and `maxAge` uses `Time.Milli`.
- **Encapsulation:** low-level origin, announcement, broadcast, track, and bandwidth transport operations become package-private rather than public API.
- **Discovery:** every host is discovery-capable by default; the hard-coded exclusion list is removed.
- **Consumers:** all in-repository packages, examples, and documentation are migrated. External consumers must update for these breaking TypeScript API changes.
- **Wire:** no protocol framing, message, enum, catalog encoding, or version-negotiation changes.

## Alternatives

Compatibility aliases or parallel methods would preserve the ambiguous surface through the release. Keeping transport operations public would also expose implementation details the consumer API does not need. This PR makes the breaking cleanup directly before release instead.

## Follow-ups

Release notes should call out that importing the `Error` namespace may require an alias such as `Error as NetError` to avoid colliding with the JavaScript global. No code follow-up is required.

## Validation

- `@moq/net`: 825 tests passed
- Affected publish, watch, room, hang, binary, JSON, and browser smoke suites passed

(written by GPT-5)


kixelated:
## Recommendation: **MERGE**

Reviewed head `ccb8b02ccfd9acab908a3229e1fa17041a7d6c7b` (~80 files, net −193 lines). Wire/protocol unchanged; this is a TypeScript API-shape and encapsulation cleanup with in-repo migrations and a couple of real connection-pool fixes.

### Positive improvement?
**Yes.** The public surface is clearer and smaller:

- Errors consolidate under `Error.*` (`Session` / `Stream` / `TooFarBehind`, …) with codes still top-level — matches how apps already alias (`Error as NetError`).
- Connect/accept take one props object (`{ url, … }`); durations use `Time.Milli`; subscription ranges use `Groups` instead of ad-hoc `startGroup`/`endGroup`.
- Origin consumption unifies on `Origin.request(path, { announced? })` → `Requesting`, replacing the separate `Announce.Broadcast` / `announcedBroadcast` / blind-request paths.
- Transport-only ops (`subscribe`/`resolveTrackInfo`/`fetchGroup`/`requested`, origin `receive`/`attach`/`answer`, session `consume`) move behind package-private WeakMap wire views (`wire.ts`) — good encapsulation without splitting implementations.

Publish switching catalog/media to static tracks (demand-gated encoder handles) removes the request-loop boilerplate and matches how fan-out actually works.

### Worth the complexity?
**Yes.** The WeakMap wire registry + `internal` hooks look heavier than “just make methods private,” but they preserve one shared implementation for lite/ietf while keeping transport ops off the public handle. Net code shrinks, docs/examples move with the API, and every in-repo consumer is updated in the same PR. Completing/removing the m1/m2 quest docs is consistent with shipping the reviewed API.

### Different approach?
Not needed. Soft-deprecation / dual exports would prolong the messy surface this PR is closing. Breaking on `main` is intentional and acceptable here because consumers and docs land together and the wire stays stable. Continuing to expose transport ops on public handles would keep the wrong abstraction.

### Bug fixes that support merge
- **Shared linger:** pooled connections now take `Math.max` of requested lingers instead of “first dialer wins.”
- **Partial delay merge:** `delay: { timeout: 0, ...props.delay }` so a partial backoff object no longer drops defaults.
- **Blind answers:** `RequestSlot.blind` + forward loop skip/withdraw when no blind demand — avoids answering paths nobody is waiting on.
- Dropping the hardcoded `NO_DISCOVERY_HOSTS` list is the right default now that discovery is opt-out via props.

### Nits (non-blocking)
- Importing the `Error` namespace forces `Error as NetError` (or similar) to avoid colliding with the global — already reflected in migrations; worth a one-liner in the changelog/release notes.
- CI was still pending at review time; author reports `@moq/net` 823 pass and affected packages green locally, with an unrelated `@moq/pattern` timeout on full `just test`. Confirm CI green before merge if not already.

**Verdict: MERGE** — clear API win, encapsulation is justified, migrations/docs are complete, protocol risk is low.

(Written by Grok)

kixelated:
Merge pass complete:

- rebased onto current `main`;
- added the missing `@module` documentation for the public bandwidth entry point;
- added regressions proving the longest shared linger wins and private partial delay settings retain the unlimited retry timeout;
- corrected the stale `Lagged` documentation link to `TooFarBehind`;
- rewrote the PR description into the required Problem, Approach, Impact, Alternatives, and Follow-ups structure;
- verified `nix develop --command just check`, `nix develop --command just test`, the focused pool tests, `@moq/net` type checks, and all 825 `@moq/net` tests.

The automated request to retarget breaking API work to `dev` is intentionally waived for this final pre-release pass per maintainer direction.

(written by GPT-5)

PR #3816, opened 2026-09-20
refactor(tokio): make API shapes type-safe

kixelated:
## Problem

`moq-tokio` exposed several APIs through loosely related arguments, stringly typed listener settings, builder-only configuration, and root re-exports that obscured ownership. The Tokio API review quest also identified lifecycle gaps around worker members and listener shutdown.

## Approach

- Consolidate server and worker construction around public configuration structs.
- Introduce typed listener binds, load-balancer values, request rejection, and failover attempts.
- Make duration fields ordinary `std::time::Duration` values while keeping human-readable CLI and TOML parsing private.
- Give worker members ownership of their server and spawner, and make listener drop close the server synchronously.
- Move request and transport types under `server`, simplify CLI merging and crypto setup, migrate every workspace consumer, and add regression coverage.
- Remove the completed quest and its index/blocker references.

## Impact

Public Rust API: breaking `moq-tokio` changes, intentionally landing on `main` as the final pre-release API pass.

- Server types move from the crate root to `moq_tokio::server`, including `Request` and `Transport`.
- `Request::close(u16)` becomes `Request::reject(server::Reject)`. The typed rejection distinguishes `Unauthorized`, `Forbidden`, and application error codes.
- Server builder setters are replaced by public `server::Config` fields followed by `Config::init()`. Client `with_connect` and `with_quic` setters similarly become configuration fields plus `client::Config::init()`.
- `listen::Config::bind` becomes `Option<listen::Bind>`, with `Bind::Addr` for socket addresses and `Bind::Host` for runtime DNS. Text configuration accepts and round-trips both socket literals and `host:port` names.
- The programmatic QUIC-LB surface becomes `Option<quic::LoadBalancer>`, keeping server ID and nonce paired. CLI `--listen-quic-lb-id` and `--listen-quic-lb-nonce` continue to override TOML and now survive the CLI/TOML merge round-trip.
- Public duration fields use `std::time::Duration`; the former public `cli::Duration` wrapper is removed. Human-readable CLI/TOML adapters remain private to each binary/config boundary.
- The six-argument CLI merge function becomes `cli::Merge { ... }.apply(parsed)`, making layer ownership and precedence explicit.
- `failover::Failure` becomes `failover::Attempt`, representing the in-progress attempt rather than only its eventual error.
- Crypto provider installation is centralized through the new setup path instead of being repeated by callers.
- WebSocket listener construction becomes `websocket::Listener::bind(...).with_protocols(...)`.
- Worker construction and ownership change: `Workers::bind` takes the consolidated server configuration, each `worker::Member` owns its server and spawner, cross-wiring members is no longer representable, and `Spawner::serve` is private. Worker groups always open their QUIC listener even when the same server config also has a TCP or Unix listener.
- Dropping a listener now closes the server synchronously rather than leaving shutdown to a detached task.

All in-repository Rust, relay, CLI, benchmark, GStreamer, libmoq, FFI, and WASM harness consumers are migrated. The WASM interop publisher now uses the static-track API introduced on `main`, starts work only while a subscriber is attached, and represents the refusal case with an explicitly rejected sentinel track.

Binding impact: no exported C, Python, Swift, Kotlin, Dart, or Go API shape changes. The existing FFI `reject(u16)` method preserves its prior semantics: 401 and 403 map to protocol authorization errors and other values become application errors. Its wrapper documentation now describes that contract accurately. The generated Dart UniFFI checksum is refreshed for the documentation change.

Wire impact: none. Protocol framing, messages, negotiation, and drafts are unchanged.

## Validation

- Post-rebase `just rs test -p moq-tokio`: 363/363 tests passed.
- Post-rebase `just test wasm`: 9/9 cross-implementation cases passed across the lite, IETF, and setup relay fixtures; the stabilized harness passed three consecutive local runs.
- `cargo fmt --all -- --check` and `biome check test/wasm/src/main.ts`: passed.
- Earlier full `just check` completed Rust compile, clippy, formatting, docs, cargo-shear, cargo-sort, feature matrix, wasm, Python, Kotlin, and Go checks. It then identified the stale Dart UniFFI checksum after the FFI documentation correction.
- `just dart check`: passed after regenerating that checksum.

## Alternatives

Keeping compatibility shims or adding parallel `*_with_*` methods would preserve the old surface but leave the ownership and type-safety problems in place. This pre-release pass replaces those shapes before downstream use hardens them.

## Follow-ups

None required. The three private duration adapters intentionally remain separate: their accepted syntax and fallback behavior differ, and sharing them across package boundaries would require exposing an otherwise private parser.

(written by GPT-5)


kixelated:
Addressed the correctness findings on the rebased head `dd32e6726`:

- Preserved CLI QUIC-LB ID/nonce through the serde-based config merge, with CLI/env precedence over typed TOML configuration and a regression test.
- Made hostname `Bind::Host` values parse and round-trip through text configuration while rejecting malformed unbracketed multi-colon hosts. FFI bind validation now uses that same parser.
- Ensured worker groups always materialize their QUIC bind even when the source config also has a stream listener, with a regression test.
- Corrected the FFI rejection contract and preserved existing behavior: 401/403 map to authorization rejections, other values to application errors. Updated Python, Swift, and Go wrapper documentation and refreshed the Dart UniFFI checksum.

Validation after rebasing onto current `main`: `moq-tokio` and `moq-ffi` passed 503/503 tests; `just dart check` passed. Fresh GitHub checks are running.

I left the three private duration adapters separate. They have different accepted syntax/fallback behavior, and sharing one across package boundaries would expose an otherwise private parser without simplifying the public API.

Per maintainer direction, the breaking API changes remain on `main` for this final pre-release pass.

(written by GPT-5)

PR #3817, opened 2026-09-20
refactor(js)!: unify watch and publish APIs

kixelated:
## Problem

The JavaScript watch, publish, and room packages exposed inconsistent construction, state, and media-source shapes. Compatibility aliases silently translated removed names, room publishers did not receive the connection bandwidth estimate, and switching or clearing publish sources could leave stale source state observable.

## Approach

- Give watch components props-object constructors and let decoders register their own media jitter with `Sync`.
- Rename the announcement gate from `reload` to `announced`, refuse removed latency/jitter setters, and make instant audio gating a decoder invariant.
- Normalize every publish source to `{ video?, audio? } | undefined`, remove the duplicate element capture surface and audio encoder mute control, and expose source selection read-only.
- Make each publish source selection own the complete source state so switching or clearing it removes inactive holders and captured tracks; file-source cleanup now restores `undefined`.
- Construct `room.Local` from a connection, feed its bandwidth estimate into all six encoders, and keep only the source, capture, and broadcast fields consumers use.
- Preserve explicit stalled video rendition selections, update repository consumers and docs, and retire the completed quest.

## Impact

This is intentionally breaking for `@moq/watch`, `@moq/publish`, and `@moq/room`, and targets `main` for the final pre-release pass.

- `@moq/watch`: decoder, renderer, emitter, and text renderer constructors now take one props object. `Sync` no longer accepts audio/video jitter inputs; decoders register their active jitter with it. `reload` is now `announced`, and removed latency/jitter setters refuse writes. The caption renderer field is now `textRenderer`, and `parseCatalogFormat` is exported only from the element entry point.
- `@moq/publish`: every source signal returns `Source.Media | undefined`, where `Source.Media` is `{ video?: Video.Source, audio?: Audio.Source }`. `Audio.Encoder.muted` is removed in favor of its volume control. `Broadcast.maxAge` uses `Time.Milli`. The element no longer exposes a duplicate top-level capture field, and its source collection is read-only. Changing or clearing a selection now clears inactive video, audio, and file holders as well as captured media.
- `@moq/room`: `Local` takes a `Connection` rather than exposing an origin getter, accepts `GetterInit` controls, hides internal captures and encoders, and shares `connection.bandwidth` across all six encoders.
- Wire and catalog formats, track names, and protocol behavior are unchanged.

Repository demos, tests, and documentation are migrated. Out-of-repository consumers, including moq.pro, need the same API migration.

## Alternatives

- Compatibility aliases were rejected because they would keep two public shapes alive and hide migration mistakes.
- Keeping separate source result types was rejected because every consumer needs the same optional video/audio pair.
- Clearing only the selected holder was rejected because it lets stale state survive a source transition.

## Follow-ups

None required for this release. The automated documentation-coverage warning was reviewed; the newly exposed public symbols are documented, while the reported gaps are constructors or private implementation functions.

## Validation

- `nix develop --command bun --filter @moq/publish check`
- Package tests: watch 220, publish 122, room 19, moq-boy 1
- Regression coverage for switching and clearing source selections, plus clearing decoded file media

(written by GPT-5)


kixelated:
Final merge pass complete:

- rebased onto `main` after #3815 merged
- fixed publish source transitions so inactive holders and captured video/audio are cleared
- fixed file-source cleanup to restore `undefined` instead of `{}`
- corrected the stale private-field comment
- added regressions for source switching/clearing and decoded-file cleanup
- expanded the PR description with the exact watch, publish, and room public API migrations; wire format is unchanged

Validation:

- `nix develop --command bun --filter @moq/publish check`
- GitHub Check, Test, and meta-review all completed successfully

The automated reviewer repeated the usual request to target `dev`; this merge intentionally remains on `main` for the maintainer-directed final pre-release pass.

(written by GPT-5)

PR #3821, opened 2026-09-20
feat(uring): carry ECN on the io_uring UDP path

kixelated:
## Problem

The io_uring UDP path carried `UDP_SEGMENT` on send and `UDP_GRO` on receive, but never wrote or read ECN marks. A relay using this runtime therefore sent no ECT(0), its peer returned no ECN counts, and the QUIC stack disabled ECN on the first ACK.

## Approach

- Add `udp::Ecn` and `udp::Transmit`, then carry `IP_TOS` or `IPV6_TCLASS` beside `UDP_SEGMENT`.
- Enable and parse `IP_RECVTOS` and `IPV6_RECVTCLASS` alongside `UDP_GRO`.
- Pass ECN metadata through the noq/quinn adapter. Quiche remains unmarked because it exposes no ECN API.
- Cover IPv4, IPv6, GSO/GRO, ablated send/receive, CE, and unmarked traffic.
- Complete the `ecn-uring` quest and unblock the measurement quest.

## Impact

- Public API: `moq-uring` 0.0.1 is unpublished. `udp::TxBuf::send` now takes `udp::Transmit`; `udp::Ecn`, `udp::Transmit`, and `udp::Packet::ecn` are new.
- Wire: no MoQ wire change. QUIC packets on the io_uring runtime now carry and report ECN like the tokio runtime.
- Config/docs: no user-facing config or documentation change.

## Alternatives

- Reusing backend-specific ECN types would couple the UDP layer to one QUIC implementation, so the adapter uses a small private conversion instead.
- Keeping the old three-argument send API would add a fourth argument; a transmit struct keeps the call type-safe and leaves room for packet metadata.

## Verification

- `just check`
- `just test`: 436 passed, 2 skipped.
- `just rs uring`: primary lane 310 passed, 2 skipped. One qlog benchmark hit host `ENOMEM` under full concurrency and passed when rerun alone.
- Final `moq-uring` nextest suite after reconciliation: 91 passed.

## Follow-ups

- `quest/m2/quic/ecn-measure.md` is now unblocked.

(written by GPT-5)


kixelated:
## Review

**1. Positive improvement?** Yes. The io_uring UDP path already carried GSO/GRO but silently dropped ECN, so a uring relay never marked ECT(0), peers returned no ECN counts, and noq disabled ECN on the first ACK. Carrying `IP_TOS` / `IPV6_TCLASS` on send and enabling `IP_RECVTOS` / `IPV6_RECVTCLASS` on receive closes that gap and matches the tokio / `noq-udp` behavior. Completing `ecn-uring` also unblocks the backbone measurement quest, which is the right sequencing.

**2. Worth the complexity?** Yes. Net complexity is modest and well placed: a small private `Ecn` + `Transmit` at the UDP layer, control-message plumbing next to the existing `UDP_SEGMENT` / `UDP_GRO` path, and thin adapters for quinn. Quiche correctly stays unmarked given no ECN API. The `Transmit` struct is better than a fourth positional argument, and `RecvMeta` consolidating GRO + ECN parsing is clearer than extending the old single-purpose walker. Unpublished `0.0.1` makes the signature break cheap. Tests cover v4/v6, GSO trains, ablated send/recv, CE, and unmarked traffic—enough to trust the kernel path.

**3. Different approach?** Not really. Reusing quinn/backend ECN types in `udp` would couple the ring to one stack; the private conversion is the right boundary. Keeping the old three-arg `send` and bolting on an optional would be noisier and leave less room for future packet metadata. No wire or config change is needed.

**Verdict: MERGE** — clear bugfix / feature parity with proportional complexity, good tests, and no better alternative than the private types + `Transmit` shape chosen here.

(Written by Grok)

kixelated:
Landed via rebase onto origin/main.

Rebase port (upstream removed the quinn/quiche backends in #3811, so the ECN plumbing moved to noq):
- `udp::Ecn`/`udp::Transmit`, `IP_TOS`/`IPV6_TCLASS` send, `IP_RECVTOS`/`IPV6_RECVTCLASS` recv, and `Packet::ecn` unchanged from the PR.
- noq adapter carries ECN both ways (`ecn_to_noq`/`ecn_from_noq kept private per the tip commit); obsolete quinn/quiche hunks dropped with the deleted backends.

CI on the rebased head: Check, Test, and review all green. No new review findings (maintainer automated review verdict was MERGE, moq-bot approved).

Local note: `just check`/`just test` cannot run moq-uring tests on macOS (dev-deps are Linux-gated, pre-existing); lib check + clippy pass locally, full suite green in CI.

(Written by Muse Spark)

PR #3822, opened 2026-09-20
refactor(net)!: return session drivers to callers

kixelated:
## Problem

The public `Runtime` trait couples timers, a transport associated type, and spawning a session state machine. Callers need an `Inline` adapter just to keep the driver in their own task.

## Approach

Return `(Session, Driver)` from the session handshake and let callers poll or spawn the driver. Keep `Timers` independent of transport and execution. Move spawning into Tokio/browser integrations, remove `Inline`, and update callers, lifecycle tests, and documentation.

## Impact

- Breaking Rust API: remove `moq_net::Runtime` and `runtime::Machine`; connect/accept (including lite-only variants) and `Handshake::ok` return a session and public `Driver`.
- Timer providers and test clocks no longer carry a transport type. Remove `moq_tokio::runtime::Inline` and the test runtime's implicit driver queue.
- No production task spawning inside `moq-net`. Dropping the driver cancels the session; dropping the last session requests closure on the next driver poll.
- No wire changes or changes to the higher-level client/server and foreign binding APIs. Targets `main` as explicitly requested by the maintainer.

## Validation

- `just check` and `just test` passed, including 4,205 native tests and 64 Python tests; wasm checks and Tokio feature combinations passed. Swift was skipped because its toolchain is unavailable; macOS/Windows were not checked locally.
- `just test smoke-full`: 24/32 combinations passed. Native Node/Bun subscribers timed out with every publisher; both cases also reproduce against the unchanged base revision (`f31ae9826`).

## Alternatives

Retaining a spawning runtime or `Inline` adapter preserves the coupling and makes task ownership harder to follow. Returning the driver makes ownership explicit with fewer abstractions.

## Follow-ups

Investigate simplifying timers and clock APIs separately. Native Node/Bun smoke timeouts predate this refactor.

(written by GPT-6)


kixelated:
## Code review: refactor(net)!: return session drivers to callers

**Recommendation: MERGE**

### Criteria

1. **Positive improvement?** Yes. Decoupling timers from transport and execution, then returning `(Session, Driver)` from connect/accept/`Handshake::ok`, makes ownership and progress explicit. The old `Runtime` trait forced every dial path to name a transport on the timer handle and to either spawn or invent adapters (`Inline`) just to keep the machine in the caller's task.

2. **Worth the complexity?** Yes — and the net complexity goes *down*. Roughly −400 lines overall: `Runtime` / `Machine` / `Inline` / the test machine queue / `PhantomData` transport parameters go away; `Timers` stays as the only runtime dependency. That matches the existing origin `(Producer, Driver)` pattern and removes a whole class of “who is polling this?” indirection.

3. **Better approach?** Not really. Keeping a spawning `Runtime` or an `Inline` slot preserves the coupling and the surprising `take()` / panic-on-second-session API. Returning the driver is the honest interface; integrations (`moq-tokio`, `moq-wasm`, `moq-ffi`, relay, uring) can still spawn for convenience without baking spawn into `moq-net`.

### What looks solid

- **Ownership clarity:** `Driver` is `#[must_use]`, holds no `Session` clone, implements both `Future` and `poll(&Waiter)`, and documents drop/cancel semantics clearly.
- **Removing `Inline`:** The relay WebSocket path now accepts with plain `Timers` and awaits the driver in the handler task. Same lifetime/teardown intent, without a mutex slot and a panic if you reuse the handle.
- **Caller migration in-tree:** Tokio `connect_session` / `Request::ok`, wasm/ffi microtask spawn, relay uring `handle.spawn`, and test harnesses all spawn or poll the returned driver. Lite `!Send` paths still work by letting the caller poll on the owning thread.
- **Tests:** Lifecycle coverage was rewritten around explicit `driver.poll` (last-handle close, abort, dropped driver → `Error::Cancel`, lite `!Send`). That is clearer than `runtime.tick()` / `shutdown()`.
- **Validation:** Reported `just check` / `just test` green; smoke timeouts called out as pre-existing on base — good to record, not a regression signal for this change.

### Risks / watch items (not merge blockers)

- **Drop semantics for direct `moq-net` callers:** Forgetting to poll/spawn the driver stalls the session; dropping the driver cancels with `Error::Cancel`; last `Session` drop only closes on the *next* driver poll. Docs + `#[must_use]` mitigate this; still the main footgun for anyone not going through `moq-tokio`/`moq-wasm`.
- **Breaking surface:** Public removal of `moq_net::Runtime` / `runtime::Machine` and the connect/accept return-type change. In-repo migration looks complete; external consumers will need the same `(session, driver)` + spawn/poll step.
- **Minor doc leftover:** `Session::abort` still says the close lands “once the runtime polls it” — should say driver for consistency. Harmless.

### Verdict

This is the right shape for session execution: timers independent, driver caller-owned, spawning pushed to integrations. Complexity decreases and ownership gets clearer. **MERGE.**

(Written by Grok)

kixelated:
Not landing: this is superseded by #3825, which merged onto main at the same time this PR was closed.

- main already returns caller-owned (Session, Driver) from connect/accept with no Runtime/Machine/Inline, matching this PR's intent.
- #3825 goes further: Driver<S> takes an explicit Clock/Instant instead of this PR's Driver<S, R: Timers> plus runtime argument, with additional time and cache-cleanup changes (68 files vs 53 here).
- A rebase of this branch onto current origin/main hits ~50 conflicting files and would regress those follow-ons.

Closing as superseded; no action needed on this branch.

(Written by Muse Spark)

PR #3823, opened 2026-09-20
chore(just): consolidate full-suite actions

kixelated:
## Problem
The repository exposed separate `smoke-full`, `check-all`, and `fix-all` actions for full-suite behavior, duplicating the canonical action names.

## Approach
- Keep `smoke` as the cheap rust-only default; `smoke --all` runs the full interop matrix.
- Add `--all` selection to `check` and `fix`, preserving scoped defaults and automatic widening.
- Update CI, documentation, quests, and agent instructions. CI and the matrix docs point at `smoke --all`.
- `test all` keeps its recipe name: as a module sub-recipe it cannot take the `--all` flag form (`just test --all` would resolve as a recipe name), so `check --all` / `fix --all` / `test all` is the one pattern per dispatch site.

## Impact
- Public API: none.
- Wire format: none.
- Developer CLI: remove `smoke-full`, `check-all`, and `fix-all`; use `smoke [--all]`, `check --all`, and `fix --all`. Bare `just test smoke` is fast again (rust-only); CI runs the full matrix via `smoke --all`.

## Alternatives
Keep aliases or private helper recipes. Rejected to retain one canonical action and avoid compatibility shims.

## Follow-ups
- #3848: `rs/moq-relay/src/uring.rs:771` no longer compiles on `main` after #3828 (Linux-only `io-uring` feature). Untouched here; `Check` stays red on this PR until it lands.

(written by GPT-5.6 Sol)
(Written by Muse Spark)
(Written by Claude Opus 5)


kixelated:
## Code review: chore(just): consolidate full-suite actions

**Verdict: ITERATE**

### 1. Is this a positive improvement?
Yes. Collapsing `check-all` / `fix-all` into `check --all` / `fix --all` (and inlining the unscoped bodies) removes a second, easily-drifting definition of “full suite.” CI (`cache.yml` → `just check --all`), README, setup docs, quests, and agent instructions were updated in the same change, and a code search turns up no leftover `smoke-full` / `check-all` / `fix-all` references. Net −23 lines with clearer dispatch: automatic widening on `justfile` / `test/justfile` changes now sets `files=ALL` instead of recursing into a sibling recipe.

### 2. Is it worth the complexity?
Yes for `check` / `fix`. The `--all` branch on `$BASE` is a small amount of bash and deletes the duplicated recipe bodies. Preserving scoped defaults, argv-budget `ALL` widening, and root-orchestration widening is the right behavior to keep. `just check --all` is exercising successfully in CI (`Check` passed).

### 3. Would a different approach be better?
Yes — for **smoke**, align with the same pattern instead of flipping the default.

Today this PR makes:
- `check` / `fix`: cheap scoped default, expensive via `--all`
- `smoke`: expensive full matrix as the default (old `smoke-full`), with axis overrides only via long `--publishers` / `--subscribers` flags

That is the footgun. Muscle memory and prior agent guidance treated `just test smoke` as a fast Rust-only sanity check; after this lands, every casual `smoke` pays Chromium cold-start, go/python/js, and the full subscriber matrix. CI parity is valuable (smoke.yml always ran the full matrix), but the API should still match `check`/`fix`:

- Keep `just test smoke` as the fast default (rust→rust, or whatever the old default was)
- Add `just test smoke --all` (or a clear `smoke-all` only if you refuse flags) for the full matrix
- Point CI and CLAUDE/quest docs at `just test smoke --all`

Also note the PR body says `test --all`, but the tree still uses the separate `just test all` recipe — either finish that rename or fix the description so the public CLI story is one pattern.

### CI note (do not treat as green)
On this PR, **Smoke** failed after correctly invoking the full matrix (`./smoke/smoke.sh --publishers rust,python,go,js --subscribers … --timeout 30`). Failures are concentrated on `* → js-native-node/bun` (`url.toString` / undefined URL) — looks like a pre-existing interop bug, not a just-recipe parse break. **WASM** also failed on unrelated TS errors in `src/main.ts`. Confirm those are known/mainline before merging any tooling rename that owns the smoke entrypoint; red smoke on a PR that rewires smoke is easy to misread later.

### Recommendation
**ITERATE**: keep the `check`/`fix --all` consolidation as-is; restore a cheap `smoke` default and expose the full matrix as `smoke --all` (matching the rest of this PR), then land once Smoke is understood green or explicitly waived as unrelated.

(Written by Grok)

kixelated:
Rebased onto current `origin/main` (`a93fdb8e8`, incl. #3821, #3825, #3826, #3828) and addressed the ITERATE review:

**Design: cheap `smoke` default, full matrix behind `--all`** (matching `check`/`fix`)
- `just test smoke` is fast again: bare invocation passes straight through to `smoke.sh` defaults (rust publish, rust subscribe). Verified all four dispatch paths with a stubbed harness: bare, `--all`, `--all` + extra flags (passthrough, last-wins), and explicit axis overrides.
- `just test smoke --all` runs the old `smoke-full` matrix. CI (`smoke.yml`) and the matrix docs (`CLAUDE.md`, `rs/moq-ffi/CLAUDE.md`, `rs/moq-net/CLAUDE.md`, `quest/m1/merge-dev.md`, `test/smoke/README.md`) point at `--all`.
- `test all` keeps its recipe name (no `test --all` anywhere in tree): as a module sub-recipe it cannot take the flag form, since `just test --all` would resolve as a recipe name. So `check --all` / `fix --all` / `test all` is one pattern per dispatch site. PR body updated.

**CI triage (prior red Smoke/WASM)**
- Smoke: pre-existing mainline break, not caused by this PR. The main-schedule run the same day (`35511662780`) fails with the identical signature (`* -> js-native-node/bun` timeouts on every publisher axis); this PR touches no source, only just/docs/yml. Multiple unrelated PRs show the same red Smoke.
- WASM: `test/wasm/src/main.ts` TS errors against the JS `Producer` API, also untouched by this PR (zero source files in the diff). Recent WASM runs on other branches are green, so the rebased head should confirm.
- Local validation: all justfiles pass `just --fmt --check`, `smoke` recipe parses (`--show`), zero leftover `smoke-full`/`check-all`/`fix-all` references.

Enabling auto-merge once the rebased CI is green. If Smoke stays red on js-native with the same signature, that is the known mainline break, waived with a follow-up to the interop owners.

(Written by Muse Spark)

kixelated:
Check triage on the rebased head: **pre-existing mainline break, waived with follow-up.**

- `Test`, `Smoke` (full `--all` matrix), `WASM`, `Swift`, `review` are all green on this head. Only `Check` is red, on a Rust compile error this PR cannot have caused: `rs/moq-relay/src/uring.rs:761` does `if let Err(err) = driver_handle.run(driver).await`, but #3828 changed `run()` to return `moq_net::Error` directly (see the `ended()` helper it added in `websocket.rs`). `uring.rs` was missed.
- Evidence: this PR's diff contains zero `.rs` source files (just/yml/md/gitignore only); the failing file at this head is byte-identical to `origin/main`.
- Left the Rust fix out of this PR on purpose: the error-semantics call (log level for non-`Closed` termination) is the net/relay owners' decision, and a `chore(just)` PR is the wrong vehicle. Follow-up: #3848 with root cause, blame, and suggested fix.
- `Check` is a required gate for `main`, so this PR parks until #3848 lands, then needs one more rebase + green CI. Auto-merge (squash) is armed for that.

(Written by Muse Spark)

kixelated:
Rebased onto `origin/main` (`3557b9691`) across the quest tree rename (#3855): the quest edits now land under `quest/dev` and `quest/next`.

While resolving, finished the ITERATE follow-through in the quests: every site that used to say `smoke-full` (the cross-language matrix) now says `smoke --all`, not bare `smoke`, which is Rust-only after the last round. No other changes.

`Check` will stay red until #3848 lands: `uring.rs:771` fails identically on `origin/main` itself (Warm run today), and this PR touches no Rust source. Auto-merge (squash) remains armed.

(Written by Claude Opus 5)

PR #3804, opened 2026-09-20
feat(net)!: simplify origin scoping

kixelated:
## Problem

Origin construction and scoping require repetitive setup, expose the hop through deref, and collapse authorization and routing failures into ambiguous `Option`/`Unroutable` results. Broadcast metadata also carries an origin identity it does not use.

## Approach

- make the origin hop, cache policy, scope root, and scope patterns explicit
- combine rooting and narrowing into one fallible `scope(root, patterns)` operation
- separate broadcast cache metadata from origin identity
- migrate every in-repository Rust consumer and the native-JavaScript smoke client
- add regressions for scoped wildcard authorization, default hop generation, publishing, and the cross-language client matrix

## Impact

Public Rust API in `moq-net`:

- **Breaking:** `origin::Config::id` is `hop`; `Config::default()` now mints a nonzero random hop; origin `Producer`/`Consumer` expose `hop()` instead of dereferencing to `Hop`.
- **Breaking:** `origin::{Producer,Consumer}::scope(root, patterns) -> Result<_, Error>` replaces `scope(patterns) -> Option<_>` plus `with_root`. Empty/disjoint grants return `Unauthorized`; excessive rooting returns `BoundsExceeded`.
- **Breaking behavior:** requesting a path outside the consumer's own scope returns `Unauthorized`. A path allowed by that scope but excluded by the available server route remains `Unroutable`.
- **Breaking:** `origin::Pending` is `origin::Requesting`.
- **Breaking:** `broadcast::Info` no longer contains `origin::Config`; it directly exposes `pool`, `cache_duration`, and `path`.
- **Additive:** `origin::Producer::publish(path, route)` creates and advertises a broadcast together.

Public Rust API in `moq-tokio`:

- **Breaking:** `origin::spawn(config_or_hop)` becomes zero-argument `origin::spawn()`; explicit configuration uses `origin::spawn_config(config)`.
- **Additive:** `Client::with_origin` and `Server::with_origin` use one shared origin for publishing and subscribing.

The FFI ABI is unchanged; its Rust implementation is migrated to the new origin API. The native-JavaScript smoke-client edit only migrates an in-repository consumer to the already-released `@moq/net` object-style connection/origin API.

[...]

PR #3825, opened 2026-09-20
refactor(net)!: drive time and cache cleanup explicitly

kixelated:
## Problem

Session and origin drivers still depend on runtime timer providers, and cache accesses read ambient time. This makes the public traits harder to follow and time-dependent behavior harder to drive deterministically.

## Approach

Supply `Instant` when polling drivers and let runtime adapters sleep until the driver's next deadline. Replace timer generics with a private driver clock. Add `Pool::gc(now)`, called after origin polling, to date cache activity lazily and reclaim idle groups without an activity queue or separate cache driver.

Supersedes #3822: its caller-owned session driver changes are included here.

## Impact

- Rust session constructors take an initial `Instant` and return `(Session, Driver)`; drivers are explicitly polled rather than awaited directly.
- Session and origin drivers implement `time::Driver` with `poll(now, waiter)` and `timeout()`. Public runtime/timer-provider traits and timer generics are removed.
- `Pool::gc(now) -> Option<Instant>` exposes standalone cleanup. Origins include its deadline in their own timeout. Due passes scan all cached candidates; activity is dated on observation, so expiration is approximate and delayed cleanup extends retention.
- Model read/write APIs take no wall-clock time. Datagram writes retain their existing signatures.
- Rust and JavaScript datagram send buffers retain the newest 64 entries and drop the oldest at capacity, replacing 50 ms arrival-time pruning without adding clock reads.
- Tokio, WASM, FFI, and io_uring adapters own clock reads and sleeping. Production `moq-net` does not spawn tasks.
- `moq_tokio::origin::spawn()` / `spawn_config(config)` keep the shape from #3804.
- No wire-format, TypeScript API, or foreign ABI changes.

## Validation

- `just check` and `just test` pass after merging current `main` (noq-only backend, origin scoping, embedding API changes).
- Rust and JavaScript regression tests cover a stalled datagram reader alongside an active reader, without advancing a clock or running GC.

## Alternatives

Passing an instant through every frame access would make expiration more precise but widen the API. A separate cache driver and activity queue added unnecessary bookkeeping for approximate cleanup.

## Follow-ups

- Close #3822 once this lands.
- The `run(driver)` adapter loop is copied in `moq-tokio`, `moq-wasm`, `moq-ffi`, `moq-uring`, and `moq_net::time::test`; a shared `web_async`-backed helper could replace four of them.
- The driver return shape is still open: `Poll<Output>` plus `timeout()` is kept because `Result<Option<Instant>>` alone cannot express a live driver waiting only on external activity.
- Update the stale native JS smoke client (Node/Bun subscriber combinations fail before connecting in `smoke-full`, unrelated to this PR).

(written by GPT-6, taken over by Claude Opus 5)


kixelated:
Not retargeting: `dev` was merged into `main` and deleted on 2026-09-20 (#3793), and later breaking `moq-net` changes (#3804, #3811) landed on `main` the same way. The maintainer asked for `main` on the predecessor #3822.

(written by Claude Opus 5)

PR #3826, opened 2026-09-20
refactor(net)!: name path roles without new types

kixelated:
## Summary

Explored splitting `Path` into `path::{Absolute, Prefix, Pattern, Relative}` newtypes. Three of the four roles already have distinct types (`Path`, `Pattern`, `PathRelative`), and the missing boundary (exact path vs prefix) cannot be enforced: every path is a valid prefix, and the announce contract lets the application treat an announced prefix as the broadcast path (13 sites do `request_broadcast(update.path)`). A `Prefix` type would convert freely in both directions, so it would document without enforcing. This PR does the naming pass instead.

- `AnnounceUpdate.path` is now `prefix`, with the doc stating an announcement is always a prefix and never a broadcast. Mirrored as `Announce.Update.prefix` (JS), `MoqAnnounceUpdate::prefix()` (ffi), `AnnounceUpdate.Prefix()` (Go), `AnnounceUpdate.prefix` (Swift); Kotlin, Python, and Dart pick it up from the generated bindings.
- `PathRelative` / `PathRelativeOwned` become `moq_net::path::Relative` / `path::RelativeOwned` and leave the crate root, per the short-names-under-a-module rule. `Path`, `PathOwned`, `AsPath`, `Pattern`, `Patterns` stay at the root.
- `PathPrefixes` is deleted (zero consumers); closes quest `m1/api-path-prefixes`.
- The `Path` doc stops saying "all paths are RELATIVE", which collided with `origin.absolute()` / `root()` and with `path::Relative`.
- JS: `Path.Relative` brand; `normalizeRelative` / `relative` return it and `resolve` / `tryResolve` require it. hang's `RelativeBroadcast` carries the same brand, and `@moq/watch` types its rendition references with it.

## Public API

Breaking on `moq-net` (field rename, type move, type deletion), `@moq/net` (field rename, `Relative` brand on `resolve`/`tryResolve`), `@moq/hang` (`RelativeBroadcast` is branded), `@moq/watch` (`relativeBroadcast(effect, rel)` takes `Path.Relative`), `moq-ffi` and the Go/Swift wrappers (`prefix` accessor). No `dev` branch exists right now, so this targets `main` like the other recent `!` commits.

Not changed: the libmoq C struct `moq_announce_update.path` stays, so `cpp/obs` and the C docs are untouched. Renaming it is a C ABI source break that belongs with #3819 if wanted.

## Wire

None.

## Verification

- `cargo check --workspace --all-targets` and `cargo clippy ... -D warnings` clean, excluding two pre-existing failures on `main` unrelated to this change: `rs/moq-hls/src/export/mod.rs:645` calls `moq_mux::catalog::Producer::new` with one argument, and `rs/moq-bench/src/duration.rs` has dead code on macOS.
- `cargo test -p moq-net` path and announce tests, doctests, and `cargo doc -D warnings` pass.
- `bun run check` clean across `js/`; `bun test` in `@moq/net`, `@moq/hang`, `@moq/watch`, `@moq/room` pass.

(Written by Claude Fable 5.1)


kixelated:
## Verdict: **ITERATE**

Positive improvement, and the decision not to add a `Prefix` newtype is the right call. Exact path vs prefix cannot be enforced here (every path is a valid prefix, and callers already treat announced prefixes as broadcast paths), so a free-converting newtype would only document. Renaming the field to `prefix`, moving `PathRelative` under `path::Relative`, deleting unused `PathPrefixes`, and fixing the “all paths are RELATIVE” doc are clearer without fake type safety. The JS `Path.Relative` brand is a cheap, useful companion.

Worth the blast radius: mostly mechanical renames across languages. Leaving the C ABI `moq_announce_update.path` for a later ABI break (#3819) is sensible.

### Blockers before merge
CI shows PR-caused gaps, not only the pre-existing `moq-hls` `Producer::new` arity issue called out in the description:

1. **Swift** — `AnnounceUpdate.prefix` is wired in `Origin.swift`, but `swift/Tests/MoqTests/SmokeTests.swift` still uses `.path` (lines ~66 and ~71). That fails the Swift job.
2. **Dart** — README/docs/tests call `announcement.prefix()`, but the generated `MoqAnnounceUpdate` still exposes `path()` (C field unchanged). Either keep Dart on `path()` until the ABI rename, or add a Dart-side `prefix` alias that forwards. Right now docs and the pub package check disagree with the binding.

### Minor notes (non-blocking once CI is green)
- JS branding is cast-based (`as Relative`); fine for TS, just don’t expect runtime enforcement.
- After this lands, callers that still do `request_broadcast(update.prefix)` are unchanged in behavior; the name only makes the prefix-vs-broadcast contract harder to miss.

**Different approach?** Hard break without deprecation aliases is fine if that’s the house style for `!` API moves. I would not resurrect a `Prefix` newtype. Finish the Swift test + Dart binding/docs alignment, then this is mergeable.

(Written by Grok)

kixelated:
Rebased onto the new `origin/main` (now includes #3825 explicit driver time and #3821 uring ECN) and fixed the fallout. PR is MERGEABLE with fresh CI running.

## Fixes on top of the rebase
- `rs/moq-hls` call sites updated for `catalog::Producer::new(broadcast, config)` (was the known compile failure).
- Three macOS-only compile gates, all pre-existing and exposed because this diff selects those crates: `mod duration` in `moq-bench-host` is now Linux-gated (only used there), `moq-uring/tests/support.rs` gets `#![cfg(target_os = "linux")]` so its standalone test target is empty on macOS (its deps are Linux-only), and the `echo_noq` non-Linux stub uses `criterion::Criterion` by path. Linux CI never saw these.
- `quest/m1/api-review-gate.md`: dropped the PathPrefixes entry since this PR lands that quest (quest deletion convention: remove every reference).

## Review findings (CodeRabbit, one round)
Both are stale after the rebase, no change needed:
- py `prefix` rename: already on main (`py/moq-rs/moq/origin.py` exposes `prefix()`, example/tests/docs use it).
- Dart binding: the checked-in generated binding already exposes `MoqAnnounceUpdate.prefix()` and callers use it; the one remaining `.path()` is `BroadcastRequest.path`, a different type where `path` is correct.

## Verification
- `just check` green (exit 0, incl. wasm32 gate) and `just test` green (exit 0), run under `nix develop`. Note: bare-macOS `just check` fails in `rs _select` on multi-crate diffs because BSD awk rejects newlines in `-v` strings; pre-existing infra quirk, nix gawk works.
- No wire change (in-memory `AnnounceUpdate` field rename + `path::Relative` move only), so no draft update; FFI change is internal (field already `prefix` on main), so no binding regen.

(Written by Muse Spark)

PR #3828, opened 2026-09-21
refactor(net)!: return the next deadline from driver polls

kixelated:
## Problem

Follow-ups from #3825. `time::Driver` split its answer across `poll(now, waiter) -> Poll<Output>` and a separate `timeout()`, so every turn took two calls (and two clock locks), and "finished" had a different shape from `Session::closed`. The adapter loop that sleeps until the next deadline was copied into moq-tokio, moq-wasm, moq-ffi, and a `test-runtime` feature.

## Approach

`poll(now, waiter) -> Result<Option<Instant>, Error>`: `Ok(Some(at))` is the next deadline, `Ok(None)` waits on external activity only, and `Err` is terminal, with `Error::Closed` for a clean finish (the same value `Session::closed` reports). One `moq_net::time::run(driver) -> Error` lives next to the trait, since moq-net already depends on `web-async` for the wasm `Instant`; moq-uring keeps its own loop on the worker timer heap.

## Impact

- Breaking `moq-net`: `time::Driver` loses `Output` and `timeout()`; `Driver::poll` and `origin::Driver::poll` return `Result<Option<Instant>, Error>`. New `time::run`. The `test-runtime` feature is removed.
- Breaking `moq-tokio`, `moq-wasm`: the `runtime` modules are removed; callers use `moq_net::time::run`. `moq-mux`, `moq-srt`, `moq-stats`, `moq-video` drop their `moq-tokio` dev-dependency.
- `moq_uring::Handle::run` resolves with the terminal `Error`.
- The relay's WebSocket handler maps `Error::Closed` back to `Ok(())` so a clean close is not logged as a failure.
- No wire, TypeScript, or foreign ABI changes.

## Validation

`just check` and `just test` pass (4737 Rust tests, JS, Python).

## Notes

`web-async` stays, but only for time: moq-net needs the `performance.now()`-backed `Instant` and `setTimeout` sleep on wasm, which is exactly the cfg split the crate exists for. Its `Lock` uses are swapped for `kio::Lock` (which gains a `Default` impl), so they now sit under loom's model too.

The native JS smoke client follow-up from #3825 was already fixed on `main` by #3819.

(written by Claude Opus 5)


kixelated:
## MERGE

**Positive improvement?** Yes. This is the right cleanup after #3825. Folding the next deadline into `poll` removes the two-call (and two-lock) turn, and collapsing four nearly identical adapter loops (`moq-tokio`, `moq-wasm`, `moq-ffi`, `time::test`) into one `moq_net::time::run` is real complexity gone, not moved. Mapping a clean finish to `Error::Closed` also aligns the driver terminal with `Session::closed`, which the relay's `ended()` helper then handles cleanly.

**Worth the complexity?** Yes — this reduces surface area. Losing `Output` and `timeout()` is a break, but it is the natural second step while the explicit-time API is still settling, and the net diff shrinks the tree (~100 lines deleted, `test-runtime` gone, several crates drop a `moq-tokio` dev-dep). Call sites become one spawn of `moq_net::time::run`; custom loops (moq-uring) keep their own heap and only change the poll return shape.

**Different approach?** A dedicated step enum (`Deadline` / `Wait` / `Done(Error)`) would read clearer than `Result<Option<Instant>, Error>`, where `Err` means "finished" rather than "failed." The chosen encoding is still fine: three outcomes, documented on the trait, and matches how callers already want to branch. Not worth another break for naming alone. Keeping separate `poll` + `timeout` would preserve the duplication this PR exists to kill.

**Nits (non-blocking):**
- `doc/lib/rs/moq-net.md` still says "`moq-tokio` and `moq-wasm` drive sessions for their callers" after those `runtime` modules are removed — worth a one-line fix to point at `moq_net::time::run`.
- CI was still running at review time; author reports `just check` / `just test` green.

Ship it.

(Written by Grok)

PR #3830, opened 2026-09-21
feat(e2ee): derive opaque broadcast paths

### LINKED ISSUE #3717: e2ee: drop the `.e2ee` suffix and the `/<epoch>` path segment

kixelated:
## Problem

The unpublished E2EE profile exposed the semantic broadcast name and encryption policy through a .e2ee path suffix. The core implementations need one opaque, epoch-discoverable path contract before adopting the profile.

## Approach

- Derive a 22-character opaque prefix from the credential and semantic broadcast name with the moq-e2ee-00 path label.
- Keep the epoch as the final cleartext segment for instance discovery, outside the prefix derivation.
- Add and verify the meeting.hang known-answer vector, align the draft and Hang docs, and retire the completed quest.

## Impact

- Public API: none. The Rust and TypeScript core quests adopt this contract later.
- Wire: none. MoQ framing and persisted bytes are unchanged; broadcast paths are application data.
- Profile interoperability: unpublished moq-e2ee-00 path derivation changes from a semantic .e2ee suffix to an opaque prefix.

## Alternatives

Keeping a .e2ee suffix would leak application policy into every path consumer. Hiding the epoch would make generation discovery depend on data that cannot be decrypted until the epoch is known.

## Follow-ups

The E2EE API quest is now unblocked and owns the Rust implementation.

Closes #3717

(written by GPT-5)


kixelated:
## Recommendation: MERGE

**Positive improvement?** Yes. Replacing the semantic `.e2ee/<epoch>` suffix with an HKDF-derived opaque prefix removes encryption-policy leakage from every path consumer (matchers, exporters, recorders) while keeping the epoch as the final cleartext segment so instance discovery still works without decrypting first.

**Worth the complexity?** Yes for an unpublished profile. The derivation (`"moq-e2ee-00 path"` over context, kid, semantic broadcast, no epoch) mirrors the existing name/key HKDF shape, the Hang concept doc and draft Security Considerations are updated honestly (opaque hides the name only if the app does not republish it; epoch remains visible), and the `meeting.hang` known-answer vector plus verifier close the interoperability loop before cores adopt the contract.

**Different approach?** Keeping `.e2ee` would bake policy into paths forever; folding the epoch into the opaque segment would force discovery to depend on ciphertext. Minor follow-up (non-blocking): JSDoc on the exported `pathInfo` / `derivePath` APIs would help the next Rust/TS implementers, but vectors + draft are enough to merge.

Closes the path quest cleanly and unblocks the E2EE API work.

(Written by Grok)

kixelated:
Merge pass summary: PR was OPEN, MERGEABLE/CLEAN with all checks passing. Addressed the coderabbit nitpick (also flagged as a non-blocking follow-up in the automated review) by adding one-line JSDoc to the exported `pathInfo` and `derivePath` APIs in `drafts/moq-e2ee-00.ts`, per the repo convention to document every exported symbol. No functional change; verified locally with `bun drafts/moq-e2ee-00.ts` (1 path, 4 derivation, 5 group, 1 datagram, 30 negative vectors, all verified). Pushed as `586f0baf1` to `quest/m0/e2ee-path` and enabled auto-merge (squash, matching recent main history). No disagreements with automated findings worth replying to.

PR #3831, opened 2026-09-21
feat(moq-mux): record the TS mux rate and pad export to it

kixelated:
Completes `quest/m2/ts-mux-rate.md` (decision 6 of #3731; the issue stays open for the MSFTS convergence questions).

## Problem

A broadcast that came in as a CBR multiplex left `moq export ts` as a VBR one: the source's null stuffing is gone by the time media rides MoQ, and nothing recorded the rate it was padded to. An IRD or groomer had to be told the rate, or measure it (see [mpegts-pacer](https://github.com/tdrapier-wbd/mpegts-pacer), whose README calls the mux rate "not recoverable from MoQ").

## Approach

- **Import** measures the whole-multiplex rate off the PCR PID: every packet (nulls, PSI, retransmissions included) over the 27 MHz ticks between clocks. Intervals pool into half-second samples, four per window; a window whose samples agree within 2% publishes its aggregate, a later stable window republishes only past 1% drift, an unstable window with no sample at the published rate clears it, and a time-base discontinuity (or a PCR PID change, or a backwards/stalled clock) clears it at once. A VBR or file-paced source never publishes. Lives in `ts/mux_rate.rs`.
- **Export** settles a fixed-point packet balance before every clock packet: each slot credits its exact allowance (`mux_rate` units, one packet = 1504 × 40 units, so nothing rounds per slot) and every packet sent debits one; `floor(max(balance, 0))` nulls go out ahead of the clock packet, in the frame that precedes it. Media is never delayed or dropped; debt and credit are both capped at one second (credit matches the clock backfill cap), and the debt cap logs once per overrun.
- `Export::with_mux_rate` overrides the catalog; `moq export ts --mux-rate <bps>` calls it. The flag lives on a new `Transport` args struct so it doesn't leak onto `flv`/`h264`/`h265`.
- New fixture `test_data/bbb_cbr.ts` (145 KB): `bbb5s.ts` remuxed with `ffmpeg -t 2.8 -c copy -muxrate 400000`, byte-locked with 42% stuffing. The Kyrion captures turned out unusable for this: `kyrion_dirtystart` stamps PCR at 0.5 ms granularity and wanders ±1.2% per half second (hence the 2% sample tolerance, which it now passes), and `kyrion_mpeg2av_ac3` is byte-locked but its PMT arrives 1 s in, leaving one interval short of a window.

## Impact

- Catalog: additive `mpegts.muxRate` (bits per second), omitted when absent. Documented in `draft-lcurley-moq-mpegts` (-00, unsubmitted) with a `Rebuilding` bullet.
- `moq_mux::container::ts::Mpegts::mux_rate: Option<u64>`, counted by `is_empty`.
- `moq_mux::container::ts::Export::with_mux_rate(u64) -> Self`.
- `moq export ts --mux-rate <bps>`; `doc/bin/cli.md` updated.
- Output of a catalog without the field and no override is byte-identical to before.
- Wire: none beyond the catalog field. No JS mirror exists for the typed `mpegts` section.

## Alternatives

- Per-interval agreement (as the quest sketched) at 1%: rejected by real hardware (see above); half-second samples keep packet rounding and clock granularity well inside tolerance while a VBR source is still off by multiples.
- Nudging PCR values to the padded byte position: the export asserts a uniform 25 ms grid on purpose (#3351), so byte position stays quantized to ±1 packet per interval; a groomer's regenerate mode rewrites the six PCR octets from byte offset, which is its job.

## Follow-ups

- Like continuity counters (#2779), the stuffing balance starts at the exporter's first clock packet, so two exporters joining at different times pad differently.
- Overrun logging is at the debt cap only; a transient keyframe burst that recovers within a second is silent by design.

(Written by Claude Opus 5)


kixelated:
## Recommendation: ITERATE

**Positive improvement?** Yes. Recording the whole-multiplex CBR rate on import and padding export with nulls closes a real IRD/groomer gap (mux rate is not recoverable from MoQ alone). The meter design is thoughtful: half-second samples, 2% agreement / 1% drift, clear-on-discontinuity, VBR stays absent, and export without the field stays byte-identical.

**Worth the complexity?** Mostly — the fixed-point `STUFFING_UNIT` balance, one-second debt/credit caps, and `Transport` CLI isolation are justified. But the export path trusts an unvalidated catalog `muxRate` (`u64`) too far.

**Different approach / required fixes before merge:**

1. **Bound the rate at the catalog / override boundary.** `stuff` does `let rate = rate as i64`, then credits `slots * rate` and emits `nulls = balance / STUFFING_UNIT` with `payload.reserve(nulls * 188)`. A hostile or corrupt catalog with an absurd `muxRate` (or wrap past `i64::MAX`) can OOM the exporter or produce nonsensical negative rates. Reject `0`, cap at a realistic max (e.g. tens of Gb/s for TS), and/or hard-cap nulls per slot to the credit ceiling already implied by `PCR_BACKFILL`.
2. **Prefer checked conversion over `as i64`.** Saturating mul helps after the cast, but the cast itself is the bug for values above `i64::MAX`.

Import measurement, fixtures (`bbb_cbr.ts`, Kyrion), and override semantics look solid. Land the validation/caps and this is a MERGE.

(Written by Grok)

kixelated:
Merge review: addressed the automated findings and rebased onto current main. All checks pass (Check, Test, review, Android) and the PR is MERGEABLE/CLEAN with no human change requests outstanding.

**Findings addressed (agreed with both moq-bot and the automated Grok review):** the untrusted catalog `mpegts.muxRate` reached the stuffing allocator unbounded. Fixed in `fix(moq-mux): refuse zero and absurd TS multiplex rates` (rs/moq-mux/src/container/ts/export.rs):

- New `MAX_MUX_RATE = 1_000_000_000` (1 Gb/s, far above any broadcast contribution multiplex) plus `sanitize_mux_rate`: zero and anything past it are refused at both entries (the `with_mux_rate` builder and the catalog snapshot in `update_catalog`), with a warning, leaving the output unpadded. An explicit-but-invalid override shadows the catalog rather than silently falling back to it.
- `stuff()` converts with `i64::try_from` instead of `as i64` (no wrap past `i64::MAX`), refusing to pad if it ever fails, and caps nulls per call to what the credited slots allow (a no-op on sanitized rates, whose balance carries less than one packet).
- Two regression tests (`export_mux_rate_override_is_bounded`, `export_refuses_an_absurd_catalog_mux_rate`): both abort in the allocator (SIGABRT) without the fix and pass with it. Full `moq-mux` (783) and `moq-cli` (95) suites pass, clippy/fmt/rustdoc clean, `quest check` passes.

**Rebase:** origin/main moved during CI (quest tree reorg #3855 renamed `quest/m2/ts-mux-rate.md` to `quest/next/` and deleted `quest/m2/README.md`). Resolved per the quest conventions (a completed quest is deleted with all its references): the quest doc is deleted in its new home and the Mux rate line is removed from `quest/next/README.md`. No code changes in the rebase.

Enabling squash auto-merge (the repo norm).

PR #3832, opened 2026-09-21
refactor(moq-sock)!: complete groups before serving

kixelated:
## Problem

`Member::bind` exposed sockets before reuseport formation and steering were complete. Callers also had to retain every socket themselves, because dropping one renumbers the kernel group and invalidates shard steering.

## Approach

- Return opaque claims while forming a reuseport group.
- Validate every claim and attach steering in `Group::complete`.
- Retain the original sockets in `Bound` while releasing duplicate serving handles.
- Migrate the Tokio and io_uring consumers.
- Add Linux regressions for partial completion and dropped serving handles.
- Complete the socket-group quest and unblock its dependents.

## Impact

- Public API: breaking `moq-sock` 0.0.1 change. `Member::bind` returns `Claim`; `Group::complete` creates `Bound`; `Bound::member` returns retained `Socket` handles.
- `moq-tokio` public worker signatures and its `bind` re-export are unchanged.
- Wire: no change.

## Alternatives

Keeping raw sockets in callers preserves the same unenforceable lifetime contract, so ownership remains in `moq-sock`.

## Follow-ups

- #2964 remains scoped to replacing `moq-tokio`'s redundant private lifetime bookkeeping after its other prerequisite lands.

(written by GPT-5)


kixelated:
## Recommendation: MERGE

**Positive improvement?** Yes. The previous API handed usable sockets before the reuseport group was complete and left "keep every socket or steering breaks" as an unenforceable caller rule. `Claim` → `Group::complete` → `Bound` makes partial formation and early drops unrepresentable: no socket serves until every member is bound and the BPF filter is attached, and `Bound` retains the originals so dropping a serving clone cannot renumber the kernel group.

**Worth the complexity?** Yes. The extra types are the complexity — they encode the kernel lifetime contract instead of documenting it. Linux regressions for incomplete completion and dropped serving handles prove the invariant. Tokio/uring/relay consumers are migrated; published moq-tokio worker signatures stay intact.

**Different approach?** Keeping raw sockets in callers would preserve the same footgun. Ownership belonging in `moq-sock` is the right place. Follow-up #2964 can now replace tokio's redundant private bookkeeping without redesigning this surface again.

This is the primitive the one-port / worker quests needed. Merge.

(Written by Grok)

kixelated:
Merge check: state OPEN, not a draft, MERGEABLE/CLEAN. Checks green: Check, Test, review all pass. CodeRabbit reported no actionable comments; moq-bot approved; automated reviewer recommended MERGE. No conflicts, no CI failures, no human change requests, so no fixes were pushed and no worktree was needed. Enabling squash auto-merge (repo's usual method per recent merged PRs).

PR #3829, opened 2026-09-21
refactor(archive)!: use inclusive group ranges

kixelated:
## Problem

Archive callers had to reverse group bounds across related APIs, and direct `Key::Groups` construction could bypass range validation.

## Approach

Use finite `RangeInclusive<u64>` values in first-to-last order throughout keys, object bounds, validation, decoding, and store reads. Validate empty, reversed, exhausted, and out-of-profile ranges before serialization or decoding while keeping the private filename codec largest-first.

## Impact

- **Public API:** `Key::Groups` now contains `range: RangeInclusive<u64>` instead of separate `largest` and `smallest` fields.
- **Public API:** `Key::groups`, `Store::get_groups`, `Object::decode_groups`, and `Object::check_bounds` now accept `RangeInclusive<u64>`.
- **Public API:** `Object::bounds` now returns `RangeInclusive<u64>`.
- **Wire/persisted format:** unchanged. Existing `largest.smallest` paths and segment bytes are preserved.

## Alternatives

A new archive-specific bounds type was rejected because the standard inclusive range expresses the finite contract directly.

## Follow-ups

None.

(written by GPT-5)

PR #3833, opened 2026-09-21
refactor(archive)!: scope listing queries to recordings

kixelated:
## Problem

Streaming archive lists accepted fully prefixed object-store paths while paginated lists accepted relative strings and the full backend option type. That mismatch allowed double-prefixing, sibling recording matches, and directory results that the archive page silently discarded.

## Approach

Introduce `store::list::Query`, `Entry`, and `Page` around recording-relative semantic scopes and offsets. Both listing paths now consume the same query. Pagination exposes a nonzero page size and returns the next query with its backend token kept private. Lexical backend prefixes end in `/`, and unexpected directory results fail explicitly.

Move track/group/segment path construction behind query and store methods, add regression coverage for track and recording boundaries, exclusive offsets, multiple pages, and every pagination control, then complete the quest.

## Impact

- Breaking `moq-archive` 0.0.1 API: replace root `List`, `Listed`, and `Page` with `store::list::{Query, Entry, Page}`; change `Store::list` and `Store::list_paginated` to consume `Query`.
- Pagination no longer accepts `object_store::PaginatedListOptions` or exposes raw continuation strings. It supports recording scopes, exclusive offsets, nonzero page sizes, and opaque next-page queries.
- Remove the public `path` module; root exports `check_id`, `encode_track`, `decode_track`, `format_id`, `parse_id`, and `ID_WIDTH`; `Key::{path, parse}`; and `Store::{groups_prefix, segments_prefix, groups_offset}`. `Store::path` and the `object_store` re-export remain available as the escape hatch.
- Add `Error::Directory` for a paginated backend violating the flat-list contract.
- Wire format and persisted object paths and bytes are unchanged.

## Alternatives

- Keep exposing backend pagination options. Rejected because delimiter and extensions are not behavior the archive result represents.
- Add a second public relative-path type. Semantic query constructors keep the smaller API while preventing recording prefixes from being supplied as offsets or scopes.

## Follow-ups

None.

Validation: `nix develop --command just check` and `nix develop --command just test` pass; all 44 `moq-archive` tests pass.

(written by GPT-5)


kixelated:
## Recommendation: ITERATE

**Positive improvement?** Yes. Unifying streaming and paginated listing behind recording-relative `store::list::Query` fixes real bugs: double-prefixing, sibling matches like `rec` vs `rec-other` (trailing `/` on paginated prefixes), and silently discarded directory results (`Error::Directory`). Opaque next-page tokens and `NonZeroUsize` page sizes are the right public surface for a 0.0.1 crate.

**Worth the complexity?** The Query/Entry/Page redesign is justified. Two gaps keep this from MERGE:

1. **`list` ignores `page_token` (and `max_keys`).** Docs say page-only controls are ignored, but `Page::next` is a full `Query` — passing it to `Store::list` silently restarts from the scope start. Fail the stream (or return an error) when `query.page_token.is_some()`, and document that continuation queries must go to `list_paginated`.
2. **No end-to-end `list_paginated` exercise.** Current tests assert prefix/option strings and drive synthetic `page()` output. Add a backend that implements `PaginatedListStore`, write siblings plus a few segments, walk `Page::next` via `list_paginated` with `page_size(1)` to `None`, and assert the union matches streaming results minus siblings. That is the regression the quest promised for trailing-slash scope, offset wiring, and token round-trip.

After those two land, MERGE. The API shape itself is good.

(Written by Grok)

kixelated:
Merge prep done in `.worktrees/merge-3833`:

- Addressed both automated review findings: `Store::list` now fails the stream with the new `Error::Pagination` when handed a `Page::next` continuation (docs point at `list_paginated`), and added an end-to-end paginated walk over a test-only `PaginatedListStore` backend (`PaginatedMemory`) covering exclusive offsets, opaque token round-trip with `page_size(1)`, and trailing-slash sibling exclusion at recording scope.
- Rebased onto `origin/main` twice: resolved the `store.rs` import overlap with #3829 (inclusive group ranges) and aligned the quest index deletions with #3855 (quest tree mirror, `m0->main` renames).

Local: 49 `moq-archive` tests pass (2 new), clippy/fmt clean, `quest check` 417 docs ok. All CI checks pass, no human change requests outstanding.

PR #3834, opened 2026-09-21
fix(moq-nvenc): roll back failed resource mapping

kixelated:
## Problem

`register_generic_resource` returned immediately when NVENC mapping failed, leaving the successful registration behind and releasing the caller's input owner too early.

## Approach

- Make register, map, and rollback one transaction behind an injectable resource API.
- Unregister exactly once when mapping fails, while retaining the input owner through cleanup.
- Keep the mapping error primary and attach an unregister failure as cleanup context.
- Retain the allocation when unregister fails because NVENC may still refer to it.
- Cover rollback success, rollback failure, and normal mapped-resource destruction without an NVIDIA driver.
- Complete the registration rollback quest and remove its dependency edge.

## Impact

- Public API: additive `EncodeError::cleanup()` access to rollback failures; the same cleanup error is exposed through `std::error::Error::source`.
- Runtime behavior: failed resource mapping now unregisters successful registration before releasing the input owner.
- Wire: no change.

## Alternatives

A temporary guard using only `Drop` could unregister, but it could not return an unregister failure alongside the primary mapping error.

## Follow-ups

None.

(written by GPT-5)


kixelated:
## Recommendation: MERGE

**Positive improvement?** Yes. `register_generic_resource` previously returned on map failure after a successful register, leaking the NVENC registration and releasing the caller's input owner too early. Making register → map → rollback one transaction fixes a real resource-lifetime bug on the GPU encode path.

**Worth the complexity?** Yes and well contained. The injectable `ResourceApi` exists so rollback success, rollback failure, and happy-path Drop can be tested without a driver. Retaining the marker via `mem::forget` when unregister also fails is the correct unsafe-adjacent choice (driver may still refer to the allocation). Primary map error + `EncodeError::cleanup` / `Error::source` preserves both failures without inventing a Drop-only guard that cannot surface the cleanup error.

**Different approach?** A Drop guard alone cannot return the unregister failure alongside the map error; the explicit transactional path is better. Additive public `cleanup()` access is minimal API surface for a useful diagnostic.

Ship it.

(Written by Grok)

PR #3835, opened 2026-09-21
fix(nvenc)!: retain resources through completion

kixelated:
## Problem

The safe NVENC facade exposed forgeable driver handles and pointer-bearing settings, while submitted buffers could be released before synchronous completion. Session-borrowed resources also prevented bitrate reconfiguration while buffers existed.

## Approach

- Make buffers and registrations own the encoder lifetime and retain external CUDA allocations.
- Seal input handles, make raw registration/config entry points explicitly unsafe, and own config storage.
- Return an owned `Submission` that waits before releasing input/output on completion or cancellation.
- Add a fake completion-driver boundary plus compile-fail coverage for lifecycle, cross-session, and handle-forging constraints.
- Adapt `moq-video` to the narrowed facade.

## Impact

- Breaking `moq-nvenc` API: removes public `EncodeAPI`, `ENCODE_API`, `EncoderOutput`, and raw codec picture parameters; resource types no longer borrow `Session`; `encode_picture` consumes buffers and returns `Submission`; raw registration/configuration require `unsafe`.
- `moq-video` retains CUDA frame storage through NVENC completion.
- Wire: none.

## Alternatives

A second codec configuration API was avoided; preset-derived SDK config stays available through an explicit unsafe escape hatch. Borrowed submissions were rejected because cancellation could still release in-flight storage.

## Follow-ups

Pooling and broader recovery remain in their existing M2 quests. Hardware encode/drain validation remains separate from the no-GPU lifecycle tests. This PR also performs the minimal unregister rollback needed for partial-initialization safety; PR #3834 remains the focused registration-rollback change and is otherwise untouched.

(written by GPT-5)


kixelated:
## Recommendation: MERGE

**Positive improvement?** Yes. The safe NVENC facade previously let callers forge input handles and drop CUDA/input storage while encode was still in flight. This seals `EncoderInput`, makes buffers/`RegisteredResource` own an `Arc<Encoder>` (so sessions can reconfigure without lifetime borrows), and turns `encode_picture` into a consuming API that returns a `Submission` which waits (or flushes) before releasing handles. `moq-video` correctly clones CUDA frames into the registration marker so the allocation outlives completion. Partial map failure now unregisters — good.

**Worth the complexity?** Yes. Ownership across a synchronous SDK completion boundary is the hard part; the `CompletionDriver` / fake-driver tests plus compile-fail coverage for forging and early `drop(input)` make the complexity pay for itself. Leaking input/output on a failed wait is the right conservative failure mode versus use-after-free. Breaking removals (`EncoderOutput`, public `ENCODE_API`, `CodecPictureParams` / `picture_type`) are justified for a pre-0.1 facade that was too open.

**Different approach?** No better alternative for the lifecycle goal. Borrowed submissions were correctly rejected (cancellation still frees in-flight storage). A second config API was rightly avoided in favor of owning `Box<NV_ENC_CONFIG>` with an explicit `unsafe encode_config` escape hatch. Follow-ups (pooling, hardware drain validation) stay appropriately out of scope.

Clear ownership fix for a real UAF-shaped hole — merge.

(Written by Grok)

kixelated:
Rebased onto latest origin/main (resolving the quest-tree restructure: applied the quest completion/de-linking to quest/main and quest/next paths) and addressed the automated review findings in f879177c8:

- NeedMoreInput submissions no longer flush EOS (which ended the stream). Completion now waits without sending end-of-stream, and encode_picture documents that the facade does not reorder B-frames (configure frameIntervalP = 1, as moq-video does).
- Sealed-handle compile_fail now implements the full EncoderInput trait, so only the seal rejects it.
- EncoderInput::encoder has its one-line doc.

Verified: cargo check/test/clippy/fmt on moq-nvenc and moq-video, quest check (415 documents ok). All CI checks pass and the automated review approves. Auto-merge (squash) enabled.

PR #3837, opened 2026-09-21
fix(video)!: confine synchronous codecs to their thread

kixelated:
## Problem

Public synchronous video `Encoder` and `Decoder` inherited `Send` from their private backend trait, even though several native codecs and Media Foundation COM state must be created, used, and dropped on one thread. Direct constructors bypassed the worker confinement that made the old unsafe backend promises valid.

## Approach

- Remove the backend `Send` bound and the codec-level unsafe `Send` implementations.
- Make direct `Encoder` and `Decoder` handles explicitly thread-bound, with compile-fail ownership examples.
- Keep async `Sink` and `Consumer` movable; the worker owns non-macOS codecs, while the macOS inline wrapper is the controlled serialized migration boundary.
- Balance `CoInitializeEx` if `MFStartup` fails, and document inherited `Consumer::read` cancellation poisoning.
- Convert the spawned transcoder test producer to the supported worker-backed `Sink`.
- Remove the completed quest and its dependency references.

## Impact

Public API: breaking auto-trait tightening for synchronous `moq_video::encode::Encoder` and `moq_video::decode::Decoder`, which are now `!Send` and `!Sync`. Async `Sink` and `Consumer` remain `Send`.

Wire: none.

## Alternatives

Keeping unsafe `Send` on every backend would preserve the accidental public contract but could not guarantee same-thread destruction for direct handles. Documentation alone would not prevent misuse.

## Follow-ups

None. The existing nightly `just rs windows` lane already compiles the Media Foundation path.

## Validation

- `just check`
- `just test` (496 passed, 6 skipped)
- `cargo test --locked -p moq-video --doc` (including both compile-fail ownership checks)

(written by GPT-5)


kixelated:
## Recommendation: MERGE

**Positive improvement?** Yes. Synchronous `Encoder` / `Decoder` were accidentally `Send` because backends carried `unsafe impl Send` to satisfy a `Backend: Send` bound, even though Media Foundation COM, VAAPI `Rc`s, and several native codecs must be created/used/dropped on one thread. Direct constructors bypassed the worker that made those promises hold. Removing the bound, dropping the unsafe `Send`s, and marking public sync handles `!Send`/`!Sync` via `PhantomData<Rc<()>>` (with compile-fail examples) matches reality. Async `Sink` / `Consumer` stay movable; macOS inline wrappers keep a controlled serialized `unsafe impl Send`. `ComGuard` now balances `CoUninitialize` if `MFStartup` fails — a real apartment leak fix.

**Worth the complexity?** Yes, and the diff is small for the safety win. Auto-trait tightening is a breaking change but the correct one: documentation alone cannot stop moving a direct handle. Transcode test correctly moved to worker-backed `Sink`. Cancellation-poisoning docs on `Consumer::read` are honest.

**Different approach?** Keeping unsafe `Send` on every backend would preserve an accidental public contract and still allow same-thread-destroy violations for direct handles. Making every backend truly thread-safe would be far more complex and unnecessary when `Sink` already owns the worker model.

Right confinement story for native codecs — merge.

(Written by Grok)

kixelated:
Addressed all 4 findings in bc18525 (docs-only): decode/sink.rs no longer claims a MediaFoundation Send impl, the MediaFoundation and VideoToolbox backend docs now state backends are !Send with confinement via Worker (macOS Sink::Inner keeps the serialized Send wrapper), and Consumer::read docs distinguish mid-stream decode cancel (error on every read) from tail-flush cancel (error then None). Verified with cargo check, moq-video lib tests (113 passed), doc tests (2 passed), and quest check (415 ok).

kixelated:
Merge summary: rebased onto origin/main (d665f83), resolving quest-tree conflicts by dropping the completed video-thread-ownership quest from quest/main and quest/next references. Addressed all 4 automated review findings with docs-only fixes (bc18525). All checks green (Check, Test, Android, Swift, review). Auto-merge (squash) enabled.

Public API impact: breaking auto-trait tightening, sync moq_video Encoder/Decoder now !Send/!Sync; async Sink/Consumer remain Send. Wire: none.

PR #3839, opened 2026-09-21
Make media backends optional

kixelated:
## Problem

`moq-video` always compiled OpenH264 and enabled rendering by default, so codec-only and hardware-only consumers paid for C++ and wgpu stacks they did not need. The split NVIDIA feature aliases also preserved unsupported pre-0.1 names, and audio host flags activated cpal without capture or playback.

## Approach

- Put OpenH264 behind the default `openh264` feature in video and transcode, remove rendering from video defaults, and remove the `nvenc`/`nvdec` aliases.
- Keep backend selection strict when software support is disabled, with regression coverage for `Software` and named OpenH264 selection.
- Make audio host flags customize cpal only when device I/O enables it.
- Add independent minimal, software, native, rendering, and audio host checks plus dependency-tree assertions to CI (`just rs media-features`). The dependency graph is resolved for Linux on every host, since `moq-nvenc` and the PipeWire cpal host are Linux-only.
- Gate the VideoToolbox and Media Foundation decoder tests on `openh264`: they encode their test stream with the software encoder, so without it `Kind::Software` correctly refuses and they cannot run. The hardware-to-hardware HEVC round trips stay ungated.
- Update explicit workspace consumers and feature documentation, and close the completed quest.

## Impact

Public API: Cargo defaults and supported feature names change for unpublished `moq-video` and `moq-transcode`; `moq-audio` host feature activation becomes conditional. Rust symbol APIs are unchanged.

Wire: none.

## Alternatives

Keeping OpenH264 or rendering unconditional would preserve simpler manifests but retain avoidable native and graphics build costs. Splitting encode and decode into separate backend features was deferred because there is not yet evidence that the extra feature surface pays for itself. A checked-in H.264 fixture would let the hardware decoder tests run without OpenH264, but the software encoder is on by default so the gate loses nothing in the merge gate.

## Follow-ups

None.

## Verification

- `nix develop --command just rs media-features` (macOS host; the minimal `moq-video` test suite runs and every graph assertion holds)

(written by GPT-5, landed by Claude Opus 5)


kixelated:
## Recommendation: MERGE

**Positive improvement?** Yes. `moq-video` always compiled OpenH264 and defaulted `render`, so codec-only / hardware-only consumers paid for vendored C++ and wgpu they did not need; `nvenc`/`nvdec` aliases preserved unsupported pre-0.1 names; audio host flags activated cpal without device I/O. This puts OpenH264 behind a default `openh264` feature, drops `render` from defaults, removes the aliases, gates selection so `Kind::Software` / named `openh264` fail cleanly when disabled, and adds a CI `media-features` matrix with dependency-tree assertions. Workspace consumers are updated explicitly.

**Worth the complexity?** Yes. Feature plumbing across video/transcode/audio plus CI gates is real work, but it is the right cost to pay once before 0.1 rather than shipping always-on native stacks. Strict selection tests for disabled software are important so Auto/Hardware do not silently claim a missing fallback. Deferring encode/decode feature splits is sensible until demand shows up.

**Different approach?** Keeping OpenH264 or rendering unconditional would preserve simpler manifests at ongoing build cost. A compatibility re-export of `nvenc`/`nvdec` would only prolong the unsupported names. No wire impact; symbol APIs unchanged — Cargo defaults and feature names are the intentional break.

Build-cost hygiene done properly — merge.

(Written by Grok)

PR #3840, opened 2026-09-21
refactor(mux)!: share media rate policy

kixelated:
## Problem

The shared media rate policy was exposed only under `moq_video::encode::rate`, even though audio and transcode senders are intended to use the same policy. `Control::new` also initialized with `max.max(min)`, so inverted bounds could start above the configured ceiling even though updates normalized them.

## Approach

- Move the single `Policy` and `Control` implementation, tests, and public docs to `moq_mux::rate`.
- Update all current video, FFI, and libmoq consumers to the shared namespace.
- Initialize `Control` at `Policy::max` and strengthen the inverted-bounds regression test.
- Remove the completed quest and its dependency references.

## Impact

Public API:
- Adds `moq_mux::rate::{Policy, Control}`.
- Removes `moq_video::encode::rate` without a compatibility shim. This is a breaking change to the pre-0.1 video API.
- Inverted policies now start and remain at their ceiling instead of initializing above it.

Wire impact: none.

## Alternatives

A compatibility re-export would preserve two public namespaces before 0.1 and undermine the purpose of settling the API now. Keeping the policy in `moq-net` would mix sender media policy into the transport allocation layer.

## Follow-ups

The existing audio, ladder, and OBS quests can now consume the settled shared namespace. Their remaining work stays separate.

## Validation

- `just check`
- `just test` (4,195 Rust tests passed, 9 skipped; the no-default relay test and 64 Python tests also passed)
- `cargo test --locked -p moq-mux --doc` (10 passed)

(written by GPT-5)


kixelated:
## Recommendation: MERGE

**Positive improvement?** Yes. Rate policy lived under `moq_video::encode::rate` while audio, ladder, and OBS need the same congestion→target mapping. Moving it to `moq_mux::rate` (with `Control` / `Policy`) and updating video/FFI/libmoq call sites settles one shared namespace before 0.1. Inverted `min > max` now normalizes to the ceiling at construction instead of clamping through an inverted range — clearer and safer. Wire unchanged.

**Worth the complexity?** Low. Mostly a rename/move with docs and a small behavioral fix for inverted bounds. Removing `moq_video::encode::rate` without a shim is the right pre-0.1 call; a compatibility re-export would leave two public homes. Keeping policy in `moq-net` would mix sender media policy into transport allocation.

**Different approach?** No. Shared mux-level policy is the natural home for senders that already depend on mux catalog bits. Follow-up quests can consume the settled API without further reshuffling.

Clean shared-policy landing — merge.

(Written by Grok)

PR #3841, opened 2026-09-21
refactor(audio)!: remove ineffective FEC flag

kixelated:
## Problem

The public `moq_audio::encode::Config::fec` and `Options::fec` flags only enabled the Opus in-band FEC control. The encoder keeps expected loss at zero, so libopus emits no redundancy, while the decoder never requests FEC recovery. The option therefore promised behavior the shipped path did not provide.

## Approach

- Remove the FEC field from both public audio encoding configuration paths.
- Stop setting the ineffective Opus control.
- Retain and test DTX behavior.
- Add a focused test that every remaining producer option reaches the encoder configuration.
- Remove the completed quest and its references.

## Impact

- Public API: removes `fec: bool` from the pre-0.1 Rust `moq_audio::encode::Config` and `Options` structs.
- Wire: no change to currently emitted packets. Opus packetization, DTX, and ordinary packet-loss concealment remain unchanged.

## Alternatives

Keeping the boolean would preserve a false capability promise. Adding expected-loss tuning or decoder recovery here would expose an unproven policy without sequencing, loss metadata, or an end-to-end dropped-packet fixture.

## Follow-ups

The existing audio loss-recovery quest remains the place to measure a complete policy before adding a replacement API.

## Validation

- `just check`
- `just test` (421 passed)

(written by GPT-5)


kixelated:
## Recommendation: MERGE

**Positive improvement?** Yes. Public `fec` on `moq_audio::encode::Config` / `Options` only flipped Opus in-band FEC while expected loss stayed at zero (so libopus emitted no redundancy) and the decoder never requested recovery — a false capability. Removing the flag and the ineffective `OPUS_SET_INBAND_FEC` call makes the API honest. DTX remains tested; new `options_preserve_encoder_configuration` checks remaining producer options reach the encoder. Wire packets unchanged.

**Worth the complexity?** Trivial and worth it. Shipping a no-op boolean is worse than a small breaking removal pre-0.1. Deferring real FEC/loss policy to the existing audio loss-recovery quest (with sequencing, loss metadata, and dropped-packet fixtures) is the right sequencing.

**Different approach?** Keeping the boolean preserves a lie. Implementing expected-loss tuning or decoder recovery in this PR would expose unproven policy without end-to-end proof. No shim needed before 0.1.

Honest API cleanup — merge.

(Written by Grok)

PR #3842, opened 2026-09-21
refactor(audio)!: expose demand without track authority

kixelated:
## Problem

`moq-audio` exposed its transport producer and standalone resampler even though callers only need subscriber demand. The capture shorthand also duplicated the fields already owned by `PublicationOptions`, and audio frame records could not grow compatibly.

## Approach

- replace `encode::Producer::track` with a watch-only `demand` handle and adapt Rust, CLI, FFI, and libmoq internals
- keep full track authority private to capture ownership and stop exporting `Resampler`
- make `publish_capture` accept `PublicationOptions`
- make `Frame` and `Encoded` non-exhaustive while retaining their constructors
- cover demand transitions, demand closure, and publication state closure
- remove the completed quest and its dependency references

## Impact

Public API: breaking Rust-only cleanup for the unpublished 0.0.x `moq-audio` API. `Producer::track` and the root `Resampler` export are removed, `Producer::demand` is added, `publish_capture` now takes `PublicationOptions`, and direct external struct literals for `Frame` and `Encoded` must use their constructors. Published libmoq and language-binding signatures and layouts are unchanged.

Wire impact: none. Audio framing, packetization, timestamps, priming, padding, and abort behavior are unchanged.

## Alternatives

Keeping `track` alongside `demand` would preserve transport write and close authority that audio callers do not need. Keeping the five capture arguments would duplicate `PublicationOptions` and make future options another signature change.

## Follow-ups

None.

(written by GPT-5)


kixelated:
**Verdict: MERGE**

Positive improvement: callers only need subscriber demand, and exposing `Producer::track` let them bypass audio framing/padding. Replacing it with a watch-only `demand()` (and keeping `track` crate-private for capture) matches the video pattern and shrinks the public surface. Bundling `publish_capture` into `PublicationOptions`, un-exporting `Resampler`, and marking `Frame`/`Encoded` `#[non_exhaustive]` are the right 0.0.x cleanups; published bindings stay unchanged and wire is untouched.

Worth the complexity: the diff is small (~+77/−69) and the new `demand_follows_subscribers_and_closes_with_the_producer` test covers the lifecycle that used to be implied. No better alternative stands out — keeping `track` alongside `demand` would leave write/close authority in the public API for no caller benefit.

Minor merge note: this branch and #3844 both touch `EncoderOutput::now()`’s `Clock` call (`now().value()` vs `now().as_micros()`); resolve when stacking them, but that is not a reason to hold this PR.

(Written by Grok)

PR #3843, opened 2026-09-21
refactor(audio)!: separate configuration contracts

kixelated:
## Problem

`moq-audio` mixed source PCM, codec settings, decoded output conversion, and subscription policy into overlapping configuration types. Raw channel counts also lost speaker meaning and made safe layout conversion ambiguous.

## Approach

- Add `Layout` with named mono/stereo and unnamed discrete channels.
- Split encoding into `Input`, exact codec `Settings`, and publication `Options`.
- Split decoding into backend `Config`, PCM `Output`, and subscription `Options`.
- Refuse unsupported spatial conversions and codec-specific settings such as PCM DTX/FEC.
- Adapt native callers and published bindings internally without changing their external shapes.
- Remove the completed quest and its absolute references.

## Impact

- **Rust API:** breaking `moq-audio` configuration and decoder-construction changes. The crate is still `0.0.x`.
- **Published bindings:** C, UniFFI, and wrapper signatures, record layouts, channel-count conventions, and zero/None sentinel behavior are unchanged.
- **Wire:** unchanged.

## Alternatives

Keeping optional channel counts and duplicated codec fields would preserve an ambiguous API and block safe backend/layout growth, so this uses one concrete codec-settings shape instead.

## Follow-ups

Existing layout and decoder-backend quests can extend the new contracts with named surround layouts, mixers, and platform backends.

## Verification

- `just check`
- `just test`

(written by GPT-5)


kixelated:
**Verdict: MERGE**

Positive improvement: channel counts alone cannot express speaker meaning, and folding source PCM, codec knobs, and subscription policy into overlapping configs made safe layout/backend growth ambiguous. Introducing `Layout` (Mono/Stereo/Discrete), splitting encode into `Input` / `Settings` / `Options` and decode into `Config` / `Output` / `Options`, and refusing PCM DTX/FEC and unsupported spatial conversions is the contract the audio stack needed before 0.1.

Worth the complexity: ~+658/−602 across 33 files is large, but it is concentrated API reshaping with internal binding migrations and no wire/ABI change. Keeping optional channel counts or compatibility aliases would preserve the ambiguity this quest exists to remove. Follow-up surround/mixer/backend quests can extend `Layout` and the new settings shapes rather than re-slicing them later.

No stronger alternative than shipping this split while `moq-audio` is still 0.0.x.

(Written by Grok)

PR #3844, opened 2026-09-21
fix(audio)!: enforce exclusive AEC ownership

kixelated:
## Problem

AEC controls were cloneable but also owned the microphone processing path, so one adaptive filter could be attached to multiple microphones. Creating another engine reference silently replaced the first. The AEC feature also no longer compiled after the broadcast clock API changed.

## Approach

- Split the public cloneable `aec::Control` from a crate-private exclusive microphone attachment.
- Refuse a second live microphone attachment or playback reference with `Error::Busy`, and release each slot on drop.
- Keep processing state across demand gaps while clearing in-flight samples during capture teardown.
- Add fake ownership, control-clone, replacement, and concurrent-teardown regressions, and update the speakerphone docs.
- Use `Clock::now()` in the capture feature so the off-by-default AEC build compiles again.

## Impact

- Public API: `aec::Canceller` becomes `aec::Control`; `Engine::canceller` now returns `Result<Control, Error>`; `capture::Config::aec` accepts `Control`; `Error::Busy` is added.
- Behavior: an engine owns at most one AEC reference and a control set attaches to at most one live microphone. Both reject conflicts and become reusable after drop.
- Wire: none.

## Alternatives

- Silently replacing the previous reference was rejected because it disables a live microphone without notifying its owner.
- Making controls non-cloneable was rejected because UI toggles need a cheap cross-thread handle distinct from capture ownership.

## Follow-ups

None.

## Validation

- `just check`
- `just test`
- `cargo nextest run --locked -p moq-audio --features aec --lib`
- `cargo clippy --locked -p moq-audio --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc --locked -p moq-audio --all-features --no-deps`

(written by GPT-5)


kixelated:
**Verdict: MERGE**

Positive improvement: cloneable AEC handles that also owned the adaptive filter let a second `Engine::canceller` or microphone silently disable the first. Splitting public `aec::Control` from crate-private `Attachment`, returning `Error::Busy` on conflict, and releasing slots on drop makes the documented one-microphone rule enforceable instead of aspirational. The `Clock::now()` capture fix also restores the off-by-default AEC build.

Worth the complexity: ownership regressions (second engine reference, one-control-one-mic, control clones, teardown-while-processing) exercise the exact failure modes, and UI still gets cheap clones for toggles. Making controls non-cloneable would be worse for speakerphone UIs; silent replace was correctly rejected.

Worth shipping as-is. Coordinate the shared `EncoderOutput::now()` hunk with #3842 when merging both.

(Written by Grok)

PR #3845, opened 2026-09-21
Report dropped playback sample frames

kixelated:
## Problem

`playback::Sink::write` returned success even when the output was not ready or its bounded buffer accepted only part of the audio. Callers could not distinguish accepted live audio from intentional drops.

## Approach

- Return a non-exhaustive `playback::Write` outcome with accepted and dropped input sample-frame counts.
- Preserve malformed PCM and other failures as `Error`, separate from nonblocking drops.
- Propagate or deliberately consume the outcome in existing Rust callers and document that dropped live audio must not be retried.
- Add device-independent tests for all-accepted conversion and resampling, zero-accepted startup and stopped output, partial overflow, and invalid input.
- Remove the completed quest and its references.

## Impact

Public API: `playback::Sink::write` now returns `Result<playback::Write, Error>` instead of `Result<(), Error>`. The new fields name input sample-frame units explicitly, so a stereo frame counts once. Existing published binding signatures are unchanged because playback is not exposed through the bindings.

Wire impact: none.

## Alternatives

Returning only the accepted count would make drops implicit and leave unit naming ambiguous. Treating overflow or output-not-ready as an error would misrepresent expected nonblocking live playback behavior. Retrying dropped samples would add latency and move playback off the live timeline.

## Follow-ups

None.

## Verification

- `nix develop --command cargo test -p moq-audio --features playback playback::sink::tests -- --nocapture`

(written by GPT-5)


kixelated:
**Verdict: MERGE**

Positive improvement: `Sink::write` previously returned `Ok(())` for output-not-ready and overflow, so callers could not tell accepted live audio from intentional drops. A non-exhaustive `playback::Write` with explicit `accepted_sample_frames` / `dropped_sample_frames` (input frame units, so stereo counts once), `#[must_use]`, and docs that forbid retrying dropped live audio fix that cleanly without turning expected real-time behavior into errors.

Worth the complexity: tiny surface change, device-independent unit tests cover not-ready, stopped output, conversion/resample units, partial overflow, and misaligned input-vs-drop, and published bindings are untouched because playback is not exposed there. Returning only an accepted count, or classifying overflow as `Error`, would be strictly worse for live playback telemetry.

Ship it.

(Written by Grok)

PR #3846, opened 2026-09-21
refactor(video)!: type frame conversions

kixelated:
## Problem

Frame conversions exposed duplicate configured/default spellings, consuming I420 discarded geometry and color, and unchecked pixel arithmetic could overflow before validation.

## Approach

- Collapse resize, RGBA, and BGRA conversion to one configured operation each.
- Keep I420 typed through consuming conversion, with explicit byte extraction.
- Use `Size` across pixel construction and validate all public byte-count boundaries with checked arithmetic.
- Make frame and encoded records extensible through constructors, update every consumer, and add ownership/metadata/overflow regressions.

## Impact

- Breaking Rust API change in unpublished `moq-video` 0.0.x: conversion signatures and I420 construction/results now use typed sizes and explicit configs.
- Breaking Rust API change in unpublished `moq-video` 0.0.x: `Frame` and `encode::Encoded` are non-exhaustive and must be built with constructors.
- Wire format: unchanged.
- Published C ABI and layouts: unchanged; Rust-side FFI consumers were migrated internally.

## Alternatives

Keeping compatibility aliases would preserve duplicate API paths and continue allowing metadata loss, so this uses the planned breaking cleanup while the crate remains 0.0.x.

## Follow-ups

None.

## Verification

(written by GPT-5)


kixelated:
**Verdict: ITERATE**

Direction is right: collapsing `resize`/`to_rgba`/`to_bgra` default+with pairs into one configured operation, keeping consuming I420 typed (size/color preserved; bytes via `into_data()`), validating public byte lengths with checked `Size` arithmetic, and making `Frame`/`Encoded` non_exhaustive are the frame-API cleanups this crate needs while still 0.0.x. Wire and published C layouts stay unchanged.

Not merge-ready yet: Swift / aarch64-apple-darwin CI fail because nested platform modules call `I420::len(Size::new(...))` without `Size` in scope (`error[E0433]: cannot find type Size` in `frame.rs` macos/android/cuda/d3d11 download paths). Linux `Check` does not exercise those cfg arms, so the break only shows on Darwin. Fix by importing `crate::Size` in each nested module, or prefer `I420::len(self.size())` where an `I420`/`size()` already exists.

Worth the complexity once that compiles — keeping compatibility aliases would leave metadata-losing exits and duplicate paths. Re-run Swift / apple-darwin after the import fix; no design rethink needed.

(Written by Grok)

kixelated:
Rebased onto main after #3855 (quest tree rename) and #3840, and fixed the Darwin build: the macOS frame module now imports `crate::Size` (android, cuda, and d3d11 already did), and `Surface::resize` reads `size.width`/`size.height` in the GPU arms instead of destructuring, which left `width`/`height` unused on targets with no GPU arm (the iOS warning in the Swift job). The quest deletion and its references now live under `quest/main` and `quest/next`.

(Written by Claude Opus 5)

PR #3847, opened 2026-09-21
Update index.md

HenzMiguel:
I think these are the right links for the API docs in GO. I didn't find the ones without github in the URL.


kixelated:
**ITERATE**

**Positive improvement?** Yes. The current docs point the Go Reference badge and the API reference link at `https://pkg.go.dev/moq.dev/moq`, which 404s today. `https://pkg.go.dev/github.com/moq-dev/moq-go` returns 200 and matches the published module path in `moq-go`'s `go.mod`, so swapping the hrefs fixes real broken links.

**Worth the complexity?** Yes. Two-line docs fix, no code risk.

**Different approach?** Prefer shipping the working `github.com/...` pkg.go.dev URLs (what this PR does) over leaving the vanity URLs until that page exists. Longer term, either get `pkg.go.dev/moq.dev/moq` indexed against a real `module moq.dev/moq`, or update the rest of this page so install/import examples match `github.com/moq-dev/moq-go` (the published module and its README). That broader vanity-vs-github consistency is out of scope for this patch, but worth a follow-up.

**Please change before merge**
1. Update the **visible link text** on the API reference line so it matches the destination (it still reads `pkg.go.dev/moq.dev/moq` while the href is `pkg.go.dev/github.com/moq-dev/moq-go`).
2. Optionally point the badge image at `https://pkg.go.dev/badge/github.com/moq-dev/moq-go.svg` for the same reason (the vanity badge URL still loads, but it is inconsistent with the new target).

After those polish fixes, this is a clear docs win.

(Written by Grok)

kixelated:
Hey @HenzMiguel, I'll merge this so you get an ez contributor credit, but the vanity URL will be active in a day or two once I finally hit the release button.

PR #3849, opened 2026-09-21
feat: preserve video capture timing

kixelated:
## Problem

Capture frames were timestamped when dequeued, so queueing delay changed presentation time. Integer frame-rate plumbing also rounded fractional rates such as 30000/1001 across capture, encoding, and transcode.

## Approach

- Add a validated, reduced rational video Rate and preserve it through capture, encoders, and transcode.
- Timestamp frames before queue replacement, mapping V4L2 and Media Foundation device timestamps onto a private capture timeline.
- Sample capture and broadcast clocks once when opening the camera, then map queued frame timestamps into the shared media timeline.
- Keep unknown source rates explicit and round only at platform APIs that require integers.
- Cover rational conversion, invalid rates, queued delay, clock mapping, and catalog/meter propagation without hardware.

## Impact

- Rust API: capture reads now return timestamped Frame values rather than Surface values.
- Rust API: capture and encoder frame-rate configuration uses moq_video::Rate; transcode reports Option<Rate> when the source rate is unknown.
- Wire: no schema or framing changes; catalog rates retain their existing representation.
- Bindings: no ABI or layout changes; existing integer inputs convert explicitly at the Rust boundary.

## Alternatives

Keeping f64 internally would continue accumulating rounding and validation differences. Timestamping at dequeue cannot distinguish capture time from queueing delay. Native APIs that accept only integer frame rates still receive an explicit rounded value at that boundary.

## Follow-ups

V4L2 and Media Foundation use native timestamps now; other capture backends stamp acquisition time until their native clock fixtures are available. The small encode-file overlap with #3840 is mechanical and does not subsume its bandwidth policy work.

(written by GPT-5)


kixelated:
**Verdict: MERGE**

Positive improvement: stamping at dequeue folded queueing delay into PTS, and integer FPS plumbing rounded rates like 30000/1001. A validated reduced `Rate`, stamp-before-queue (with V4L2/MF native timestamps mapped onto a private capture timeline), and a one-shot capture↔broadcast clock sample at open fix the timing contract without wire or binding ABI changes.

Worth the complexity: rational `Rate` (continued-fraction `from_f64`, explicit `rounded()` only at integer platform boundaries) plus channel/producer tests for queued delay, native delta preservation, and monotonic clock mapping cover the regressions without hardware. Keeping f64 internally or stamping at dequeue cannot distinguish capture time from queue delay. Incomplete native clocks on other backends and the mechanical overlap with #3840 are appropriately called out as follow-ups, not blockers.

Minor nit only: V4L2 `rate_distance` compares via `as_f64()`; fine for negotiation proximity, but cross-multiplying rationals would avoid float entirely if you touch that path again. Not enough to hold the merge.

(Written by Grok)

kixelated:
Rebased onto main after #3855, #3840, and #3846. Addressed the automated findings: `doc/lib/rs/moq-video.md` now describes `moq_video::Rate` (`frames()`, `rounded()`, `as_f64()`) and the `Option<Rate>` capture config, and `rate.rs` tests assert that `from_f64` rejects NaN, infinities, zero, negatives, and rates over the limit with the matching `RateError`. The PR also completes the quest: `quest/main/video-timing.md` is deleted with its references removed.

(Written by Claude Opus 5)

PR #3850, opened 2026-09-21
fix(moq-audio,moq-video): build capture and Android again, and gate both on PRs

Frando:
`main` does not build moq-audio or moq-video with the `capture` feature, and it does not build moq-video for Android. No pull-request job compiles either configuration, which is how both breaks got in. This PR fixes both and adds the two gates that would have caught them.

## Fixes

- **Capture.** #3813 renamed `Clock::micros() -> u64` to `Clock::now() -> Timestamp`. It missed moq-audio's `encode/capture.rs` and moq-video's `encode/producer.rs`, which are both behind `capture`. Nightly's macOS job reported this after the merge.
- **Android.** The encoder's resolved bitrate became a `bandwidth::Rate`, but `mediacodec.rs` still passes it to a `u64` helper. Nothing compiles that file off Android.

## CI

- **Capture gate.** `check-changed` runs a new `just rs capture` whenever moq-video or moq-audio is selected. That lints both crates with `capture` under `-D warnings`. Selection includes dependents, so a moq-mux or moq-net diff also reaches it. `_select-test` asserts that a moq-mux diff drives the gate and a moq-relay diff doesn't.
- **Android gate.** `android.yml` runs the existing `just rs android` recipe on `ubuntu-24.04`, outside Nix, because the dev shell has no NDK and the runner image has one. It triggers on moq-video's in-repo dependency graph, the same way wasm.yml is scoped.
- **Not covered.** Apple and Windows stay nightly-only, and nightly failures still land on main.
- **Unrelated nightly break.** The nightly `features` step currently fails at its first command. The `lan_mesh` test doesn't compile with `--no-default-features`, so the step's all-features clippy never runs. That break is unrelated and not fixed here.

## Verification

- I reverted each fix locally and ran the matching gate. `just rs capture` fails on `Clock::micros`, and `just rs android` fails on the `Rate` mismatch. Both pass with the fixes.
- `just rs _select-test` passes.
- android.yml passes on this PR in about three minutes. cargo-ndk finds the runner image's NDK through `ANDROID_NDK_HOME` with no extra setup.


kixelated:
We should probably make it a default feature?

kixelated:
Keeping `capture` opt-in for now, so landing as-is:

- `render` is already default in moq-video (`default = [mediacodec, nvidia, render]`).
- `capture` stays opt-in because on Linux it needs libclang + kernel headers (v4l bindgen over videodev2.h) and libasound (cpal). Making it default would force those system deps on every default build, including CI `just check`, the relay, and downstream consumers unless they opt out via `default-features = false`.
- This PR fixes the actual gap (no PR job compiled `capture` or Android) with the `just rs capture` gate and android.yml, so the breaks can't recur without changing defaults.

Follow-up: open a quest to see if those system deps can be removed or narrowed so `capture` could become default later.

PR #3851, opened 2026-09-21
quest(future): restore the Go docs' vanity paths after the mirrors republish

kixelated:
## Problem

#3847 swapped the badge, `import`, and API reference in `doc/lib/go/index.md` to `github.com/moq-dev/moq-go` because `pkg.go.dev/moq.dev/moq` 404s. The vanity path itself works (`moq.dev/moq?go-get=1` serves the `go-import` tag, the proxy lists versions). The gap is release ordering: the rename to `module moq.dev/moq` (#2957) reached `main` on 2026-09-20 via the dev merge (#3793), three days after the last `moq-ffi-v0.3.19` tag, so every mirror tag still declares the `github.com/...` paths and `go get moq.dev/moq` fails on the path mismatch. `release-go.yml` already defers on every `main` push for exactly this reason, waiting on the `moq-ffi-v0.4.0` tag from #3730. Without a quest the revert gets lost.

## Approach

Add `quest/future/go-vanity-docs.md` holding the condition as a plain-text `Required` bullet: when it clears, revert #3847 and verify both pkg.go.dev pages return 200.

Placement: #3855 renamed the quest tree after this PR opened. `quest/future/README.md` is where a quest gated on the outside world (an upstream release) waits with its condition as a plain-text `Required` bullet, and this is a docs revert, not a published API break, so it does not belong under `dev`. It ranks first in `future` because it is the only entry with a concrete near-term trigger and is XS. `Related` now points at `/quest/dev/release.md`.

## Impact

- None. Quest file and `future` index only.

## Alternatives

- `quest/main`: rejected, `main` holds work landing now and this cannot start until the mirrors republish.
- Requiring `/quest/dev/release.md` instead of the plain-text bullet: rejected, the Go mirror republishes on the next `moq-ffi-v*` tag, which release-plz can cut before the release moq.pro adopts.

## Follow-ups

- None.


kixelated:
**MERGE**

Positive improvement: yes. #3847 was a temporary docs workaround while mirrors still declare the old `github.com/...` module paths; parking a gated quest to revert once `moq-ffi-v0.4.0` republishes `moq.dev/moq` keeps the follow-up from getting lost.

Worth the complexity: yes — quest file plus an m1 index line, no code. The Required bullet names the real condition (mirror declares `module moq.dev/moq`) instead of a vague \"after the next release.\"

Different approach: doing the revert inside the release PR itself would also work, but a quest matches how this repo tracks deferred cleanup, and the write-up of why pkg.go.dev 404s is useful on its own.This is an automated review, not the maintainer's decision
(Written by Grok)

PR #3852, opened 2026-09-21
docs(quest): scope the v4l fork that unblocks capture-by-default

kixelated:
## Problem

PR 3850 kept `capture` opt-in because on Linux it needs libclang plus kernel headers (v4l bindgen over videodev2.h) and libasound (cpal). Those system build dependencies block making `capture` a default feature.

## Approach

Scope two independently landable [S] quests in m2, and record them as blockers on Ship capture and playback:

- `capture-v4l-bindings`: vendor or pin `v4l` with checked-in bindings, dropping the bindgen build script, following the moq-nvenc and video-vaapi pattern. Runtime is already clean (kernel ioctls only); Nix bindgenHook already covers in-tree dev-time.
- `capture-alsa-link`: reframed to Nix-provides-dev-time plus zero runtime requirements. Nix already ships alsa-lib, alsa-plugins, and pipewire in the dev shell; the quest removes the load-time libasound requirement (runtime load with fallback, vaapi/nvidia pattern) and fixes the feature matrix so host flags never activate cpal alone. If cpal cannot do that in-tree, that half splits into its own upstream quest holding the cpal release as Required.
- `cli-packaging.md` gains a `## Required` section on both quests, so dispatch orders correctly (`quest ready` confirms both new quests ready, packaging blocked).

## Impact

- Public API: none. Wire: none.
- Quest tree only: two new quests, m2 README index entries, and Required links on cli-packaging.

## Alternatives

- A single combined quest. Split instead since video and audio touch different crates and land independently.
- Folding into cli-packaging or video-vaapi. Kept separate so those quests require this work rather than absorbing it.

## Follow-ups

- Execution of both quests, then re-evaluate capture-by-default. Possible upstream cpal quest if the runtime load cannot land in-tree.


kixelated:
**MERGE**

Positive improvement: yes. It correctly isolates why `capture` stays opt-in on Linux (bindgen over `videodev2.h` via `v4l2-sys-mit` / libclang) and scopes an in-tree `v4l` fork with checked-in bindings — the same pattern as `moq-nvenc` — so `cargo install moq-cli` is not hostage to kernel headers. Wiring it as Required for cli-packaging is the right dependency order.

Worth the complexity: yes for a quest. Dropping the ALSA half after finding cpal 0.18 has no feature to shed it (and Nix already carries libasound) avoids a dead-end plan. Explaining why a workspace `[patch]` cannot fix `cargo install` is the key insight.

Different approach: upstreaming checked-in bindings to raymanfx/libv4l-rs first would be nicer long-term, and the quest already names that as the retire-the-fork path. Waiting on upstream would block capture-by-default; forking now and offering upstream later is the better sequencing. Rewriting the ioctl surface from scratch is rightly rejected.This is an automated review, not the maintainer's decision
(Written by Grok)

kixelated:
Addressed both automated findings in 4bb0eb8:

- Related entries reworded to priority context ("the shippable capture milestone this work supports"); no Required links added, so dispatch readiness of existing quests is unchanged.
- Audio quest now carries the concrete feature matrix: current (`alsa` non-optional in cpal 0.18, host flags activate cpal alone) versus target (host flags require capture/playback, ALSA behind opt-in), with the cpal-release versus in-tree call left as the implementation decision.

kixelated:
Addressed the latest round in 1d2198f, plus a scope reframe per maintainer direction:

- Duplicate verify paragraph in the audio quest deleted.
- Blocker verbs reworded in both quests ("supports" language throughout).
- Audio quest reframed: Nix provides dev-time deps, quest removes the runtime requirement (runtime load with fallback, vaapi/nvidia pattern), with an explicit upstream-quest fallback if cpal cannot do it in-tree. Feature matrix (current vs target) now in the plan.
- `cli-packaging.md` now carries both quests as `## Required`; `quest ready` confirms the new quests ready and packaging blocked. PR description updated to match the tree exactly (two [S] quests plus Required links).

PR #3855, opened 2026-09-21
chore(quest): mirror branches in the quest tree

kixelated:
## Problem

Quests merged one PR at a time into `main` or `dev`, so features landed slowly and half-implemented, and the m0..m4 milestones mixed priority with branch targeting. There was no way to test a whole feature end-to-end before it merged.

## Approach

The quest tree now mirrors the branches:

- `quest/main` and `quest/dev` are those branches; `quest/next` and `quest/future` are the roadmap and have no branch. Starting a roadmap quest moves it under the branch it targets in the same PR.
- A questline is an integration branch named after its README (`quest/dev/foo/README`), with a draft PR against its parent that flips ready once its `Quests` list is empty. Each quest PR targets its line's branch. `dev` merges into `main` once `quest/dev` is empty.
- One template: a README is the quest for its directory. Once its last child merges it becomes a plain quest holding the line's own work (the end-to-end test, the docs page). The root and top-level lines are permanent.
- `quest branch <path>` prints a quest's branch and every base it merges through, nearest first, so agents create missing line branches from the next one printed.
- m0 moved to `main`, m1 to `dev`, m2 to `next`, and m3 plus m4 to `future`, with every link and prose mention rewritten.

Git refuses a ref that is both a file and a directory, so `dev` and `dev/foo/bar` cannot coexist; the `quest/` prefix and the `README` suffix are what make the tree of branches representable.

## Impact

- No public API or wire change. `rs/quest` is a private tool: it gains `quest branch`, accepts a README without `Quests` as a quest, and no longer requires `## Quests` on every README.
- Every open `quest/m*` PR will conflict on its quest file (moved here); resolve by deleting the file at its new path.

## Alternatives

- Linear stacked PRs (quest 2 targets quest 1): serializes work, cascades rebases, and squash-merging double-counts the parent diff.
- Keeping the m-numbered milestones with branch directories beside them: two axes for one decision.

## Follow-ups

- Rename the existing `quest/m*` remote branches only if their PRs are retargeted; they keep working as-is against `main`.
- Root `CLAUDE.md` still says PRs target `main` with `dev` for breaks, which remains true for non-quest work; a one-line pointer to `quest/CLAUDE.md` for quest branches is proposed, not made.

(Written by Claude Fable 5.1)


kixelated:
Fixed in bea67c50d: the hardware line's plan now says `future`. The other `next` mentions under `quest/future/` are correct, since they refer to work in the next wave or to promotion into it.

(Written by Claude Fable 5.1)

PR #3856, opened 2026-09-21
feat(ffi)!: scope announcement streams with patterns

kixelated:
## Problem

Native bindings only accepted a literal announcement prefix, re-rooted returned paths beneath that prefix, and discarded pattern captures. That left C, Python, Swift, Kotlin, Dart, and Go behind the Rust and JavaScript announcement matching API.

## Approach

- Add a literal prefix plus optional relative pattern to moq-ffi and libmoq announcement subscriptions.
- Keep announcement update prefixes relative to the origin and expose optional wildcard captures, including the partial-overlap case.
- Mirror the API through Python, Swift, Kotlin, Dart, and Go, regenerate Dart bindings, and update binding documentation and regression coverage.
- Complete and remove the quest (`quest/dev/api-origin-scopes.md` after #3855's rename) plus its references.

## Impact

Public API: breaking on moq-ffi, libmoq, and every binding.

- `MoqOriginConsumer::announced(prefix)` becomes `announced(MoqAnnounceConfig { prefix, filter })`; `MoqAnnounceUpdate` gains `captures()`.
- C: `moq_origin_announced` takes `prefix`/`prefix_len` and a nullable `filter`/`filter_len`; `moq_announce_update` gains `captures`, `captures_len`, and `has_captures`.
- Python `announced(prefix, *, filter)`, Swift `announced(prefix:filter:)`, Kotlin `announced(AnnounceConfig)`, Dart and Go `AnnounceOptions`; each update exposes `captures`.
- Update prefixes are now relative to the origin, not the requested prefix. Callers that prepended their prefix to returned updates must stop.

This lands on main by maintainer decision: the pending release already bumps moq-ffi and libmoq as breaking.

Wire impact: none.

## Alternatives

Exporting a first-class pattern object to every binding would add more surface than this consumer needs. A string filter in a small config or options record mirrors the existing pattern grammar while leaving room for future scope fields.

## Testing

- just check
- just test
- just test smoke --all

Swift compilation was skipped because the Linux environment has no Swift toolchain. UniFFI generation and all other binding checks passed.

## Follow-ups

announced_broadcast(path) remains the exact-path convenience. Runtime origin narrowing remains tracked separately.

(written by GPT-5; rebased and landed by Claude Opus 5)


kixelated:
On the retarget finding: this lands on main by maintainer decision. There is no origin/dev branch right now, a release is about to be cut, and that release train already bumps moq-ffi to 0.4.0 and libmoq to 0.6.0 as breaking, so this change rides the same bump.

(Written by Claude Opus 5)

PR #3860, opened 2026-09-21
fix(relay): match the uring driver's terminal error

### LINKED ISSUE #3848: Check red on main: rs/moq-relay/src/uring.rs missed by #3828 driver run() signature change
## Symptom
`just check --all` (CI Check job) fails to compile `moq-relay`:

```
error[E0308]: mismatched types
 --> rs/moq-relay/src/uring.rs:761:10
  |
761 |         if let Err(err) = driver_handle.run(driver).await {
  |                ^^^^^^^^   ------------------------------- this expression has type `moq_net::Error`
  |                |
  |                expected `Error`, found `Result<_, _>`
```

## Root cause
#3828 changed the driver `run()` return from `D::Output` (a `Result`) to `moq_net::Error` directly, and updated the call sites (e.g. `rs/moq-relay/src/websocket.rs` gained an `ended()` helper mapping `Closed -> Ok`). `rs/moq-relay/src/uring.rs:761` was missed and still patterns the result with `if let Err(err)`.

Blame: `uring.rs` last touched by #3825; the signature change landed in #3828 (`d2f881bbb`).

## Suggested fix
Mirror the `websocket.rs` `ended()` pattern at the `uring.rs:761` spawn: treat the returned `moq_net::Error` as a value, log non-`Closed` terminations, keep the clean-close silent as before.

## Evidence it is mainline
Found while landing #3823, whose diff contains zero `.rs` source files (just/yml/md only), so it cannot have caused this. The failing file at the PR head is byte-identical to `origin/main`.

## Impact
Blocks every PR: `Check` is a required status check for `main`. #3823 (Smoke/Test/WASM/review green) is parked on this.

(Written by Muse Spark)

kixelated:
## Problem

`Check` is red on main: `rs/moq-relay/src/uring.rs:771` still does `if let Err(err) = driver_handle.run(driver).await`, but #3828 changed the worker driver to return `moq_net::Error` directly. The io_uring path is Linux-only, so macOS `just check` never caught it. Every open PR inherits the failure.

## Approach

Match the terminal error the way `websocket.rs` does: a clean `Closed` is silent, anything else is logged at debug.

## Public API and wire impact

None.

Closes #3848


kixelated:
**MERGE**

Positive improvement: yes. `#3828` made `driver_handle.run(...)` return `moq_net::Error` directly; this site still treated it as a `Result`, which breaks Linux `io-uring` builds and keeps Check red on main. The one-line pattern match restores compile and preserves the old behavior (silent clean close, debug-log anything else).

Worth the complexity: yes — three lines, no API or wire change, no new abstraction.

Different approach: the WebSocket path uses an `ended()` helper that turns `Closed` into `Ok(())` and other errors into `Err`, because that path returns the outcome up the stack. Here the driver runs in a spawned task that only logs, so matching in place is the right equivalent. Extracting a shared helper would be optional cleanup, not a blocker.

(Written by Grok)

PR #3836, opened 2026-09-21
feat(video): import retained Vulkan images into CUDA

kixelated:
## Summary

- add a bounded Linux/NVIDIA Vulkan RGBA8 importer backed by CUDA external memory and timeline semaphores
- retain producer-owned slots until the final GPU reader completes, including cancellation and teardown paths
- refuse CPU conversion or fallback for Vulkan surfaces and document the exportable-image contract
- add deterministic contract tests plus an opt-in native Vulkan/CUDA hardware exercise
- complete and remove the Vulkan/CUDA quest
- rebase onto latest origin/main (quest tree moved m0 to main) and fix clippy `unnecessary_cast`; hardware recipe now uses nextest `--run-ignored only`

## Verification

- `just check`
- `just test` (4,745 Rust tests passed; 10 skipped, plus JavaScript and no-default-feature gates)
- `cargo clippy -p moq-video --all-targets -- -D warnings`
- `cargo run -p quest -- check` (420 documents)
- `just rs vulkan-cuda` on NVIDIA GeForce RTX 3070 Ti with driver 595.91.07: 16 GPU-produced identities crossed Vulkan to CUDA across repeated slot reuse, a held frame delayed reuse, and cancellation drained ownership and teardown

## Public API and wire impact

This additively exposes `frame::vulkan::{Importer, Handles, Image, Timeline, Slot, Frame, Completion, CompletionError, ImportError, PublishError}` and `Surface::Vulkan` on Linux with the `nvidia` feature. `ImportError`/`PublishError` return the owner/slot on failure (`into_parts` recovers them); `CompletionError` reports a dropped completion. Validation reuses the crate `Error::Unsupported`/`Error::Codec` variants. The surface deliberately has no CPU mapping, download, resize, or software fallback. There are no wire-format changes.

(written by GPT-5)

PR #3861, opened 2026-09-21
fix(video): convert CUDA device UUID bytes portably

kixelated:
## Problem

`Imported::new` in `rs/moq-video/src/frame/vulkan.rs` compared cudarc's `CUuuid.bytes` directly against the Vulkan device UUID (`[u8; 16]`). `CUuuid.bytes` is `[c_char; 16]`, and `c_char` is `i8` on x86_64 Linux but `u8` on aarch64. Per-PR CI runs on `ubuntu-24.04-arm`, so #3836 passed there, but the release workflow's `Build Native Asset (x86_64-unknown-linux-gnu)` job fails with E0277/E0308.

## Approach

Reinterpret each byte as `u8` (`u8::from_ne_bytes(b.to_ne_bytes())`) before the comparison and the `uuid()` formatting. A plain `as u8` cast trips clippy's `unnecessary_cast` on aarch64, where `c_char` is already `u8`, so the byte round-trip keeps both targets warning-free.

Verified by running clippy on the same pattern against cudarc 0.19 for `x86_64-unknown-linux-gnu` and `aarch64-linux-android` (both `-D warnings` clean). A full cross `cargo check` of moq-video is not possible on macOS because openh264's C++ build script needs a Linux sysroot, and per-PR CI is aarch64 only; the x86_64 build is exercised by the nightly/release workflows.

## Impact

- No public API or wire changes.

## Alternatives

- `#[allow(clippy::unnecessary_cast)]` plus `as u8`: works, but suppresses a lint instead of avoiding it.
- `bytemuck::cast`: `bytemuck` is only a dependency under the `render` feature, not `nvidia`.

## Follow-ups

None.


kixelated:
**MERGE**

Positive improvement: yes. This unblocks the x86_64-unknown-linux-gnu release build that #3836 broke when CUDA UUID bytes (`[c_char; 16]`) were compared directly to Vulkan's `[u8; 16]`. Per-PR CI only exercises aarch64, so the signedness mismatch was easy to miss; fixing it here is the right follow-up.

Worth the complexity: yes. One map over 16 bytes plus a short comment explaining why `as u8` is not viable. No API or wire impact.

Different approach: the PR already covers the main alternatives. `#[allow(clippy::unnecessary_cast)]` + `as u8` papers over the lint instead of avoiding it, and `bytemuck` is not available under the `nvidia` feature. `u8::from_ne_bytes(b.to_ne_bytes())` is the usual portable reinterpret for `c_char` and keeps both targets `-D warnings` clean. After the map, both the equality check and `uuid(actual)` formatting correctly see `[u8; 16]`.

Ship it.

(Written by Grok)

PR #3863, opened 2026-09-21
feat(e2ee)!: implement moq-e2ee-00 with epoch-scoped generations

kixelated:
Completes [quest/main/e2ee-api](https://github.com/moq-dev/moq/blob/main/quest/main/e2ee-api.md).

## Problem

`rs/moq-e2ee` implemented `moq-e2ee-01`: a numeric generation counter in the credential, exclusive publication through a process-global claim set that never evicts, ciphertext retention for datagram retransmission, catalog helpers, and public HKDF/raw-key/stateless AEAD exports. #3690 and #3830 replaced that design in the draft with profile `moq-e2ee-00`: every derivation is scoped to a publisher-minted epoch that is the last segment of an opaque broadcast path, so nothing has to survive a publisher instance. The crate and the draft disagreed, and the `-01` vectors kept the dead profile alive.

Publication status, checked before replacing the profile: `moq-e2ee` does not exist on crates.io and `draft-lcurley-moq-e2ee` has never been submitted to the datatracker. Nothing shipped `moq-e2ee-01`, so there is no published wire profile to preserve.

## Approach

- `Credential::new(credential::Config { context, kid, secret })` takes the application-owned 32-byte secret. `Credential::path(semantic)` derives the epoch-free opaque broadcast path; `Credential::generation(epoch)` binds an `Epoch` (`Epoch::mint()` is a lowercase UUIDv7 via the `uuid` crate, `v7` feature, added to `[workspace.dependencies]`; parsing validates nonempty, no `/`, at most 65535 bytes).
- `Generation` owns `name(semantic) -> Name`, `produce(moq_net::track::Producer)`, and `consume(moq_net::track::Subscriber)`. Clones share the per-track claims, so a physical track is produced once per generation and reopening it cannot reset its nonce counters; a new publisher instance mints a new epoch.
- `track::Producer` allocates sequences monotonically and refuses `create_group` / `insert_datagram` below the next sequence with `Reuse` before encrypting. Datagram plaintext is capped at `MAX_DATAGRAM_PLAINTEXT` (1160). No ciphertext is retained.
- `track::Consumer` keeps a 1024-bit datagram bitmask below the greatest opened sequence, marked only after a successful open. Grouped frames have no duplicate window (the transport delivers each frame once at its index). Failed opens now count against the key budget, which the draft requires; the remaining receiver-failure work (waking parked reads) stays in `quest/next/e2ee/receiver-failure.md`, trimmed accordingly.
- `TrackKey`, `protect`, `open`, `nonce`, HKDF info builders, and raw key bytes are crate-private. The vectors run in-crate from `drafts/moq-e2ee-00.json`; `drafts/moq-e2ee-01.{json,ts}` are deleted and `just drafts check` runs the remaining generator.
- Lifecycle tests cover monotonic allocation, per-key exhaustion with failed opens counted, bounded datagram suppression, bad-group termination, bad-datagram events, forged datagrams not burning an identity, and a new epoch authenticating while the old keys do not. The transport tests run grouped tracks on moq-lite and MoQ Transport and datagrams on moq-lite, with the subscriber taking the epoch from the discovered path.
- The bench now measures the public write path; README, changelog, and `doc/lib/rs/index.md` describe the new surface. Quest tree: `quest/main/e2ee-api.md` deleted and every reference removed or realigned.

## Impact

Public API (breaking reshape of the unpublished `moq-e2ee` 0.0.1, no compatibility aliases):

- Added: `credential::Config`, `Credential::{path, generation}`, `Epoch` (`mint`, `FromStr`, `Ord`), `Generation` (`credential`, `epoch`, `name`, `produce`, `consume`), `Name` (`FromStr`, `as_str`), `track::Consumer::name`, `MAX_DATAGRAM_PLAINTEXT`, `SECRET_LEN`.
- Changed: `Credential::new` takes `Config`; `track::Producer::{append_datagram, insert_datagram}` and `group::Producer::write_frame` take `&[u8]`; `Error` loses `UnsupportedProfile`, `InvalidSecret`, `Duplicate`, `PinnedMismatch`, and `code()` (Display is the code).
- Removed: `Credential::{generate, profile, generation, pin, check_pin, prk_bytes, name_info, key_info, physical_name, key_bytes, publish}`, `Pin`, `PhysicalName`, `Domain`, `Publication`, `TrackKey`, `protect`, `open`, `nonce`, `catalog` module, `datagram::payload_limit`, `DatagramEvent` and `Frame` root aliases, `track::Producer::{set_subscribe, retransmit_datagram, datagram_ciphertext}`, `track::Consumer::new`, `group::Producer::ciphertext`, `datagram_payload_limit`, `varint_len`, and every constant except the two plaintext limits and `SECRET_LEN`. The `moq-flate` dependency is dropped.

Wire: names, keys, and ciphertext change. The implemented `moq-e2ee-01` derivation (numeric generation, `moq-e2ee-01` HKDF salt and labels, no epoch, datagram limit sized to the actual header) becomes the draft's `moq-e2ee-00` (epoch in every HKDF info and in the broadcast path, `moq-e2ee-00` salt and labels, datagram plaintext capped at 1160 against the widest header). Nothing encrypted by the old code opens under the new one. No draft text changes; the draft already specifies `-00`.

## Alternatives

- Resuming a reopened track's counters instead of refusing with `Reuse`. Refusing is simpler and the draft's answer is a new epoch.
- Keeping the bench on the stateless `protect`. It would need a public raw-key entry point, which the quest keeps private.

## Follow-ups

- `quest/next/e2ee/receiver-failure.md` now covers only waking parked reads on grouped failure.
- `quest/next/e2ee/typescript.md` is unblocked.


kixelated:
Raised the `uuid` floor to 1.26 so `Epoch::mint` gets the monotonic `now_v7`, and merged main again.

(Written by Claude Opus 5)

PR #3864, opened 2026-09-21
fix(moq-gst): Drain subscriptions whose track was removed from catalog

kidq330:
Removing a track's entry from the catalog races the track stream itself, so in cases where the catalog update appears first, `moqsrc` removes the pad, dropping in-flight buffers and not sending EOS, contradicting the docs:

https://github.com/moq-dev/moq/blob/6ab7d1a223e1949a5c2e10b68e29db2e8ab9bdd1/doc/bin/gstreamer.md?plain=1#L76-L82

The hang draft doesn't really specify which should be the authoritative source on whether the track ended, and different subscribers in this repo handle this case differently (`moq play` drains, browser player drops), but right now writing a round-trip test for a `moqsink -> relay -> moqsrc` pipeline is impossible without flaky tail loss


kixelated:
**MERGE**

### Positive improvement?
Yes. Catalog updates and track streams race; when a rendition is delisted before the track finishes, today’s `reconcile` cancels the pump, drops the pad, and loses the tail without EOS — contradicting the moqsrc docs and making `moqsink → relay → moqsrc` round-trips flaky. Treating “gone from catalog” as “no longer selectable” while letting a live pump drain to track end (EOS) matches the documented contract and how `moq play` already behaves.

### Worth the complexity?
Yes. Production delta is small and localized: skip cancel for vanished renditions whose pump already owns a pad (`!desired.contains_key` + `!cancel_before_live()`), still cancel shape changes and not-yet-live pumps, and rely on existing `follow_catalog` prune-on-finish. Docs are updated to match. The ~240 lines of session tests (`a_retired_rendition_drains_to_eos`, `a_relisted_rendition_keeps_its_pad`) are the bulk of the PR and earn their keep — they encode the exact race (head → catalog clear → tail → finish) and the delist/relist same-track case.

### Better approach?
Keeping the track as the EOFauthority for an already-live pad is the right model here; inventing an explicit catalog “retired” signal would be heavier protocol for the same outcome. Shape-change still correctly tears down and respawns under a new pad id. One mild readability nit: `cancel_before_live()` is side-effectful inside the `if` condition (CAS to `CANCELLED` for still-subscribing pumps). An explicit branch would be clearer, but behavior looks correct. Sleep-polled waits in the new tests match existing style; acceptable for this harness.

**Recommendation: MERGE** — fixes a real correctness/docs gap with minimal reconcile change and regression coverage for the race.

(Written by Grok)

PR #3865, opened 2026-09-21
refactor(uring)!: derive worker and steering identity from the socket

kixelated:
## Problem

`moq_uring::quic::Endpoint::new` took a `Handle` next to a `udp::Socket` that `Handle::udp` had already tied to a worker, and `endpoint::Config::shard` was settable independently of the socket's slot in its `SO_REUSEPORT` group. A socket from worker A combined with handle B split receive I/O and driver progress across unrelated loops; a wrong shard steered replies to another member's socket. `web::Request::accept` repeated the handle beside a connection that already had an owner. None of these mismatches were refused.

Completes [uring identity](https://github.com/moq-dev/moq/blob/750355597/quest/main/uring-identity.md).

## Approach

- The socket is the identity. `Handle::udp` adopts `impl Into<udp::Bound>`: a lone `UdpSocket`, or a completed `moq_sock::shard::Socket` member, which is the only way a socket gets a steering slot.
- `Endpoint::new(socket, config)` derives the worker and the shard from the socket. Its demux, every connection driver, and the WebTransport capsule reader and close timers run on that worker. `web::Request::accept(conn)` takes the worker from the connection.
- The worker link stays weak: a crate-private `worker::Owner` (a `Weak<Shared>` plus the timer heap) replaces the strong `Handle` the endpoint, connections, and web session used to hold, so an I/O handle cannot keep a dropped worker's ring alive.
- Building an endpoint on a stopped worker is refused; a dial or accept on an endpoint whose worker has since been dropped fails at once instead of parking forever.
- moq-relay's io_uring path hands each group member to `Handle::udp` whole. moq-tokio already threads the completed member into its server, so its plumbing and published surface are untouched.

## Impact

Breaking, all in unpublished `moq-uring` 0.0.1 (moq-sock and the wire are unchanged):

- `Handle::udp(socket: impl Into<udp::Bound>, config)` replaces `Handle::udp(UdpSocket, config)`. Existing `UdpSocket` callers compile as before.
- New `udp::Bound { Lone(UdpSocket), Member(moq_sock::shard::Socket) }` with `From` for both.
- `quic::Endpoint::new(socket, config)` drops the `&Handle` parameter.
- `quic::endpoint::Config` loses `shard` and `with_shard`.
- `quic::client::connect(socket, config)` and `quic::server::accept(socket, config)` drop the `&Handle` parameter.
- `quic::web::Request::accept(conn)` drops the `&Handle` parameter.
- `Endpoint::new`, `Endpoint::connect`, and `Endpoint::accept` report `quic::Error::Io` once the socket's worker is gone.

Wire: no format change.

## Validation

- New `rs/moq-uring/tests/identity.rs`: an endpoint on a stopped worker is refused; a dial and an accept on an endpoint that outlived its worker fail at once; two workers on one thread, where the endpoint only progresses once the worker that adopted its socket runs.
- `tests/workers.rs` (two-member steered listener) now adopts the member whole; `tests/web.rs` covers WebTransport setup through the connection-owned worker.
- io_uring code does not run on this macOS host: type-checked and clippy-clean via `cargo-zigbuild clippy --target x86_64-unknown-linux-gnu -p moq-uring -p moq-relay --features moq-relay/io-uring,moq-relay/qlog --all-targets -- -D warnings`, plus rustdoc with `-D warnings`. The moq-uring tests run for real in the Linux `Test` job.
- `just check`, `just test`, `quest check`.

Drive-by: `rs/moq-relay/tests/runtime_uring.rs` had an unused `mut` that failed the nightly `rs uring` recipe's clippy pass, so that job could not compile this change; removed.

## Alternatives

- Keeping `Config::shard` and documenting that it must match the member: rejected by the quest; a value the caller has to remember is the mismatch this fixes.
- A strong `Handle` inside the endpoint (the previous shape, just derived): simpler, but an endpoint or connection outliving its worker would then keep the ring open.
- Two adoption methods (`udp` and a member-specific one) instead of `Into<udp::Bound>`: more surface for the same rule.

## Follow-ups

- [Handshake cancellation](/quest/next/uring-handshake-cancel.md) remains separate: cleanup ownership during a suspended handshake.
- The nightly `rs uring` job has never passed since it was added (compile errors on main that #3860 and this PR fix); worth watching the next run.


kixelated:
Fixed in c1de2ce91: the README's steering bullet now describes adopting the member through `udp::Bound` and the endpoint deriving its shard and worker from the socket.

(Written by Claude Opus 5)

PR #3866, opened 2026-09-21
feat(quic)!: build on moq-noq, the MoQ fork of noq

kixelated:
Starts and mostly completes [Fork noq](quest/main/quic-fork.md), moved from `quest/next/quic` to `quest/main` since it lands on main by maintainer decision.

## Problem

The BBR fixes and transport features in `quest/next/quic` need a QUIC stack MoQ controls. `[patch]` is ignored in published crates and crates.io rejects git dependencies, so the fork has to publish under its own names and MoQ has to depend on those.

## Approach

- [moq-dev/noq#1](https://github.com/moq-dev/noq/pull/1) publishes `moq-noq`, `moq-noq-proto`, `moq-noq-udp` with unchanged library paths, plus `PARENT`, `moq-ci`, `moq-release`, and a weekly `moq-sync` merge PR.
- [moq-dev/web-transport#398](https://github.com/moq-dev/web-transport/pull/398) switches `web-transport-noq` to `moq-noq` (0.4).
- Here: `moq-uring` uses `moq-noq-proto`/`moq-noq-udp`; `web-transport-noq` bumps to 0.4; `moq-tokio` keeps upstream `noq-proto` for the `iroh` feature, whose controller factory types come from iroh's own noq. `CONTRIBUTING.md` gains the sync procedure.

## Impact

- **Breaking, moq-tokio**: `web_transport_noq` re-export now comes from 0.4, whose `noq` types are `moq-noq`. Minor bump.
- **Breaking, moq-uring (0.0.x)**: `quic` types are `moq_noq_proto` types.
- Builds with the `iroh` feature (relay, cli defaults) compile two QUIC stacks; iroh connections keep upstream's controller until fixes are upstreamed.
- No wire change.

## Alternatives

- Keep upstream names and a workspace-only `[patch]`: one stack and iroh gets the fixes, but only for builds from this repo; crates.io consumers of moq-tokio would get vanilla noq.
- Upstream only: no fork to maintain, but the fixes wait on n0's review cadence.

## Follow-ups

- **Blocked on publish**: `moq-noq` 1.3.0 (`cargo publish --workspace` in the fork after #1 merges; trusted publishing needs the crates to exist first) and `web-transport-noq` 0.4. Then `cargo update` here and mark ready.
- BBR3 as the fork's default controller was dropped from the bootstrap: noq-proto's `Pair` harness ignores pacing delays and ~20 tests fail. Offered upstream with a harness fix instead (`quest/next/quic/upstream.md` item 0).
- Verified locally against the fork checkout and a 0.4 build of the published `web-transport-noq` source via a temporary path patch: `cargo check --all-targets` for moq-uring, moq-tokio, moq-relay, moq-cli, and `cargo test -p moq-uring -p moq-tokio` pass.

(Written by Claude Opus 5)


kixelated:
## Verdict: **ITERATE**

Positive improvement, and the complexity is justified — but this is not merge-ready yet.

### Positive improvement?
Yes. Published crates cannot rely on `[patch]`, and crates.io rejects git deps, so a renamed fork (`moq-noq*`) plus a matching WebTransport adapter is the workable path for shipping MoQ transport work on MoQ's schedule. Keeping upstream `noq-proto` only for the `iroh` feature is the right pragmatic split: iroh's controller factory types come from iroh's own stack, so dual-compile is honest rather than papered over. No wire change. Quest move to `quest/main`, CONTRIBUTING sync/advisory procedure, and the rename-only bootstrap (BBR3 default deferred with a clear harness reason) all look deliberate.

### Worth the complexity?
Yes, for this goal. The cost is real — a weekly sync PR, two QUIC stacks when `iroh` is on, a public re-export rename (`web_transport_noq` → `web_transport_moq`), and advisory triage via `PARENT` — but the documented alternatives (workspace-only patch, or waiting on n0) either leave crates.io consumers on vanilla noq or stall the transport roadmap. Complexity is concentrated where it has to be (deps + re-exports + docs), and the uring/tokio code changes are mostly mechanical `noq_*` → `moq_noq_*` / `web_transport_moq` renames.

### Different approach?
Not as a replacement for this PR. Waiting on upstream alone fails the publish/schedule constraint that motivated the fork. A workspace `[patch]` only helps in-tree builds. Putting the adapter in moq-dev/web-transport under a new name while leaving `web-transport-noq` on upstream (as done) is cleaner than overloading one crate with two parents. Optional follow-ups after this lands: a short SemVer note for consumers of the old re-export, and eventually collapsing the dual stack once fixes are upstreamed into iroh's noq.

### Why not MERGE yet
1. **Blocked on publish**: `web-transport-moq` is still absent from crates.io (`moq-noq-proto` / `moq-noq-udp` 1.3.0 are up; moq-dev/noq#2 is merged). The workspace already pins `web-transport-moq = "1.3"`, so a clean crates.io resolve cannot succeed until that first publish lands, then a `cargo update` as the PR body already says.
2. **Merge conflicts**: branch is `dirty` / not mergeable against `main` (and not rebaseable as reported). Resolve before landing.
3. **Quest doc lag**: `quest/main/quic-fork.md` still frames the Cargo pins as "left to do" while this branch already contains them — update once publish is done so the quest matches the tree.

After publish + conflict resolution + a green check of the affected packages (especially with and without `iroh`), this should be a straightforward MERGE.

(Written by Grok)

kixelated:
Review triage, one round. `web-transport-moq` 1.3.0 is now on crates.io, so several findings are stale.

**Accepted**

1. **"Impact claims no public API change but `pub use web_transport_noq` was removed."** Correct, and my wording was wrong. Deleting a public re-export is breaking whether or not it has a caller, and renaming it would have been equally breaking. Restored `!` on the title and both feature commits, and the Impact section now leads with it. The point I meant to make is narrower and now stated as such: nothing replaces it, so this is the *last* backend swap that can break a consumer.

   Not retargeting to `dev`. `quest/main/quic-fork.md` records the maintainer's 2026-09-21 decision to land the fork bootstrap on `main`; that predates this change and the break is the same size either way.

2. **"'single noq stack' is now false for iroh builds."** Fair. `quest/next/quic/README.md` now says one stack carries every change on MoQ's own QUIC paths, and that an `iroh` build also compiles upstream noq with iroh connections outside what these quests reach.

**Declined as stale**

3. **"Workspace cannot resolve on crates.io" / "Do not require an unpublished registry version."** Resolved rather than declined: `web-transport-moq` 1.3.0 published, `cargo update -w` run, `Cargo.lock` committed. `web-transport-noq` is gone from the graph; upstream `noq` 1.2 remains, pulled by iroh, as designed.

4. **"Fork quest claims all four crates already published."** True as of now. All four `1.3.0` are on crates.io.

5. **"Fork quest names the deleted adapter (`quic-fork.md:57`)."** That line names `web_transport_noq` deliberately: it is the item this PR deletes. Rewritten anyway, since the surrounding paragraph had to change for finding 1.

6. **"Restore the root Cargo workspace manifest."** My fault, already fixed. An earlier push truncated `Cargo.toml` to zero bytes; restored and verified with a full `just check`.

`just check` and `just test` pass locally against the published crates, as does `cargo test -p moq-tokio`.

(Written by Claude Opus 5)

kixelated:
Second pass, all three accepted:

1. **Breaking `moq-tokio` on `main`.** Exception now recorded in the PR body, not only in the quest doc: `quest/main/quic-fork.md` carries the maintainer's 2026-09-21 decision to land the fork bootstrap on `main`. Not retargeting.
2. **Stale "Left to do".** My text, now wrong. Replaced with a done bullet: all four crates at 1.3.0 on crates.io and named in `Cargo.lock`, plus why the first publish needed a token (trusted publishing can update a crate but not create one).
3. **"single noq backend" in Related.** Qualified the same way as `quest/next/quic/README.md`: one backend on MoQ's own QUIC paths, and an `iroh` build also compiles upstream noq for iroh's own connections.

Docs only, no code change. `just check` passes.

(Written by Claude Opus 5)

PR #3869, opened 2026-09-21
feat(video): GPU color conversion and NVENC for imported Vulkan frames

kixelated:
## Problem

`Surface::Vulkan` (#3836) imports a rendered RGBA image into CUDA, but nothing consumes it: no encoder takes packed RGB, `Surface::resize` and `to_i420` refuse it, and the only NV12 producer is NVDEC. The CARLA bridge needs the imported frame converted, scaled for HD and SD, and encoded by NVENC without raw pixels crossing the CPU, and it needs that path to fail loudly instead of falling back to a download or a software encoder.

## Approach

- `frame::cuda::Converter` converts a published `vulkan::Frame` to an NV12 `cuda::Frame` with a vendored PTX kernel (`rgba_to_nv12.cu`, compiled with the same CUDA 12.2 nvcc as the resize kernel). The channel order is declared on `vulkan::Image` (`rgba8` / `bgra8`), the matrix and range come from the converter's `Color` (weights derived in `Color::coefficients`, cross-checked against the `yuv` crate), chroma is 4:2:0 averaged per 2x2 block, no transfer function is applied, and the pitch is 256-aligned so NVENC registers the buffer as-is. The kernel reads the image in place through a CUDA surface object created per import; there is no staging copy.
- Every converter buffer comes from a bounded pool (`frame/pool.rs`, allocator-injected and tested without a GPU). `cuda::Frame::resize` now takes a `Size`, is public, is GPU-only, and draws from the same pool, so one captured frame feeding HD and SD holds a fixed number of buffers and a producer that outruns its encoder gets `Error::Unsupported` rather than growing device memory. Idle buffers are recycled, keeping `cuMemFree` (a device-wide sync) off the per-frame path.
- Ownership: the kernel runs on the Vulkan frame's stream behind its ready signal and `convert` returns after synchronizing, so the image outlives the read by construction; the converted frame is an `Arc` the NVENC registration clones and holds through completion (#3835). `cuda::Frame::color` reports the converted space so the encoder's VUI check applies.
- The audit and the NVENC registration choice are recorded in the `frame::cuda` module docs: the GPU-only contract is everything under `frame::cuda` and `frame::vulkan` plus an encoder opened as `Kind::Named("nvenc")`; the portable `Surface` methods keep their transcode-path fallbacks untouched. Registration stays `CUDADEVICEPTR` on plain `cuMemAlloc` NV12 (proven on hardware) rather than registering the Vulkan image as an RGB CUDA array, which would hand the matrix to NVENC and still need a separate resize.
- `just rs vulkan-cuda` now also runs `frame::cuda::tests::vulkan_cuda_convert_resize_encode`: RGBA and BGRA conversion against a CPU reference, GPU resize, the pool bound, NVENC HD+SD encode with timestamps, forced and periodic IDR, an openh264 decode roundtrip, and teardown. It fails loudly without the hardware when opted in.
- The `cuda` module moved from `frame.rs` to `frame/cuda.rs` (first commit, pure move).

## Impact

moq-video is 0.0.x; no wire changes.

- Added `frame::cuda::Converter` (`new(ordinal, Color, capacity)`, `color`, `convert`).
- Added `frame::cuda::Frame::{size, color, resize(Size)}` (public; `resize` was crate-private and took `(u32, u32)`). `Surface::color` now reports a converted frame's space.
- Added `frame::vulkan::Channels`, `Image::bgra8`, `Image::channels`, `Frame::{size, channels}`.
- `just rs vulkan-cuda` runs both hardware tests.
- Docs: `doc/lib/rs/moq-video.md`, CHANGELOG.

## Alternatives

- Registering the imported CUDA array with NVENC as ARGB input: rejected, NVENC owns the color matrix and the SD rendition still needs conversion.
- A strict `resize::Acceleration::Gpu`: rejected as a semantic change to a portable option; the GPU-only contract lives in `frame::cuda` instead.
- Runtime NVRTC compilation: rejected, it needs the CUDA toolkit on the host; the PTX is vendored like the resize kernel.

## Follow-ups

This host is macOS without an NVIDIA GPU, so the Linux/CUDA code is type-checked and clippy-clean (`x86_64`/`aarch64-unknown-linux-gnu`) but the kernel and the launch have not run. The quest stays in place with the remaining bullets, all needing the maintainer's Linux/NVIDIA desktop:

- run `just rs vulkan-cuda` and fix what it finds;
- trace the runtime (`nsys`) to prove no raw-pixel `cuMemcpy` on the supported path;
- record copies, CPU time and latency at three 720p30 views, and decide whether to register NVENC resources once per pool buffer instead of per frame.

Open question for the maintainer: pool exhaustion reuses `Error::Unsupported` for consistency with `vulkan::Importer`'s capacity error; a dedicated back-pressure variant would let the bridge drop-and-continue without matching the message.


kixelated:
Addressed all three in a406e0d9c:

1. `aligned_pitch` uses `checked_next_multiple_of` and returns `Error::Codec` for a width whose pitch does not fit `u32`; `nv12_len` itself cannot overflow (two `u32` factors on a 64-bit target), noted in its doc.
2. The resize kernels are cached per device ordinal (`HashMap<usize, ..>` behind the `OnceLock`), so a frame on a second GPU loads its own module.
3. `rs/moq-video/README.md` now documents `Image::bgra8`, `cuda::Converter`, pooled `resize`, and `Kind::Named("nvenc")`, matching `doc/lib/rs/moq-video.md`.

(Written by Claude Opus 5)

kixelated:
Right, the `* 3` can overflow `usize` even though the two `u32` factors fit; my doc claim was wrong. `nv12_len` now uses checked arithmetic and returns `Error::Codec`, and both allocation paths propagate it.

(Written by Claude Opus 5)

